PR #41, opened 2026-04-17
feat(discovery): auto-discover repos by GitHub topic tag

### LINKED ISSUE #39: feat: auto-discover repositories by GitHub topic tag
## Summary

Replace the static `HEIMDALLM_REPOSITORIES` list with automatic discovery based on a GitHub **topic tag** (e.g. `heimdallm-review`). Repositories tagged with the configured topic are automatically monitored — no need to maintain a comma-separated list that must be updated every time a repo is added or removed.

> Migrated from theburrowhub/heimdallr-docker#19 (archived). Adapted to Heimdallm naming and aligned with the filtering model discussed in #24.

## Motivation

Currently, monitored repos must be explicitly listed in `HEIMDALLM_REPOSITORIES` or `config.toml`. This doesn't scale well: every new repo requires updating the config and restarting/reloading the service. A tag-based approach lets teams self-service by simply adding a topic to their repo.

## Proposed Behavior

1. New config option: `HEIMDALLM_DISCOVERY_TOPIC` (env var) / `[github] discovery_topic` (TOML)
   - Example: `HEIMDALLM_DISCOVERY_TOPIC=heimdallm-review`
2. On each poll cycle (or a separate slower cycle), Heimdallm queries the GitHub Search API for repos with that topic within the configured org(s)
   - `GET /search/repositories?q=topic:<tag>+org:<org>`
3. Discovered repos are **merged** with any explicitly listed in `HEIMDALLM_REPOSITORIES` (static list still works as override/addition)
4. The explicit list (`HEIMDALLM_REPOSITORIES`) remains supported for repos where you can't control topics

## Configuration

```toml
[github]
# Static list (existing behavior, always respected)
repositories = ["org/repo-legacy"]

# Topic-based auto-discovery (new)
discovery_topic = "heimdallm-review"

# Optional: limit discovery to specific orgs (empty = infer from static repos)
discovery_orgs = ["freepik-company", "theburrowhub"]

# Discovery refresh interval (separate from poll_interval to avoid rate-limiting)
discovery_interval = "15m"
```

Env vars:
- `HEIMDALLM_DISCOVERY_TOPIC=heimdallm-review`
- `HEIMDALLM_DISCOVERY_ORGS=freepik-company,theburrowhub`
- `HEIMDALLM_DISCOVERY_INTERVAL=15m`

## Usage

```bash
# Instead of listing every repo:
# HEIMDALLM_REPOSITORIES=org/repo1,org/repo2,org/repo3,...

# Just set the topic:
HEIMDALLM_DISCOVERY_TOPIC=heimdallm-review
```

Then tag repos in GitHub Settings > Topics with `heimdallm-review`.

## Interaction with issue tracking filters (#24)

This feature discovers **which repos to monitor**. The issue tracking filters proposed in #24 (`develop_labels`, `review_only_labels`, `organizations`, `assignees`, `filter_mode`) determine **how to act on issues within those repos**. Both systems are complementary and independent:

- **Discovery** (this issue): _"which repos does Heimdallm watch?"_ — via topic tag + static list
- **Issue filtering** (#24): _"which issues within those repos does Heimdallm process, and how?"_ — via labels, org, assignee, filter mode

A repo discovered by topic is subject to the same issue tracking filters as a statically configured one. No special handling needed.

## Tasks

- [ ] Add `discovery_topic`, `discovery_orgs`, `discovery_interval` to config (TOML + env vars)
- [ ] Implement `FetchReposByTopic(topic, orgs)` in `internal/github/client.go`
- [ ] Merge discovered repos with static list in poll cycle (deduplicate by `org/repo`)
- [ ] Run discovery on separate interval (`discovery_interval`) to avoid rate-limiting
- [ ] Cache discovered repos between cycles (only update on discovery tick, not every poll)
- [ ] Update README and `docker/.env.example` with new config options
- [ ] Add tests for discovery + merge logic

## Notes

- Discovery is additive: it never removes repos from the static list
- If `discovery_topic` is empty, behavior is identical to today (static list only)
- This can be implemented independently of #24 (issue tracking) — both modify config but in separate sections

ivanmunozruiz:
## Summary

Implements #39. Opt-in repository auto-discovery: tag a repo on GitHub with a configured **topic** and Heimdallm starts monitoring it — no more manual `HEIMDALLM_REPOSITORIES` updates every time a team adds a repo.

- New config: `discovery_topic`, `discovery_orgs`, `discovery_interval` (TOML + env vars `HEIMDALLM_DISCOVERY_*`).
- The discovered list is merged with the static `repositories` list at poll time; the static list still works and takes precedence in ordering.
- Runs on its own `discovery_interval` loop (default `15m`) — independent from the PR poll, so the stricter Search API rate limit (30 req/min) can't starve the PR pipeline.
- Reload picks up changes to topic/orgs/interval and restarts the loop without restarting the daemon.

## Design decisions

- **`discovery_orgs` is mandatory when `discovery_topic` is set.** Avoids accidentally scanning all of GitHub if someone sets the topic without thinking about scope. Fails fast at config validation.
- **Topic format validated** against GitHub's rules (lowercase, digits, hyphens, ≤50 chars) to prevent silent 422s from the API.
- **Graceful degradation:** if the Search API fails (rate limit, 5xx, network), the last known-good cache is preserved — an outage never silently empties the monitored set. Partial failures update with what we got and still surface the error in logs.
- **Per-org queries** so one failing org does not poison the others. Results are deduplicated and sorted deterministically.
- **Service-style separation:** discovery lives in its own `internal/discovery` package, decoupled from the GitHub client via a small `ReposFetcher` interface → trivial to unit-test without an HTTP server.

## Tests

New tests covering:

- **Config validation** — topic format, orgs required, interval parsing, env var overrides.
- **GitHub client** `FetchReposByTopic` — single/multi-org, dedup, archived/disabled filtering, pagination (>100 repos), partial and total failures.
- **Discovery service** — cache preserved on total failure, partial failure updates, `Discovered()` returns a copy (no external mutation), ticker loop, context cancellation, zero-interval guard.
- **`MergeRepos`** helper — static ordering preserved, dedup, empty cases.

All new and existing tests pass:
```
ok  github.com/heimdallm/daemon/internal/config
ok  github.com/heimdallm/daemon/internal/discovery
ok  github.com/heimdallm/daemon/internal/github
```
(The pre-existing `TestDetect_Fallback` failure in `internal/executor` is environment-dependent and unrelated — reproduces on `main`.)

## Docs updated

- `README.md` — new "Auto-discover repositories by topic" subsection.
- `docker/.env.example` — all three `HEIMDALLM_DISCOVERY_*` vars documented.
- `docker/config.example.toml` — commented TOML example.

## Test plan

- [ ] `go test ./...` in CI (should be green on whatever runner doesn't have stray CLI binaries in PATH).
- [ ] Manual: set `HEIMDALLM_DISCOVERY_TOPIC=heimdallm-review` + `HEIMDALLM_DISCOVERY_ORGS=theburrowhub` on a dev daemon, tag one repo, confirm it appears in the poll loop within `discovery_interval`.
- [ ] Manual: revoke the token mid-discovery and confirm the cached list is preserved (log should say `discovery: refresh failed, keeping previous cache`).
- [ ] Manual: edit `config.toml` to toggle discovery on/off, hit `POST /reload`, confirm the loop starts/stops.

Closes #39

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,95 @@
+# AGENTS.md
+
+Instructions for AI coding agents (Claude Code, Cursor, Copilot, …) and human
+collaborators working on this repo. These rules override the agent's default
+workflow — follow them exactly.
+
+## Go tests must run in Docker
+
+**Always use `make test-docker` for the Go daemon test suite.**
+
+```bash
+make test-docker
+# narrow down to specific packages/tests:
+make test-docker GO_TEST_ARGS="-run TestDiscovery ./internal/discovery/..."
+```
+
+### Why this is non-negotiable
+
+`go test` compiles a transient binary named `<pkg>.test` in
+`/var/folders/.../go-build/` (macOS) or `/tmp/go-build...` (Linux), then runs
+it. These tests:
+
+- open random local ports via `httptest.NewServer`,
+- send HTTP requests with `Authorization: Bearer <token>` headers,
+- exit within milliseconds.
+
+That signature matches heuristics used by corporate EDR agents
+(Elastic Security, CrowdStrike, SentinelOne, Defender for Endpoint, …).
+The EDR will kill the process mid-test — surfaced to the user as
+`signal: killed` — and send a **malware alert** to the security team, who
+will then open a ticket with the developer. **This has already happened
+once in this repo (2026-04-17).** We do not want to do it again.
+
+Running the tests inside a Docker container means the ephemeral binaries
+live in the container's tmpfs, not the host filesystem. The host EDR never
+sees them.
+
+### What `make test-docker` guarantees
+
+Hardening details are inlined in the Makefile target. Summary:
+
+| Concern | Guarantee |
+|---|---|
+| Image | Official `golang:1.21-alpine`, **pinned by SHA256 digest** (not a mutable tag) |
+| Host filesystem access | Single mount: repo at `/src`, **read-only** (`:ro`) |
+| Build cache | Redirected to `/tmp/heimdallm-gocache/` — does not touch `~/.cache` or `~/go` |
+| Privileges | Runs as invoking user (`--user $(id -u):$(id -g)`) — no root in the container |
+| Container persistence | `--rm` — destroyed on exit, no `*.test` artefacts on the host |
+| Network | Default bridge (needed for `go mod download` on first run) |
+| Ports | None exposed |
+
+If IT/Security asks how we run Go tests on their laptops, point them at
+this section and the `test-docker` target.
+
+## When `go test` on the host is OK
+
+Never, on a corporate laptop with an EDR, unless Security has explicitly
+whitelisted `$GOCACHE` and `/var/folders/.../go-build/`. Raising a ticket
+for that exception is preferable to bypassing the rule.
+
+On personal machines, CI runners, or Docker-in-Docker sandboxes, `go test`
+directly is fine. The CI workflow (`.github/workflows/tests.yml`) runs on
+`macos-14` GitHub Actions runners and uses `setup-go` natively — that
+environment has no EDR.
+
+## Flutter tests
+
+Flutter tests do not trip EDR heuristics (Dart VM, no ephemeral native
+binaries). Run them normally:
+
+```bash
+cd flutter_app && flutter test
+# or combined with Go tests (note: `make test` runs Go on the host):
+make test
+```
+
+**Prefer `make test-docker && cd flutter_app && flutter test`** over
+`make test` on EDR-protected endpoints.
+
+## Conventional commits
+
+`release-please` reads commit prefixes to bump semver and generate the
+changelog. Stick to: `feat:`, `fix:`, `refactor:`, `docs:`, `chore:`,
+`ci:`, `test:`. A `feat!:` or trailer `BREAKING CHANGE:` forces a major
+bump.
+
+## Do not
+
+- Run `go test` / `go build` on the host on a corporate laptop with EDR.
+- Stage `.DS_Store` in commits.
+- Push directly to `main` — always open a PR.
+- Skip hooks (`--no-verify`) without an explicit user instruction.
+- Invent a different Docker image/tag for tests — the pinned digest in the
+  Makefile is the canonical one. Bump it deliberately, in its own commit,
+  when upgrading Go.
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ else
   APP_BUNDLE       := $(FLUTTER_BUILD)/bundle
 endif
 
-.PHONY: build-daemon build-app test dev dev-daemon dev-stop \
+.PHONY: build-daemon build-app test test-docker dev dev-daemon dev-stop \
         release-local package-macos install-service verify-linux clean
 
 # ── Build ─────────────────────────────────────────────────────────────────────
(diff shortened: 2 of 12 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured feature with good test coverage, graceful degradation on API errors, and clean integration with the existing reload lifecycle. A few thread-safety, resource-management, and validation concerns worth addressing before merge.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:268** — startDiscovery is documented as 'must be called with cfgMu held', but the discovery loop's goroutine calls discoverySvc.Refresh which does network I/O. Meanwhile makePollFn reads discoverySvc.Discovered() outside cfgMu. This is safe only because Service uses its own internal RWMutex. The comment is misleading — cfgMu protects the cancel-func swap, not the service itself. Clarify the comment to avoid future confusion.
⚠️ **daemon/cmd/heimdallm/main.go:282** — On reload, the old discovery loop is cancelled AFTER starting the new one. Between startDiscovery(newCfg) and oldDiscoveryCancel(), two goroutines are running the same topic query concurrently, doubling Search API rate-limit consumption. Cancel the old loop before starting the new one (still outside cfgMu if blocking is a concern).
🟡 **daemon/cmd/heimdallm/main.go:270** — startDiscovery copies DiscoveryOrgs with `append([]string(nil), ...)`, which is correct, but DiscoveryTopic is a string (immutable in Go), so that's fine. However, the cfg pointer itself is later mutated on reload, and makePollFn closes over `c` (a *Config). If makePollFn's closure captures the old Config pointer and a reload replaces cfg, the poll function created before reload still references the old config's static repos, which is the intended behavior — but this subtlety deserves a comment since it's easy to break.
🟡 **daemon/internal/discovery/discovery.go:97** — Run() returns silently when interval <= 0 without performing even one refresh. This means if someone misconfigures the interval to 0 (after bypassing validation somehow), discovery is completely disabled with only a warning log. Consider whether one initial refresh should still happen.
⚠️ **daemon/internal/github/client.go:289** — fetchReposForOrg builds the search query using fmt.Sprintf with user-supplied topic and org values injected directly into the GitHub Search API query string. While these values are validated by config (topic via regex, orgs as org names), if validation is ever bypassed (e.g., programmatic use), a malicious topic like `org:evil-org` could widen the search scope. Consider URL-encoding or quoting the values defensively.
🟡 **daemon/internal/github/client.go:270** — FetchReposByTopic collects errors as formatted strings in a []string slice and joins them. Using a proper error slice ([]error) with errors.Join (Go 1.20+) would be more idiomatic and allow callers to unwrap individual errors.
🟡 **daemon/internal/github/client.go:313** — The pagination loop breaks when len(result.Items) < 100. If the API returns exactly 100 items on the last page (total_count is an exact multiple of 100), the loop makes one extra request that returns 0 items. This is harmless but wasteful — consider also checking against result.TotalCount.
🟡 **daemon/internal/config/config.go:189** — HEIMDALLM_DISCOVERY_ORGS env override splits on comma and trims spaces, but does not validate individual org names (e.g., reject names with spaces or special characters). The topic is validated via regex in validateDiscovery(), but org names are accepted as-is. A malformed org name would silently produce zero results rather than failing fast.
🟡 **daemon/internal/discovery/discovery.go:80** — Refresh() updates both cache and lastErr under the same lock, which is correct. However, the success log in Run() calls s.Discovered() immediately after Refresh(), acquiring the read lock a second time. This is fine for correctness but the count could be captured from the repos slice returned by the fetcher to avoid the extra lock round-trip.
🟡 **daemon/cmd/heimdallm/main.go:259** — discoverySvc is created unconditionally via NewService(ghClient) even when discovery is disabled (no topic configured). This is harmless (Discovered() returns nil), but it means there is always an allocated Service object. Minor — not a real issue, just noting it.

### Suggestions

- Cancel the old discovery loop before starting the new one on reload to avoid concurrent Search API queries that double rate-limit usage.
- Consider adding a maximum bound or validation on DiscoveryOrgs length to prevent accidentally querying dozens of orgs and burning through the Search API rate limit (30 req/min).
- Add an integration-style test (or at least a test in discovery_test.go) that exercises the reload path: cancel old loop, start new loop with different topic/orgs, verify the cache updates correctly.
- The README documents HEIMDALLM_DISCOVERY_ORGS as 'required' but only when topic is set — consider adding a note that it is ignored when topic is empty, to avoid user confusion.
- Add a metric or structured log field for the discovery cache size on each poll cycle, so operators can monitor whether discovery is working without digging through logs.

---

PR #72, opened 2026-04-20
fix(daemon): make shared api_token readable to web container

### LINKED ISSUE #71: bug(docker): web UI cannot read /data/api_token due to UID mismatch + 0600 perms
## Síntoma

Tras \`make up\` limpio, la web UI arranca y muestra:

> Daemon unreachable — check \`HEIMDALLM_API_URL\`. ({\"message\":\"daemon token missing: set HEIMDALLM_API_TOKEN or mount /data/api_token\"})

El \`/data/api_token\` **sí está montado** en el contenedor web (shared volume \`heimdallm-data:/data:ro\`) pero el código del web no puede leerlo.

## Causa

UID mismatch cruzado con permisos \`0600\`:

\`\`\`bash
# Daemon container
$ id
uid=100(heimdallm) gid=101(heimdallm)
$ ls -la /data/api_token
-rw------- 1 heimdallm heimdallm 65 ... /data/api_token

# Web container (en el mismo volumen)
$ id
uid=1000(node) gid=1000(node)
$ cat /data/api_token
cat: can't open '/data/api_token': Permission denied
\`\`\`

El daemon (imagen \`node:20-alpine\` con user \`heimdallm\` UID 100) escribe el token con perms \`0600\`. La web (imagen \`node:22-alpine\` con user \`node\` UID 1000) no puede leerlo porque los UIDs difieren y los perms son owner-only.

## Impacto

El workaround actual (\`make setup\`) **siempre** es obligatorio, no opcional como prometía el README. Newcomers siguen los 5 pasos del README, tienen todo \"healthy\" y aun así la UI no conecta.

### A. Abrir perms del token en el daemon (mi recomendación)
Cambiar la escritura a \`0644\` (o \`0604\`). El volumen es privado al compose y sólo contiene este token + la SQLite. No hay exposición a terceros.

Donde buscar: el daemon genera/rota el token y lo escribe en \`/data/api_token\`. Probablemente en \`daemon/internal/keychain/\` o en el arranque del \`server/\`.

### B. Coordinar UIDs entre imágenes
Hacer que \`heimdallm\` y \`node\` compartan UID. Feo — obliga a parchear una de las dos base images o crear un usuario custom en ambas.

### C. Compartir via grupo
Perms \`0640\`, añadir \`node\` al grupo \`heimdallm\` (GID 101). Obliga a modificar la web image y sigue dependiendo de que los GIDs casen.

Claramente A. No hay vector de ataque realista contra un token que sólo vive dentro de un volumen Docker privado.

## Criterios de aceptación

- [ ] \`docker compose -f docker/docker-compose.yml up -d\` → la web UI conecta al daemon sin necesitar \`make setup\`.
- [ ] \`ls -la /data/api_token\` desde el contenedor web es legible.
- [ ] \`make setup\` sigue funcionando como fallback (no regresión).
- [ ] El README no necesita mencionar \`make setup\` como paso obligatorio.

## Relacionados

- Descubierto setting up un checkout limpio tras #68.
- El README actual ya explica que \`make setup\` es \"opcional\"; ese texto sólo se mantiene fiel si este issue se cierra.

ivanmunozruiz:
Closes #71.

## Summary

The daemon wrote `<dataDir>/api_token` with mode `0600`. The SvelteKit web container mounts the same data volume and tries to read that file to authenticate against the daemon, but it runs as `node` (UID 1000) while the daemon runs as `heimdallm` (UID 100) — `0600` blocks any UID other than the owner, so the web UI always failed with *\"daemon token missing: set HEIMDALLM_API_TOKEN or mount /data/api_token\"*. The workaround was `make setup`, which the README advertises as optional.

Fix: write the token with `0644` (new `tokenFileMode` constant), and chmod existing `0600` tokens on startup so upgrades don't need the file to be deleted.

## Why 0644 is safe here

- The volume (`heimdallm-data`) is private to the compose stack.
- The only consumers are two first-party services (`heimdallm` daemon and `heimdallm-web`) that we ship.
- The sibling file in the same volume (`heimdallm.db`) already uses the default `0644`.
- The token is already rotated on demand via `make setup`.

No credential crosses a trust boundary that `0600` was actually defending.

## What changed

- `daemon/cmd/heimdallm/main.go`
  - New `tokenFileMode = 0644` constant with a comment pointing at #71.
  - `loadOrCreateAPIToken`:
    - `O_CREATE|O_EXCL` opens with `tokenFileMode`.
    - Explicit `os.Chmod` after creation so the final mode is deterministic regardless of daemon umask.
    - Best-effort `os.Chmod` after a successful read of an existing token, to upgrade older installs in place. Failures are logged and non-fatal (`make setup` stays a viable fallback).
- `daemon/cmd/heimdallm/main_test.go`
  - `TestLoadOrCreateAPIToken_NewFileIsWorldReadable` — fresh create produces `0644`.
  - `TestLoadOrCreateAPIToken_LegacyFileIsUpgradedTo0644` — seeds a `0600` token, asserts the loader returns the same token *and* leaves the file at `0644`.
  - `TestLoadOrCreateAPIToken_ShortFileIsRegenerated` — documents current behaviour around malformed tokens; skips if the loader returns an error (pre-existing bug, not in scope here).

## Out of scope (deliberately not touched)

- `docker-compose.yml` — the `heimdallm-data:/data:ro` mount is correct; the fault was at the producer.
- Web UI token-loading code (`web_ui/src/lib/server/token.ts`) — the error it surfaces is legitimate, we're just making sure it no longer fires under the happy path.
- `make setup` — still useful for scripts / CI / local curl from the host. Not regressed.

## Test plan

- [x] `make test-docker` — all packages green, including the three new tests.
- [ ] Reviewer: `make down -v && make up` on a fresh clone (no pre-existing volume) — web container connects without `make setup`.
- [ ] Reviewer: retain an existing `0600` token volume, pull the fix, restart daemon — token is upgraded to `0644` in place, web connects.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -719,10 +719,21 @@ func (n *notifyWithSSE) Notify(title, message string) {
 	n.notifier.Notify(title, message)
 }
 
+// tokenFileMode is the permission mask for <dataDir>/api_token.
+//
+// 0644 (world-readable) is deliberate: the file lives on a Docker volume
+// that is private to the compose stack and is consumed by two services we
+// control (the daemon that writes it and the SvelteKit web UI that reads
+// it). Those services run under different UIDs in their respective images
+// (daemon: heimdallm UID 100; web: node UID 1000), so the previous 0600
+// blocked the web container from reading the token via the shared volume,
+// forcing operators to run `make setup` as a manual workaround. See #71.
+const tokenFileMode = 0644
+
 // loadOrCreateAPIToken reads an existing token from <dataDir>/api_token, or
-// generates a new cryptographically-random one and writes it with mode 0600.
-// The token is used by the HTTP server to authenticate all mutating requests
-// (POST/PUT/DELETE) — see security issue #3.
+// generates a new cryptographically-random one and writes it with
+// tokenFileMode. The token is used by the HTTP server to authenticate all
+// mutating requests (POST/PUT/DELETE) — see security issue #3.
 //
 // SECURITY (M-4): Uses O_CREATE|O_EXCL to create the file atomically. If two
 // daemon instances race, only one will win the exclusive create; the other reads
@@ -736,6 +747,13 @@ func loadOrCreateAPIToken(dir string) (string, error) {
 	if err == nil {
 		tok := strings.TrimSpace(string(data))
 		if len(tok) >= 32 {
+			// Best-effort upgrade for tokens written by older daemons with
+			// mode 0600 — see tokenFileMode comment above. Errors are logged
+			// but non-fatal: the daemon itself can still read the token, and
+			// `make setup` remains a viable fallback.
+			if err := os.Chmod(path, tokenFileMode); err != nil {
+				slog.Warn("api_token: could not upgrade permissions", "path", path, "err", err)
+			}
 			return tok, nil
 		}
 	}
@@ -750,7 +768,7 @@ func loadOrCreateAPIToken(dir string) (string, error) {
 	// Use O_CREATE|O_EXCL for atomic creation: if another process created the
 	// file between our ReadFile and here, os.OpenFile returns an error that
 	// satisfies os.IsExist — we then read the file created by the other process.
-	f, err := os.OpenFile(path, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0600)
+	f, err := os.OpenFile(path, os.O_CREATE|os.O_EXCL|os.O_WRONLY, tokenFileMode)
 	if err != nil {
 		if os.IsExist(err) {
 			// Another process created the file first — read their token.
@@ -768,6 +786,12 @@ func loadOrCreateAPIToken(dir string) (string, error) {
 	defer f.Close()
 	if _, err := fmt.Fprintf(f, "%s\n", tok); err != nil {
 		return "", fmt.Errorf("api_token: write %s: %w", path, err)
+	}
+	// os.OpenFile's mode arg is masked by the process umask (typically 0022),
+	// which would leave the file 0644 anyway — but chmod explicitly so the
+	// final mode is deterministic regardless of the daemon's umask at startup.
+	if err := os.Chmod(path, tokenFileMode); err != nil {
+		slog.Warn("api_token: could not set permissions", "path", path, "err", err)
 	}
 	slog.Info("api_token: created new token", "path", path)
 	return tok, nil
--- a/daemon/cmd/heimdallm/main_test.go
+++ b/daemon/cmd/heimdallm/main_test.go
@@ -1,6 +1,9 @@
 package main
 
 import (
+	"os"
+	"path/filepath"
+	"strings"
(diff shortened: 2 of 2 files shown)


Muriano (approved):
## 🤖 Heimdallr AI Review

Clean, well-documented PR that relaxes api_token file permissions from 0600 to 0644 so the web container (different UID) can read the token from a shared Docker volume. The security trade-off is justified by the private-volume context and thoroughly explained in comments. Tests cover all three code paths (new file, legacy upgrade, short-token edge case). No blocking issues found.

### Suggestions

- The short-token path (file exists with <32 chars) silently fails because O_EXCL refuses to overwrite. The test documents this with t.Skipf, which is fine for now, but consider opening a follow-up issue to handle truncated/corrupt tokens gracefully (e.g., truncate+rewrite instead of exclusive create).
- Consider using 0640 with a shared group instead of 0644 to limit readability to only the intended containers rather than world-readable, even though the volume is private to the compose stack — defense in depth.
- Minor: the two os.Chmod calls (upgrade path at ~line 755 and new-file path at ~line 792) share identical error-handling logic. If this pattern grows, a small helper could reduce duplication, but for two call sites it's fine as-is.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-reasoned fix for a real cross-container permission issue. The 0600→0644 change is justified by the Docker-private volume constraint, and the code is thoroughly commented and tested. Main concern is that 0644 (world-readable) is broader than necessary—0640 with a shared GID would be tighter. The test suite transparently documents a pre-existing short-token regeneration gap.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:729** — 0644 makes the API token world-readable on the host filesystem, not just within Docker. If the data directory is ever bind-mounted or accessible to other users on the host, any local user can read the token. Consider 0640 with a shared GID between the daemon and web container for defense-in-depth; this keeps cross-UID access while blocking 'other'.
🟡 **daemon/cmd/heimdallm/main.go:750** — The best-effort Chmod on the read path is non-atomic with the ReadFile: another process could replace the file between the read and the chmod. In practice this is harmless (the file is long-lived and replaced only on regeneration), but worth noting the TOCTOU gap.
🟡 **daemon/cmd/heimdallm/main_test.go:197** — TestLoadOrCreateAPIToken_ShortFileIsRegenerated uses t.Skipf to document that short-token regeneration silently fails (O_EXCL refuses to overwrite). This is a pre-existing bug but the PR makes it visible—consider tracking it as a separate issue rather than leaving it as a skipped test that could mask future regressions.

### Suggestions

- Prefer 0640 with a shared GID (e.g., add both containers to a common group in the Dockerfiles) instead of 0644. This keeps cross-UID readability while not exposing the token to every user on the host.
- Track the short-token regeneration gap (O_EXCL preventing overwrite of malformed tokens) as a follow-up issue so it doesn't stay buried in a t.Skipf.
- Extract the magic number 32 (minimum token length) into a named constant alongside tokenFileMode for clarity and single-source-of-truth.

---

PR #76, opened 2026-04-20
fix(daemon): make /logs stream work under Docker

### LINKED ISSUE #75: bug(docker): web UI /logs never shows daemon output in Docker mode
## Síntoma

En la UI Docker, \`/logs\` muestra permanentemente:

> Reconnecting…  
> (log file not found at /home/heimdallm/.local/share/heimdallm/heimdallm.log — daemon may be running in dev mode or log path differs)

No hay manera de ver los logs del daemon desde la web UI — hay que ir a \`docker compose logs -f heimdallm\` para verlos.

## Causa

\`daemon/internal/server/handlers.go:702 handleLogsStream\` tail-ea un fichero en disco:

\`\`\`go
func daemonLogPath() string {
  home, _ := os.UserHomeDir()
  switch runtime.GOOS {
  case "darwin":
    return filepath.Join(home, "Library", "Logs", "heimdallm", "heimdallm-daemon-error.log")
  default:
    return filepath.Join(home, ".local", "share", "heimdallm", "heimdallm.log")
  }
}
\`\`\`

Pero \`setupLogging()\` configura slog contra **\`os.Stderr\`**:

\`\`\`go
slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{...})))
\`\`\`

En Docker, stderr va al pipe que Docker redirige a \`docker logs\`. Ningún fichero se crea. La vista \`/logs\` quedó diseñada para el deployment nativo (macOS LaunchAgent / Linux systemd con redirección) y nunca funcionó bajo contenedor.

## Fix propuesto

1. **Escribir logs a fichero además de stderr** — \`setupLogging\` usa \`io.MultiWriter(os.Stderr, file)\` contra \`<dataDir>/heimdallm.log\`. Stderr sigue llegando a \`docker logs\` (nada se pierde), y el fichero lo puede tail-ear el endpoint HTTP.
2. **Actualizar \`daemonLogPath()\`** para preferir \`\$HEIMDALLM_DATA_DIR/heimdallm.log\` (o \`/data/heimdallm.log\` cuando la env var no está pero \`/data\` existe), cayendo al path XDG/macOS si no hay nada.
3. El volumen \`heimdallm-data:/data\` ya persiste \`/data\`, así que el log sobrevive restarts. No hace falta tocar compose.

Rotación queda fuera de alcance de este bug (follow-up si el fichero crece demasiado).

## Criterios de aceptación
- [ ] \`make up\` → abrir \`http://localhost:3000/logs\` muestra líneas del daemon en vivo, sin reconnect loop.
- [ ] \`docker compose logs heimdallm\` sigue mostrando las mismas líneas (no se pierden).
- [ ] Deployment nativo (macOS LaunchAgent / Linux host) sigue usando el path XDG anterior cuando \`HEIMDALLM_DATA_DIR\` no está set.
- [ ] Tests para \`daemonLogPath()\` cubren ambos modos.

ivanmunozruiz:
Closes #75.

## Summary

The web UI's \`/logs\` view kept showing *\"log file not found at …heimdallm.log — daemon may be running in dev mode or log path differs\"* and reconnecting in a loop. Root cause: \`setupLogging()\` wired slog to \`os.Stderr\` only, but Docker redirects stderr to \`docker logs\`, never to a file — while \`handleLogsStream\` tails a file path. The two ends disagreed about where \"logs\" live.

Fix: have \`setupLogging()\` mirror slog output to \`<dataDir>/heimdallm.log\` via \`io.MultiWriter(os.Stderr, file)\`, and teach \`daemonLogPath()\` to prefer that same location. \`docker logs heimdallm\` keeps working — nothing is lost.

## What changed

- \`daemon/cmd/heimdallm/main.go\`
  - \`setupLogging\` now takes a \`dataDir\` and opens \`<dataDir>/heimdallm.log\` in append mode (0640) alongside stderr. I/O failures degrade to stderr-only with a \`slog.Warn\` — the daemon never refuses to start because of disk logging.
  - Exported \`DaemonLogFileName\` so the writer and the HTTP handler agree on the name.
  - Call site reordered: resolve \`dataDir()\` **before** \`setupLogging\` so the log file lands in the same volume as \`heimdallm.db\`.

- \`daemon/internal/server/handlers.go\`
  - \`daemonLogPath()\` priority now matches \`dataDir()\` exactly:
    1. \`\$HEIMDALLM_DATA_DIR/heimdallm.log\`
    2. \`/data/heimdallm.log\` (when \`/data\` exists as a directory — Docker convention)
    3. macOS LaunchAgent path
    4. \`\$XDG_STATE_HOME/heimdallm/heimdallm.log\`, with \`~/.local/share/heimdallm/heimdallm.log\` as final fallback.

- \`daemon/internal/server/handlers_internal_test.go\` (new)
  - Three regression tests for \`daemonLogPath()\` in the internal package so the unexported helper stays covered without widening its visibility.
  - Env vars saved/restored via a \`withEnv\` helper to keep the suite hermetic under parallel runs.

## Why not rotate?

Deliberately out of scope. The log file will grow indefinitely under long-lived Docker deployments. A follow-up issue can add either a built-in rotator or a documented \`logrotate\` config — whichever the team prefers — but fixing visibility is the blocker today.

## Out of scope

- \`docker-compose.yml\` — untouched. The existing \`heimdallm-data:/data\` mount already persists the log across restarts.
- \`docker/Dockerfile\` — untouched. \`/data\` is already \`chown heimdallm:heimdallm\` so the new \`O_CREATE\` succeeds as the runtime user.
- Flutter / macOS LaunchAgent path — same path as before, just now reached via a clearer precedence chain.

## Test plan

- [x] \`make test-docker\` — all packages green, including the three new tests.
- [x] \`docker compose up -d --build heimdallm\` on a fresh volume: \`/data/heimdallm.log\` is created with mode \`0640\`, owner \`heimdallm\`, contains the \`"daemon started"\` line.
- [x] \`docker compose logs heimdallm\` shows the same lines — no regression for \`docker logs\` users.
- [ ] Reviewer: hit \`/logs\` on \`http://localhost:3000\` after rebuild — lines stream live, connection pill stays green.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -55,7 +55,12 @@ func main() {
 	// /logs stream reads that file; writing only to stderr (as we used
 	// to) left the stream empty under Docker — see #75.
 	logDir := dataDir()
-	setupLogging(logDir)
+	logFile := setupLogging(logDir)
+	if logFile != nil {
+		// Flush buffered writes on shutdown so the last lines reach
+		// disk even when the daemon is killed mid-log.
+		defer logFile.Close()
+	}
 
 	cfgPath := configPath()
 	var cfg *config.Config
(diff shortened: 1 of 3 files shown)


Muriano:
## ⚠️ MEDIUM Issue

DaemonLogFileName constant defined in main.go is not used here — 'heimdallm.log' is hardcoded as a string literal in multiple places (lines 700, 703). If the filename ever changes in main.go, this handler will look for the wrong file and the /logs stream will silently break. Import and use the constant, or move it to a shared internal package. Severity: medium because it's a maintainability trap that would silently break a user-facing feature.

**Location:** `daemon/internal/server/handlers.go` line 700

---

Muriano:
## ⚠️ MEDIUM Issue

The log file is opened with O_APPEND|O_CREATE but never rotated or truncated. On a long-running daemon, heimdallm.log will grow without bound. For a desktop app this is unlikely to be critical short-term, but in Docker deployments (the very scenario this PR targets) containers often run for weeks. Consider adding a size check on startup and truncating if above a threshold, or document a cron/logrotate expectation.

**Location:** `daemon/cmd/heimdallm/main.go` line 685

---

Muriano (approved):

Solid fix for #75. The approach of dual-writing to stderr + file via io.MultiWriter is correct, and the graceful fallback (stderr-only if file open fails) is good defensive coding. Two confirmed issues: the exported constant DaemonLogFileName in main.go is never referenced in handlers.go (which hardcodes the same string), creating drift risk; and the log file grows unbounded with no rotation mechanism, which will eventually exhaust disk on long-running deployments.

### Suggestions

- The /data directory detection via os.Stat("/data") is fragile — any host with a /data directory (e.g., data science workstations) will trigger Docker-mode logging. Consider using a more explicit signal like a HEIMDALLM_DOCKER=1 env var or checking for /.dockerenv instead.
- The opened log file handle (f) is never closed. While the OS reclaims it on process exit, explicitly deferring f.Close() (or storing f for a clean shutdown path) would be more robust and prevent potential buffered-write loss on abrupt termination.
- Tests in handlers_internal_test.go use hardcoded "heimdallm.log" strings — these should also reference the shared constant to stay in sync.

---

ivanmunozruiz:
Thanks @Muriano. Response to the review:

**Applied:**
- ✅ **DaemonLogFileName drift** — moved the constant from \`cmd/heimdallm\` (package main, unimportable) to \`internal/server\` so the writer and \`/logs\` reader share one symbol. Tests also updated to use it. Commit \`141bec0\`.
- ✅ **Unclosed log file handle** — \`setupLogging\` returns the \`*os.File\`; main defers \`Close()\` so buffered writes flush on shutdown. Returns \`nil\` in the stderr-only fallback path, defer is a no-op in that case.
- ✅ **Tests using hardcoded \"heimdallm.log\"** — all three cases now reference \`DaemonLogFileName\`.

**Not applied, with justification:**
- ❌ **\`os.Stat(\"/data\")\` Docker detection fragility** — fair concern, but the exact same heuristic lives in \`cmd/heimdallm dataDir()\`. Changing only \`daemonLogPath\` here would make the log path disagree with the DB path when both fire. The correct fix is a coordinated refactor to \`HEIMDALLM_DOCKER=1\` / \`/.dockerenv\` across both sites; it deserves its own PR so the reviewer can evaluate the container-detection heuristic in isolation. Happy to open a follow-up issue if you want.
- ❌ **Log rotation** — explicitly out of scope in the original PR body (also Muriano raised it). Will open a follow-up issue to track it before this PR merges.

ivanmunozruiz:
Rotation follow-up tracked in #77.

PR #78, opened 2026-04-20
feat(daemon): size-based rotation for heimdallm.log

### LINKED ISSUE #77: chore: rotate $dataDir/heimdallm.log to avoid unbounded growth
Follow-up to #75 / #76.

## Context

#76 made the daemon mirror its slog output into \`\$dataDir/heimdallm.log\` so the web UI's \`/logs\` view works under Docker. The file is opened \`O_APPEND|O_CREATE\` and never rotated — on a long-lived deployment it grows without bound. A chatty daemon hitting INFO per poll will fill disk eventually.

## Options

1. **Built-in rotator** — e.g. \`lumberjack.v2\` wrapping the \`io.Writer\` setupLogging feeds to slog. Pros: zero operator config, works identically on any host. Cons: new dep, one more moving piece inside the daemon.
2. **Document \`logrotate\`** — shipped, reliable, cross-distro. Cons: Docker users need to cook their own, doesn't help macOS.
3. **Hybrid** — built-in size cap (e.g. 50 MB, keep 3) because operator-configurable rotation is overkill for a single-file log. Simple copy-truncate inside setupLogging's goroutine.

Recommendation: **option 3** — keep it self-contained so both native and Docker deployments get sane defaults out of the box, no operator setup.

## Criterios de aceptación

- [ ] \`/data/heimdallm.log\` is capped at a documented size (default configurable via env var, e.g. \`HEIMDALLM_LOG_MAX_MB=50\`).
- [ ] After rotation, older files are named \`heimdallm.log.1\`, \`.2\`, …, up to a documented keep count (e.g. 3).
- [ ] The \`/logs\` endpoint continues to tail the current \`heimdallm.log\` without missing lines across a rotation.
- [ ] Unit test that simulates the rotation boundary.

ivanmunozruiz:
Closes #77.

## Summary

#76 wired the daemon's slog output into \`<dataDir>/heimdallm.log\` so the web UI's \`/logs\` view could see it under Docker. But the file was opened \`O_APPEND|O_CREATE\` and never rotated — a chatty INFO-per-poll daemon would eventually fill the data volume on any long-lived deployment.

This PR adds a lightweight in-process rotator. No new dependency (single-process log, a \`sync.Mutex\` is all we need), ~170 lines of code covered by 5 unit tests.

## How it works

\`\`\`
heimdallm.log       ← active, capped at HEIMDALLM_LOG_MAX_MB (50 MiB)
heimdallm.log.1     ← most recently rotated
heimdallm.log.2
…
heimdallm.log.<KEEP>
\`\`\`

When a write would push the active file past the cap:
1. Close active.
2. Delete \`heimdallm.log.<KEEP>\` if present.
3. Rename \`.<KEEP-1>\` → \`.<KEEP>\`, …, active → \`.1\` in reverse order so we never clobber a backup we still need.
4. Open a fresh active file.

Rename failures fall back to truncating the active file — losing history is preferable to blocking the daemon, and the event is logged to stderr.

## Config

Two new env vars (both optional, documented in \`docker/.env.example\`):

| Var | Default | Meaning |
|-----|---------|---------|
| \`HEIMDALLM_LOG_MAX_MB\` | \`50\` | Active-file cap in MiB before rotation fires. \`0\` disables rotation. |
| \`HEIMDALLM_LOG_KEEP\` | \`3\` | Rotated backups to retain. Worst-case disk use = \`(KEEP+1) × MAX_MB\`. |

## /logs tail loop

The SSE handler in \`handleLogsStream\` saves an offset between polls. After rotation the old active file is renamed and a fresh one replaces it at byte 0 — leaving the saved offset pointing past EOF. The fix: on every poll, if \`stat.Size() < savedOffset\`, reset offset to 0 so we stream the new content from the start instead of stalling.

## What changed

- \`daemon/internal/server/rotating_writer.go\` (new) — \`RotatingWriter\`, thread-safe \`io.Writer\` + \`io.Closer\`, with \`DefaultLogMaxBytes\` / \`DefaultLogKeep\` constants.
- \`daemon/cmd/heimdallm/main.go\` — \`setupLogging\` wraps the file in \`RotatingWriter\` and returns \`io.Closer\`. New \`logRotationConfig()\` reads the env vars with safe fallbacks on parse errors.
- \`daemon/internal/server/handlers.go\` — \`handleLogsStream\` rotation detection.
- \`daemon/internal/server/rotating_writer_test.go\` (new) — 5 cases covering the boundary transitions.
- \`docker/.env.example\` — documents the new env vars.

## Out of scope

- **Time-based rotation** (\`MaxAge\`). Can be a follow-up if operators need it. Size-based is sufficient for the stated goal of not filling disk.
- **Compression of backups**. Nice-to-have; skipped to keep scope tight.
- **\`os.Stat(\"/data\")\` Docker heuristic refactor** — still pending per Muriano's comment on #76. Independent of rotation.

## Test plan

- [x] \`make test-docker\` — all packages green, 5 new tests pass.
- [x] \`docker compose up -d --build heimdallm\`: \`/data/heimdallm.log\` is created with mode 0640, content mirrors \`docker logs heimdallm\`.
- [ ] Reviewer: set \`HEIMDALLM_LOG_MAX_MB=1\`, \`HEIMDALLM_LOG_KEEP=2\`, run \`make restart\`, wait for a couple of polls, confirm \`heimdallm.log.1\` and \`.2\` appear on disk and \`/logs\` keeps streaming across the rotation boundary.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -55,11 +56,11 @@ func main() {
 	// /logs stream reads that file; writing only to stderr (as we used
 	// to) left the stream empty under Docker — see #75.
 	logDir := dataDir()
-	logFile := setupLogging(logDir)
-	if logFile != nil {
+	logCloser := setupLogging(logDir)
+	if logCloser != nil {
 		// Flush buffered writes on shutdown so the last lines reach
 		// disk even when the daemon is killed mid-log.
-		defer logFile.Close()
+		defer logCloser.Close()
 	}
 
 	cfgPath := configPath()
@@ -668,14 +669,39 @@ func main() {
 	broker.Stop()
 }
 
+// logRotationConfig reads HEIMDALLM_LOG_MAX_MB and HEIMDALLM_LOG_KEEP from
+// the environment, falling back to the package defaults. Invalid values
+// fall back silently rather than refusing to start — logging is
+// non-critical enough that a typo in an env var shouldn't take the daemon
+// down.
+func logRotationConfig() (maxBytes int64, keep int) {
+	maxBytes = server.DefaultLogMaxBytes
+	keep = server.DefaultLogKeep
+	if v := os.Getenv("HEIMDALLM_LOG_MAX_MB"); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			maxBytes = int64(n) * 1024 * 1024
+		}
+	}
+	if v := os.Getenv("HEIMDALLM_LOG_KEEP"); v != "" {
+		if n, err := strconv.Atoi(v); err == nil && n > 0 {
+			keep = n
+		}
+	}
+	return
+}
+
 // setupLogging configures slog to write to stderr and, when possible, also
 // to <dataDir>/heimdallm.log — the file the web UI's /logs endpoint tails
-// (see #75). Returns the opened file handle so the caller can Close it on
-// shutdown; returns nil when we're running stderr-only (either dataDir is
-// empty or the file open failed). Either way the daemon never refuses to
-// start because logging to disk failed; `docker logs` / the host terminal
-// continue to work.
-func setupLogging(dataDir string) *os.File {
+// (see #75). Returns an io.Closer so the caller can flush on shutdown;
+// returns nil when we're running stderr-only (either dataDir is empty or
+// the file open failed). The daemon never refuses to start because
+// logging to disk failed; `docker logs` / the host terminal continue to
+// work.
+//
+// The log file is wrapped in a size-based rotator (see #77). MaxBytes
+// and Keep come from HEIMDALLM_LOG_MAX_MB / HEIMDALLM_LOG_KEEP with the
+// server package defaults.
+func setupLogging(dataDir string) io.Closer {
 	handlerOpts := &slog.HandlerOptions{Level: slog.LevelInfo}
 
 	if dataDir == "" {
@@ -683,14 +709,9 @@ func setupLogging(dataDir string) *os.File {
 		return nil
 	}
 
-	// O_APPEND so restarts extend the log rather than truncating it;
-	// 0640 so the web container (different UID, same data volume) could
-	// read via group membership, and the host user keeps full write
-	// access. The web container today reads this over HTTP via the
-	// daemon, not directly, but keeping the mode narrow is still the
-	// right posture.
 	logPath := filepath.Join(dataDir, server.DaemonLogFileName)
-	f, err := os.OpenFile(logPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0640)
+	maxBytes, keep := logRotationConfig()
+	w, err := server.NewRotatingWriter(logPath, maxBytes, keep)
 	if err != nil {
 		// Warn via a temporary logger that is visible on stderr even
 		// before SetDefault runs below.
@@ -701,8 +722,8 @@ func setupLogging(dataDir string) *os.File {
 		return nil
 	}
 
-	slog.SetDefault(slog.New(slog.NewTextHandler(io.MultiWriter(os.Stderr, f), handlerOpts)))
-	return f
+	slog.SetDefault(slog.New(slog.NewTextHandler(io.MultiWriter(os.Stderr, w), handlerOpts)))
+	return w
 }
 
 // dataDir resolves the data directory.
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -671,20 +671,27 @@ func main() {
 
 // logRotationConfig reads HEIMDALLM_LOG_MAX_MB and HEIMDALLM_LOG_KEEP from
 // the environment, falling back to the package defaults. Invalid values
-// fall back silently rather than refusing to start — logging is
-// non-critical enough that a typo in an env var shouldn't take the daemon
-// down.
+// fall back to the default *and* warn to stderr so operators notice typos
+// instead of silently losing the override they thought they had set.
+// Logging is non-critical enough that a bad env var should never take the
+// daemon down.
 func logRotationConfig() (maxBytes int64, keep int) {
 	maxBytes = server.DefaultLogMaxBytes
 	keep = server.DefaultLogKeep
 	if v := os.Getenv("HEIMDALLM_LOG_MAX_MB"); v != "" {
 		if n, err := strconv.Atoi(v); err == nil && n > 0 {
 			maxBytes = int64(n) * 1024 * 1024
+		} else {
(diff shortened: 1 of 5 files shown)


sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Well-structured size-based log rotation with good tests and documentation. One high-severity bug: rotateLocked can leave w.f nil on mid-rotation filesystem errors, causing a nil-pointer panic in Write.

### Issues

🔴 **daemon/internal/server/rotating_writer.go:97** — Write will panic with nil pointer dereference if rotateLocked fails after closing the file. After line 114 sets w.f = nil, early returns from os.Remove (line 121) or os.Rename failures in the backup-shift loop (line 131) leave w.f nil. Back in Write, line 97 (w.f.Write(p)) dereferences nil. The rename-to-.1 fallback (line 142) correctly reopens the file, but the other error paths do not. Fix: reopen the file in append mode before returning on those error paths, or add a nil guard in Write that reopens the file.
🔴 **daemon/internal/server/rotating_writer.go:155** — If opening the fresh active file fails (line 155), rotateLocked returns an error with w.f still nil. The old active file was already renamed to .1, so it is gone. Write then panics on w.f.Write(p). Unlike the rename-to-.1 failure, there is no fallback here. Consider attempting to reopen or recreate the file, or rename .1 back as a recovery step.
🟡 **daemon/internal/server/handlers.go:779** — The rotation detection (stat.Size() < offset) also triggers on manual truncation, not just rotation. This is probably fine for log tailing, but worth a comment noting the intentional conflation. More importantly, there is a TOCTOU gap between the Stat and the Seek — a second rotation could fire in between — though this is inherent to the polling design and unlikely in practice.
⚠️ **daemon/internal/server/rotating_writer.go:90** — No nil check on w.f in Write. After Close() is called (setting w.f to nil), a concurrent or late Write call will panic. The doc says 'intended for shutdown only', but a defensive nil check returning an error (e.g., os.ErrClosed) would be safer than a panic, especially since slog could still be flushing.

### Suggestions

- Add a recovery path in rotateLocked that re-opens the file in append mode (or truncate mode as last resort) before returning on any error after w.f has been closed and set to nil. The rename-active-to-.1 fallback already demonstrates this pattern — apply it consistently to the remove-oldest and shift-backup error paths as well.
- Add a nil guard at the top of Write (if w.f == nil, return 0, os.ErrClosed) to convert a panic into a recoverable error for any path that leaves the file handle nil.
- Consider adding a test that simulates a rotation failure (e.g., read-only directory for the backup path) to verify the writer degrades gracefully rather than panicking.

---

Muriano:
## 🔴 HIGH Issue

Nil-pointer panic: rotateLocked() sets w.f = nil after closing (line 118), then if os.Remove at line 124 or os.Rename in the shift loop (lines 131-139) fails, it returns an error with w.f still nil. Write() only logs the rotation error and proceeds to w.f.Write(p), which panics. Fix: add a recovery path on these early returns that reopens or truncates the active file (similar to the fallback at lines 146-153), or restructure so w.f is never nil when rotateLocked returns.

**Location:** `daemon/internal/server/rotating_writer.go` line 128

---

Muriano (changes requested):

Well-designed size-based log rotator with good tests and documentation. One confirmed crash bug: early-return paths in rotateLocked() leave w.f nil, causing a panic in Write() on the next w.f.Write(p) call. The rename-to-.1 fallback correctly reopens the file, but the Remove and shift-loop failures do not.

### Suggestions

- Add a concurrency test (multiple goroutines writing simultaneously) to validate the mutex protection under race detector.
- Consider logging a warning when HEIMDALLM_LOG_MAX_MB or HEIMDALLM_LOG_KEEP contain invalid values, so operators notice typos instead of silently getting defaults.
- In handleLogsStream rotation detection (handlers.go), the offset reset to 0 is a good heuristic but could briefly re-emit lines already sent if the new file already has content. Consider tracking an inode or open-time marker for more precise detection — low priority since the /logs endpoint is best-effort.

---
*Severity: **HIGH** · Reviewed by Heimdallr*

ivanmunozruiz:
Thanks @sergiotejon @Muriano — both caught the same HIGH bug, and they're right.

**Applied (HIGH fix):**
- ✅ \`rotateLocked\` no longer leaves \`w.f == nil\` on the three error paths sergio and muriano called out (Remove, shift-Rename, fresh OpenFile). Each now calls a new \`recoverActive()\` helper that tries \`O_APPEND\` first (so the shift-failure path keeps any content at \`w.path\` that was never moved) and falls back to \`O_TRUNC|O_CREATE\`.
- ✅ \`Write\` has a nil-guard returning \`os.ErrClosed\` as a safety net for the remaining \"even recovery failed\" edge case — e.g. the data directory being read-only.
- ✅ Documented the post-rotate invariant at the top of \`rotateLocked\`.
- ✅ New \`TestRotatingWriter_RotationRecoversFromRenameFailure\` chmods the parent dir 0500 to force \`EACCES\` on rename and asserts subsequent writes do not panic.
- ✅ New \`TestRotatingWriter_ClosedWriterReturnsErrNotPanic\` covers the \"write after Close()\" case.
- ✅ New \`TestRotatingWriter_ConcurrentWritesStaySafe\` — 20 goroutines × 50 writes with a tight cap that forces many rotations, exercising the mutex.

**Applied (muriano LOW):**
- ✅ Invalid \`HEIMDALLM_LOG_MAX_MB\` / \`HEIMDALLM_LOG_KEEP\` now warn to stderr instead of silently defaulting.

**Not applied:**
- ❌ **Inode tracking for \`/logs\` rotation detection** (muriano 🟡) — size-based reset is deliberately a heuristic. At worst we briefly re-emit a few lines after an extremely fast rotation burst, which is harmless for a log tail. Switching to inode tracking would complicate the handler for a case I can't reproduce. Can revisit if it ever causes real issues.

Pushed to \`4633990\`. Re-requesting review.

Muriano (approved):
## 🤖 Heimdallr AI Review

Well-implemented size-based log rotation with thorough error recovery, good test coverage, and sensible defaults. The RotatingWriter handles edge cases (rename failures, concurrent writes, closed handles) defensively. The log-stream rotation detection in handlers.go is a pragmatic heuristic. No blocking issues found.

### Suggestions

- handlers.go:779 — The rotation detection heuristic `stat.Size() < offset` has a narrow TOCTOU window: if the daemon rotates the file and immediately writes enough to exceed the old offset before the next poll, the stream handler won't detect the rotation and will continue from a stale offset, potentially missing lines from the new file. This is unlikely in practice but could be made more robust by also comparing the file's inode (via os.SameFile or Stat().Sys()) against the previous iteration's.
- rotating_writer.go:88 — Consider calling w.f.Sync() before Close in rotateLocked() so the last writes are guaranteed on disk before the rename shuffle begins. Currently the OS page cache is trusted, which is fine for most filesystems but a Sync would make the durability guarantee explicit.
- rotating_writer_test.go — A test verifying that data written by multiple goroutines (TestRotatingWriter_ConcurrentWritesStaySafe) is not silently lost could strengthen the suite — e.g., count total bytes across all surviving files (active + backups) and compare against expected totals, allowing for evicted backups.
- cmd/heimdallm/main.go:696 — logRotationConfig accepts arbitrarily large values for HEIMDALLM_LOG_MAX_MB (e.g., 999999999 → ~1 PB). An upper-bound sanity check (say, 10 GiB) with a stderr warning would help operators catch typos without silently accepting unreasonable values.

---

PR #80, opened 2026-04-20
fix(daemon): persist PUT /config — store > env > TOML precedence

### LINKED ISSUE #79: fix(config): PUT /config writes are dead — store values never applied on reload
## Summary

The web UI's `Save & reload` on `/config` returns HTTP 200 but nothing persists. Values vanish on refresh.

## Root cause

`PUT /config` writes to SQLite via `store.SetConfig` (`daemon/internal/server/handlers.go:397`), but no consumer reads those rows:

1. `store.GetConfig` is defined (`daemon/internal/store/store.go:144`) but never called in production code.
2. `srv.SetReloadFn` (`daemon/cmd/heimdallm/main.go:462`) only calls `config.Load(cfgPath)` → TOML + env vars. The `configs` table is never merged.
3. `writeConfigTOML` runs once in `LoadOrCreate` at bootstrap (`config.go:553`); never re-written after.

Result: `PUT /config` is a dead write for every key, not just `issue_tracking`. The user-visible symptom today is that `issue_tracking` fields (develop_labels, review_only_labels, skip_labels, organizations, assignees, filter_mode, default_action) silently revert on refresh.

Additional smaller bug: `issue_tracking` is missing from the `validConfigKeys` allowlist in `handlePutConfig`, so the request actually returns 400. The UI shows the error at the top of the form (easy to miss when editing labels at the bottom).

## Fix (agreed approach)

**Precedence: store > env > TOML** — UI saves always win, env vars become bootstrap defaults, TOML is the base layer.

1. New `(c *Config) ApplyStore(s Storer)` in `config.go`: scans `configs` table, unmarshals into matching fields by key.
2. Call `ApplyStore` after `config.Load` both at bootstrap and in `reloadFn`.
3. Add `issue_tracking` to `validConfigKeys` + validate nested body before persisting (decode into `IssueTrackingConfig`, run `Validate()`).
4. Add `issue_tracking` to `configFn` map in `main.go` so `GET /config` returns it.
5. Add `json:\"...\"` tags alongside existing `toml:\"...\"` tags on `IssueTrackingConfig` (and any other Config struct we serialize to the wire) so the snake_case keys the web UI sends round-trip correctly.

## Tests (TDD)

- `TestApplyStore_MergesOverTOML`
- `TestApplyStore_StoreWinsOverEnv`
- `TestApplyStore_InvalidJSON_ReturnsError`
- `TestHandlerPutConfig_IssueTracking_RoundTrip`
- `TestHandlerPutConfig_IssueTracking_InvalidFilterMode`
- `TestHandlerPutConfig_AfterReload_ValuesPersist`

Run with `make test-docker`.

## Out of scope

- Preserving TOML comments/formatting (we stop writing TOML after bootstrap).
- Endpoint to reset a key back to env/TOML (delete row manually for now).
- Migration of existing `configs` rows (assume empty in current deployments).

ivanmunozruiz:
## Summary

- `PUT /config` now actually persists. `ApplyStore` merges the SQLite `configs` table on top of TOML + env vars at both bootstrap and `POST /reload`, so web-UI saves survive refresh and daemon restart.
- `issue_tracking` added to the PUT allowlist with typed validation (rejects bad `filter_mode` / `default_action` before they reach the store).
- `IssueTrackingConfig` dual-tagged with `json:"..."` so the snake_case keys the web UI sends round-trip cleanly.
- `GET /config` now returns `issue_tracking`, so the form prefills with whatever was last saved.

## Why

Before this change, `PUT /config` wrote to `configs` but nothing read those rows — the reload path only re-read TOML + env. Saves returned 200 but silently vanished on refresh. The `issue_tracking` allowlist gap compounded the problem: the body was rejected with HTTP 400 shown at the top of the form, where operators editing labels at the bottom never saw it.

Precedence is now **TOML < env < store**. The UI is the primary surface for operator edits, so it wins; env vars remain the bootstrap default. If someone wants an env var to re-take control, delete the matching row from `configs` (a small follow-up can add a reset endpoint — intentionally out of scope here).

## Test plan

- [x] `make test-docker` — all packages green (vet + tests).
- [x] New tests:
  - `config.TestApplyStore_*` (4 cases: merge, store-wins-over-env, bad JSON, unknown key)
  - `store.TestConfigs_ListReturnsAllRows` + empty-table case
  - `server.TestHandlerPutConfig_IssueTracking_*` (4 cases: accepted, invalid filter_mode, invalid default_action, end-to-end round-trip)
- [ ] Manual smoke in Docker: `make up` → open `/config` → edit `issue_tracking` labels → Save & reload → refresh page → values persist. (Recommended reviewer check.)

## Out of scope

- Preserving TOML comments/formatting on rewrite (we never rewrite TOML after bootstrap).
- Endpoint to reset a key back to env/TOML (manual DELETE from `configs` works for now).
- Migration of existing `configs` rows (none expected in live deployments — the endpoint was dead-write until this PR).

Closes #79

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -90,6 +90,18 @@ func main() {
 	}
 	defer s.Close()
 
+	// Merge PUT /config values on top of TOML+env. This is the third and
+	// highest-precedence layer: UI saves win over env vars, env vars win
+	// over TOML. See daemon/internal/config/store.go for the key mapping.
+	if rows, lcErr := s.ListConfigs(); lcErr != nil {
+		slog.Warn("config: list store rows failed, continuing with TOML+env", "err", lcErr)
+	} else if asErr := cfg.ApplyStore(rows); asErr != nil {
+		slog.Warn("config: apply store failed, continuing with TOML+env", "err", asErr)
+	} else if vErr := cfg.Validate(); vErr != nil {
+		slog.Error("config invalid after applying store, refusing to start", "err", vErr)
+		os.Exit(1)
+	}
+
 	if err := s.PurgeOldReviews(cfg.Retention.MaxDays); err != nil {
 		slog.Warn("retention purge failed", "err", err)
 	}
@@ -430,6 +442,7 @@ func main() {
 			"ai_fallback":    c.AI.Fallback,
 			"review_mode":    c.AI.ReviewMode,
 			"retention_days": c.Retention.MaxDays,
+			"issue_tracking": c.GitHub.IssueTracking,
 			"repo_overrides": repoOverrides,
 			"agent_configs":  agentConfigs,
 		}
@@ -463,6 +476,13 @@ func main() {
 		newCfg, err := config.Load(cfgPath)
 		if err != nil {
 			return fmt.Errorf("reload: %w", err)
+		}
+		if rows, lcErr := s.ListConfigs(); lcErr != nil {
+			slog.Warn("config: list store rows on reload failed", "err", lcErr)
+		} else if asErr := newCfg.ApplyStore(rows); asErr != nil {
+			return fmt.Errorf("reload: apply store: %w", asErr)
+		} else if vErr := newCfg.Validate(); vErr != nil {
+			return fmt.Errorf("reload: validate after store: %w", vErr)
 		}
 
 		cfgMu.Lock()
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -101,36 +101,36 @@ const (
 // highest-precedence match. This is a fail-safe ordering: when in doubt,
 // review before developing, and skip before either.
 type IssueTrackingConfig struct {
-	Enabled bool `toml:"enabled"`
+	Enabled bool `toml:"enabled" json:"enabled"`
 
 	// FilterMode decides how the org / assignee / label dimensions are
 	// combined ("exclusive" = AND, "inclusive" = OR). Applied by the
 	// pipeline; not consulted by Classify itself.
-	FilterMode FilterMode `toml:"filter_mode"`
+	FilterMode FilterMode `toml:"filter_mode" json:"filter_mode"`
 
 	// Organizations limits processing to issues belonging to these orgs.
 	// Empty = no org filter.
-	Organizations []string `toml:"organizations"`
+	Organizations []string `toml:"organizations" json:"organizations"`
 
 	// Assignees limits processing to issues assigned to these GitHub users.
 	// Empty = no assignee filter.
-	Assignees []string `toml:"assignees"`
+	Assignees []string `toml:"assignees" json:"assignees"`
 
 	// DevelopLabels are labels that mark an issue as "please implement".
-	DevelopLabels []string `toml:"develop_labels"`
+	DevelopLabels []string `toml:"develop_labels" json:"develop_labels"`
 
 	// ReviewOnlyLabels are labels that mark an issue as "please analyse and
 	// comment only". Takes precedence over DevelopLabels when both are
 	// present on the same issue — fail-safe default.
-	ReviewOnlyLabels []string `toml:"review_only_labels"`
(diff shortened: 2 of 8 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured PR that adds a third config precedence layer (store > env > TOML) with good test coverage and input validation. The main concern is that ApplyStore mutates the Config in-place key-by-key, so a mid-iteration JSON parse error leaves the config in a partially-applied state that the caller's warning message ('continuing with TOML+env') does not accurately describe.

### Issues

⚠️ **daemon/internal/config/store.go:18** — ApplyStore mutates the Config receiver in-place as it iterates. If an error occurs partway through (e.g. bad JSON on the third key), the first two keys are already applied. The caller in main.go logs 'continuing with TOML+env' but the config is actually in a hybrid TOML+env+partial-store state. Consider a two-pass approach (validate all values first, then apply) or apply to a shallow copy and swap on success.
⚠️ **daemon/cmd/heimdallm/main.go:96** — On reload, if ListConfigs fails (e.g. transient DB issue), the code warns but proceeds with newCfg that has no store overrides, silently reverting all user UI customizations back to TOML+env. Since this is a reload (not cold start), consider returning an error to keep the old config intact rather than silently losing store values.
🟡 **daemon/internal/config/store.go:42** — server_port is handled in ApplyStore but is absent from validConfigKeys in handlers.go, so it cannot be set via PUT /config. If a row is manually inserted into the DB, it would be applied on reload and could change the listening port mid-flight. Either add it to validConfigKeys or document why it is excluded.
🟡 **daemon/internal/config/config.go:454** — ValidateIssueTracking constructs an empty Config with zero-value fields, then calls validateIssueTracking. If validateIssueTracking is ever extended to reference other Config fields (e.g. checking IssueTracking against GitHub.Repositories), this will silently validate against zero values. A comment noting this coupling constraint would help prevent future regressions.
⚠️ **daemon/cmd/heimdallm/main.go:93** — At startup, ApplyStore failure is treated as a warning (continue without store), but Validate failure causes os.Exit(1). If partial application (see first issue) corrupted a field, Validate might pass on the half-applied state when it should not, or fail and kill a daemon that would have been fine with pure TOML+env. The Validate call should run against a known-good state.

### Suggestions

- Refactor ApplyStore to use a two-pass strategy: first decode and validate all values into local variables, then assign them to the Config only if every key succeeds. This eliminates the partial-mutation problem at startup and reload.
- On reload, treat ListConfigs failure as a hard error (return fmt.Errorf) rather than a warning, so the daemon keeps its previous config—including store overrides—instead of silently reverting to TOML+env.
- Add an integration test that simulates a partial ApplyStore failure (e.g. one valid key and one malformed JSON key) and asserts that the Config is unchanged afterward.
- Consider adding server_port to validConfigKeys or removing its case from ApplyStore to keep the two maps in sync and avoid dead code.

---

Muriano (approved):
## 🤖 Heimdallr AI Review

Clean, well-structured fix that adds store-based config persistence with correct TOML < env < store precedence. Good test coverage including round-trip, precedence, and invalid-input scenarios. No confirmed bugs or security issues.

### Suggestions

- main.go:93-101 — The else-if chain means a ListConfigs failure silently skips Validate on the store-merged config. Consider splitting into independent steps (list, apply, validate) so validation always runs on the final config state, even if the store layer is empty or unavailable.
- config/store.go — ApplyStore accepts bare strings for poll_interval, ai_primary, ai_fallback, and review_mode without any validation (e.g. malformed duration, empty string). The handler validates some of these before persisting, but a manually-edited DB or future writer could store invalid values. Consider lightweight validation inside ApplyStore or at least a comment documenting the trust boundary.
- handlers.go:382-404 — The json.Marshal → json.Unmarshal round-trip on issue_tracking works but is indirect. Since the body is already parsed as map[string]interface{}, you could use json.Marshal once and Unmarshal directly into IssueTrackingConfig, removing one step.
- config/config.go:454-459 — ValidateIssueTracking constructs a zero-valued Config to call validateIssueTracking(). If that method ever grows to inspect other Config fields, validation will silently pass with zero values. A brief comment noting this coupling would help future maintainers.

---

PR #82, opened 2026-04-20
fix(daemon): harden store-layer merge — atomic, strict reload, drop server_port

### LINKED ISSUE #81: fix(daemon): harden ApplyStore — two-pass merge, strict reload, allowlist sync, validation coupling
Follow-up to #79 / #80. Both AI reviewers (`sergiotejon`, `Muriano`) approved the merge but flagged four legitimate concerns that were left out of the initial fix to keep scope tight. Tracking them here as a single hardening pass.

## 1. `ApplyStore` mutates Config in-place during iteration

**Where:** `daemon/internal/config/store.go:17-57`

`ApplyStore` writes each decoded key directly onto the Config receiver as it walks the rows map. If key N has malformed JSON we return an error, but keys 1..N-1 are already applied. The caller in `main.go` logs `"continuing with TOML+env"` — that message is now a lie, the Config is in a hybrid TOML+env+partial-store state.

**Fix:** two-pass merge. First pass decodes every known key into local variables and surfaces any JSON errors. Second pass assigns to the Config only if the first pass succeeded. Side effect: atomic semantics — either every store row applies or none do.

**Test to add:** `TestApplyStore_PartialFailure_LeavesCfgUnchanged` — rows with one valid key + one malformed key, assert the valid key did NOT land and the original cfg values are intact.

## 2. Reload treats `ListConfigs` failure as a warning → silent revert

**Where:** `daemon/cmd/heimdallm/main.go` inside `srv.SetReloadFn`

A transient SQLite error on `s.ListConfigs()` currently `slog.Warn`s and keeps going with `newCfg` as TOML+env only. On a **reload** (not cold start) that silently reverts every operator customisation made via the web UI back to bootstrap state. The user sees no error; labels, repositories etc. vanish.

**Fix:** on reload, a `ListConfigs` error should be a hard error (`return fmt.Errorf("reload: list store: %w", err)`). The reload path is atomic — either we successfully build a new config that includes store overrides, or we keep serving the previous in-memory cfg.

Bootstrap behaviour (warn + continue with TOML+env) is fine to keep as-is: at cold start there is no previous cfg to preserve.

## 3. `server_port` in `ApplyStore` switch but absent from `validConfigKeys`

**Where:**
- `daemon/internal/config/store.go:42-47` handles `case "server_port":`
- `daemon/internal/server/handlers.go:305-314` `validConfigKeys` does NOT include `server_port`.

Today this is dead code through the PUT path — the allowlist rejects a `server_port` body before it ever reaches `SetConfig`. But a manually-inserted row in the `configs` table (debugging, future code path, migration script) would be picked up on reload and could swap the listening port mid-flight.

**Fix:** pick one —
- (a) add `"server_port": {}` to `validConfigKeys` with the numeric range check that already exists at `handlers.go:362-368`, making it a first-class configurable, **or**
- (b) remove the `server_port` case from `ApplyStore` and document why (port is bootstrap-only for safety).

(b) is simpler and matches the fact that the web UI has no input for it.

## 4. `ValidateIssueTracking` uses a zero-valued `Config`

**Where:** `daemon/internal/config/config.go:452-459`

```go
func ValidateIssueTracking(it IssueTrackingConfig) error {
    c := &Config{}
    c.GitHub.IssueTracking = it
    return c.validateIssueTracking()
}
```

Today `validateIssueTracking` only reads `c.GitHub.IssueTracking`, so the zero-value wrapper is fine. But if the method ever grows to cross-check other Config fields (e.g. validating assignees against `GitHub.Repositories`), this helper will silently pass because every other field is zero.

**Fix:** add a one-line comment above `ValidateIssueTracking` noting that `validateIssueTracking` must stay self-contained (only reads `c.GitHub.IssueTracking`) for this helper to remain correct. Future maintainer sees it, knows the coupling, thinks twice.

---

## Scope

All four changes are internal-only, no new endpoints, no wire-format changes. A single PR should be enough. Same test discipline as #80 (`make test-docker`).

ivanmunozruiz:
## Summary

Follow-up to #80 addressing the four review points raised by `sergiotejon` (MEDIUM) and `Muriano` (LOW). All internal-only, no wire-format changes.

1. **Atomic `ApplyStore`** — shadow-copy pattern. A malformed row no longer leaves earlier keys half-applied. The bootstrap warning "continuing with TOML+env" is finally truthful when an error fires.
2. **Strict reload** — new `Config.MergeStoreLayer(StoreLister)` helper; the reload path hard-errors on `ListConfigs`/`ApplyStore`/`Validate` failure so the running daemon keeps its previous cfg instead of silently reverting operator customisations. Bootstrap intentionally stays soft.
3. **`server_port` removed from `ApplyStore`** — was dead through the PUT path (absent from `validConfigKeys`) but a manually-inserted row could swap the listening port mid-flight. Bootstrap-only is safer and the web UI has no surface for it anyway.
4. **Coupling comment on `ValidateIssueTracking`** — documents the zero-valued `Config` wrapper so a future extension that cross-checks other fields doesn't silently pass on zeros.

## Test plan

- [x] `make test-docker` — all packages green (`go vet` + tests).
- [x] New tests:
  - `TestApplyStore_PartialFailure_LeavesCfgUnchanged` — atomicity contract
  - `TestApplyStore_ServerPort_IsIgnored` — bootstrap-only guarantee
  - `TestMergeStoreLayer_Success`
  - `TestMergeStoreLayer_ListConfigsFailure_ReturnsError` — fake `StoreLister` returning a DB error is surfaced
  - `TestMergeStoreLayer_InvalidStoreValue_ReturnsError`
  - `TestMergeStoreLayer_FailsValidationOnBadMergedCfg` — merged cfg that fails `Validate` is rejected
- [ ] Manual smoke in Docker: `make up`, corrupt a row in `heimdallm.db` (`UPDATE configs SET value = 'garbage' WHERE key = 'repositories'`), `POST /reload` → expect 500, daemon keeps serving old cfg. (Recommended reviewer check.)

Closes #81

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -93,13 +93,12 @@ func main() {
 	// Merge PUT /config values on top of TOML+env. This is the third and
 	// highest-precedence layer: UI saves win over env vars, env vars win
 	// over TOML. See daemon/internal/config/store.go for the key mapping.
-	if rows, lcErr := s.ListConfigs(); lcErr != nil {
-		slog.Warn("config: list store rows failed, continuing with TOML+env", "err", lcErr)
-	} else if asErr := cfg.ApplyStore(rows); asErr != nil {
-		slog.Warn("config: apply store failed, continuing with TOML+env", "err", asErr)
-	} else if vErr := cfg.Validate(); vErr != nil {
-		slog.Error("config invalid after applying store, refusing to start", "err", vErr)
-		os.Exit(1)
+	//
+	// Bootstrap treats any failure here as a warning: with no previous
+	// in-memory cfg to fall back to, rejecting a startup over a corrupted
+	// configs row would lock the operator out. Reload is stricter (below).
+	if err := cfg.MergeStoreLayer(s); err != nil {
+		slog.Warn("config: store layer not applied, continuing with TOML+env", "err", err)
 	}
 
 	if err := s.PurgeOldReviews(cfg.Retention.MaxDays); err != nil {
@@ -477,12 +476,12 @@ func main() {
 		if err != nil {
 			return fmt.Errorf("reload: %w", err)
 		}
-		if rows, lcErr := s.ListConfigs(); lcErr != nil {
-			slog.Warn("config: list store rows on reload failed", "err", lcErr)
-		} else if asErr := newCfg.ApplyStore(rows); asErr != nil {
-			return fmt.Errorf("reload: apply store: %w", asErr)
-		} else if vErr := newCfg.Validate(); vErr != nil {
-			return fmt.Errorf("reload: validate after store: %w", vErr)
+		// On reload we have a working cfg already — a transient DB error or
+		// a corrupted row must NOT silently revert the running daemon to
+		// TOML+env and wipe operator customisations. Propagate the error;
+		// handleReload returns 500 and the in-memory cfg is untouched.
+		if err := newCfg.MergeStoreLayer(s); err != nil {
+			return fmt.Errorf("reload: %w", err)
 		}
 
 		cfgMu.Lock()
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -453,6 +453,13 @@ func (c *Config) Validate() error {
 // Used by the PUT /config handler to pre-check a standalone IssueTrackingConfig
 // without having to assemble a full Config (which would trip over other
 // required fields like ai.primary).
+//
+// COUPLING: this helper wraps the struct in a zero-valued Config. It stays
+// correct only as long as validateIssueTracking reads exclusively from
+// c.GitHub.IssueTracking. If you ever extend it to cross-check other Config
+// fields (e.g. assignees against GitHub.Repositories), this wrapper must
+// take the extra fields as parameters too or future validations will pass
+// silently against zero values.
 func ValidateIssueTracking(it IssueTrackingConfig) error {
 	c := &Config{}
 	c.GitHub.IssueTracking = it
--- a/daemon/internal/config/store.go
+++ b/daemon/internal/config/store.go
@@ -51,9 +51,14 @@ func (c *Config) MergeStoreLayer(s StoreLister) error {
 // single malformed row therefore leaves the receiver untouched, so the
 // caller's error-path ("continuing with TOML+env") is truthful.
 //
-// `server_port` is intentionally NOT supported: mutating the listening port
-// at runtime would invalidate every in-flight connection and the web UI has
-// no surface for it. Bootstrap-only.
+// INVARIANT — shallow copy + wholesale replacement: `shadow := *c` is a
+// shallow copy, so `shadow.AI.Agents` and `shadow.AI.Repos` (both maps)
+// still share backing storage with the receiver. Today every case below
+// *replaces the whole field* (slice/struct/string assignment) rather than
+// mutating it in place, so the atomicity guarantee holds. If you ever add
+// a case that writes into an existing map (e.g. `shadow.AI.Agents[k] = v`)
+// you MUST deep-copy that map into the shadow first, or the mutation will
+// leak through to the receiver even when a later row fails.
 func (c *Config) ApplyStore(rows map[string]string) error {
 	shadow := *c
 	for key, raw := range rows {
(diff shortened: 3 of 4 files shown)


Muriano:
## ⚠️ MEDIUM Issue

Shallow copy (`shadow := *c`) shares underlying map references with the receiver — specifically `AI.Agents` (map[string]CLIAgentConfig) and `AI.Repos` (map[string]RepoAI). No current switch case mutates these maps so atomicity holds today, but the doc comment claims 'the merge happens on a shadow copy' without noting the shallow-copy caveat. A future contributor adding a map-mutating case (e.g. `shadow.AI.Agents[k] = v`) would silently break the atomicity guarantee. Recommend either (a) adding a comment noting the shallow-copy limitation or (b) deep-copying the two map fields in the shadow initialization.

**Location:** `daemon/internal/config/store.go` line 48

---

Muriano (approved):

Clean hardening PR. Atomic shadow-copy pattern in ApplyStore is correct for all currently-handled keys. MergeStoreLayer properly DRYs bootstrap (lenient) vs reload (strict) semantics. Tests cover key scenarios including partial-failure atomicity. Two minor suggestions below; nothing blocks merge.

### Suggestions

- server_port now falls through to the `default` branch and logs 'unknown store key'. Since it's a deliberately unsupported key (not unknown), consider adding an explicit `case "server_port":` with a more descriptive log like `slog.Warn("config: server_port is bootstrap-only, ignoring store override", "key", key)` — this gives operators a clear signal instead of the generic 'unknown key' message.
- TestApplyStore_PartialFailure_LeavesCfgUnchanged depends on map iteration visiting `poll_interval` before `repositories` to exercise the 'valid row rolled back' path. In practice Go map iteration is random so the test is still sound (both orderings test useful things), but a brief comment noting this would clarify intent for future readers.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured refactor that consolidates the store-merge flow into a single MergeStoreLayer method, adds atomicity via shadow copy, and improves bootstrap-vs-reload error semantics. The changes are well-tested and the design rationale is clearly documented. A few minor issues worth addressing.

### Issues

⚠️ **daemon/internal/config/store.go:48** — Shallow copy (`shadow := *c`) shares underlying slice/map backing arrays with the receiver. While no current switch case mutates shared maps in-place (all cases reassign), the doc comment claims full shadow-copy atomicity. A future contributor adding e.g. `shadow.AI.Agents[k] = v` would silently corrupt the original on error paths. Agree with the existing reviewer: at minimum add a comment documenting the shallow-copy limitation, or deep-copy the map fields.
🟡 **daemon/internal/config/store.go:27** — Double 'config:' prefix in error messages. MergeStoreLayer wraps with `config: apply store:` but ApplyStore already returns errors prefixed `config: apply store key %q:`. This produces messages like `config: apply store: config: apply store key "repositories": ...`. Either drop the prefix from MergeStoreLayer or from ApplyStore to avoid redundancy.
🟡 **daemon/internal/config/store.go:48** — Go map iteration is non-deterministic: if the `rows` map contains multiple malformed values, which error surfaces first is random. This makes debugging harder if operators report different errors for the same corrupted state. Consider sorting keys before iterating, or collecting all errors.
⚠️ **daemon/internal/config/store_test.go:165** — TestApplyStore_PartialFailure_LeavesCfgUnchanged relies on map iteration hitting `repositories` (the bad key) after `poll_interval` (the good key), but Go map iteration order is non-deterministic. If `repositories` is visited first, the test never exercises the atomicity guarantee (the good key was never applied to the shadow). Use a deterministic ordering or add enough valid keys to make accidental all-first-error ordering statistically negligible.

### Suggestions

- Add a brief comment at `shadow := *c` noting the shallow-copy caveat: map/slice fields share backing storage, so future cases must reassign (not mutate in-place) to preserve atomicity.
- Normalize the error prefixes: have ApplyStore return un-prefixed errors (e.g. `apply store key %q: %w`) and let MergeStoreLayer add the single `config:` prefix, or vice versa.
- For the partial-failure atomicity test, consider using a helper that iterates keys in a fixed order (e.g. sorted), or restructure the test to assert atomicity regardless of iteration order by including multiple valid keys that would all need to revert.
- The removal of `server_port` is a behavior change for operators who manually inserted it into the configs table. Consider logging at Warn level with a specific migration hint (e.g. 'server_port is now bootstrap-only; set it in TOML or env') rather than the generic 'unknown store key' message.

---

PR #87, opened 2026-04-20
fix(daemon): accept read-only fields in PUT /config so web UI saves don't 400

### LINKED ISSUE #86: fix(daemon): PUT /config rejects fields GET /config returns — save from web UI fails on every refresh
## Symptom

After saving config in the web UI (or hitting refresh without editing anything), the page shows

> `unknown config key: "agent_configs"`

## Root cause

`web_ui/src/routes/config/+page.svelte:109` composes the PUT body as `{ ...raw, …overrides }`, deliberately round-tripping every field from `GET /config` so a future field the daemon starts persisting isn't clobbered by an old UI. But `GET /config` returns three fields that `PUT /config`'s `validConfigKeys` allowlist (`daemon/internal/server/handlers.go:305`) does not accept:

| Key | Source | Writable? |
|---|---|---|
| `non_monitored` | `c.GitHub.NonMonitored` (list of repos excluded from polling, Flutter-UI managed) | ❌ rejected with 400 |
| `repo_overrides` | `[ai.repos.<name>]` block (per-repo AI config) | ❌ rejected with 400 |
| `agent_configs` | `[ai.agents.<name>]` block (per-CLI agent config) | ❌ rejected with 400 |

Which one triggers the 400 is a lottery — Go `for k := range body` iterates randomly — so you might see `non_monitored` one refresh and `agent_configs` the next.

The earlier `issue_tracking` bug (fixed in #80) was masking this: the handler returned 400 on `issue_tracking` first, so these three never got a chance to trip. Now that `issue_tracking` is allowlisted, the next unknown key wins the race.

## Bonus

`server_port` is IN `validConfigKeys` (writable) but `ApplyStore` explicitly ignores it as bootstrap-only (documented in #81 follow-up). So every UI save today silently writes a redundant `server_port` row into the `configs` table that the reload immediately ignores. Not a bug, but cruft — we should move it to the read-only group too.

## Fix

Introduce a second allowlist next to `validConfigKeys`:

```go
// readOnlyConfigKeys are returned by GET /config for client round-tripping
// but are NOT persisted by PUT. The handler accepts them (no 400) and skips
// the SetConfig call.
var readOnlyConfigKeys = map[string]struct{}{
    "non_monitored":  {},
    "repo_overrides": {},
    "agent_configs":  {},
    "server_port":    {}, // bootstrap-only, see ApplyStore
}
```

- `handlePutConfig` pre-check accepts keys in EITHER map; 400s only for truly unknown keys.
- The write loop skips keys that live in `readOnlyConfigKeys`.
- `server_port` is moved out of `validConfigKeys`; its existing numeric-range validation follows it (keeps "range check still runs when present" behaviour).

Preserves the security goal of the original `fix(security): validate config values in PUT /config` (commit 406a4b0) — still an explicit allowlist, arbitrary keys still rejected — while respecting the Svelte page's "round-trip unknown fields for forward compat" intent.

## Tests (TDD)

- `TestHandlerPutConfig_ReadOnlyKeys_Accepted` — body with `non_monitored`, `repo_overrides`, `agent_configs` (and `server_port`) → 200.
- `TestHandlerPutConfig_ReadOnlyKeys_NotPersisted` — after that PUT, `ListConfigs` does NOT contain those keys.
- `TestHandlerPutConfig_WritableKeys_StillPersisted` — mixed body with read-only + writable keys → only writable lands in the store.
- `TestHandlerPutConfig_UnknownKey_StillRejected` — e.g. `{"not_a_field": 1}` → 400. Security behaviour preserved.
- `TestHandlerPutConfig_ServerPort_RangeStillValidated` — range check still fires on out-of-range port, even though key is now read-only.

Run with `make test-docker`.

## Out of scope

- Refactoring the Flutter app to skip round-tripping (its config endpoints are separate).
- Adding a write path for `repo_overrides` / `agent_configs` — those have dedicated config surfaces (`config.toml`, `/agents`) and the web UI doesn't edit them directly.

ivanmunozruiz:
## Summary

Web UI's `Save & reload` was returning 400 with `unknown config key: "agent_configs"` (or `non_monitored` / `repo_overrides`, randomly) because GET /config returns three fields that PUT /config's allowlist never accepted. The Svelte page intentionally round-trips the full GET payload as a forward-compat safeguard, so the first rejected key made the whole save fail.

Two allowlists now:

- **`validConfigKeys`** — keys that PUT persists. Unchanged behaviour; `server_port` moved out (see below).
- **`readOnlyConfigKeys`** — `non_monitored`, `repo_overrides`, `agent_configs`, `server_port`. Accepted at the endpoint boundary (200), dropped before `SetConfig`. Still a strict allowlist — unknown keys still 400.

`server_port` moves out of the writable set because ApplyStore already ignored it as bootstrap-only (#81), so every UI save was writing a redundant row the reload immediately discarded. Its numeric-range pre-check stays intact; clients still get 400 on out-of-range values.

## Test plan

- [x] `make test-docker` — full suite green.
- [x] New tests:
  - `TestHandlerPutConfig_ReadOnlyKeys_Accepted` (5 subcases: each key individually + all-at-once)
  - `TestHandlerPutConfig_ReadOnlyKeys_NotPersisted` — no leakage into `configs` table
  - `TestHandlerPutConfig_WritableAndReadOnly_Mixed` — realistic save: writable lands, read-only dropped
  - `TestHandlerPutConfig_UnknownKey_StillRejected` — security regression guard
  - `TestHandlerPutConfig_ServerPort_RangeStillValidated` — range check still fires post-move
- [ ] Manual smoke: open web UI `/config`, change any editable field, Save & reload. Expected: 200, no red banner; refresh keeps the change.

## Out of scope

- Flutter app PUT path (has its own config endpoints).
- Adding write paths for `repo_overrides` / `agent_configs` — those have dedicated surfaces (`config.toml`, `/agents` endpoint).

Closes #86


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-documented fix. The two-allowlist approach (writable + read-only) is sound and maintains the security invariant from issue #4. Tests are thorough, covering acceptance, non-persistence, mixed payloads, unknown-key rejection, and validation-still-fires for server_port.

### Issues

🟡 **daemon/internal/server/handlers.go:339** — The server_port range validation (referenced at ~line 362) now runs on a key that will be silently dropped in the persistence loop. This coupling is implicit: a future refactor that reorders validation to run after or inside the persistence loop would silently swallow bad port values instead of returning 400, and the only guard is the test. Consider adding an inline comment near the port-range check noting it intentionally covers a read-only key.
🟡 **daemon/internal/server/handlers.go:363** — Go map iteration is unordered, so when a request contains both an unknown key and a read-only key, which error surfaces first is nondeterministic. This is pre-existing behavior and low-impact (the request is still rejected), but worth noting if deterministic error reporting ever matters for the UI.

### Suggestions

- Consider a combined lookup structure (e.g., a struct with a `readOnly bool` field per key) to avoid the risk of the two maps drifting out of sync as keys are added over time.
- Add an inline comment near the server_port numeric-range pre-check explicitly noting it validates a read-only key, so future maintainers don't accidentally move or skip it during refactors.

---

Muriano (approved):
## 🤖 Heimdallr AI Review

Clean, well-designed fix. Introduces a read-only key allowlist so the web UI can round-trip GET /config payloads without 400s, while still enforcing strict key validation and preventing persistence of read-only values. Test coverage is thorough across all meaningful scenarios (acceptance, non-persistence, mixed payloads, unknown-key rejection, and range validation). No confirmed bugs or security issues found.

### Suggestions

- TestHandlerPutConfig_ServerPort_RangeStillValidated references a hardcoded line number in its comment ("handlers.go:362") which will drift with future edits — consider removing the line reference or using a more stable anchor.
- Consider adding a compile-time or init-time assertion that validConfigKeys and readOnlyConfigKeys have no overlapping keys, to prevent a future contributor from accidentally placing a key in both maps (which would silently allow writes to a supposedly read-only field).

---

PR #93, opened 2026-04-20
feat(issues): issue dependencies + auto-promote when blockers close

### LINKED ISSUE #91: feat(issues): issue dependencies + auto-promote when blockers close
## Motivation

Today the issue pipeline classifies each issue in isolation: an issue either carries a `develop_label` (→ auto_implement) or it doesn't. For multi-step work — "land the schema migration before the API change", "stand up the endpoint before the UI starts consuming it" — the only ways to keep Heimdallm from running all issues in parallel are (a) don't label the downstream ones until you remember to come back, or (b) label them and hope the AI agent figures out the ordering from context. Both lose the "set-and-forget" value prop.

We need a first-class notion of **issue dependencies**: mark downstream work as blocked, let Heimdallm promote it automatically when its blockers close.

### 1. Dependency declaration

Parse a structured section in the issue body. Suggested format (human-readable, machine-parseable, tolerant):

```markdown
## Depends on
- #42
- owner/other-repo#57
```

Cross-repo via `owner/repo#N`. Same-repo via `#N`. The pipeline scans monitored-repo issue bodies once per poll and caches `(issue, [deps])` in the store.

Alternative / companion formats to also accept (all low-cost with a shared regex):
- `Depends on #42` / `Blocked by #42` inline (common ad-hoc conventions).
- HTML comment marker: `<!-- heimdallm-deps: #42, owner/r#57 -->` for users who don't want the section visible.

Open question for brainstorm: pick one canonical form or accept all?

### 2. New label dimension — blocked

Introduce a fourth label list next to the existing three:

```
# Issues carrying a label from this list are NOT processed while any
# declared dependency is still open. Classified as a new IssueMode
# "blocked" (below).
HEIMDALLM_ISSUE_BLOCKED_LABELS=blocked,needs-unblock
```

In `Classify()` precedence: `skip > blocked > review_only > develop > default`. An issue with a blocked label plus a develop label stays blocked until promotion.

### 3. Promotion

On every poll cycle, for each issue that carries a blocked label:
- Re-parse its deps.
- For each dep, look up issue/PR state via the GitHub API (or a cache — the daemon already fetches both).
- If ALL deps are in state `closed` (issues) or `merged`/`closed` (PRs), run a promotion action:
  - Remove every label that appears in `HEIMDALLM_ISSUE_BLOCKED_LABELS`.
  - Add the label configured in a new `HEIMDALLM_ISSUE_PROMOTE_TO_LABEL` (defaults to first entry in `HEIMDALLM_ISSUE_DEVELOP_LABELS` when unset).
  - Record the promotion in the issue timeline (a single comment: "All dependencies closed, promoting to `<develop-label>`.") for audit.
- Next poll cycle classifies the promoted issue normally and the existing pipeline picks it up.

No new webhook wiring. Everything rides the existing poll loop.

### 4. Edge cases worth thinking through

- Cross-repo deps where the target repo is NOT monitored by this Heimdallm instance → fetch the state directly via the GitHub API (one extra call per unmonitored dep). If the token can't see it, skip promotion and log a warning, don't loop.
- Dependency loops (A depends on B depends on A) → deterministic: neither promotes, log warning once.
- A dep being `reopened` after promotion → do NOT demote back (users manually reopening is rare, surprise demotion would be worse than the stale classification). Document this.
- Partial-parse failures (malformed section) → keep the issue blocked, log the parse error, don't crash.
- Rate limits: REST has 5000 req/h authenticated; one poll cycle over ~100 issues × 3 deps average = 300 req, fine.

### 5. Storage

Add a `issue_deps` column to the existing issues table (JSON array of `{repo, number}`). Populated at fetch time, consumed at promotion time. Keeps the dep graph close to where it's used and avoids a separate table.

## Scope / non-goals

- **Not** a general-purpose project-management board. No "set priority", no "assign sprint", no drag-and-drop.
- **Not** a replacement for GitHub's "linked issues" — users can still use that manually; this is Heimdallm-specific ordering.
- **Not** automatic dep inference. Users declare them explicitly.
- **Not** enabled by default — users opt in by setting `HEIMDALLM_ISSUE_BLOCKED_LABELS`.

## Flow before coding

This is ~200-300 LOC across config + pipeline + github client + store, with multiple design choices (canonical dep format, promotion-label precedence, where the promotion hook lives). Suggest brainstorming pass first before cutting an implementation plan.

## Acceptance criteria

- [ ] An issue carrying a label from `HEIMDALLM_ISSUE_BLOCKED_LABELS` is NOT picked up by the existing auto_implement / review_only pipelines.
- [ ] When all its declared deps reach `closed`/`merged`, Heimdallm replaces the blocked label(s) with the configured promote-to label within one poll cycle.
- [ ] The promotion is recorded as a comment on the issue for audit.
- [ ] Cross-repo deps work for repos where the token has read access, monitored or not.
- [ ] Full integration test covers: blocked issue with open dep → not processed; dep closes → promoted; promoted issue runs the normal pipeline.
- [ ] Config knobs documented in `docker/.env.example` and `docs/` if it's where the pipeline is documented today.

ivanmunozruiz:
## Summary

Issue-to-issue dependency tracking + automatic label-based promotion. Downstream issues declare blockers in their body under a `## Depends on` heading and carry a "blocked" label; Heimdallm holds them out of the pipeline until every referenced issue/PR closes, then flips the label automatically so the next poll cycle picks them up through the normal dispatch.

Opt-in via `HEIMDALLM_ISSUE_BLOCKED_LABELS` — installs that don't set it see zero behavioural change.

## Design choices

- **Canonical dep format**: `## Depends on` section with `- #N` or `- owner/repo#N` bullets only. Case-insensitive heading, optional colon. No HTML-comment marker or inline `Depends on #N` prose parsing in v1 (additive, can land later without migration).
- **Precedence**: `skip > blocked > review_only > develop > default_action`. An issue tagged `blocked` + `feature` stays blocked until promoted.
- **Promotion runs BEFORE fetch** in the same poll cycle so a freshly promoted issue is processed immediately, not next tick.
- **Cross-repo deps** resolved via a direct `GET /repos/{repo}/issues/{N}` call — works for any repo the token can read, even outside the monitored set.
- **Audit comment** on by default ("All declared dependencies are closed, promoting to \`ready\`"). Single-line, non-fatal if it fails after the label flip.

## Config

```bash
HEIMDALLM_ISSUE_BLOCKED_LABELS=blocked,queued
# optional — defaults to first HEIMDALLM_ISSUE_DEVELOP_LABELS entry:
HEIMDALLM_ISSUE_PROMOTE_TO_LABEL=ready
```

Validation refuses `BlockedLabels` without a resolvable promote target — blocked issues with nowhere to land would stick forever.

## What's new

- `config.IssueModeBlocked` + `BlockedLabels` + `PromoteToLabel` + `ResolvePromoteToLabel()`.
- `issues.ParseDependencies(body, defaultRepo)` — forgiving parser.
- `issues.PromoteReady(ctx, client, cfg, repos)` — orchestrator.
- `github.Client.ListOpenIssues` (raw paginated list), `.GetIssue`, `.RemoveLabels`.
- Main poll loop calls `PromoteReady` before the existing `ProcessRepo` loop when `BlockedLabels` is set.

## Docs

- `README.md`: new Headline feature bullet, updated Issue triage description, dedicated subsection with config + body format + promotion flow explanation.
- `docker/.env.example`: new commented-out env block with the body-format example.
- (No AGENTS.md update — it's about dev workflow only; this is a user feature.)

## Test plan

- [x] `make test-docker` — all packages green (`go vet` + tests).
- [x] TDD coverage: classification precedence (5 subcases), parser (9 cases incl. cross-repo/Markdown-link/stops-at-next-heading), `RemoveLabels` (happy + noop + 404-tolerated + error), `GetIssue`, `PromoteReady` (8 cases incl. disabled-noop, open-dep-not-promoted, all-deps-closed, no-deps-stays-blocked, cross-repo, multiple blocked labels, per-repo failure isolation, missing promote target).
- [ ] Manual smoke: create an issue body with `## Depends on\n- #N`, label it `blocked`, confirm it doesn't get processed. Close #N, wait one poll cycle, confirm the issue is relabelled to the promote target and an audit comment appears.

## Out of scope (documented)

- Demotion on reopen (surprise-demoting is worse than the stale classification).
- Dep-graph visualisation in the web UI (can land later).
- Automatic dep inference (users declare explicitly).
- `## Depends on` alternative syntaxes beyond bullets — additive if requested.

Closes #91


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured feature addition for issue dependencies with good test coverage, clear separation of concerns (parser, promoter, config), and proper opt-in design. A few issues around API call volume, missing context cancellation propagation, and a subtle regex edge case.

### Issues

⚠️ **daemon/internal/issues/promoter.go:99** — allDepsClosed calls GetIssue sequentially for every dependency of every blocked issue, each a separate HTTP request. For repos with many blocked issues or many deps, this could hit GitHub rate limits quickly. Consider batching with the GraphQL API or at minimum caching results within a single promotion pass (the same dep issue may be referenced by multiple blocked issues).
⚠️ **daemon/internal/issues/promoter.go:42** — The ctx parameter is accepted but never passed to ListOpenIssues, GetIssue, AddLabels, RemoveLabels, or PostComment. If the parent context is cancelled (e.g. daemon shutdown), in-flight GitHub API calls won't be interrupted. The PromoteIssueClient interface methods should accept context.Context, or at minimum ctx.Err() should be checked before each GitHub call inside the inner loop, not just the outer repo loop.
🟡 **daemon/internal/issues/deps.go:36** — The issueRef regex `([\w.-]+/[\w.-]+)?#(\d+)` will match refs embedded in URLs, code blocks, or quoted text within bullet lines. For example, a bullet like `- See https://github.com/org/repo/issues/42#issuecomment-123` would extract a spurious ref. Consider anchoring more tightly or filtering out matches that appear inside URLs.
⚠️ **daemon/internal/issues/promoter.go:62** — ListOpenIssues fetches ALL open issues for a repo (up to maxIssuePages * issuesPerPage) just to find the ones with blocked labels. This is wasteful for repos with many open issues. Consider using the GitHub label filter parameter (`?labels=blocked`) in the API call to fetch only relevant issues, which would dramatically reduce API calls and response size.
🟡 **daemon/internal/github/repos.go:155** — In RemoveLabels, the response body is read even on success (200) but never used. Minor inefficiency — consider discarding the body with io.Copy(io.Discard, ...) for the 200 case, or at least only reading it on error status codes.
⚠️ **daemon/cmd/heimdallm/main.go:357** — PromoteReady is called with context.Background() instead of the daemon's main context. If the daemon is shutting down, this promotion pass will run to completion (potentially making many GitHub API calls) instead of being cancelled gracefully. Use the daemon's context here.
🟡 **daemon/internal/issues/deps.go:22** — The dependsHeading regex uses `##+` (two or more #), but the lower-cased line is matched. However, the nextHeading regex on line 26 matches the original (non-lowered) line. This inconsistency is fine for headings (# is case-irrelevant) but the code flow is confusing: line 55 matches `lower` against dependsHeading, while line 60 matches `line` against nextHeading but also checks `lower` against dependsHeading. Consider making this consistent for readability.
🟡 **daemon/internal/issues/promoter.go:80** — There's a TOCTOU race: between checking allDepsClosed and calling applyPromotion, a dependency could be reopened. This is inherently unavoidable with polling, but the audit comment posted by applyPromotion should mention the dependency states at time of check, so operators can debug if a premature promotion occurs.

### Suggestions

- Add a dedup cache for GetIssue calls within a single PromoteReady pass — the same dependency (e.g. a shared blocker issue) may be referenced by multiple blocked issues, causing redundant API calls.
- Use GitHub's label filter query parameter in ListOpenIssues (or add a new ListIssuesByLabel method) to avoid fetching all open issues when only blocked ones are needed.
- Thread the daemon's context.Context through PromoteReady and into the PromoteIssueClient interface methods so shutdown is respected.
- Consider adding a simple integration/end-to-end test that exercises the full flow: blocked issue → deps close → promotion triggers → issue reclassified. The unit tests are good but the interaction between PromoteReady and the main loop's FetchIssues in the same cycle is only tested implicitly.
- Document the GitHub API rate-limit implications in the README: each blocked issue with N deps costs N+1 API calls per poll cycle. Users with many blocked issues on short poll intervals should be aware.

---

Muriano:
## ⚠️ MEDIUM Issue

applyPromotion performs remove → add → comment sequentially. If AddLabels fails after RemoveLabels succeeds, the issue loses both its blocked label and never receives the promote-to label, becoming invisible to both the promotion pass and the normal pipeline. No compensating action (re-apply blocked label, post error comment) is taken. The issue is effectively orphaned until a human notices via GitHub label history. Fix: attempt to restore the blocked label(s) on AddLabels failure, or at minimum post an audit comment noting the partial failure.

**Location:** `daemon/internal/issues/promoter.go` line 130

---

Muriano:
## ⚠️ MEDIUM Issue

allDepsClosed issues one sequential GetIssue HTTP call per dependency with no concurrency limit or context deadline. An issue with N dependencies makes N serial API calls, each bounded only by the HTTP client timeout (30s). For issues with many deps (e.g., 20), a single promotion check could block for up to 10 minutes on timeouts. Additionally, ctx.Err() is checked per-repo but not per-dependency call, so cancellation during dep checking is not responsive. Fix: check ctx.Err() inside the dep loop, and consider adding a per-issue timeout or parallel dep checks with bounded concurrency.

**Location:** `daemon/internal/issues/promoter.go` line 81

---

Muriano (approved):

Well-structured opt-in feature adding issue dependency tracking and auto-promotion. Clean architecture with interface-scoped client, thorough tests, and proper validation. Two medium-severity issues around partial failure during promotion and missing context propagation.

### Suggestions

- The issueRef regex `(?:[\w.-]+/[\w.-]+)?#(\d+)` can match URL fragment anchors like `github.com/org/repo#L42` as cross-repo refs. Consider adding a word boundary or negative lookbehind for `/` to avoid false positives from GitHub file URLs in bullet lines.
- promoter_test.go uses `package issues` (white-box) while the existing pipeline_test.go uses `package issues_test` (black-box). Consider aligning to the same convention for consistency.
- RemoveLabels in repos.go does not accept or check a context — if the daemon is shutting down during a multi-label removal, there's no way to cancel in-flight requests. Threading ctx through or adding a ctx.Err() check between iterations would improve shutdown responsiveness.
- The audit comment body in applyPromotion is a hardcoded string. Consider listing which specific dependencies were resolved (the IssueRef list is available in the caller) to give operators a quick link trail without checking each dep manually.

---

PR #99, opened 2026-04-20
feat: RepoDetailScreen + per-repo issue tracking config

Muriano:
## Summary

Replaces the inline repo expansion tile with a dedicated full-page settings screen per repository, and adds per-repo issue tracking with field-level override/merge.

### Daemon (Go)
- `RepoAI` gains `IssueTracking *IssueTrackingConfig` for per-repo overrides
- `IssueTrackingForRepo(repo)` merges per-repo non-zero fields over global config (3 tests)
- Poll cycle uses per-repo config instead of global for each repo
- `GET /config` response includes per-repo `issue_tracking` in `repo_overrides`

### Flutter (Dart)
- **RepoDetailScreen** — full-page settings with 4 section cards: General, AI Agent, Issue Tracking, PR Metadata
- **OverrideTextField / OverrideDropdown** — reusable widgets for per-field override pattern (green border + "overridden" badge + reset button when overriding, muted "global" display when inheriting)
- **RepoConfig** extended with 11 nullable fields (IT + PR metadata)
- **repos_screen.dart** simplified: ExpansionTile removed, repos navigate to `/repos/:name`
- **TOML writer** generates `[ai.repos."org/repo".issue_tracking]` sections
- **Global issue tracking** configurable in Settings screen

## Test Plan

- [x] `make test-docker` — all daemon packages pass (incl. 3 new merge tests)
- [x] `flutter analyze` — 0 new issues
- [x] `flutter build macos --release` — 48.9MB OK
- [ ] Manual: open repo detail, set develop_labels override, save — verify config.toml
- [ ] Manual: verify daemon uses per-repo labels (check triage classification)
- [ ] Manual: reset an override field — verify it reverts to global value

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -300,10 +300,20 @@ func main() {
 			p.PublishPending()
 
 			// ── Issue tracking cycle ─────────────────────────────────────
+			// Check if ANY repo has issue tracking enabled (global or per-repo).
 			cfgMu.Lock()
 			globalIT := c.GitHub.IssueTracking
+			anyITEnabled := globalIT.Enabled
+			if !anyITEnabled {
+				for _, r := range c.AI.Repos {
+					if r.IssueTracking != nil && r.IssueTracking.Enabled {
+						anyITEnabled = true
+						break
+					}
+				}
+			}
 			cfgMu.Unlock()
-			if globalIT.Enabled {
+			if anyITEnabled {
 				loginMu.Lock()
 				authUser := cachedLogin
 				loginMu.Unlock()
@@ -381,6 +391,9 @@ func main() {
 					cfgMu.Lock()
 					repoIT := c.IssueTrackingForRepo(repo)
 					cfgMu.Unlock()
+					if !repoIT.Enabled {
+						continue
+					}
 					n, err := issueFetcher.ProcessRepo(context.Background(), repo, repoIT, authUser, optsFor)
 					if err != nil {
 						slog.Error("poll: issue fetch failed", "repo", repo, "err", err)
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -301,9 +301,9 @@ func main() {
 
 			// ── Issue tracking cycle ─────────────────────────────────────
 			cfgMu.Lock()
-			itCfg := c.GitHub.IssueTracking
+			globalIT := c.GitHub.IssueTracking
 			cfgMu.Unlock()
-			if itCfg.Enabled {
+			if globalIT.Enabled {
 				loginMu.Lock()
 				authUser := cachedLogin
 				loginMu.Unlock()
@@ -378,7 +378,10 @@ func main() {
 				}
 
 				for _, repo := range repos {
-					n, err := issueFetcher.ProcessRepo(context.Background(), repo, itCfg, authUser, optsFor)
+					cfgMu.Lock()
+					repoIT := c.IssueTrackingForRepo(repo)
+					cfgMu.Unlock()
+					n, err := issueFetcher.ProcessRepo(context.Background(), repo, repoIT, authUser, optsFor)
 					if err != nil {
 						slog.Error("poll: issue fetch failed", "repo", repo, "err", err)
 						continue
@@ -439,14 +442,18 @@ func main() {
 		cfgMu.Lock()
 		c := cfg
 		cfgMu.Unlock()
-		repoOverrides := make(map[string]map[string]string)
+		repoOverrides := make(map[string]map[string]any)
 		for repo, ai := range c.AI.Repos {
-			repoOverrides[repo] = map[string]string{
+			ro := map[string]any{
 				"primary":     ai.Primary,
 				"fallback":    ai.Fallback,
 				"review_mode": ai.ReviewMode,
 				"local_dir":   ai.LocalDir,
 			}
+			if ai.IssueTracking != nil {
+				ro["issue_tracking"] = ai.IssueTracking
+			}
+			repoOverrides[repo] = ro
 		}
 		agentConfigs := make(map[string]map[string]any)
 		for name, ac := range c.AI.Agents {
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -438,6 +438,8 @@ func main() {
 	}()
 
 	// Expose live config for GET /config
+	srv.SetRepoMetaFns(ghClient.FetchLabels, ghClient.FetchCollaborators)
+
 	srv.SetConfigFn(func() map[string]any {
 		cfgMu.Lock()
 		c := cfg
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -369,8 +369,8 @@ func main() {
 				// same poll cycle rather than waiting for the next one.
 				// No-op when BlockedLabels is unset, so default installs
 				// don't pay for the extra calls.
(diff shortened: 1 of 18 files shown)


ivanmunozruiz (approved):
## ✅ LGTM

Clean mirror of the daemon's `IssueTrackingConfig` — JSON keys, defaults, and TOML path all match the Go side exactly. UI composition follows the same `copyWith` + form-state pattern used by the rest of Settings, so there's no new abstraction to learn.

### Evaluation

| Dimension | Verdict |
|-----------|---------|
| JSON keys ↔ daemon Go struct | ✅ exact match (verified against `daemon/internal/config/config.go:103-134`) |
| TOML path `[github.issue_tracking]` | ✅ correct nested sub-table |
| Defaults (`exclusive` / `ignore` / `false`) | ✅ match Go zero values |
| Codebase patterns (`copyWith`, `toJson`, `fromJson`, `_stringList`) | ✅ idiomatic mirror |
| TOML escape on label values | ✅ uses `_tomlEscapeString` everywhere |
| UI composition (Switch collapses sub-fields) | ✅ consistent with the rest of `ConfigScreen` |
| Server accepts `issue_tracking` at root of `PUT /config` | ✅ confirmed (`daemon/internal/config/store.go:86`) |

### Non-blocking notes for future polish

- **No unit tests for the new serialization.** The 15 existing tests still pass, but a tiny round-trip (`IssueTrackingConfig → toJson → fromJson → equal`) and a TOML-emission smoke test would catch accidental key renames. Worth a quick follow-up PR if issue tracking becomes the primary surface people touch.
- **Manual test plan items unchecked.** The PR body leaves "toggle on, set labels, save" and "daemon picks up changes after reload" as TODO. Worth confirming on a real build before relying on it.
- **`Switch` lacks an explicit accessibility label.** Screen readers hear "switch, on/off" without context. A `Semantics(label: 'Enable issue tracking')` wrap is a one-liner.
- **`TextFormField.initialValue` only applies on first build.** If the form state is ever reset externally (e.g., server-side reload pushed via SSE), the label fields won't reflect it. Not hit by the current flow, but worth noting.
- **Labels containing commas won't round-trip** through the CSV parser. GitHub almost never allows commas in labels, so this is a theoretical edge.

Nothing here is a blocker — ship it.

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Clean addition of issue tracking config model, TOML serialization, and settings UI. The model and serialization are solid. Main concerns are a potential crash from unvalidated dropdown values and a known Flutter gotcha with TextFormField initialValue.

### Issues

🔴 **flutter_app/lib/features/config/config_screen.dart:384** — DropdownButtonFormField will throw an assertion error if `_issueTracking.filterMode` (or `defaultAction`) loaded from JSON contains a value not in the `items` list (e.g., a typo or future value in the config file). Clamp unknown values to the default in `fromJson`, or add the current value to the items list dynamically.
⚠️ **flutter_app/lib/features/config/config_screen.dart:480** — TextFormField with `initialValue` only reads the value on first mount. If config is reloaded from the server while the issue tracking section is visible and expanded, the text fields will show stale data. Consider using a `TextEditingController` that you update in the config-reload path, or use a `key` (e.g., `ValueKey(values)`) to force rebuild.
⚠️ **flutter_app/lib/core/models/config_model.dart:140** — `filterMode` and `defaultAction` are stringly-typed with no validation. The model will happily accept any string from JSON, which can then crash the dropdown (see config_screen issue). Consider validating in `fromJson` and falling back to the default for unknown values, or use an enum.
🟡 **flutter_app/lib/core/models/config_model.dart:200** — `_stringList` uses `.cast<String>()` which throws an opaque TypeError if any list element is not a String. `.map((e) => e as String).toList()` gives a better stack trace, or `.whereType<String>().toList()` silently filters non-strings — pick based on desired strictness.

### Suggestions

- Validate filterMode/defaultAction in IssueTrackingConfig.fromJson by clamping unknown values to the default. This prevents a crash in the dropdown UI when reading a hand-edited or future-version config file.
- Consider using enums for filterMode and defaultAction instead of raw strings to get compile-time safety across the model, serialization, and UI layers.
- Add a `key: ValueKey(_issueTracking.hashCode)` or similar to the _labelListField calls so the TextFormFields rebuild when the backing state changes externally.

---

Muriano:
## Response to AI Review

**Fixed (b028c98):**
- **Item 1 + 3 (dropdown crash on unknown values)**: `IssueTrackingConfig.fromJson` now validates `filterMode` and `defaultAction` against `validFilterModes` / `validDefaultActions` static lists, clamping to defaults on unknown input. Covers hand-edited config, future daemon versions, and corrupted JSON.
- **Item 2 (stale TextFormField on reload)**: `_labelListField` now uses `ValueKey('$label:${values.join(',')}')` so the widget rebuilds when the backing data changes externally.

**Not changing:**
- **Item 4 (`.cast<String>()`)**: Standard Dart idiom. The resulting `TypeError` includes the value and expected type. `whereType<String>()` would silently swallow non-strings, which is worse for debugging.
- **Enums for filterMode/defaultAction**: Clamping in `fromJson` gives the same runtime safety without adding type boilerplate across model + serialization + UI. YAGNI.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid backend implementation with clean field-level merge logic and good test coverage. A few concerns around lock granularity in the poll loop and the inability to override `Enabled` per-repo due to bool zero-value ambiguity. The 872-line plan document should likely not be merged into the main branch.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:374** — cfgMu.Lock() is acquired per-repo inside the loop to call c.IssueTrackingForRepo(repo), but `c` was already captured at the start of the poll cycle. If `c` is a pointer snapshot, the lock protects reads through a stale pointer (same config object), creating within-cycle inconsistency if a config reload mutates the underlying struct mid-iteration. If `c` is a value copy, the lock is redundant. Consider either (a) building all per-repo configs once before the loop under a single lock, or (b) re-reading `cfg` (not `c`) inside the lock if mid-cycle freshness is intended.
🟡 **daemon/internal/config/config.go:301** — IssueTrackingForRepo does not merge the `Enabled` field because bool zero-value (false) is indistinguishable from 'not set'. This means per-repo overrides cannot disable issue tracking for a specific repo when it is globally enabled. If this is intentional, document it; if per-repo opt-out is desired, use *bool for the override's Enabled field.
🟡 **daemon/cmd/heimdallm/main.go:435** — Changing repoOverrides from map[string]map[string]string to map[string]map[string]any alters the JSON shape of the GET /config response — `issue_tracking` is now a nested object instead of a flat string. This is a breaking API contract change for any consumers not updated in this PR. Ensure the Flutter model and any other clients handle both the old (no issue_tracking key) and new shapes gracefully for rolling deploys.
🟡 **docs/superpowers/plans/2026-04-20-repo-detail-screen.md:1** — 872-line implementation plan document with step-by-step instructions and code snippets committed to the repo. This will become stale immediately and adds maintenance burden. Consider keeping implementation plans outside the repo or in a separate docs PR, and removing this file before merging.

### Suggestions

- Pre-compute all per-repo IssueTrackingConfig values before the loop under a single lock to avoid both redundant locking and mid-cycle inconsistency: `repoITs := make(map[string]IssueTrackingConfig); cfgMu.Lock(); for _, r := range repos { repoITs[r] = c.IssueTrackingForRepo(r) }; cfgMu.Unlock()`
- Add a brief doc-comment or inline note on IssueTrackingForRepo stating that Enabled is always inherited from global and cannot be overridden per-repo
- Consider adding a test case where the per-repo override struct is non-nil but all fields are zero-valued, verifying it returns global config unchanged (documents the zero-value passthrough behavior)

---

PR #100, opened 2026-04-20
fix: config save + reload resilience (three linked bugs)

vbuenog:
Three independent fixes for the config-save + reload flow that the web UI exercises. Each commit stands alone; together they close the chain of errors surfaced while trying to save the Configuration screen in a dockerised deployment.

## The chain

1. Web UI saves config → PUT sends `{ ...raw, ...edits }`.
2. Daemon rejects an unknown key → web UI shows "unknown config key: ..." and never gets to the reload step.
3. Once the PUT is accepted, web UI fires POST /reload.
4. Daemon reload hard-calls `config.Load(/config/config.toml)` → ENOENT because the bind-mounted `/config` dir was created as root and the daemon user could never write the auto-generated TOML → 500.

Each commit fixes one link, but none is load-bearing on the others — #87 already added a server-side accept-and-drop for the read-only keys, so Fix 1 below is defence-in-depth, not a blocker.

### `3e8854e` — `fix(web_ui): send only writable keys in config PUT payload`

The save flow used `{ ...raw, ...edits }` to forward-compat unknown fields. But the daemon's PUT /config has a strict allowlist; `server_port`, `non_monitored`, `repo_overrides` and `agent_configs` round-trip back from GET but are rejected on write (currently tolerated via `readOnlyConfigKeys`, but that contract is daemon-side and can drift if a new read-only key is added to GET).

Build the payload explicitly from the known-writable keys only.

**Files:** `web_ui/src/routes/config/+page.svelte` (+23 / -18)

### `36bc9d7` — `fix(docker): use a named volume for /config so the daemon user can write`

`./config:/config` was a bind mount; Docker creates the host directory as root when the mount first fires, so the `heimdallm` user (uid 100 inside the image) could read `/config` but not write `config.toml` on first boot. `writeConfigTOML` silently failed, no TOML was ever created, and reload blew up downstream.

Switch to a named `heimdallm-config` volume. Docker initialises named volumes with the ownership the image already `chown`'d in — `heimdallm:heimdallm` — so the auto-generated `config.toml` persists as intended.

**Migration note:** operators with hand-edited `docker/config/config.toml` lose those edits on the first `docker compose up` after this merges. Replicate them via `docker/.env` env vars or the web UI's Configuration screen (both persist to the store layer, which survives the volume swap).

**Files:** `docker/docker-compose.yml`, `.gitignore` (+11 / -5)

### `eef3d88` — `fix(daemon): tolerate missing config.toml on reload, mirroring startup`

Cold-start uses `LoadOrCreate` when `HEIMDALLM_DATA_DIR` is set (Docker), rebuilding from env vars if no TOML exists. Reload hard-called `config.Load`, so POST /reload returned 500 the moment `config.toml` was absent — even though the daemon was happily serving traffic on the same env + store state.

Mirror the startup branch in `SetReloadFn`. In Docker we use `LoadOrCreate`; everywhere else we keep `Load` so the desktop flow doesn't auto-create a config at an unexpected path.

**Files:** `daemon/cmd/heimdallm/main.go` (+15 / -1)

## Tests

- [x] `make test-docker` — all packages pass (server, config, pipeline, issues, store, …). No regressions.
- [x] `cd web_ui && npm run check && npm test && npm run lint && npm run build` — 81/81 tests green, 0 type errors, lint clean, build succeeds.
- [x] `LoadOrCreate` already covered by `TestLoadOrCreate_Creates`, `TestLoadOrCreate_LoadsExisting`, `TestLoadOrCreate_FailsWithoutPrimary` in `daemon/internal/config/config_test.go`. No new unit test needed for the reload wrapper's env-check branch — it's two lines choosing between those two functions.

## Manual verification plan (for the reviewer)

After checkout:

```bash
# 1. Force a full rebuild — Docker's layer cache is the common reason
# these fixes "don't apply" on a laptop that's been running for a while.
docker compose -f docker/docker-compose.yml down --volumes
docker compose -f docker/docker-compose.yml build --no-cache
docker compose -f docker/docker-compose.yml up -d

# 2. Verify no "could not persist generated config" warning in daemon logs
docker logs heimdallm 2>&1 | grep -i "could not persist" || echo OK

# 3. Save config through the web UI, then hit reload
#    → should return HTTP 200, not 500
TOKEN=\$(docker exec heimdallm cat /data/api_token)
curl -sS -X POST http://127.0.0.1:7842/reload \\
  -H "X-Heimdallm-Token: \$TOKEN" -w "\\nHTTP %{http_code}\\n"
```

Expected: `HTTP 200 {"status":"reloaded"}`.

## Scope

Does **not** touch:
- PR list UI (the approved/changes-requested badge is a separate follow-up — different feature, different shape).
- Flutter app.
- Daemon handlers beyond the reload wrapper.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,4 @@ daemon/bin/
 
 # Local Docker runtime state
 # - docker/.env holds GITHUB_TOKEN + provider credentials; never commit.
-# - docker/config/ is the mounted config volume (config.toml auto-written
-#   by the daemon on first run).
 docker/.env
-docker/config/
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,9 @@ daemon/bin/
 
 # Local Docker runtime state
 # - docker/.env holds GITHUB_TOKEN + provider credentials; never commit.
+# - docker/config/ is a legacy directory from the pre-named-volume setup.
+#   The mount switched to a named volume; the directory is no longer created
+#   on fresh installs, but old checkouts may still have one sitting around
+#   — keeping the ignore entry prevents accidental commits.
 docker/.env
+docker/config/
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -495,8 +495,22 @@ func main() {
 	// Wire the reload callback: re-read config from disk, restart scheduler
 	// and the discovery loop so changes to discovery_topic / orgs / interval
 	// take effect without a daemon restart.
+	//
+	// Mirrors the startup load logic (line 66 above): in Docker deployments
+	// HEIMDALLM_DATA_DIR is set and we use LoadOrCreate so a missing TOML
+	// is not fatal — the daemon rebuilds from env vars and persists a
+	// fresh config.toml. Without this, `POST /reload` after a deployment
+	// that lost its config volume (or on fresh docker-compose startups
+	// where writeConfigTOML warned but didn't error) returned 500 even
+	// though startup itself had succeeded on the same env vars.
 	srv.SetReloadFn(func() error {
-		newCfg, err := config.Load(cfgPath)
+		var newCfg *config.Config
+		var err error
+		if os.Getenv("HEIMDALLM_DATA_DIR") != "" {
+			newCfg, err = config.LoadOrCreate(cfgPath)
+		} else {
+			newCfg, err = config.Load(cfgPath)
+		}
 		if err != nil {
 			return fmt.Errorf("reload: %w", err)
 		}
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -64,13 +64,20 @@ func main() {
 	}
 
 	cfgPath := configPath()
-	var cfg *config.Config
-	var err error
+
+	// loadConfig is captured once at startup so the reload path further
+	// down cannot drift: both read the same env-var and select the same
+	// loader. Docker deployments (HEIMDALLM_DATA_DIR set) use LoadOrCreate
+	// so a missing config.toml is not fatal — the daemon rebuilds from env
+	// vars. Desktop deployments use Load; the Flutter app is expected to
+	// have written the TOML before the daemon starts, so ENOENT is a real
+	// error there.
+	loadConfig := config.Load
 	if os.Getenv("HEIMDALLM_DATA_DIR") != "" {
-		cfg, err = config.LoadOrCreate(cfgPath)
-	} else {
-		cfg, err = config.Load(cfgPath)
+		loadConfig = config.LoadOrCreate
 	}
+
+	cfg, err := loadConfig(cfgPath)
 	if err != nil {
 		slog.Error("config load failed", "path", cfgPath, "err", err)
 		os.Exit(1)
@@ -494,23 +501,11 @@ func main() {
 
 	// Wire the reload callback: re-read config from disk, restart scheduler
 	// and the discovery loop so changes to discovery_topic / orgs / interval
-	// take effect without a daemon restart.
-	//
-	// Mirrors the startup load logic (line 66 above): in Docker deployments
-	// HEIMDALLM_DATA_DIR is set and we use LoadOrCreate so a missing TOML
-	// is not fatal — the daemon rebuilds from env vars and persists a
-	// fresh config.toml. Without this, `POST /reload` after a deployment
-	// that lost its config volume (or on fresh docker-compose startups
-	// where writeConfigTOML warned but didn't error) returned 500 even
-	// though startup itself had succeeded on the same env vars.
+	// take effect without a daemon restart. Reuses the `loadConfig` closure
+	// captured at startup so the two paths cannot drift on which loader
+	// they pick — both see the same HEIMDALLM_DATA_DIR snapshot.
 	srv.SetReloadFn(func() error {
-		var newCfg *config.Config
-		var err error
-		if os.Getenv("HEIMDALLM_DATA_DIR") != "" {
-			newCfg, err = config.LoadOrCreate(cfgPath)
-		} else {
-			newCfg, err = config.Load(cfgPath)
-		}
+		newCfg, err := loadConfig(cfgPath)
 		if err != nil {
 			return fmt.Errorf("reload: %w", err)
 		}
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -68,6 +68,20 @@ services:
       # POST /reload return 500 whenever config.toml was missing. The
       # image already declares VOLUME ["/config"] and chowns the path to
       # heimdallm:heimdallm during build.
+      #
+      # Migration (pre-v0.1.8 -> this): if you had a customised
+      # docker/config/config.toml under the old bind mount, copy it into
+      # the new named volume BEFORE your first `docker compose up` with
+      # this version, otherwise the daemon will regenerate from env vars
+      # and your manual edits are lost silently:
+      #
+      #   docker compose up --no-start heimdallm     # create container
+      #   docker cp docker/config/config.toml heimdallm:/config/config.toml
+      #   docker compose up -d                        # start with the file in place
+      #
+      # Env-driven setups (HEIMDALLM_* in docker/.env) and saves you made
+      # via the web UI Configuration screen are unaffected; both flow
+      # through the store layer and survive the volume swap.
       - heimdallm-config:/config
       #
       # Gemini OAuth tokens from host (if using browser-based auth):
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -60,8 +60,15 @@ services:
     volumes:
       # Persistent data (SQLite DB + API token)
       - heimdallm-data:/data
-      # Configuration (optional — env vars work without this)
-      - ./config:/config
+      # Configuration. Named volume (not a bind mount) so the daemon user
+      # inside the container owns the directory and can persist the
+      # generated config.toml on first boot. A `./config:/config` bind
+      # mount inherits the host directory's ownership (root, since Docker
+      # creates it on demand), which blocked writeConfigTOML and made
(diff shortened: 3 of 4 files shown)


sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Solid three-bug fix that aligns reload logic with startup, fixes container volume ownership, and stops the UI from round-tripping read-only fields. The main concern is the bind-mount-to-named-volume migration path for existing deployments.

### Issues

🔴 **docker/docker-compose.yml:67** — Breaking change for existing users: switching from bind mount `./config:/config` to named volume `heimdallm-config:/config` silently drops any custom config.toml that users had in their local `./config/` directory. There are no documented migration steps. Users upgrading via `docker-compose pull && docker-compose up` will lose their configuration without warning.
🟡 **daemon/cmd/heimdallm/main.go:505** — The reload closure re-derives the load strategy by checking `os.Getenv("HEIMDALLM_DATA_DIR")` independently of the startup path (line ~66). If the startup logic ever changes its detection mechanism, this branch will silently drift. Consider capturing the load strategy once at startup (e.g., a `loadConfig func(string) (*config.Config, error)` variable closed over by both paths) so the two remain in lockstep.
🟡 **.gitignore:8** — Removing `docker/config/` from .gitignore means existing checkouts that still have a leftover `docker/config/` directory from previous deployments could accidentally stage and commit it. Consider keeping the ignore entry with a comment noting the bind mount is deprecated, or add a note in upgrade docs to delete the directory.
🟡 **web_ui/src/routes/config/+page.svelte:108** — The explicit allowlist payload is safer but creates a maintenance coupling: any new writable field added to the daemon's PUT /config must also be added here or it will be silently dropped on save. A brief code comment noting this contract (or a shared type/constant for the writable keys) would reduce the risk of future field omissions.

### Suggestions

- Add a migration note (CHANGELOG, README, or docker-compose comment) telling existing users how to copy their config into the new named volume before upgrading, e.g. `docker cp ./config/config.toml heimdallm-heimdallm-1:/config/`.
- Extract the config-load strategy into a closure or variable set once at startup and reused in the reload callback, eliminating the duplicated env-var check.
- Keep `docker/config/` in .gitignore (even as a comment-only entry) to protect users who still have the directory locally from accidental commits.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Three-bug fix covering config save (UI sending rejected fields), reload resilience (using Load vs LoadOrCreate), and Docker volume ownership. The core intent is sound, but each change has a notable concern worth addressing before merge.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:506** — HEIMDALLM_DATA_DIR is used as a proxy for 'should we tolerate a missing config file', but these are semantically different things. Any non-Docker deployment that happens to set HEIMDALLM_DATA_DIR will silently get LoadOrCreate semantics on reload, potentially masking a genuinely missing config. A dedicated HEIMDALLM_CONFIG_OPTIONAL=true env var (or simply always using LoadOrCreate in the reload path, mirroring the rationale that reload should be at least as resilient as startup) would be less surprising.
⚠️ **docker/docker-compose.yml:63** — Switching from a bind mount (./config:/config) to a named volume (heimdallm-config:/config) is a silent breaking change for existing deployments. Any config.toml previously written to ./config will be invisible after the upgrade — the daemon will regenerate from env vars, but custom manual edits are silently lost. No migration note is provided in the compose file or PR description.
🟡 **web_ui/src/routes/config/+page.svelte:124** — The issue_tracking block still spreads `raw.issue_tracking` before overriding known fields. If the server returns unrecognized sub-keys inside issue_tracking (e.g. from a newer daemon version), they will be forwarded to PUT /config and could trigger the same 400 the PR is fixing for top-level fields. The spread should either be removed or the daemon's allowlist extended to cover this nested object explicitly.
🟡 **web_ui/src/routes/config/+page.svelte:111** — `aiFallback || undefined` uses loose falsiness: an empty string is intentionally cleared by the user, but a value of '0' or any other falsy-but-valid string would be incorrectly dropped. Prefer `aiFallback !== '' ? aiFallback : undefined` or `aiFallback ?? undefined` depending on the intended semantics.

### Suggestions

- Consider always using LoadOrCreate in the reload path rather than branching on HEIMDALLM_DATA_DIR — reload should be at least as resilient as startup regardless of deployment style.
- Add a comment or upgrade note in docker-compose.yml (or a CHANGELOG entry) explaining that users with existing ./config bind-mount data should copy config.toml into the new named volume before upgrading.
- Strip the raw.issue_tracking spread entirely from the payload and enumerate only the known keys, consistent with how the top-level payload is now constructed.
- Verify in the Dockerfile that /config is actually declared as VOLUME and chowned to heimdallm:heimdallm — the compose comment asserts this but the Dockerfile is not in the diff.

---

Muriano:
## 🟡 LOW Issue

Code duplication: the HEIMDALLM_DATA_DIR env-var conditional that selects LoadOrCreate vs Load is now duplicated between startup (~line 70) and the reload closure (~line 504). If one branch is updated (e.g., a new env var or a third loading mode), the other can silently drift. Consider extracting a small helper like `loadConfig(cfgPath string) (*config.Config, error)` that encapsulates the conditional, so both call sites stay in sync. Impact: maintenance risk, not a runtime bug today.

**Location:** `daemon/cmd/heimdallm/main.go` line 504

---

Muriano (approved):

Clean, well-documented fix for three linked bugs (reload 500 in Docker, bind-mount permissions, UI 400 on save). The reload function correctly mirrors startup logic, the Svelte payload now sends exactly the 7 keys in validConfigKeys, and the named volume resolves ownership issues. No confirmed bugs or security issues found.

### Suggestions

- Document the bind-mount → named-volume migration in release notes or CHANGELOG: existing users with a customized `./config/config.toml` will need to either (a) set equivalent env vars or (b) `docker cp` their TOML into the new `heimdallm-config` volume after the first `docker compose up`.
- The `.gitignore` removal of `docker/config/` means that directory can now be tracked if someone creates it locally (e.g., old checkout that still has it). Consider adding a one-line comment noting the directory is no longer used, or keep the gitignore entry with a 'legacy' note to prevent accidental commits.

---

vbuenog:
## Addressing the review feedback

Three follow-up commits pushed (`b796578`, `56eda50`, `1a1da46`). Mapping each to the reviewer concerns:

### 🔴 Migration path for the bind-mount → named-volume swap (sergiotejon / ivanmunozruiz / Muriano)

- **`b796578` docs(docker): migration path for /config named-volume swap**
  - `docker/docker-compose.yml`: inline migration block at the named-volume declaration with the exact `docker cp` one-liner. Anyone reading the compose file at the site of the change sees the instructions without needing to hunt elsewhere.
  - Commit message carries a `BREAKING CHANGE:` trailer so release-please surfaces the note in the next CHANGELOG entry.
  - `.gitignore`: re-added `docker/config/` with a "legacy" comment so leftover directories on old checkouts can't be accidentally committed.

### 🟡 Reload loader duplication / drift risk (sergiotejon / ivanmunozruiz)

- **`56eda50` refactor(daemon): share loader between startup and reload via closure**
  - Capture `loadConfig := config.Load` at startup, swap to `config.LoadOrCreate` when `HEIMDALLM_DATA_DIR` is set, and close over it in `SetReloadFn`. One decision, two call sites, no drift.
  - The alternative "just always use LoadOrCreate on reload" (ivanmunozruiz) was considered but rejected in the commit message: on desktop, the Flutter app owns config creation, and having the daemon auto-create a TOML on the first reload would collide with that setup flow. The env-var split stays, but lives in exactly one place now.

### 🟡 `issue_tracking` spread + loose `aiFallback` fallback (ivanmunozruiz)

- **`1a1da46` fix(web_ui): drop issue_tracking spread + stricter aiFallback clear check**
  - Enumerate the eight known `issue_tracking` sub-keys explicitly — no `...raw.issue_tracking`. Fixes the same class of bug as the outer payload, one level deeper.
  - `aiFallback !== '' ? aiFallback : undefined` instead of `|| undefined`.
  - Contract comment above the payload points at `validConfigKeys` in `daemon/internal/server/handlers.go` so future authors know the two must stay in lockstep.
  - Side-effect: the `raw: Config` state variable is no longer referenced — removed (lint caught it).

## Verification

- `make test-docker`: all daemon packages pass.
- `cd web_ui && npm run check && npm test && npm run lint && npm run build`: 81/81 tests, type-check + lint clean, build OK.

Ready for re-review.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured fix for three linked bugs: Docker volume permissions, reload loader drift, and UI payload round-tripping. The closure approach for loadConfig is clean, the named-volume swap is correct, and explicitly enumerating the save payload is safer than spreading raw server state. Comments and migration docs are thorough. No runtime bugs found; only minor maintenance concerns.

### Issues

🟡 **web_ui/src/routes/config/+page.svelte:107** — The lockstep contract between this explicit payload object and `validConfigKeys` in daemon/internal/server/handlers.go is documented only by a comment. If either side adds or removes a field, the other silently breaks (dropped writes or 400s). Consider generating the key list from a shared schema or adding an integration test that asserts the two sets match.
🟡 **docker/docker-compose.yml:63** — The 20-line inline migration comment is valuable but heavy for a compose file. Users who deploy via automated tooling or who copy-paste compose snippets may never see it. Consider also surfacing the migration steps in a top-level UPGRADING.md or release notes (the BREAKING CHANGE trailer helps, but changelogs are often skimmed).
🟡 **web_ui/src/routes/config/+page.svelte:130** — The `aiFallback !== '' ? aiFallback : undefined` change is more precise than `|| undefined`, but the inline comment explaining why is 5 lines for a one-character semantic difference. The reasoning is sound; the comment weight is disproportionate.

### Suggestions

- Add a lightweight integration or build-time test that compares the set of keys the Svelte save payload sends against `validConfigKeys` in handlers.go, so the lockstep requirement is enforced automatically rather than by comment.
- Extract the docker-compose migration instructions into a dedicated UPGRADING.md or a clearly-labeled section in the project README, and reference it from the compose comment with a one-liner instead of embedding the full procedure inline.

---

PR #101, opened 2026-04-20
feat: show GitHub review decision badge on PR list + detail

vbuenog:
Adds a second pill next to the existing severity badge showing the GitHub review decision the daemon recorded — sourced from GitHub itself, not derived locally from severity. The two answer different questions:

- **Severity** (existing): how bad is what the AI found?
- **Decision** (new): what did the daemon actually tell GitHub? (APPROVED, CHANGES_REQUESTED, COMMENTED, DISMISSED, PENDING)

Keeping both preserves the at-a-glance severity sort while surfacing the published outcome alongside it.

## Why not derive from severity

The daemon's `severityToEvent` maps severity to the submit event (`high` → REQUEST_CHANGES, else APPROVE), so deriving would look equivalent today. But:

- GitHub's response is authoritative for what was recorded. If the mapping ever changes — or if dismissals get tracked here later — the UI reflects that automatically rather than drifting.
- The response body always includes `state`; we were throwing it away. Capturing it is free at the wire level.

### `b37d962` — `feat(daemon): capture GitHub review state alongside review ID`

- `github.SubmitReview(...)` signature: `(int64, error)` → `(int64, string, error)`. Parses `state` from the response body alongside `id`.
- `store.Review.GitHubReviewState` added; schema column + ALTER migration (empty string default for legacy rows, so old reviews just don't render a badge).
- `store.UpdateGitHubReviewID` replaced with `MarkReviewPublished(reviewID, ghReviewID, ghReviewState)` — one call, two fields, same transaction. All three pipeline call-sites updated; the orphan-PR sentinel path passes `("", -1)` to silence retry noise without polluting the state.
- Pipeline fake `fakeGH.SubmitReview` now mirrors GitHub's event → state mapping so existing pipeline tests see realistic values.
- New `TestMarkReviewPublished_RoundTripsStateAndID` locks in the field's round-trip through the store.

### `75b2540` — `feat(web_ui): render review-decision badge sourced from GitHub`

- `types.ts` Review gains `github_review_state: string`.
- New `$lib/review.ts#decisionLabel(state)` — centralised state → pill mapping. Returns `null` for empty/unknown values so the badge simply doesn't render (legacy rows + orphan sentinel stay clean).
- New `ReviewDecisionBadge.svelte` — thin rendering wrapper. Exposes `data-state` for future E2E selectors.
- Wired into `PRTile.svelte` (dashboard tiles) and `PRReviewCard.svelte` (PR detail history).
- 9 new unit tests in `review.test.ts` — each mapped state, fallback cases, dark-mode variants, deliberate case-sensitivity (GitHub emits uppercase; lowercasing would mask upstream drift).

## Migration story

No backfill. Reviews that landed before this merge keep `github_review_state = ''` and show no decision badge. The next time the daemon reviews them (manual trigger or next poll cycle) the state will populate. Keeps us honest: the badge means "GitHub said this" — if it's blank, we don't know.

## Tests

- [x] `make test-docker` — all daemon packages pass, including the new store round-trip test.
- [x] `cd web_ui && npm run check` — 0 type errors.
- [x] `cd web_ui && npm test` — 90/90 (81 pre-existing + 9 new).
- [x] `cd web_ui && npm run lint` — prettier + eslint clean.
- [x] `cd web_ui && npm run build` — adapter-node bundle OK.

## Manual verification

After merge + rebuild:

```bash
# Trigger a fresh review so github_review_state populates.
curl -sS -X POST http://127.0.0.1:7842/prs/<id>/review \\
  -H "X-Heimdallm-Token: \$TOKEN"

# Wait a few seconds, then inspect the stored state:
curl -sS http://127.0.0.1:7842/prs/<id> -H "X-Heimdallm-Token: \$TOKEN" \\
  | jq '.reviews[0].github_review_state'
# Expected: "APPROVED" or "CHANGES_REQUESTED" (depending on severity).

# Open the PR list at http://localhost:3000/prs — decision pill
# should appear next to the severity pill on the reviewed PR.
```

## Stays out of scope

- PR B (config payload / named volume / reload tolerance) landed separately.
- No Flutter-side equivalent yet; the Dart `Review` model can add the same field when someone needs it there.
- No backfill of historical reviews — intentional, per above.

### DIFF
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -194,8 +194,12 @@ func (c *Client) fetchByQualifier(username, qualifier string, repos []string) ([
 
 // SubmitReview posts an AI-generated review to GitHub as a PR review.
 // event should be "REQUEST_CHANGES", "COMMENT", or "APPROVE".
-// Returns the GitHub review ID.
-func (c *Client) SubmitReview(repo string, number int, body, event string) (int64, error) {
+// Returns the GitHub review ID and the review state reported by the API —
+// typically "APPROVED", "CHANGES_REQUESTED", or "COMMENTED" depending on the
+// event and on GitHub's server-side rules. We pass the state through to the
+// store so the web UI can show a review-decision badge sourced from GitHub
+// rather than derived locally from severity.
+func (c *Client) SubmitReview(repo string, number int, body, event string) (int64, string, error) {
 	path := fmt.Sprintf("/repos/%s/pulls/%d/reviews", repo, number)
 
 	payload := map[string]any{
@@ -206,7 +210,7 @@ func (c *Client) SubmitReview(repo string, number int, body, event string) (int6
 	data, _ := json.Marshal(payload)
 	req, err := http.NewRequest("POST", c.baseURL+path, strings.NewReader(string(data)))
 	if err != nil {
-		return 0, fmt.Errorf("github: submit review: %w", err)
+		return 0, "", fmt.Errorf("github: submit review: %w", err)
 	}
 	req.Header.Set("Authorization", "Bearer "+c.token)
 	req.Header.Set("Accept", "application/vnd.github+json")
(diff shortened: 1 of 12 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured feature that threads GitHub's review state from the API response through the store to the web UI. The approach of storing the state verbatim rather than deriving it from severity is sound. Tests are thorough on both the Go and TypeScript sides. Only minor issues found.

### Issues

🟡 **web_ui/src/lib/components/PRTile.svelte:33** — Badge ordering is inconsistent between views: PRTile renders ReviewDecisionBadge before SeverityBadge, while PRReviewCard renders SeverityBadge before ReviewDecisionBadge. This will look inconsistent to users navigating between the list and detail views.
🟡 **daemon/internal/github/client.go:228** — Pre-existing: `resp.StatusCode != http.StatusOK && resp.StatusCode != 200` is a redundant check — http.StatusOK is 200. Not introduced by this PR but touched in this diff; worth a drive-by fix.
⚠️ **daemon/internal/pipeline/pipeline.go:272** — Errors from MarkReviewPublished are silently discarded (`_ = p.store.MarkReviewPublished(...)`). If the DB write fails, the in-memory `rev` struct is still updated with the new state, creating a divergence between memory and the database that will persist until the next restart. Consider at least logging the error. (This pattern was inherited from UpdateGitHubReviewID but now covers more data.)
🟡 **daemon/internal/github/client.go:233** — The `State` field is extracted from the GitHub API response without any validation. If the API returns an unexpected or empty string on a 200 response, the store will persist it and the badge will silently not render. Consider logging a warning when the state is empty or not in the known set, to surface API contract drift early.

### Suggestions

- Unify badge ordering: pick a consistent order (e.g., SeverityBadge first, then ReviewDecisionBadge) in both PRTile.svelte and PRReviewCard.svelte so the list and detail views feel cohesive.
- Log MarkReviewPublished errors instead of discarding them — even a slog.Warn would help diagnose issues where the badge doesn't appear despite a successful GitHub publish.
- Consider adding a brief comment or a Go-side constant list for the known GitHub review states (APPROVED, CHANGES_REQUESTED, COMMENTED, DISMISSED, PENDING) to make the contract between daemon and web UI explicit in one place.

---

PR #106, opened 2026-04-20
fix: per-repo IT implicit enable + LED consistency + PR metadata persistence

Muriano:
## Summary

Fixes multiple issues with per-repo issue tracking configuration:

### Daemon
- `IssueTrackingForRepo`: implicitly enables IT when per-repo labels are configured (no separate toggle needed)
- Poll cycle checks per-repo enabled status, not just global
- 4 new tests for merge behavior

### Flutter
- LED indicators: vertical layout, reflect labels-as-active (matches daemon implicit-enable)
- Toggles: `itEnabled`/`devEnabled` derived from labels in `fromJson` (consistent with LEDs)
- TOML writer: now persists `pr_reviewers`, `pr_assignee`, `pr_labels`, `pr_draft` per-repo
- TOML writer: PR metadata written BEFORE issue_tracking sub-table (TOML ordering requirement)
- `isMonitored` considers repos with labels as monitored
- `save()` restored to writeConfig + reload (PUT /config doesn't support per-repo overrides)

## Test Plan

- [x] `make test-docker` — all packages pass (7 IssueTrackingForRepo tests)
- [ ] Manual: configure PR reviewers/labels on a repo, restart app — verify they persist
- [ ] Manual: repo with labels shows green LEDs + ON toggles


sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

The PR fixes a UX pitfall where users configure per-repo labels but forget to set Enabled=true. However, it introduces a regression: repos can no longer explicitly disable issue tracking via per-repo override when the global setting is on, because `ov.Enabled = false` is now ignored rather than honored.

### Issues

🔴 **daemon/internal/config/config.go:317** — Regression: lost ability to explicitly disable issue tracking per-repo. Previously `merged.Enabled = ov.Enabled` allowed a repo to set `Enabled: false` to override a global `Enabled: true`. Now, if `ov.Enabled` is false and no labels are configured, `merged.Enabled` retains the global value (which may be true), so there is no way for a per-repo override to turn off issue tracking. Consider a tri-state or pointer (`*bool`) for `Enabled` to distinguish 'explicitly false' from 'not set'.
🟡 **daemon/internal/config/config.go:322** — The two branches (`if ov.Enabled` and `else if len(ov.DevelopLabels) > 0 || len(ov.ReviewOnlyLabels) > 0`) both assign `merged.Enabled = true`. This can be simplified to a single condition: `if ov.Enabled || len(ov.DevelopLabels) > 0 || len(ov.ReviewOnlyLabels) > 0`.
⚠️ **daemon/internal/config/config.go:317** — No test coverage is included in this diff. The original behavior was nuanced (per-repo override wins unconditionally); changing it without tests for both the new implicit-enable path and the disable-override path risks silent regressions.

### Suggestions

- Use a pointer type (`*bool`) or a tri-state enum for `IssueTrackingConfig.Enabled` so you can distinguish 'explicitly set to false' from 'not specified'. This preserves the implicit-enable-on-labels behavior while still allowing explicit opt-out per repo.
- Add unit tests covering: (1) global on + per-repo explicitly disabled, (2) global off + per-repo has labels but no Enabled flag, (3) global off + per-repo has Enabled=true, (4) global on + per-repo has no overrides.
- Simplify the two identical branches into a single `if` with an OR condition for readability.

---

Muriano:
## Response to AI Review

**Fixed (f67497e):**
- **Item 2 (simplify)**: Single OR condition.
- **Item 3 (no tests)**: 3 new tests — labels-imply-enabled, empty-override-disabled, global-on-inherits. Total 7 IssueTrackingForRepo tests.

**Documented, not changed:**
- **Item 1 (regression: can't disable per-repo)**: Valid observation. Documented as comment in the code. The `bool` zero value makes "explicitly false" indistinguishable from "not set". A `*bool` refactor would fix it but changes TOML parsing + JSON serialization — scope beyond this fix. No current user has global IT enabled + needs to exclude specific repos. If that use case materializes, the `*bool` path is clear.

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid fix for the implicit-enable and TOML ordering bugs. Three targeted issues remain: a daemon/Flutter parity gap when only develop_labels is configured (IT LED shows wrong state), devLedStatus now reports 'repo' even when no local dir mapping exists, and the isMonitored additions are partially redundant. The known can't-disable regression is correctly documented.

### Issues

⚠️ **flutter_app/lib/core/models/config_model.dart:174** — itLedStatus only checks reviewOnlyLabels for the implicit-active path, but the daemon enables IT when either ReviewOnlyLabels OR DevelopLabels is non-empty. A repo configured with only develop_labels will have the daemon report IT enabled while the Flutter LED falls through to 'global' or 'off'. Fix: add || (developLabels != null && developLabels!.isNotEmpty) to the labels check in itLedStatus.
⚠️ **flutter_app/lib/core/models/config_model.dart:418** — In AppConfig._parseRepos, itEnabled is derived as (itExplicit || hasReviewLabels) but does not include hasDevLabels. Since the daemon treats develop_labels as implicitly enabling IT (Enabled=true), the model should use (itExplicit || hasReviewLabels || hasDevLabels). With only develop_labels configured, itEnabled stays null and the derived LED state diverges from the daemon.
⚠️ **flutter_app/lib/core/models/config_model.dart:180** — devLedStatus removed the hasLocalDir guard from the devEnabled == true branch. A repo with devEnabled=true but no local_dir mapping now returns 'repo' (active LED) even though dev mode cannot function without a local dir. The old guard was intentional; this change may mislead users about feature availability.
🟡 **flutter_app/lib/core/setup/first_run_setup.dart:272** — pr_draft = true is only written when prDraft == true. If a future default flips to true globally and a per-repo override of false is needed, this will silently drop the override. Low risk given current defaults, but worth a comment or handling the false case explicitly.
🟡 **flutter_app/lib/core/models/config_model.dart:143** — The reviewOnlyLabels and developLabels conditions added to isMonitored are now largely redundant: _parseRepos already sets itEnabled=true / devEnabled=true when labels are present, so the earlier (itEnabled ?? false) and (devEnabled ?? false) clauses will already fire. Not harmful, but adds dead logic paths that will diverge if parsing logic changes.

### Suggestions

- In itLedStatus, mirror the daemon condition exactly: treat developLabels being non-empty as an implicit 'repo' return, matching the daemon's `len(ov.DevelopLabels) > 0` branch.
- In _parseRepos, derive itEnabled as (itExplicit || hasReviewLabels || hasDevLabels) ? true : null to keep Flutter and daemon semantics in sync.
- Restore the hasLocalDir guard in devLedStatus for the devEnabled == true branch, or add a comment explaining why showing 'repo' without a local dir is intentional.
- Consider a follow-up issue for the *bool refactor so the can't-disable regression has a tracked path forward before a user hits it.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

PR addresses previous feedback well (tests added, limitation documented). The remaining concerns are a behavioral inconsistency in devLedStatus and a potential round-trip data loss in TOML parsing. Overall quality is good.

### Issues

⚠️ **flutter_app/lib/core/models/config_model.dart:172** — devLedStatus: removing `hasLocalDir` from the `devEnabled == true` branch means the LED shows 'repo' (active) even when no local directory is mapped — but develop mode cannot function without one. The implicit-enable path (labels) still correctly gates on `hasLocalDir`. This inconsistency could mislead users into thinking develop mode is operational when it isn't. Consider restoring `&& hasLocalDir` on line 172, or removing it from the labels path too and handling 'configured but non-functional' separately.
🟡 **flutter_app/lib/core/models/config_model.dart:414** — TOML parsing collapses explicit `enabled = false` into `null` (not set). If a user explicitly sets `enabled = false` in TOML, it parses as itEnabled=null. On next save, that null won't emit `enabled = false` to the file, silently dropping the user's intent. This is the same bool zero-value limitation as the Go side, but the Flutter round-trip (read → save → read) can lose data. Worth a code comment at minimum so the behavior is documented here too.

### Suggestions

- Add a unit test for devLedStatus verifying the LED state when devEnabled=true but hasLocalDir=false — this documents the intended behavior either way.
- Consider adding a brief comment near the TOML parsing (line 414) noting the round-trip limitation for explicit false, mirroring the Go-side documentation.

---

PR #114, opened 2026-04-20
feat: multi-tier polling pipeline (#107)

### LINKED ISSUE #107: feat: multi-tier polling — global, per-repo, and per-issue/PR cycles
## Summary

Replace the single monolithic poll cycle with a multi-tier architecture where each tier watches for different types of changes at the appropriate granularity.

## Motivation

Currently, one global timer (poll_interval) runs everything sequentially: PR fetching, issue promotion, and issue pipeline processing for all repos. This doesn't scale well and misses reactive scenarios (e.g., a change request comment should trigger re-review faster than waiting for the next full poll).

### Tier 1: Global (slow cycle, e.g. 15m)
- **Repo discovery** — topic-based auto-discovery of new repos (already exists via discovery_interval)
- **Issue/PR discovery** — find new issues and PRs across all monitored repos
- **Repo metadata refresh** — labels, collaborators cache refresh

### Tier 2: Per-repo (medium cycle, e.g. 5m)
- **New issues to triage** — fetch open issues for a specific repo, classify, run review_only pipeline
- **New PRs to review** — fetch review-requested PRs for a specific repo
- **Issue promotion** — check if blocked issues can be promoted (dependencies closed)
- **Develop candidates** — issues with develop labels that haven't been auto-implemented yet

### Tier 3: Per-issue / Per-PR (fast cycle, e.g. 1m, or event-driven)
- **State transitions** — issue moved from triage to develop (label change)
- **Change request comments** — someone requested changes on a Heimdallm-generated PR → re-review
- **New comments on triaged issues** — context changed, may need re-triage
- **PR merge/close** — update internal state, clean up
- **Issue close** — stop processing, update dependents

## Design Considerations

- Each tier should have its own configurable interval
- Tiers should be independent — a slow Tier 1 cycle shouldn't block Tier 3 reactivity
- Consider GitHub webhooks as an alternative to Tier 3 polling (requires inbound HTTP, may not work for desktop deployments)
- Rate limiting budget should be distributed across tiers, not shared equally
- Per-repo cycles could be staggered to avoid API burst

## Configuration (tentative)

```toml
[github]
# Tier 1: global discovery
discovery_interval = "15m"

# Tier 2: per-repo processing
poll_interval = "5m"

# Tier 3: per-item reactivity
watch_interval = "1m"
```

## Out of Scope (for now)

- Webhook support (could replace Tier 3 in server deployments)
- Per-repo poll intervals (all repos share the tier interval)

## Notes

This is an architectural change to the daemon's core scheduling. The current single-loop design in `main.go/makePollFn` would be split into separate scheduler instances or a priority-based task queue.

Muriano:
## Summary

Replaces the monolithic poll cycle with a 3-tier pipeline architecture:

- **Tier 1 (Discovery, 15m)**: discovers repos via topic + PR activity, sends to Tier 2 via channel
- **Tier 2 (Per-repo, 5m)**: fetches PRs, processes issues, runs promotion, enqueues active items to Tier 3
- **Tier 3 (Per-item watch, 1m base + backoff)**: watches active items for state changes (label changes, new comments, merge/close) with exponential backoff (1m → 2m → 4m → 8m → 15m cap, evict after 1h inactivity)

### Architecture
```
Tier 1 → repos channel → Tier 2 → WatchQueue → Tier 3 (backoff)
```

Shared rate limiter (4500 req/h) with priority: T3 > T2 > T1.

### New files in `daemon/internal/scheduler/`
- `ratelimit.go` — token bucket with tier-based priority wait times
- `queue.go` — priority queue (container/heap) with exponential backoff
- `tier1.go` — discovery goroutine
- `tier2.go` — per-repo processing goroutine
- `tier3.go` — per-item watch goroutine
- `pipeline.go` — orchestrator, wires channels, manages shutdown

### main.go changes
- Removed `makePollFn`, `startScheduler`, `startDiscovery` (~238 lines)
- Added `tier2Adapter` bridging concrete types to scheduler interfaces (~286 lines)
- Simplified reload: stop pipeline → update config → start new pipeline

### Config
- New: `watch_interval` (default "1m") + `HEIMDALLM_WATCH_INTERVAL` env var

## Test Plan
- [x] `make test-docker` — all 14 packages pass
- [x] `go build` — compiles cleanly
- [x] Rate limiter: 4 tests (budget, priority, cancellation, refill)
- [x] Queue: 8 tests (push/pop, dedup, backoff, cap, reset, evict, ordering)
- [ ] Manual: start daemon, verify logs show "pipeline: started" with 3 tier intervals
- [ ] Manual: verify Tier 3 watches active items and detects changes

Closes #107

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -180,7 +180,8 @@ func main() {
 	srv := server.New(s, broker, p, apiToken)
 
 	// cfgMu protects cfg and the pipeline so reload is safe from any goroutine.
-	var cfgMu sync.Mutex
+	var cfgMu   sync.Mutex
+	var reloadMu sync.Mutex // serialises config reloads to prevent duplicate pipelines
 
 	// discoverySvc holds the discovered repo cache.
 	discoverySvc := discovery.NewService(ghClient)
@@ -415,6 +416,13 @@ func main() {
 	// startup so the two paths cannot drift on which loader they pick — both
 	// see the same HEIMDALLM_DATA_DIR snapshot.
 	srv.SetReloadFn(func() error {
+		// Serialise reloads: without this, two concurrent /reload calls
+		// could each read the same oldPipe, both stop it, both build a
+		// new pipeline, and leave two pipelines running against the same
+		// GitHub API budget.
+		reloadMu.Lock()
+		defer reloadMu.Unlock()
+
 		newCfg, err := loadConfig(cfgPath)
 		if err != nil {
 			return fmt.Errorf("reload: %w", err)
@@ -736,9 +744,14 @@ func parseDiscoveryInterval(s string) time.Duration {
 }
 
 func parseWatchInterval(s string) time.Duration {
+	const minWatch = 10 * time.Second
 	d, err := time.ParseDuration(s)
 	if err != nil || d <= 0 {
 		return 1 * time.Minute
+	}
+	if d < minWatch {
+		slog.Warn("watch_interval too small, clamping to minimum", "configured", d, "minimum", minWatch)
+		return minWatch
 	}
 	return d
 }
@@ -977,7 +990,15 @@ func (a *tier2Adapter) HandleChange(ctx context.Context, item *scheduler.WatchIt
 		}
 		a.runReview(ghPR, aiCfg)
 	}
-	// For issues, Tier 2 will pick up the change on its next cycle.
+	// For issues we cannot trigger an immediate re-triage from Tier 3,
+	// but returning nil signals the caller (RunTier3) to call
+	// queue.ResetBackoff — which resets the item's backoff to 1m so it
+	// gets re-checked quickly rather than waiting at a potentially 15m
+	// backoff for Tier 2's next full cycle.
+	if item.Type == "issue" {
+		slog.Info("tier3: issue change detected, backoff will reset",
+			"repo", item.Repo, "number", item.Number)
+	}
 	return nil
 }
 
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -325,7 +325,16 @@ func main() {
 
 	pipe := buildPipeline(cfg)
 	pipe.Start(context.Background())
-	defer pipe.Stop()
+	// Use a closure so the defer reads the current pipe variable at shutdown
+	// time, not the initial pointer captured at defer-statement time. After a
+	// reload, pipe points to a new pipeline — the bare defer would stop the
+	// already-halted original pipeline and leak the post-reload one.
+	defer func() {
+		cfgMu.Lock()
+		p := pipe
+		cfgMu.Unlock()
+		p.Stop()
+	}()
 
 	// Expose live config for GET /config
 	srv.SetRepoMetaFns(ghClient.FetchLabels, ghClient.FetchCollaborators)
@@ -418,17 +427,26 @@ func main() {
 			return fmt.Errorf("reload: %w", err)
 		}
 
-		// Stop the old pipeline before starting a new one. The Search API
-		// rate limit (30 req/min) is tight enough that running two loops in
-		// parallel — even briefly during reload — risks throttling.
-		pipe.Stop()
-
+		// Read the current pipe under cfgMu, then stop it OUTSIDE the lock.
+		// Holding cfgMu across Stop() risks deadlock: if Stop() blocks
+		// waiting for in-flight goroutines that also acquire cfgMu (e.g.
+		// tier2Adapter callbacks), both sides block forever.
 		cfgMu.Lock()
-		cfg = newCfg
+		oldPipe := pipe
 		cfgMu.Unlock()
 
-		pipe = buildPipeline(newCfg)
-		pipe.Start(context.Background())
+		oldPipe.Stop()
+
+		newPipe := buildPipeline(newCfg)
+
+		// Swap cfg + pipe atomically under the lock so readers never see
+		// a half-updated state.
+		cfgMu.Lock()
+		cfg = newCfg
+		pipe = newPipe
+		cfgMu.Unlock()
+
+		newPipe.Start(context.Background())
 		return nil
 	})
 
@@ -738,8 +756,6 @@ type tier2Adapter struct {
 	cfg       **config.Config
 	loginMu   *sync.Mutex
 	login     *string
-	reviewMu  *sync.Mutex
-	inFlight  map[int64]bool
 	runReview func(pr *gh.PullRequest, aiCfg config.RepoAI)
 }
 
@@ -760,6 +776,10 @@ func (a *tier2Adapter) FetchPRsToReview() ([]scheduler.Tier2PR, error) {
 			ID:        pr.ID,
 			Number:    pr.Number,
 			Repo:      pr.Repo,
+			Title:     pr.Title,
+			HTMLURL:   pr.HTMLURL,
+			Author:    pr.User.Login,
+			State:     pr.State,
 			UpdatedAt: pr.UpdatedAt,
 		})
 	}
@@ -773,13 +793,14 @@ func (a *tier2Adapter) ProcessPR(ctx context.Context, pr scheduler.Tier2PR) erro
 	aiCfg := c.AIForRepo(pr.Repo)
 	a.cfgMu.Unlock()
 
-	// Reconstruct the full gh.PullRequest from the minimal Tier2PR.
-	// FetchPRsToReview already resolved the repo and set UpdatedAt; we
-	// need at minimum ID, Number, Repo, UpdatedAt for the pipeline.
 	ghPR := &gh.PullRequest{
 		ID:        pr.ID,
 		Number:    pr.Number,
 		Repo:      pr.Repo,
+		Title:     pr.Title,
+		HTMLURL:   pr.HTMLURL,
+		User:      gh.User{Login: pr.Author},
+		State:     pr.State,
 		UpdatedAt: pr.UpdatedAt,
 	}
 	a.runReview(ghPR, aiCfg)
@@ -882,7 +903,11 @@ func (a *tier2Adapter) PromoteReady(ctx context.Context, repos []string) (int, e
 	globalIT := c.GitHub.IssueTracking
 	a.cfgMu.Unlock()
 
-	if !globalIT.Enabled || len(globalIT.BlockedLabels) == 0 {
+	// Promotion only makes sense when blocked labels are configured.
+	// Intentionally NOT gated on globalIT.Enabled — per-repo IT configs
+	// can enable issue tracking independently while global is disabled.
+	// Gating on Enabled here would silently regress promotion for those users.
+	if len(globalIT.BlockedLabels) == 0 {
 		return 0, nil
 	}
 	return issuepipeline.PromoteReady(ctx, a.ghClient, globalIT, repos, a.broker)
@@ -921,15 +946,34 @@ func (a *tier2Adapter) CheckItem(ctx context.Context, item *scheduler.WatchItem)
 // HandleChange implements scheduler.Tier3ItemChecker.
 func (a *tier2Adapter) HandleChange(ctx context.Context, item *scheduler.WatchItem) error {
 	if item.Type == "pr" {
+		// Apply the same dedup gate as Tier 2: skip dismissed PRs and
+		// recently-reviewed PRs. Note: there is a small TOCTOU window
+		// between this check and the actual runReview call — the worst
+		// case is an occasional harmless duplicate review.
+		if a.PRAlreadyReviewed(item.GithubID, item.LastSeen) {
+			slog.Debug("tier3: PR already reviewed, skipping", "pr", item.Number, "repo", item.Repo)
+			return nil
+		}
+
 		a.cfgMu.Lock()
 		c := *a.cfg
 		aiCfg := c.AIForRepo(item.Repo)
 		a.cfgMu.Unlock()
 
+		// Fetch the full PR from the store so runReview receives all
+		// fields (Title, Author, URL, etc.) instead of a sparse struct.
+		stored, _ := a.store.GetPRByGithubID(item.GithubID)
 		ghPR := &gh.PullRequest{
 			ID:     item.GithubID,
 			Number: item.Number,
 			Repo:   item.Repo,
+		}
+		if stored != nil {
+			ghPR.Title = stored.Title
+			ghPR.HTMLURL = stored.URL
+			ghPR.User = gh.User{Login: stored.Author}
+			ghPR.State = stored.State
+			ghPR.UpdatedAt = stored.UpdatedAt
 		}
 		a.runReview(ghPR, aiCfg)
 	}
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -179,15 +179,11 @@ func main() {
 	issueFetcher := issuepipeline.NewFetcher(ghClient, s, issuePipe)
 	srv := server.New(s, broker, p, apiToken)
 
-	// cfgMu protects cfg, sched and the discovery loop so reload is safe from any goroutine.
+	// cfgMu protects cfg and the pipeline so reload is safe from any goroutine.
 	var cfgMu sync.Mutex
-	var sched *scheduler.Scheduler
 
-	// discoverySvc holds the discovered repo cache; it is nil when topic-based
-	// discovery is disabled. discoveryCancel stops the background loop so reload
-	// can restart it with fresh config.
+	// discoverySvc holds the discovered repo cache.
 	discoverySvc := discovery.NewService(ghClient)
-	var discoveryCancel context.CancelFunc
 
 	// loginMu guards cachedLogin against concurrent reads/writes from the
 	// poll cycle and HTTP goroutines.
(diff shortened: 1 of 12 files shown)


vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Solid architectural refactor introducing a 3-tier polling pipeline with clean abstractions and a priority-queue watch layer. The core design is sound, but there are two behavioral regressions, a data race on the pipe variable during reload, and a subtle test bug that silently passes despite not exercising the intended behavior.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:384** — The `pipe` variable is read and written in the reload closure without holding any lock. If two concurrent reloads race (e.g., rapid successive HTTP /reload calls), one goroutine could call pipe.Stop() on the already-replaced pipeline while another is starting a new one. cfgMu guards `cfg` but not `pipe`.
🔴 **daemon/cmd/heimdallm/main.go:554** — The initial poll (go makePollFn(cfg)()) was removed with no equivalent in the new pipeline. Tier 2 waits a full PollInterval (default 5m) before first run. Similarly, the reload path no longer fires an immediate poll with the new config. This is a behavioral regression — PRs and issues won't be checked until the first ticker fires.
🔴 **daemon/cmd/heimdallm/main.go:858** — HandleChange for type 'pr' calls runReview directly without going through PRAlreadyReviewed. The Tier 2 path gates on dismissed status and the 30-second grace period; Tier 3 skips both checks entirely, risking duplicate or unnecessary re-reviews triggered by minor GitHub metadata updates (e.g., label syncs).
⚠️ **daemon/internal/scheduler/queue_test.go:54** — TestWatchQueue_BackoffDoubles calls q.PopReady() to 'drain' but the item's NextCheck is now+2m so PopReady returns nothing and the item stays in the queue with seen[100]=true. The second ReEnqueue is therefore a no-op (dedup check fires) yet still mutates item.Backoff before the check. The test asserts item.Backoff==4m which is true only because the field is mutated unconditionally before the dedup guard, masking that the item was never actually re-enqueued. The test should push with a past NextCheck so PopReady actually removes the item.
⚠️ **daemon/cmd/heimdallm/main.go:683** — tier2Adapter carries reviewMu and inFlight fields that are never read or written in any adapter method. If they are only used inside the runReview closure captured externally, storing them in the struct is misleading and creates confusion about ownership. Either use them in the adapter or remove them from the struct.
⚠️ **daemon/cmd/heimdallm/main.go:856** — HandleChange for issues is explicitly a no-op ('Tier 2 will pick up the change on its next cycle'). This means Tier 3 performs an API check, detects a change, and then does nothing actionable for issues — the change won't be processed for up to PollInterval (5m). The primary benefit of Tier 3 (low latency on active items) is lost for issues.
⚠️ **daemon/cmd/heimdallm/main.go:683** — Using **config.Config (pointer to pointer) to share config state between main and the adapter is fragile. If cfg is ever reassigned to a stack-allocated value or the outer pointer changes scope, the double indirection silently breaks. A sync/atomic.Pointer[config.Config] or a callback function would be safer and idiomatic.
🟡 **daemon/cmd/heimdallm/main.go:780** — ProcessPR accepts a context.Context but never passes it to runReview. If the pipeline is stopped mid-review (e.g., during reload), the in-flight review cannot be cancelled. This was not possible before (reviews were fire-and-forget goroutines), but the new interface contract implies cancellability.
🟡 **daemon/cmd/heimdallm/main.go:800** — ProcessRepo re-derives anyITEnabled on every per-repo call by iterating all c.AI.Repos under cfgMu. The scheduler already decides which repos to process; this per-call re-check under the mutex adds contention proportional to the number of configured repos and is redundant with the caller's filtering.
🟡 **daemon/internal/scheduler/pipeline.go:72** — reposChan is buffered at 1. If Tier 2 is slow processing a large repo list, Tier 1's next discovery run will block on the send until Tier 2 drains the channel. This effectively serializes discovery and polling, which may be intentional but is not documented. A non-blocking send with a drop-and-log fallback would be safer.

### Suggestions

- Protect the `pipe` variable with cfgMu (or a dedicated pipelineMu) in the reload closure to close the data race window, even if concurrent reloads are unlikely in practice.
- Add an immediate first cycle trigger after pipeline Start — either by having RunTier2 fire once before the first ticker, or by exposing a Trigger() method on Pipeline that the caller invokes after Start() and after reload.
- Route Tier 3 PR changes through PRAlreadyReviewed before calling runReview, matching the Tier 2 gate, to prevent duplicate reviews from minor GitHub metadata touches.
- Fix TestWatchQueue_BackoffDoubles to use a past NextCheck on the initial Push so PopReady actually removes the item and the second ReEnqueue exercises the real code path.
- For issue HandleChange, consider immediately queuing the repo for a Tier 2 scan (e.g., via a priority channel) rather than waiting for the next full poll cycle — otherwise Tier 3 watch provides no latency benefit for issues.
- Replace **config.Config with sync/atomic.Pointer[config.Config] for safer, lock-free config sharing between main and the adapter.

---

Muriano:
## Response to AI Review

**Fixed (c425a75):**
- **Item 1 (data race on pipe)**: `pipe.Stop()` / `pipe = buildPipeline()` / `pipe.Start()` now wrapped in `cfgMu.Lock/Unlock` in the reload closure
- **Item 2 (no immediate first poll)**: Tier 2 now blocks on `reposChan` for Tier 1's initial send, then runs `processTick()` immediately. Replaced the `time.Sleep(2s)` hack.
- **Item 3 (Tier 3 HandleChange skips dedup)**: Added `PRAlreadyReviewed` check before calling `runReview` for PR-type watch items
- **Item 4 (broken test)**: `TestWatchQueue_BackoffDoubles` rewritten with two separate queues to properly exercise the dedup path
- **Item 5 (unused struct fields)**: Removed `reviewMu` and `inFlight` from `tier2Adapter`
- **Item 6 (reposChan buffer)**: Added comment documenting the buffer-1 semantics

**Not changing:**
- **Item 6 (issue HandleChange no-op)**: Correct observation — Tier 3 detects issue changes but Tier 2 processes them. Adding a priority re-queue channel is a good idea but adds complexity beyond this PR. Opened as a follow-up concern.
- **Item 7 (**config.Config)**: The pattern matches the existing codebase (cfgMu + pointer swap). An atomic.Pointer refactor would touch the entire config sharing model — out of scope.
- **Item 8 (context not passed to runReview)**: `runReview` predates context support. Adding it requires changes to the PR pipeline interface. Follow-up.
- **Item 9 (redundant anyITEnabled check)**: It's a safety guard — cheap under mutex. Removing it saves microseconds but risks processing issues for a disabled repo if the caller's filtering changes.
- **Item 10 (reposChan blocking)**: Now documented. Non-blocking send would lose repo updates silently — blocking is the safer default.

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Well-structured refactor that cleanly separates polling concerns into a 3-tier pipeline. The queue and rate-limiter abstractions are solid and well-tested. Two high-severity issues remain: a stale pointer in `defer pipe.Stop()` leaks the post-reload pipeline on daemon exit, and there is an unguarded data race on the `pipe` variable itself. A behavior regression in `PromoteReady` may silently break dependency promotion for users with per-repo-only issue tracking.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:302** — `defer pipe.Stop()` evaluates the receiver at declaration time, capturing the initial *Pipeline pointer. After a config reload `pipe` is reassigned to a new Pipeline, but the defer still holds the old pointer—so the post-reload pipeline is never stopped on daemon exit. Fix: `defer func() { pipe.Stop() }()` (with proper mutex protection around the read).
🔴 **daemon/cmd/heimdallm/main.go:393** — Data race on the `pipe` variable: the reload handler writes `pipe = buildPipeline(newCfg)` from an HTTP goroutine while the deferred Stop reads `pipe` from the main goroutine. `cfgMu` guards `cfg` but nothing guards `pipe` itself. This is a concurrent read/write of a pointer with no synchronisation.
⚠️ **daemon/cmd/heimdallm/main.go:825** — `PromoteReady` now returns early when `!globalIT.Enabled`, but the old code ran promotion whenever `len(globalIT.BlockedLabels) > 0` regardless of `Enabled`. Users who rely on per-repo issue tracking (global `enabled=false`, per-repo `enabled=true`) will silently lose dependency promotion—a quiet regression.
⚠️ **daemon/cmd/heimdallm/main.go:870** — `HandleChange` for `item.Type == "issue"` is a complete no-op: Tier 3 detects the change, does nothing, and the item is simply not re-enqueued. This means the detected activity is discarded and Tier 2 won't pick it up until its next scheduled poll cycle, nullifying the latency benefit of Tier 3 watching for issues.
🟡 **daemon/internal/scheduler/queue.go:100** — `ReEnqueue` mutates `item.Backoff` and `item.NextCheck` before acquiring the lock, then silently drops the item if it is already in `seen` (e.g., error-path double re-enqueue). The mutation has already happened, leaving the item with a doubled backoff but no queue entry and no log line.
🟡 **daemon/internal/scheduler/pipeline.go:85** — Rate-limiter refill fires on a rolling 1-hour `time.Ticker` anchored to daemon start time, not to GitHub's actual quota reset (top of each UTC hour). After exhaustion near the boundary the daemon may wait up to ~59 extra minutes for a refill that GitHub already granted.

### Suggestions

- Guard `pipe` with `cfgMu` (or a dedicated mutex) at every site—initial assignment, reload write, and deferred Stop read—to eliminate the data race: `defer func() { cfgMu.Lock(); p := pipe; cfgMu.Unlock(); p.Stop() }()`.
- Add a `sync.Once` inside `Pipeline.Stop()` to make it idempotent; the current reload path stops the old pipeline once in the reload handler and the stale defer would call it again, making a safety guard worthwhile.
- Restore the pre-refactor `PromoteReady` guard (`len(globalIT.BlockedLabels) > 0` only, without the `globalIT.Enabled` gate) or add an explicit release note documenting that global `enabled=false` now also disables promotion.
- For Tier 3 issue-type changes in `HandleChange`, at minimum call `q.ResetBackoff(item)` and re-enqueue so the item is rechecked soon, or introduce a lightweight priority channel to wake Tier 2 immediately rather than waiting a full poll interval.
- Consider computing the time-until-next-GitHub-reset on startup (next top-of-hour UTC) and using that as the first ticker duration, then switching to `time.Hour` ticks, to keep the limiter roughly in phase with GitHub's actual quota.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

Ambitious refactor into a 3-tier pipeline with good test coverage on the queue primitives. However, there are several real correctness concerns: a heap-invariant violation in ReEnqueue when an item is already enqueued, a deferred Stop that captures the original pipe pointer (so a post-reload pipe leaks on shutdown), a discrepancy between the author's 'fix' claims and the actual diff (HandleChange still skips dedup, pipe swap is not inside cfgMu), and reconstructed gh.PullRequest structs missing fields runReview likely needs.

### Issues

🔴 **daemon/internal/scheduler/queue.go:73** — ReEnqueue mutates item.Backoff and item.NextCheck BEFORE checking q.seen. If the item is already in the heap, the function returns early — but NextCheck on the live heap entry has already been mutated, corrupting the heap invariant (Less reads NextCheck from the pointer). This can cause out-of-order pops and wrong scheduling. The test TestWatchQueue_BackoffDoubles only passes because it asserts on the mutated field, not on actual queue ordering. Fix: check seen first, or take a lock and only mutate if actually reinserting.
🔴 **daemon/cmd/heimdallm/main.go:296** — `defer pipe.Stop()` evaluates the receiver `pipe` at defer-statement time, not when the deferred call runs. After a reload replaces `pipe = buildPipeline(newCfg)`, the deferred Stop still targets the original pipeline — which was already stopped during reload — while the current pipeline leaks on shutdown (goroutines never canceled, no clean flush). Wrap in `defer func() { pipe.Stop() }()` so the current variable is read at execution time.
⚠️ **daemon/cmd/heimdallm/main.go:390** — Reload claims pipe.Stop / pipe = buildPipeline / pipe.Start are wrapped in cfgMu, but the diff shows Stop() and the new assignment+Start() outside the lock. Only `cfg = newCfg` is protected. Since the adapter closures read pipe indirectly via cfg only, this may not race today, but the inconsistency between the PR comment and the code is concerning and leaves `pipe` unsynchronized.
⚠️ **daemon/cmd/heimdallm/main.go:870** — tier2Adapter.HandleChange for PR items calls a.runReview directly without checking PRAlreadyReviewed / Dismissed — contradicting the author's Item 3 'fix' claim. Tier 3 activity can therefore trigger reviews on PRs the user dismissed, or immediately after a fresh review (no 30s grace). Add the same dedup Tier 2 uses.
⚠️ **daemon/cmd/heimdallm/main.go:770** — ProcessPR and HandleChange reconstruct a gh.PullRequest using only ID/Number/Repo/UpdatedAt (HandleChange omits UpdatedAt). Previously runReview received the full PR from FetchPRsToReview with fields like Title, Body, Head/Base refs, User, Draft, etc. If runReview or downstream AI prompt building reads those fields, they will be zero/empty. Either fetch the full PR before review or pass the original object through.
⚠️ **daemon/internal/scheduler/queue.go:110** — ResetBackoff has the same early-return-after-mutation pattern as ReEnqueue: NextCheck/LastSeen on the live item are mutated, then the function silently returns if seen[] is set. Same heap-invariant risk.
🟡 **daemon/cmd/heimdallm/main.go:820** — tier2Adapter.ProcessRepo re-computes anyITEnabled per repo by iterating c.AI.Repos — this duplicates work every tick per repo. Move the global/any-enabled decision up to Tier 2's repo loop (or to a cached flag refreshed on reload) rather than doing it inside each adapter call.
🟡 **daemon/internal/scheduler/queue.go:128** — Evict rebuilds the heap via heap.Init even when no items were removed beyond rewriting the slice. Correct, but O(n) per call; if Evict is invoked frequently on a large queue this churns. Minor — just worth documenting call cadence.

### Suggestions

- In ReEnqueue/ResetBackoff, take the lock first and only mutate item fields when the item is actually going to be pushed; if already seen, either update via heap.Fix(index) to repair ordering, or leave the item alone.
- Replace `defer pipe.Stop()` with `defer func() { pipe.Stop() }()` so shutdown always stops the latest pipeline after a reload.
- Make HandleChange reuse the PRAlreadyReviewed check (and a 30s grace) before invoking runReview, matching Tier 2 semantics.
- Either fetch the full PR in ProcessPR/HandleChange (ghClient.GetPullRequest) or thread the original *gh.PullRequest through Tier 2PR so runReview keeps receiving a fully-populated struct.
- Consider moving the pipe swap fully inside cfgMu (Stop + assign + Start) to match the stated invariant, or explicitly document that pipe is only touched by the single reload goroutine and deferred shutdown.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Author has addressed all critical and major findings from the first review. The multi-tier pipeline architecture is clean. Two remaining concerns: one about lossy PR reconstruction in the adapter layer, one pre-existing silent error from the prior review that remains unaddressed.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:768** — ProcessPR and HandleChange reconstruct a gh.PullRequest with only 4 fields (ID, Number, Repo, UpdatedAt) — HandleChange omits even UpdatedAt. The original code passed the full PR struct from FetchPRsToReview to runReview. If runReview or the downstream review pipeline reads fields beyond these four (Title, Author, HTMLURL, HeadRef, etc.), they will be zero-valued and silently produce incomplete reviews or broken GitHub links. Either widen Tier2PR to carry all fields runReview needs, or document/assert that runReview re-fetches the full PR internally.
🟡 **daemon/internal/scheduler/queue.go:88** — ReEnqueue and ResetBackoff mutate item fields (Backoff, NextCheck, LastSeen) outside the queue's mutex. This is safe today because Tier 3 processes items sequentially, but the WatchQueue API presents as a general-purpose thread-safe queue. A concurrent caller (e.g. a future Tier 2 calling ResetBackoff while Tier 3 calls ReEnqueue on the same item) would race on the item's fields. Consider either documenting the single-writer constraint or moving the mutations inside the lock.

### Suggestions

- Consider storing the full gh.PullRequest in Tier2PR (or a dedicated struct) so the adapter doesn't need to reconstruct a sparse object — this avoids a class of silent-zero-value bugs as the review pipeline evolves.
- Add a brief doc comment on WatchQueue noting the expected concurrency model (single writer for item mutation, queue lock only protects the heap and seen map).

---

Muriano:

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

Solid refactor from single-tick scheduler to 3-tier pipeline with rate limiting and watch queue. Architecture is cleanly separated behind interfaces and the WatchQueue has good test coverage. Main concerns: information loss when rebuilding gh.PullRequest from Tier2PR (only 4 fields preserved — downstream reviewers may rely on Title/Body/Head/Base/Author), lack of mutex protection on the reload-closure assignment to pipe variable, and residual unused struct fields. Author claims several follow-ups were fixed in c425a75/ae16c1e but the diff under review still shows the original issues.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:390** — Data race on `pipe` variable in reload closure: `pipe.Stop()` / `pipe = buildPipeline(newCfg)` / `pipe.Start(...)` are performed outside `cfgMu`, while startup assigns `pipe` in the main goroutine. Concurrent reload calls — or a reload racing with the deferred `pipe.Stop()` — can read a half-assigned pointer. Author claims fixed in c425a75; verify the fix landed.
🔴 **daemon/cmd/heimdallm/main.go:740** — `ProcessPR` reconstructs `gh.PullRequest` from only `{ID, Number, Repo, UpdatedAt}`. The previous code passed the full PR object (with Title, Body, Head, Base, Author, Labels, etc.) to `runReview`. If `runReview` or its pipeline dependencies read any other field (e.g. to build the review prompt or decide eligibility), the behavior silently changes post-refactor. Either re-fetch the full PR inside `runReview` or widen `Tier2PR` to carry all required fields.
🔴 **daemon/cmd/heimdallm/main.go:895** — `HandleChange` for Tier 3 PR items calls `runReview` without invoking `PRAlreadyReviewed`, bypassing the dismissed-PR and 30s-grace checks that Tier 2 enforces. A watched PR that was dismissed by the user will still be re-reviewed whenever Tier 3 detects activity. Author says fixed in c425a75 — confirm.
🟡 **daemon/cmd/heimdallm/main.go:720** — `tier2Adapter` still carries `reviewMu` and `inFlight` fields that are never read inside the adapter (concurrency is enforced inside `runReview`). Dead state increases surface area for future bugs; author said removed in c425a75 but the diff still shows them.
⚠️ **daemon/internal/scheduler/queue.go:88** — `ReEnqueue` mutates `item.Backoff` and `item.NextCheck` before acquiring `q.mu`. Any concurrent reader of the item (e.g. Tier 3 still logging fields from a recent pop) races. Safer to update the item under the lock, or document the contract that callers must own the item exclusively between Pop and ReEnqueue.
🟡 **daemon/internal/scheduler/queue.go:110** — `Evict` iterates `q.items` as a flat slice (heap underlying storage). When `evicted > 0`, the code reassigns `q.items = keep` and re-initializes the heap, but `keep` can accumulate in non-heap order, which is fine because of heap.Init — however `WatchItem.index` on kept items is stale until `Swap` rebuilds it during heapify. Since `index` is only used by heap ops themselves, this is correct but brittle; a short comment would help future maintainers.
🟡 **daemon/internal/scheduler/pipeline.go:74** — `RateLimitPerHour: 4500` is hard-coded in main.go (buildPipeline) and not exposed via config. GitHub REST tokens have 5000/hr; Search API is 30/min. A single fleet-wide 4500 budget is a guess — worth either exposing to `GitHubConfig` or adding a TODO. Also: the refiller replaces the budget once per hour but the first refill fires 1h after start, so the initial budget must be preloaded in `NewRateLimiter` (not visible in diff — worth double-checking).
🟡 **daemon/cmd/heimdallm/main.go:700** — `cfg **config.Config` (double pointer) lets the adapter observe config swaps, but every call site does `c := *a.cfg` under `cfgMu` then reads through `c`. If `cfg` is swapped between reads on different fields of the same logical operation, the operation sees a mix of old and new config. Low-risk in practice because swaps are rare, but an `atomic.Pointer[config.Config]` would both simplify the code and eliminate this window. Flagged only as a future improvement; author has already punted on this.

### Suggestions

- Add a smoke test that starts the Pipeline, injects one PR via a fake Tier2PRFetcher, and asserts runReview was called with all non-empty fields the old code depended on — this would have caught the Tier2PR information loss.
- Document the Tier 3 HandleChange contract (what dedup/filter checks it must share with Tier 2) to avoid drift.
- Expose `RateLimitPerHour` and the rate-limiter tier weights via `GitHubConfig` for operators who hit unusual traffic shapes.
- Consider a single `atomic.Pointer[config.Config]` in place of the `cfgMu` + `**config.Config` pattern — removes most of the lock/unlock noise in the adapter.
- Add an integration test for pipeline reload that exercises `Stop → buildPipeline → Start` under concurrent HTTP traffic to verify no goroutine leak or double-start.

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Solid architectural refactor introducing a clean 3-tier pipeline with good interface abstractions, backoff/eviction logic, and unit tests. Three issues need attention before merge: a defer captures the wrong pipeline pointer (goroutine leak on reload), a PromoteReady regression silently drops per-repo promotion, and the reload handler blocks indefinitely on in-flight AI reviews.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:0** — `defer pipe.Stop()` evaluates the *Pipeline pointer immediately at the defer statement, not when shutdown occurs. After reload reassigns `pipe` to a new pipeline, the deferred call stops the already-halted original pipeline while the post-reload pipeline's goroutines run unmanaged until process exit. Fix: `defer func() { pipe.Stop() }()` so the closure reads the current `pipe` variable at shutdown time.
⚠️ **daemon/cmd/heimdallm/main.go:0** — `tier2Adapter.PromoteReady` returns early when `!globalIT.Enabled`, even if individual repos have `IssueTracking.Enabled` with `BlockedLabels` set. The original code only guarded on `len(globalIT.BlockedLabels) > 0`, so per-repo promotion silently stops working when global IT is disabled — a behavior regression.
⚠️ **daemon/cmd/heimdallm/main.go:0** — `pipe.Stop()` in the reload callback calls `wg.Wait()`, blocking until all tier goroutines finish. An in-flight AI review (which can run for minutes) will cause the reload HTTP handler to block for the same duration. Consider adding a stop deadline or draining with a timeout context to bound reload latency.
🟡 **daemon/internal/scheduler/pipeline.go:0** — The rate limiter refills on a rolling 1-hour ticker from process start, not aligned with GitHub's actual rate limit reset window. Under sustained load, the daemon can exhaust its budget just as GitHub resets or hold back requests unnecessarily. Syncing to GitHub's `X-RateLimit-Reset` header would be more accurate.
🟡 **daemon/cmd/heimdallm/main.go:0** — `HandleChange` for issue-type watch items returns nil without doing anything — Tier 3 detects the change but issues wait for the next Tier 2 poll cycle anyway. Acknowledged as a known limitation, but a priority re-queue signal to Tier 2 would close the gap for high-frequency issue updates during long poll intervals.

### Suggestions

- Change `defer pipe.Stop()` to `defer func() { pipe.Stop() }()` — this is the minimal fix to ensure the current post-reload pipeline is always stopped at shutdown, not the original one captured at defer time.
- In `PromoteReady`, walk per-repo AI configs to find enabled blocked labels when `globalIT.Enabled` is false, matching the original promotion logic and avoiding a silent regression.
- Document or enforce a maximum block time for `pipe.Stop()` during reload (e.g., a `context.WithTimeout`) so the reload HTTP endpoint has a bounded response time even when reviews are in flight.
- Consider a lightweight priority channel from Tier 3 to Tier 2 so detected issue changes can trigger a fast-path re-check rather than waiting a full poll interval.

---

Muriano:
## Fixes applied (round 3)

All review feedback addressed in 3ab2154:

| Issue | Fix |
|---|---|
| `defer pipe.Stop()` captures wrong pointer | → closure reads `pipe` under `cfgMu` at shutdown time |
| Data race on `pipe` during reload | Read old pipe under lock, Stop() outside lock, swap under lock — avoids deadlock |
| `Pipeline.Stop()` not idempotent | Added `sync.Once` |
| `HandleChange` bypasses PRAlreadyReviewed | Added dedup gate + TOCTOU comment |
| Sparse PR in ProcessPR/HandleChange | Widened `Tier2PR` with Title/HTMLURL/Author/State; HandleChange fetches from store |
| `PromoteReady` ignores per-repo IT | Removed `Enabled` gate — only checks `BlockedLabels`, restoring pre-refactor behavior |
| `ReEnqueue`/`ResetBackoff` mutate outside lock | Moved all item mutations inside `q.mu.Lock()` |
| `queue_test` doesn't test pop-then-reenqueue | Rewritten to use real Pop→ReEnqueue on same queue |
| Unused `reviewMu`/`inFlight` on tier2Adapter | Removed from struct (concurrency handled by `runReview` closure) |

All tests pass with `-race`.

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Well-structured refactor that correctly introduces a 3-tier polling pipeline and resolves the previous deadlock-on-reload issue. The adapter pattern cleanly bridges the concrete types to the new interfaces, and the WatchQueue implementation is sound. Three meaningful issues remain: a concurrent-reload race that can start duplicate pipelines, an effectively no-op Tier 3 handler for issues, and a missing floor on the watch interval that could saturate the GitHub API.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:391** — Concurrent-reload race: `newPipe.Start()` is called after `cfgMu` is released. If two goroutines enter the reload closure simultaneously, both read `pipe = oldPipe`, both call `oldPipe.Stop()` (idempotent, fine), both build independent `newPipeX`, the second writer wins the `cfgMu` swap, and then both call `Start()` — leaving two pipelines running against the same GitHub API budget. A per-reload mutex (or serialising reloads at the HTTP handler level) is needed.
⚠️ **daemon/cmd/heimdallm/main.go:947** — `HandleChange` for issue-type items is a documented no-op: Tier 3 detects the change but returns without resetting the item's backoff or triggering any expedited processing — the issue won't be re-queued with `ResetBackoff`, so it sits at its current (potentially 15-minute) backoff until Tier 2 picks it up normally. This negates the latency benefit of Tier 3 for issues entirely and should either be fixed or gated so issues are not added to the WatchQueue at all.
⚠️ **daemon/cmd/heimdallm/main.go:699** — `parseWatchInterval` has no minimum floor. A misconfigured `watch_interval = "1ms"` or similarly tiny value passes the `d <= 0` guard and is used as-is, causing Tier 3 to spin-poll the GitHub Issues API thousands of times per minute and exhaust the 4500 req/h budget almost immediately. A minimum of e.g. 10 seconds should be enforced.
🟡 **daemon/internal/scheduler/pipeline.go:78** — The rate-limiter refill ticker fires every calendar hour from the moment `Start()` is called, not aligned to any wall-clock boundary. A pipeline started at T+59min gets a full refill at T+1h19min (80 minutes after start), then every 60 minutes. In the 80-minute window the budget appears fully available from the refill perspective, so a burst of 4500 calls early followed by 4500 more after the first tick is possible — effectively doubling the budget in that window. Consider tracking the refill wall-clock time or using a token-bucket with leaky fill.
🟡 **daemon/internal/scheduler/pipeline.go:0** — No integration tests exist for `pipeline.go` or any of the three tier runners. The queue unit tests are solid, but the orchestration logic (tier hand-off via `reposChan`, rate-limiter consumption, eviction timing, reload lifecycle) is entirely untested. Given the concurrency surface area of the new component, at minimum a race-detector test for Start/Stop and a test for the Tier1→Tier2 channel hand-off would be valuable.

### Suggestions

- Serialise reloads with a dedicated `reloadMu sync.Mutex` held for the duration of the reload closure (read-old, stop-old, build-new, swap, start-new). This is safe because `Stop()` is already called outside `cfgMu` to avoid the deadlock, so holding `reloadMu` across that call does not reintroduce the original problem.
- For Tier 3 issue items, call `q.ResetBackoff(item)` after detecting a change and optionally emit a log line so operators can observe that the acceleration path is active — or, if immediate processing is not feasible in this PR, remove issues from the WatchQueue population to avoid giving a false impression of lower latency.
- Add a `minWatchInterval = 10 * time.Second` guard in `parseWatchInterval` alongside the existing `d <= 0` check, and log a warning when the configured value is clamped.
- Consider adding a `Pipeline.Started() bool` helper (backed by an atomic bool set in `Start`) so that `Stop` and the reload path can assert correct lifecycle ordering and surface misuse during tests.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid refactor from a single-tier scheduler to a 3-tier polling pipeline with a watch queue for active items. The reload path has been correctly reworked to avoid the deadlock risk, Stop() is idempotent via sync.Once, and the deferred shutdown now reads the current pipe pointer under the mutex. Code is well-commented and prior review rounds addressed the main correctness concerns. A few subtleties remain around Tier 3 behavior for issues, LastSeen semantics, and Evict efficiency.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:940** — HandleChange returns nil for issue-type items, relying on Tier 2 to pick up the change on its next 5-minute cycle. This silently discards Tier 3's detection work for issues — the whole point of the watch tier is faster reaction. The author acknowledged this as a follow-up, but at minimum a slog.Debug should record that an issue change was detected-but-ignored so operators can see the gap in telemetry.
🟡 **daemon/cmd/heimdallm/main.go:920** — CheckItem calls a.ghClient.GetIssue(item.Repo, item.Number) without propagating ctx. If the Tier 3 loop is cancelled (e.g. reload or shutdown), in-flight GetIssue calls will not be interrupted, potentially delaying pipeline Stop() by the HTTP timeout. Pre-existing limitation of gh.Client, but worth noting — consider a GetIssueContext variant or at least documenting that Stop() blocks on in-flight GitHub calls.
⚠️ **daemon/internal/scheduler/queue.go:44** — WatchQueue.Push sets LastSeen = time.Now() when zero, but the semantic meaning of LastSeen is 'last detected activity time' (used by CheckItem to detect UpdatedAt.After(LastSeen) changes). On initial push, callers should ideally pass the item's real UpdatedAt so that activity that occurred between the upstream fetch and the Tier 3 check is not missed. Setting to now() here means any activity whose UpdatedAt < push-time is treated as 'no change' on the first check. Consider requiring callers to set LastSeen explicitly or documenting this contract.
🟡 **daemon/internal/scheduler/queue.go:107** — Evict always allocates and populates the 'keep' slice even when nothing is evicted (the common case). For large queues this is wasteful. Do a first pass to count evictions, and only rebuild when evicted > 0 — or append to keep only after the first eviction is found.
🟡 **daemon/internal/scheduler/queue.go:72** — ReEnqueue/ResetBackoff silently drop the update when seen[id] is true (because a concurrent Push re-added the item). This is documented as 'semantically correct' but means the backoff doubling from a failed check can be lost if another producer races in. In practice this might cause the queue to retry at the fresh backoff instead of the doubled one — benign but worth an explicit comment that the drop is intentional.
🟡 **daemon/cmd/heimdallm/main.go:770** — tier2Adapter.cfg is **config.Config — a double-pointer chosen to track reassignments of the outer cfg variable under cfgMu. This works but is non-obvious for readers and easy to misuse (e.g. storing *a.cfg in a local variable under one lock and dereferencing it later without the lock re-captures the same snapshot correctly, but modifying through it would be unsafe). Consider adding a doc comment on the field explicitly calling out the intended access pattern, or migrating to atomic.Pointer in a follow-up as noted.
🟡 **daemon/cmd/heimdallm/main.go:820** — ProcessRepo computes anyITEnabled by iterating all AI.Repos, but then checks !anyITEnabled || !repoIT.Enabled. The anyITEnabled branch is strictly redundant: if repoIT.Enabled is true for this repo, then at least one repo has IT enabled, so anyITEnabled would also be true. The author chose to keep it as a safety guard, but it's dead logic — the extra lock acquisition and iteration happen on every Tier 2 cycle for every monitored repo.
🟡 **daemon/internal/scheduler/pipeline.go:102** — The rate limiter refill ticker fires every 1 hour from pipeline start, not on wall-clock hour boundaries. If RateLimiter is meant to model GitHub's hourly budget (which resets on a GitHub-controlled hourly boundary), this drift means the local refill and GitHub's actual budget reset will diverge after multiple reloads. Minor — and pre-existing in the sense that RateLimiter is outside this diff — but worth considering aligning to a wall-clock boundary.

### Suggestions

- Add a priority re-queue channel from Tier 3 to Tier 2 so that issue changes detected by the watch tier trigger immediate reprocessing instead of waiting up to PollInterval. This is the main value proposition of the 3-tier split for issues.
- Document or enforce the LastSeen contract: callers of WatchQueue.Push should set LastSeen to the item's real UpdatedAt (not now()) to avoid missing activity between fetch and first check.
- Consider replacing the **config.Config pattern with atomic.Pointer[config.Config] in a follow-up; it would eliminate the need for cfgMu on read paths and reduce contention in high-cardinality Tier 2 iteration.
- Add a metrics hook (or at least slog.Debug) when Tier 3 HandleChange is a no-op for issues, so operators can observe the gap between detection and Tier 2 pickup.
- WatchItem.index field is maintained but never used externally (heap.Fix/heap.Remove aren't called). Either remove it or add a comment explaining it's reserved for future index-based removal.

---

Muriano:
## Fixes applied (round 4)

Addressing vbuenog's latest review in 197a333:

| Issue | Fix |
|---|---|
| 🔴 Concurrent-reload race (duplicate pipelines) | Added `reloadMu` — held for the entire reload closure, serialises all reloads |
| ⚠️ HandleChange issues no-op | Added log line for observability. Note: `RunTier3` already calls `ResetBackoff(item)` after `HandleChange` returns for ALL changed items — so issue backoff IS reset to 1m on change detection. The "no-op" is only that we don't trigger immediate re-triage (that's Tier 2's job). |
| ⚠️ parseWatchInterval no minimum | Added 10s floor with warning log when clamped |

All tests pass with `-race`.

vbuenog (approved):
## 🤖 Heimdallm AI Review

Round 4 successfully addresses all previously-flagged HIGH and MEDIUM issues. `defer pipe.Stop()` now uses a closure reading `pipe` under `cfgMu`; the reload closure is serialised by `reloadMu` and properly swaps `pipe`/`cfg` under lock while calling `Stop()` outside the lock to avoid deadlock; `Pipeline.Stop()` is idempotent via `sync.Once`. `PromoteReady` no longer gates on `globalIT.Enabled`, restoring per-repo IT behaviour. `ReEnqueue`/`ResetBackoff` mutations are now inside the queue lock. `parseWatchInterval` has a 10s floor. One LOW finding from the prior round (rate-limiter refill not aligned to UTC quota boundary) is still unaddressed.

### Issues

🟡 **daemon/internal/scheduler/pipeline.go:85** — Rate-limiter refill still fires on a rolling 1h ticker anchored to daemon start, not GitHub's UTC-hour quota reset. After exhaustion near a boundary the daemon can wait up to ~59 extra minutes for a refill GitHub has already granted. Consider aligning the first tick to the next top-of-hour UTC.
🟡 **daemon/cmd/heimdallm/main.go:870** — HandleChange for issues still performs no actionable work — it only logs. The author's response relies on RunTier3 unconditionally calling ResetBackoff after HandleChange returns, which is plausible but leaves the latency benefit for issues limited to faster re-checking rather than faster re-triage. Not a regression vs. previous review, but worth verifying RunTier3 actually resets backoff on the changed path (not the error path) to avoid a silent no-op.

### Suggestions

- Consider computing the first rate-limiter refill tick as `time.Until(nextUTCHourBoundary)` and then switching to a 1h ticker, so refills track GitHub's actual quota reset.
- Add a brief comment in HandleChange (issue branch) pointing to where backoff is reset in RunTier3, so future readers don't misread it as a dead path.

---

Muriano:
Merged to main — includes #115 and #116.

PR #116, opened 2026-04-20
feat(flutter): unified Activity view with advanced filters (#109)

### LINKED ISSUE #109: feat: unified Activity view with advanced filters
## Summary

Merge the current Activity and Issues tabs into a single Activity view with powerful filtering.

## Current state

Activity tab shows PRs + Issues in a unified feed but with minimal filtering (sort only). The Issues tab is a separate dedicated view. Having both is redundant.

## Proposed changes

Remove the separate Issues tab. The Activity tab becomes the single view for all items with these filters:

### Filters
- **Type** — multi-select: PR, Issue Tracking, Development
- **Organization** — multi-select from known orgs
- **Repository** — multi-select from monitored repos
- **Search** — free text, matches issue/PR number, title, description, author
- **Sort** — Priority / Newest (existing, now persisted per #108)
- **Reset** — clears all filters (but not sort)

### Visual

Between the status indicator and the text block on each item, add a square badge showing the item type: `PR`, `IT`, `DEV` — with distinct colors.

## Backwards compatibility

- Existing SSE events and providers remain unchanged
- The Issues tab route (`/issues`) can redirect to Activity with type=IT pre-selected

Muriano:
## Summary

Merged Activity and Issues tabs into a single unified Activity view with:

- **Type filter**: PR (blue), IT (orange), DEV (green) as toggleable FilterChips
- **Organization multi-select**: popup derived from known repos
- **Repository multi-select**: popup filtered by selected orgs
- **Free text search**: matches title, repo, number, author
- **Reset button**: clears all filters (keeps sort)
- **Type badge**: 32x32 colored square (PR/IT/DEV) on each item between severity bar and text

Dashboard now has 5 tabs (Activity, Repositories, Prompts, Agents, Stats). Issues tab removed — all items shown in Activity with type filtering.

### New files
- `activity_filters.dart` — `ActivityFilters` model + `activityFiltersProvider`
- `activity_filter_bar.dart` — compact filter bar widget
- `type_badge.dart` — PR/IT/DEV colored badge

### Modified
- `dashboard_screen.dart` — unified `_ActivityItem` sealed class, filter/sort logic, removed Issues tab

## Test Plan
- [ ] Manual: filter by type, org, repo, search — verify results
- [ ] Manual: reset clears filters but keeps sort

Closes #109

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -910,10 +910,12 @@ func (a *tier2Adapter) CheckItem(ctx context.Context, item *scheduler.WatchItem)
 // HandleChange implements scheduler.Tier3ItemChecker.
 func (a *tier2Adapter) HandleChange(ctx context.Context, item *scheduler.WatchItem) error {
 	if item.Type == "pr" {
-		// Apply the same dedup gate as Tier 2: skip dismissed PRs and
-		// recently-reviewed PRs. Note: there is a small TOCTOU window
-		// between this check and the actual runReview call — the worst
-		// case is an occasional harmless duplicate review.
+		// Dedup: skip if the PR was already reviewed at or after the last
+		// detected change — mirrors the same check Tier 2 performs.
+		// NOTE (TOCTOU): between this check and the runReview call below, another
+		// goroutine could complete the same review. The impact is a rare harmless
+		// duplicate review; the in-flight guard in runReview prevents concurrent
+		// reviews of the same PR.
 		if a.PRAlreadyReviewed(item.GithubID, item.LastSeen) {
 			slog.Debug("tier3: PR already reviewed, skipping", "pr", item.Number, "repo", item.Repo)
 			return nil
--- a/daemon/internal/scheduler/queue.go
+++ b/daemon/internal/scheduler/queue.go
@@ -26,6 +26,11 @@ type WatchItem struct {
 }
 
 // WatchQueue is a thread-safe priority queue ordered by NextCheck.
+// The mutex protects the heap and seen map. Item field mutations (Backoff,
+// NextCheck, LastSeen) happen outside the lock in ReEnqueue/ResetBackoff —
+// this is safe because Tier 3 processes items sequentially (single writer).
+// If concurrent callers ever mutate the same item, the fields should be
+// moved inside the lock.
 type WatchQueue struct {
 	mu    sync.Mutex
 	items watchHeap
--- a/docs/superpowers/specs/2026-04-21-flutter-web-replaces-svelte-design.md
+++ b/docs/superpowers/specs/2026-04-21-flutter-web-replaces-svelte-design.md
@@ -1,0 +1,307 @@
+# Flutter Web replaces SvelteKit web_ui
+
+**Date:** 2026-04-21
+**Status:** Draft
+**Scope:** Replace the SvelteKit `web_ui/` with a Flutter Web build of `flutter_app/`, served by Nginx behind a token-injecting reverse proxy.
+
+---
+
+## Motivation
+
+Heimdallm ships two independent UI codebases:
+
+| UI | Tech | Purpose |
+|---|---|---|
+| `flutter_app/` | Flutter (macOS/Linux) | Desktop app |
+| `web_ui/` | SvelteKit + Node.js | Web dashboard for Docker deployments |
+
+Both implement the same screens (dashboard, PR detail, issues, config, logs, agents) against the same daemon REST + SSE API. Every new feature must be built twice.
+
+**Goal:** eliminate `web_ui/` by compiling `flutter_app/` for the web platform, unifying both UIs into one Dart codebase. The web target is internal-only (accessed when the daemon runs in Docker), so CanvasKit bundle size and SEO are non-concerns.
+
+---
+
+## Architecture
+
+### Before
+
+```
+browser :3000 ──► SvelteKit (Node.js) ──proxy──► daemon :7842
+                  reads /data/api_token
+                  injects X-Heimdallm-Token
+```
+
+### After
+
+```
+browser :3000 ──► Nginx
+                  ├─ /           → Flutter Web static assets (build/web/)
+                  ├─ /api/*      → proxy_pass daemon :7842, injects token
+                  └─ /events     → proxy_pass daemon :7842/events, injects token
+```
+
+- **Nginx** replaces both the SvelteKit server and the Node.js runtime.
+- **Token injection** stays server-side: Nginx reads the token from env var (or file) and sets the `X-Heimdallm-Token` header on every proxied request. The browser never sees the token.
+- **The daemon is unchanged** — no CORS, no new auth endpoints.
+
+---
+
+## Tasks
+
+### Phase 1: Platform abstraction layer
+
+Create a thin abstraction that isolates desktop-only code from the shared codebase.
+
+#### 1.1 Create `lib/core/platform/` service layer
+
+```
+lib/core/platform/
+  platform_services.dart          # abstract interface + conditional export
+  platform_services_desktop.dart  # real: tray, window, notifier, PID, signals
+  platform_services_web.dart      # no-op stubs
+```
+
+**Interface (abstract class `PlatformServices`):**
+- `Future<void> initWindow()`
+- `Future<void> initTray(ApiClient api)`
+- `Future<void> initNotifications()`
+- `Future<bool> ensureSingleInstance()`
+- `void listenForActivationSignal()`
+- `void sendNotification({String title, String body, int? prId})`
+- `void onWindowClose()` / `void setPreventClose(bool)`
+- `bool get isDesktop`
+
+**Desktop implementation:** moves existing code from `main.dart` lines 20-156 and `core/tray/tray_menu.dart` into the desktop implementation.
+
+**Web implementation:** all methods are no-ops or return immediately. `isDesktop` returns `false`.
+
+**Conditional export** in `platform_services.dart`:
+```dart
+export 'platform_services_web.dart'
+    if (dart.library.io) 'platform_services_desktop.dart';
+```
+
+#### 1.2 Adapt `ApiClient` for web
+
+Currently `ApiClient` (`lib/core/api/api_client.dart`) reads the token from `~/.local/share/heimdallm/api_token` using `dart:io`. On web:
+
+- **No token needed** — Nginx injects it server-side.
+- **Base URL is relative** — `/api/` instead of `http://127.0.0.1:7842/`.
+
+Changes:
+- Extract token-loading and URI building into overridable methods or use conditional import.
+- On web: `_uri(path)` returns `Uri.parse('/api$path')`, `_apiToken()` returns `null`.
+- On desktop: unchanged (reads token from file, uses `localhost:7842`).
+
+#### 1.3 Adapt `SseClient` for web
+
+Same pattern as ApiClient:
+- On web: connect to `/events` (relative), no token header.
+- On desktop: unchanged (`http://127.0.0.1:7842/events` with file-based token).
+
+#### 1.4 Adapt `main.dart` bootstrap
+
+Current boot sequence (desktop):
+1. `_ensureSingleInstance()` — PID file, SIGUSR1
+2. `_listenForActivationSignal()` — process signals
+3. `_setupWindow()` — window_manager
+4. `_setupTray()` — tray_manager
+5. `localNotifier.setup()` — native notifications
+6. `_BootstrapApp._boot()` — detect token → write config → find daemon binary → spawn daemon → health check
+
+**Web boot sequence:**
+1. Health check (`GET /api/health` via Nginx) → success → go to `/`.
+2. If unhealthy → show error screen ("daemon not reachable").
+
+Steps 1-5 and daemon spawn are skipped entirely. The `_BootstrapApp` widget uses `PlatformServices.isDesktop` to choose the desktop vs web boot path.
+
+#### 1.5 Remove `dart:io` from shared code
+
+12 files import `dart:io`. After the platform layer:
+- `main.dart` — uses PlatformServices, no direct `dart:io`.
+- `api_client.dart` — conditional import for token/URI.
+- `sse_client.dart` — conditional import for token/URI.
+- `daemon_lifecycle.dart` — only imported on desktop (behind conditional).
+- `first_run_setup.dart` — only imported on desktop (behind conditional).
+- `gh_cli.dart` — only imported on desktop (behind conditional).
+- `core/tray/tray_menu.dart` — moved into `platform_services_desktop.dart`.
+- `config_providers.dart` — guard daemon-spawn code with `PlatformServices.isDesktop`.
+- `config_screen.dart` — verify; may only need Platform for display logic.
+
+### Phase 2: Enable Flutter Web platform
+
+#### 2.1 Add web platform support
+
+```bash
+cd flutter_app && flutter create --platforms web .
+```
+
+This generates `web/index.html`, `web/manifest.json`, `web/favicon.png`.
+
+#### 2.2 Customize `web/index.html`
+
+- Set title to "Heimdallm".
+- Set favicon/icons to Heimdallm branding (reuse `assets/icon.png`).
+- Use CanvasKit renderer (default) — internal tool, no bundle-size concern.
+
+#### 2.3 Verify `flutter build web --release` compiles
+
+Gate: zero compilation errors. Run and verify the app loads in a browser against a local daemon.
+
+### Phase 3: Nginx + Docker
+
+#### 3.1 Create `flutter_app/nginx.conf.template`
+
+```nginx
+server {
+    listen 3000;
+
+    location / {
+        root /usr/share/nginx/html;
+        try_files $uri $uri/ /index.html;
+    }
+
+    location /api/ {
+        proxy_pass http://heimdallm:${HEIMDALLM_PORT}/;
+        proxy_set_header X-Heimdallm-Token "${HEIMDALLM_API_TOKEN}";
+        proxy_set_header Host $host;
+    }
+
+    location /events {
+        proxy_pass http://heimdallm:${HEIMDALLM_PORT}/events;
+        proxy_set_header X-Heimdallm-Token "${HEIMDALLM_API_TOKEN}";
+        proxy_http_version 1.1;
+        proxy_set_header Connection "";
+        proxy_buffering off;
+        proxy_cache off;
+        proxy_read_timeout 86400s;
+    }
+}
+```
+
+Token is injected via `envsubst` at container start.
+
+#### 3.2 Create `flutter_app/Dockerfile.web`
+
+```dockerfile
+# ── Build stage ──────────────────────────────────────────────────────
+FROM ghcr.io/cirruslabs/flutter:stable AS build
+WORKDIR /app
+COPY . .
+RUN flutter build web --release
+
+# ── Runtime stage ────────────────────────────────────────────────────
+FROM nginx:alpine AS runtime
+COPY --from=build /app/build/web /usr/share/nginx/html
+COPY nginx.conf.template /etc/nginx/templates/default.conf.template
+
+# envsubst in the official nginx:alpine image reads *.template from
+# /etc/nginx/templates/ and writes the resolved output to
+# /etc/nginx/conf.d/ at container start — no custom entrypoint needed.
+
+ENV HEIMDALLM_API_TOKEN=""
+ENV HEIMDALLM_PORT="7842"
+
+EXPOSE 3000
+```
+
+Entrypoint reads `/data/api_token` as fallback if `HEIMDALLM_API_TOKEN` is empty. Add a small `docker-entrypoint.d/` script:
+
+```sh
+#!/bin/sh
+# 10-load-token.sh — read token from shared volume if not set via env
+if [ -z "$HEIMDALLM_API_TOKEN" ] && [ -f /data/api_token ]; then
+  export HEIMDALLM_API_TOKEN=$(cat /data/api_token)
+fi
+```
+
+Result: image ~30 MB (vs ~180 MB for the Node.js image).
+
+#### 3.3 Update `docker/docker-compose.yml`
+
+Replace the `web` service:
+
+```yaml
+web:
+  build:
+    context: ../flutter_app
+    dockerfile: Dockerfile.web
+  container_name: heimdallm-web
+  restart: unless-stopped
+  ports:
+    - "${HEIMDALLM_WEB_PORT:-3000}:3000"
+  environment:
+    - HEIMDALLM_API_TOKEN=${HEIMDALLM_API_TOKEN:-}
+    - HEIMDALLM_PORT=${HEIMDALLM_PORT:-7842}
+  volumes:
+    - heimdallm-data:/data:ro
+  depends_on:
+    heimdallm:
+      condition: service_healthy
+  healthcheck:
+    test: ["CMD", "wget", "-qO-", "http://localhost:3000/"]
+    interval: 30s
+    timeout: 5s
+    start_period: 10s
+    retries: 3
+```
+
+Healthcheck switches from Node.js script to `wget` (available in nginx:alpine).
+
+### Phase 4: Cleanup
+
+#### 4.1 Delete `web_ui/` directory
+
+Remove entirely: `web_ui/`, including its `Dockerfile`, `package.json`, `src/`, `svelte.config.js`, etc.
+
+#### 4.2 Update CI workflows
+
+- If any workflow references `web_ui/`, update or remove those steps.
+- Add `flutter build web` validation to PR checks (optional, can be a follow-up).
+
+#### 4.3 Update `docs/index.html` (landing page)
+
+No change — the static landing page in `docs/` is independent of the web dashboard.
+
+---
+
+## Files created
+
+| File | Purpose |
+|---|---|
+| `lib/core/platform/platform_services.dart` | Abstract interface + conditional export |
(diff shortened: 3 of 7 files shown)


vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Solid architectural move unifying PRs and issues into one activity stream with a clean filter model. Three correctness/quality issues need attention before merge: a Riverpod reactivity bug, dead code in the filter predicate, and a duplicated type-detection expression.

### Issues

🔴 **flutter_app/lib/features/dashboard/activity_filter_bar.dart:33** — `_filteredRepos` calls `ref.read(activityFiltersProvider)` inside a getter invoked during `build`. `ref.read` does not subscribe, so when the org filter changes the repo dropdown list will not update until the next unrelated rebuild. This should be `ref.watch`, or the filtered repos should be derived from the `filters` value already watched in `build` and passed in.
⚠️ **flutter_app/lib/features/dashboard/activity_filter_bar.dart:165** — `Navigator.of(ctx).pop()` is called immediately after every checkbox toggle, closing the popup after a single selection. The `StatefulBuilder` wrapper (and the `setMenuState` callback) is therefore dead code — it is never called. Multi-select requires reopening the menu for each item, which makes the UX identical to a single-select. Either remove the close-on-toggle so users can tick multiple items and dismiss manually, or rename the widget to clarify it is single-select.
⚠️ **flutter_app/lib/features/dashboard/dashboard_screen.dart:0** — The `me` parameter of `_matchesFilters` is accepted but never read inside the function body. It is a remnant of the previous 'My Reviews / My PRs' split. Dead parameters are misleading and will cause a lint warning. Remove the parameter and update all call sites.
⚠️ **flutter_app/lib/features/dashboard/dashboard_screen.dart:0** — The logic `issue.latestReview != null && issue.latestReview!.actionTaken == 'develop'` that resolves an issue to type 'dev' vs 'it' is duplicated verbatim in both `_itemType()` (top-level function) and `_IssueActivityTile._type` (getter). `_IssueActivityTile` should delegate to `_itemType` to keep the classification in one place.
🟡 **flutter_app/lib/features/dashboard/activity_filter_bar.dart:47** — Setting `_searchController.text` directly inside `build` (to sync external resets) resets the cursor to the end on every rebuild while the user is typing. Prefer `TextEditingController.value = _searchController.value.copyWith(text: filters.search)` which preserves the selection, or better, guard the assignment with `WidgetsBinding.instance.addPostFrameCallback` so it does not run mid-frame.
🟡 **flutter_app/lib/features/dashboard/activity_filters.dart:37** — `ActivityFilters.reset()` returns `const ActivityFilters()` but is never called anywhere; all reset sites construct `const ActivityFilters()` directly. Either remove the method or use it consistently.
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:0** — `_sortItems` sorts the list in-place (mutating the caller's `filtered` list). The function name does not signal mutation and it returns void. Rename to `_sortItemsInPlace` or have it return a sorted copy to avoid surprising future callers.

### Suggestions

- Consider replacing `StateProvider<ActivityFilters>` with a dedicated `StateNotifier` / `Notifier` subclass. Direct `.state = filters.copyWith(...)` mutations scattered across the UI make the filter transitions hard to test and trace; a notifier would centralise all mutations (setOrgs, setRepos, setSearch, reset) and expose them as explicit methods.
- The `_itemPriorityKey` for issues maps 'critical' to 0, the same bucket as 'no review yet' for PRs. This is intentional (both sort to the top) but the shared numeric space makes it easy to accidentally break priority ordering when adding new severity levels. A comment explaining the shared scale, or separate sort keys per item type, would prevent future regressions.
- Now that PRs and issues share a single flat list there is no visual grouping. Consider adding a lightweight section divider or a small type-badge-based count summary at the top so users can still quickly see how many PRs vs issues vs dev items are visible without scrolling.

---

Muriano:
## Response to AI Review

**Fixed (9101949):**
- **Item 1 (ref.read in build)**: `_filteredRepos` now takes `filters` as parameter instead of calling `ref.read` during build
- **Item 3 (dead `me` param)**: Removed from `_matchesFilters` and all call sites
- **Item 4 (duplicate _itemType)**: `_IssueActivityTile._type` now delegates to `_itemType`
- **Item 5 (cursor reset)**: Search controller sync guarded by `!_searchFocus.hasFocus`
- **Item 6 (unused reset())**: Removed from `ActivityFilters`

**Already fixed in previous commit (754f37d):**
- **Item 2 (Navigator.pop on every toggle)**: Replaced `PopupMenuButton` with custom Dialog that allows multi-select + Apply button

**Not changing:**
- **Item 7 (_sortItems in-place)**: Consistent with Flutter's `List.sort()` convention. Renaming adds noise for no safety gain.

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Solid PR that meaningfully improves both the Go daemon (deterministic Tier 2 startup, atomic cfg+pipe swap, dedup at Tier 3) and the Flutter side (unified activity view with multi-filter bar). The Go changes are the riskier half; the Flutter side is clean but has a few correctness gaps worth addressing.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:388** — pipe.Stop() is now called while cfgMu is held. If Stop() blocks waiting for goroutines (e.g. Tier 2 ticker loop) that themselves try to acquire cfgMu (e.g. via the config callback path), this becomes a deadlock. The comment justifies holding the lock for the cfg+pipe swap, but Stop() should complete before the lock is taken, or Stop() must be guaranteed non-blocking under this lock.
⚠️ **flutter_app/lib/features/dashboard/activity_filter_bar.dart:7** — SortMode is declared here AND as _SortMode in dashboard_screen.dart, and the comment admits the duplication. The two enums must be kept in sync manually. If a third sort mode is added to one, the filter bar silently breaks. Extract to a shared file (e.g. activity_sort_mode.dart) and use it from both.
⚠️ **flutter_app/lib/features/dashboard/dashboard_screen.dart:261** — When active filters produce zero results, the code renders an empty ListView with no feedback. The existing 'No activity yet' guard only fires when the raw data is empty. A user who sets filters that match nothing will see a blank screen with no indication why. Add an empty-filtered-state widget (e.g. 'No items match the current filters').
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:96** — _itemPriorityKey assigns bucket 0 to 'critical' severity for issues but PRs have no 'critical' branch and fall through to the default bucket 3. If a PR ever carries 'critical' severity it silently gets the lowest priority instead of the highest. Either add 'critical' to the PR branch or document that PRs cannot have critical severity.
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:84** — _SortNotifier._loadAsync silently swallows any SharedPreferences error. A failed read leaves the sort mode at the default without any log entry, making it hard to diagnose persistence issues on constrained devices. Wrap with try/catch and at least log the error.
🟡 **daemon/internal/scheduler/queue_test.go:44** — Using two independent queues (q1, q2) sidesteps the dedup map instead of testing around it. The test no longer verifies that a real queue correctly re-enqueues after an item is drained. Consider draining q1 with PopReady() and then re-enqueuing into the same q1 to keep coverage of the actual production code path.

### Suggestions

- Extract SortMode (and its persistence key) to a dedicated activity_sort_mode.dart so both dashboard_screen.dart and activity_filter_bar.dart share a single source of truth — eliminates the duplication the author flagged in the comment.
- Move pipe.Stop() back outside cfgMu (before the lock), then acquire cfgMu only for the cfg and pipe pointer swap. This preserves the atomicity guarantee for readers while eliminating the risk of Stop() blocking under the lock.
- Add a 'No results' empty state in _ActivityTabState.build when filtered.isEmpty but the raw items list is non-empty, so users know filters are active rather than assuming there is no data.
- In _SortNotifier._loadAsync, add a try/catch around the SharedPreferences call and log any exception with slog-style structured logging to aid diagnostics.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

PR unifies the Activity view in Flutter with solid filter infrastructure, and makes several backend improvements (Tier 2 blocking on initial repo list, Tier 3 dedup, mutex scope changes). The Flutter changes are well-structured and the previous AI-review feedback has been addressed. However, the new locking pattern in the pipeline reload in main.go widens the critical section around pipe.Stop()/Start() in a way that risks deadlock with Tier 2 adapter callbacks that acquire the same mutex.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:389** — Holding cfgMu across pipe.Stop() is deadlock-prone. The tier2Adapter methods (CheckItem, HandleChange, PRAlreadyReviewed path) acquire a.cfgMu, which is the same cfgMu. If a Tier 2/3 goroutine is blocked trying to lock cfgMu while pipe.Stop() waits for that goroutine to drain, the reload goroutine (holding cfgMu) will deadlock with the worker (waiting for cfgMu). The prior ordering (Stop before Lock) avoided this. Consider reloading cfg under the lock, but calling Stop()/Start() outside it, or use an RWMutex with only the cfg swap under write-lock.
🟡 **daemon/internal/scheduler/tier2.go:65** — Blocking on the first reposChan receive removes the 2s fallback. If Tier 1 never sends (startup error, panic, long API call), Tier 2 will never begin its ticker loop and only exits on ctx cancellation. Worth adding a bounded timeout or at least a warning log after N seconds to aid debugging.
🟡 **flutter_app/lib/features/dashboard/activity_filter_bar.dart:62** — Mutating a TextEditingController inside build() (even guarded by focus check) is an anti-pattern: during a rebuild triggered by focus loss while filters changed externally, this executes synchronously in build. Prefer a ref.listen on activityFiltersProvider in initState/didChangeDependencies to sync the controller imperatively, keeping build pure.
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:95** — _SortNotifier._loadAsync is declared `void ... async` (fire-and-forget). If SharedPreferences throws, the error is silently swallowed. A try/catch or `unawaited(...)` with explicit error handling would make failures observable.
🟡 **daemon/internal/scheduler/queue_test.go:46** — The rewritten test no longer exercises PopReady + ReEnqueue on the same queue; it just calls ReEnqueue on two fresh queues with the same item. This verifies that Backoff doubles per ReEnqueue call, but the comment concedes that ReEnqueue is a no-op for already-tracked IDs on the same queue — meaning the original behavior the old test intended to cover (re-enqueue-after-pop doubling) is now untested. Consider adding a test that actually exercises the pop-then-reenqueue path on a single queue.
⚠️ **daemon/cmd/heimdallm/main.go:888** — Tier 3 HandleChange dedup calls a.PRAlreadyReviewed(item.GithubID, item.LastSeen). Verify that item.LastSeen reflects the latest change timestamp (not a stale value); if LastSeen is updated only on scheduler insertion, a PR reviewed after first scheduling but before the change was seen could be incorrectly skipped. Quick comment or cross-reference to where LastSeen is maintained would help reviewers confirm the invariant.

### Suggestions

- Narrow the critical section in the config-reload callback: lock only for the cfg/pipe field swap, or use sync.RWMutex with writes only during the swap, so pipe.Stop()/Start() don't run under the same lock that Tier 2/3 goroutines acquire.
- Add a diagnostic log in RunTier2 if the initial reposChan receive is still pending after, e.g., 10 seconds, so a stuck Tier 1 is obvious in logs.
- Move the TextEditingController sync out of build() and into a ref.listen callback to avoid side effects in the build phase.
- Add an integration-style test (or a targeted unit test) covering PopReady followed by ReEnqueue to replace the coverage that the refactored TestWatchQueue_BackoffDoubles lost.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean unification of Activity view with solid daemon concurrency fixes. The previous review items (1–7) from the AI review are all addressed per the author's response. The daemon-side changes (lock scope fix, blocking Tier-1 receive, Tier-3 dedup guard) are correct improvements. A few minor issues remain in the new Flutter code.

### Issues

⚠️ **flutter_app/lib/features/dashboard/activity_filter_bar.dart:8** — SortMode enum is duplicated from dashboard_screen.dart's private _SortMode. The comment acknowledges this, but the two can silently diverge if a third sort mode is added to only one of them — the callback-based bridge will compile fine but misbehave at runtime. Extract a single public enum into a shared file (e.g., activity_filters.dart) to close this gap.
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:83** — _SortNotifier._loadAsync is an async-void fire-and-forget called from build(). If the notifier is disposed before the await completes (e.g., rapid hot-restart), assigning `state` throws a StateError that is silently swallowed. Guard the assignment with a mounted/active check, or use AsyncNotifier which handles this lifecycle natively.
🟡 **daemon/cmd/heimdallm/main.go:888** — PRAlreadyReviewed is called outside cfgMu (and presumably outside any lock). Between this check returning false and the subsequent runReview call, another goroutine could complete the same review — a classic TOCTOU window. The impact is limited to an occasional duplicate review (harmless), but worth a brief comment noting the accepted race.
🟡 **daemon/cmd/heimdallm/main.go:390** — Moving pipe.Stop() inside cfgMu.Lock() is the correct fix for the half-swap race, but it changes the blocking profile: if Stop() drains in-flight work, all readers of cfg/pipe (HTTP handlers, Tier-2 callbacks) block until the drain completes. Consider documenting the expected drain latency or adding a timeout to Stop() so a stuck goroutine can't deadlock the reload path.

### Suggestions

- Promote the SortMode enum to a public type in activity_filters.dart and import it from both activity_filter_bar.dart and dashboard_screen.dart to eliminate the duplication.
- Consider using Riverpod's AsyncNotifier for _SortNotifier — it handles the async load lifecycle safely and avoids the fire-and-forget async-void pattern.

---

Muriano:

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

PR introduces a unified Activity view with filters in Flutter and tightens concurrency guarantees in the Go daemon. The Flutter code is well-structured with good use of sealed classes and Riverpod. The main concern is a potential deadlock risk in main.go where cfgMu is now held across pipe.Stop/Start, which could block if any goroutine the pipeline is waiting on also tries to acquire cfgMu (e.g., tier2Adapter.HandleChange).

### Issues

🔴 **daemon/cmd/heimdallm/main.go:391** — cfgMu is now held across pipe.Stop(), buildPipeline(), and pipe.Start(). If pipe.Stop() blocks waiting for in-flight tier2/tier3 goroutines that call into tier2Adapter methods which also take cfgMu (e.g., HandleChange does a.cfgMu.Lock()), this creates a deadlock. The lock should be narrowed to protect only the cfg/pipe assignment, or a separate reload mutex should serialize reloads while cfgMu is released during Stop().
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:0** — SortNotifier._loadAsync is declared as `void _loadAsync() async` (fire-and-forget with no error handling). If SharedPreferences throws, the error is swallowed. Should return Future<void> and handle errors, or at minimum add a try/catch.
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:0** — SortNotifier.set fires SharedPreferences.getInstance().then(...) without awaiting or error handling. Persistence failures are silently dropped.
🟡 **flutter_app/lib/features/dashboard/activity_filter_bar.dart:58** — Mutating `_searchController.text` inside build() is an anti-pattern even guarded by the focus check — it can still trigger controller listeners mid-build. Consider using `ref.listen(activityFiltersProvider, ...)` in initState/build to react to filter resets, or move the sync to didChangeDependencies/post-frame callback.
⚠️ **daemon/internal/scheduler/tier2.go:66** — The blocking receive on reposChan at startup means RunTier2 is stuck until Tier 1 sends. If Tier 1 has a transient error and never emits the initial list (prior code worked with an empty repos slice after 2s), tier2 will never start ticking. Worth confirming Tier 1 guarantees at least one send or ctx cancellation.
⚠️ **daemon/internal/scheduler/queue_test.go:45** — The refactored test uses two separate queues, which no longer validates that backoff doubles across successive ReEnqueue calls on the same queue — it only validates the doubling arithmetic in isolation. The original intent (verify backoff grows across real enqueue cycles) is lost. Consider using PopReady or a helper that bypasses the seen-set instead.

### Suggestions

- In main.go reload handler, call pipe.Stop() before acquiring cfgMu, then acquire cfgMu only for the cfg/pipe pointer swap, then release and call pipe.Start(). Use a separate reloadMu if you need to serialize reloads.
- Consider turning _loadAsync into an explicit Future<void> and document the race between the default value and the async load (early UI interactions before load completes could be lost).
- For the search controller sync, switch to ref.listen on activityFiltersProvider to imperatively sync the controller only when the filter changes externally — cleaner than a build-time check.
- Add a test for tier2 startup that verifies ctx cancellation unblocks the initial receive, and document in the RunTier2 comment that Tier 1 must always emit or cancel.

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Solid unification of the Activity view with good filter UX. The Go backend fixes (pipeline lock ordering, Tier 2 init, TOCTOU docs) are directionally correct but the new lock scope introduces a deadlock risk. Flutter side has a subtle anti-pattern in build() and an unguarded async state write. Overall the PR is close to mergeable but the lock issue should be addressed first.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:388** — pipe.Stop(), buildPipeline(), and pipe.Start() are all held inside cfgMu.Lock(). If pipe.Stop() blocks waiting for goroutines to finish, and any of those goroutines attempt to acquire cfgMu (e.g. HTTP handlers, Tier 2 callbacks), this is a deadlock. The original code held the lock only around the cfg swap, which was safe. The comment justifies protecting the pipe reference, but the fix should be to store pipe in an atomic/separate guard rather than widening the critical section to include potentially blocking operations.
⚠️ **flutter_app/lib/features/dashboard/dashboard_screen.dart:96** — SortNotifier._loadAsync() assigns to state after an await without checking whether the notifier is still alive. If the provider is disposed between the SharedPreferences.getInstance() call and the state assignment, Riverpod throws a StateError. Riverpod's Notifier has no 'mounted' guard; the standard fix is to use AsyncNotifier or wrap with a try/catch on StateError.
⚠️ **flutter_app/lib/features/dashboard/activity_filter_bar.dart:63** — Assigning _searchController.text during build() (the search sync guard) notifies TextEditingController listeners synchronously, which can schedule a second build in the same frame. This is a known Flutter anti-pattern. The correct place is didUpdateWidget or a post-frame callback via WidgetsBinding.instance.addPostFrameCallback.
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:133** — Item types ('pr', 'it', 'dev') and the action value ('develop') are bare string literals with no constant definitions. A typo in any call site is a silent bug. Define them as constants or an enum value.
🟡 **daemon/internal/scheduler/queue_test.go:46** — TestWatchQueue_BackoffDoubles uses two separate queues to sidestep dedup, but in production the same queue is reused across re-enqueues. The test validates the backoff arithmetic on the item struct but does not exercise the queue's behavior when the same item is re-enqueued after being drained. Consider adding a separate test that drains the queue before re-enqueuing to confirm the full production path.

### Suggestions

- For the cfgMu scope problem: consider introducing a separate pipeMu (or an atomic pointer) that guards only the pipe reference, keeping cfgMu's scope limited to the cfg struct swap. This avoids holding any lock across blocking Start/Stop calls.
- SortNotifier._loadAsync() swallows all SharedPreferences errors silently. Add a try/catch and log the error so preference load failures are observable.
- The Issues tab removal is a breaking UX change for any user who bookmarked that tab by index. If tab order is persisted anywhere (SharedPreferences, URL state), ensure old indices are migrated or reset gracefully.
- The diff is truncated before the tile-rendering code. Ensure the full _ActivityTabState build method and tile widgets were also covered in the previous review round, particularly around list virtualization for large combined PR+issue lists.

---

Muriano:
## Fixes applied (round 3)

Rebased on updated `fix/persist-sort-preference` (which includes all #114 daemon fixes). All review feedback addressed:

| Issue | Fix |
|---|---|
| cfgMu held across pipe.Stop() (deadlock risk) | Fixed in #114: Stop() outside lock, swap under lock |
| SortMode enum duplicated | Already deduplicated to `activity_filters.dart` (round 2) |
| `_loadAsync` no error handling | Added try/catch + debugPrint (from #115 Notifier fix) |
| Dead `me` parameter in `_matchesFilters` | Removed parameter and `meAsync`/`meProvider` watch |
| Empty filtered state | Already fixed (round 2): "No items match the current filters." |
| `onSortChanged` used `_saveSortPreference` | Replaced with `notifier.set()` — persistence centralized |

Daemon: all scheduler fixes inherited from rebased #114 (queue lock ordering, PRAlreadyReviewed dedup, Pipeline.Stop idempotent, Tier2PR widened, PromoteReady per-repo).

All tests pass (Go + Flutter analyze).

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid implementation of the unified Activity view with well-structured filter state. The previous review rounds have addressed the major issues. A few remaining concerns around filter cascade logic, StateProvider deprecation, and minor UX inconsistencies.

### Issues

⚠️ **flutter_app/lib/features/dashboard/activity_filter_bar.dart:153** — Stale repo filter when all orgs are deselected: the `if (orgs.isNotEmpty)` guard in the org `onChanged` callback means that when the user clears all org selections (orgs becomes `{}`), any previously selected repos are NOT cleared. A user could end up with no orgs selected but still have repo filters active, producing an empty or misleading result set. The cascade should also run when orgs is empty: `repos = repos.where((r) { final org = r.contains('/') ? r.split('/').first : r; return orgs.isEmpty || orgs.contains(org); }).toSet();` — or unconditionally clear repos when orgs is cleared.
🟡 **flutter_app/lib/features/dashboard/activity_filters.dart:38** — `StateProvider<ActivityFilters>` is soft-deprecated in Riverpod 2.x in favour of `NotifierProvider`. The PR discussion notes a `NotifierProvider` fix was applied in a sibling branch for sort persistence, but this diff still uses `StateProvider`, making state mutation scattered across multiple call sites via `.notifier.state = ...`. Consolidating into a `Notifier` would centralise all mutation logic and enable cleaner testing.
🟡 **flutter_app/lib/features/dashboard/activity_filter_bar.dart:185** — `GestureDetector` wrapping a `Chip` for the multi-select popup loses Material ink/ripple feedback and does not respond to keyboard/accessibility activation. Prefer setting `onPressed` on the `ActionChip` (or using `InkWell`) to retain the built-in interaction model and accessibility semantics. The Reset button at line 218 has the same issue.
🟡 **flutter_app/lib/features/dashboard/activity_filter_bar.dart:44** — `_allOrgs` is a getter that allocates a new `Set` and runs `.map().toSet()` on every call. It is accessed in `build` and as a condition for the Org popup child. For a desktop tool with dozens of repos this is harmless, but caching it (e.g. with a local variable at the top of `build`) avoids redundant work and makes the dependency on `widget.allRepos` explicit.
🟡 **docs/superpowers/specs/2026-04-21-flutter-web-replaces-svelte-design.md:0** — 307-line Flutter-Web migration spec is unrelated to the unified Activity view feature this PR implements. Including it here makes the PR scope ambiguous and conflates a speculative future design (Draft status) with a concrete feature delivery. Consider tracking this in a dedicated issue or a separate docs PR to keep the review focused.

### Suggestions

- Fix the org→repo cascade to also clear stale repo selections when all orgs are deselected, either by removing the `if (orgs.isNotEmpty)` guard or by always intersecting repos with the active org set.
- Replace `StateProvider` with a `Notifier`/`NotifierProvider` to centralise all filter mutations in one place — this is especially valuable now that sort persistence has been added.
- Use `ActionChip(onPressed: ...)` instead of `GestureDetector` wrapping `Chip` for the Org, Repo, and Reset chips to preserve Material ripple feedback and keyboard/screen-reader accessibility.
- Cache `_allOrgs` and `_filteredRepos(filters)` as local variables at the top of `build` rather than recomputing them as getters or inline expressions to avoid redundant allocations per frame.
- Move the Flutter-Web spec document out of this PR into a dedicated tracking issue or a standalone docs PR so the activity-filter feature can be reviewed and merged independently.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Re-review focused on new changes since round 3. The previous round's findings (promoter.go, activity.go) are not in this diff — the author rebased onto #114 which addresses them. Current diff is comment updates in the daemon (TOCTOU/queue mutex docs) plus the new unified Activity filter UI. The Flutter code is clean; a couple of minor nits only.

### Issues

🟡 **flutter_app/lib/features/dashboard/activity_filter_bar.dart:52** — Mutating `_searchController.text` inside `build()` is a code smell — it can race with controller listeners and cause an unexpected rebuild or cursor jump on the same frame. The `!_searchFocus.hasFocus` guard mitigates the common case, but the idiomatic fix is to sync the controller in `didUpdateWidget`/a `ref.listen` callback rather than during build. Low priority since the empirical behavior is acceptable.
🟡 **flutter_app/lib/features/dashboard/activity_filter_bar.dart:37** — `_allOrgs` is a getter that rebuilds a `Set` from `widget.allRepos` on every build call, and the popup path additionally calls `_filteredRepos(filters)` which iterates again. For small repo counts this is a non-issue, but if `allRepos` ever grows, cache it in state (recompute on didUpdateWidget when `allRepos` changes).
🟡 **flutter_app/lib/features/dashboard/activity_filter_bar.dart:218** — Ternary used for side effects: `selected ? types.remove(type) : types.add(type);`. Works, but using the ternary's return value as a statement reads poorly — prefer an `if/else`. Same pattern at line ~330 inside the dialog (`checked ? current.remove(v) : current.add(v);`). Style-only.

### Suggestions

- Consider extracting the sort/type/multi-select chip builders into small private widgets so build() is flatter and each widget rebuilds independently.
- The multi-select dialog does not offer a 'Select all' action — with many orgs/repos this can be tedious. Low priority UX polish.
- The updated TOCTOU and WatchQueue comments are clear and load-bearing — good. Consider linking the queue comment to the Tier 3 single-writer contract in code (e.g., mention `scheduler.Tier3` by name) so future refactors hit the invariant.

---

Muriano:
Merged to main via #114 (single integration merge).

PR #118, opened 2026-04-20
fix: make run-linux usable end-to-end (HOME, token, gh CLI)

vbuenog:
Three linked fixes to `make run-linux` so the Flutter desktop app actually runs, auto-detects the GitHub token, and repo discovery works — all from inside the `heimdallm-verify` container with no user intervention beyond `make verify-linux && make run-linux`.

Each commit stands alone and could be reviewed / reverted independently; together they close the chain of "why isn't the app working in Docker?"

## The chain

1. Launch `make run-linux` → container starts → **PathAccessException** on `//.local` → app crashes before the window appears. (Commit 1)
2. After fixing that, app launches → first-run screen → no token pre-filled, user has to paste one despite `gh auth login` being set up on the host. (Commit 2)
3. After fixing token detection, app works → **Discover** button in Repositories silently returns empty because `gh` isn't installed in the container and the API fallback has a bad query. (Commit 3)

### `97315c4` — `fix(make): set HOME and mount data dir so run-linux doesn't crash`

`docker run --user $UID:$GID` leaves `HOME` unset because the UID has no entry in the image's `/etc/passwd`. Flutter's `_ensureSingleInstance` reads `$HOME/.local/share/heimdallm` for its PID file, which resolves to `//.local/share/heimdallm` and EACCESes before the GTK window comes up.

- Inject `HOME=$HOME` into the env-file the container already reads.
- `mkdir -p ~/.local/share/heimdallm` on the host pre-mount and bind it into the container at the same path (single well-scoped directory, not `~/.local` wholesale).

**Files:** `Makefile` (+3 / -1).

### `114ee1e` — `fix(make): resolve GITHUB_TOKEN from host gh CLI as fallback`

`FirstRunSetup.detectToken()` tries `gh auth token` → Keychain → `GITHUB_TOKEN` env. Inside the container `gh` wasn't installed (commit 3 fixes that), and in the meantime nothing populates `GITHUB_TOKEN` if the user hasn't exported it in their shell — so the app always lands on the first-run "paste your token" screen, even when their host `gh` has been logged in for months.

Resolve the token on the **host** in the Makefile before starting the container: if `GITHUB_TOKEN` is already set we keep it, otherwise we call `gh auth token` on the host and inject the result into the same env-file. Falls through silently if both fail — no regression.

Kept as belt-and-suspenders even after commit 3 installs `gh` inside the container: useful for users who've configured `gh` with a non-default `GH_CONFIG_DIR`, or who only want to pass a token without mounting their gh config.

**Files:** `Makefile` (+5).

### `c9f1434` — `feat(docker): install gh CLI and mount host auth so discovery works`

The Repositories screen's Discover button calls `RepoDiscovery.discoverFromPRs` → prefers `gh search prs` (not installed in the container) → falls through to `_viaApi` → issues a **single query** `(review-requested:@me OR assignee:@me OR author:@me)` that GitHub's Search API rejects with HTTP 422 "The listed users cannot be searched" (OR-joined user qualifiers aren't supported, regardless of `@me` vs explicit username). Result: every Linux user without host `gh` has had a silently-broken Discover button. The bug was masked for everyone who had `gh` on PATH because the gh-CLI path short-circuits before the broken API path runs.

- Install `gh` from GitHub's official apt repo in `Dockerfile.linux-verify`'s runtime-deps layer (kept out of the build layer — CI doesn't need it). Adds `curl` + `gnupg` for the keyring dance.
- Conditionally bind-mount `~/.config/gh` read-only if the host has it, so the container reuses the user's existing authentication. Read-only guards against the container mutating host auth state. Silently skipped if the host has no gh config.

Scope deviation from "minimum mount" is intentional: fixing this via the client-side API path would require non-trivial query restructuring in the Flutter code (a three-query union like `_viaGhCli` already does), and the bug affects any Linux user without `gh`, not just us.

**Files:** `Dockerfile.linux-verify` (+19 / -3), `Makefile` (+5).

## Combined footprint

- `Makefile` — three targeted additions to the `run-linux` recipe (13 lines, 4 lines replaced).
- `Dockerfile.linux-verify` — one new apt layer for gh + keyring (~20 lines).
- **No source code changes.** Zero Flutter, zero daemon, zero web UI.
- Image rebuild adds ~30 MB (gh + curl + gnupg). Verified by inspecting the layer diff.

## Test plan

- [x] `make verify-linux` — full build pipeline passes including the new runtime layer.
- [x] `make run-linux` — desktop window opens, no PathAccessException.
- [x] First-run token field auto-populated; no manual paste needed.
- [x] Repositories → Discover returns the full list including `theburrowhub/heimdallm` (where I'm review-requested on #106).
- [x] `docker exec heimdallm-run which gh` → `/usr/bin/gh`.
- [x] `docker exec heimdallm-run gh api user -q .login` → `vbuenog`.

## Behaviour matrix

| Scenario | Before | After |
|---|---|---|
| Host has `gh` + logged in, `~/.config/gh` present | App crashes, then broken Discover | Works end-to-end |
| Host has `gh` but no `~/.config/gh` (unusual) | Crash + unauth'd | `gh auth token` on host still returns null → env-var fallback → token unset → first-run prompt. No regression. |
| Host has no `gh` at all | Crash + broken Discover | App runs; token has to be pasted manually (same as before, minus the crash). |
| `GITHUB_TOKEN` pre-exported in shell | Honoured | Honoured (env wins over `gh auth token`). |

## Security notes

- `~/.config/gh` is mounted **read-only**. The container cannot modify the host's gh auth state.
- Token values exist briefly in a tempfile (`$ENV_FILE`) the recipe already `trap`-cleans on exit — no change to that lifetime.
- `gh` running inside a container with the user's auth has the same blast radius as running `gh` on the host directly — the token's scope and expiry apply identically.

### DIFF
--- a/Dockerfile.linux-verify
+++ b/Dockerfile.linux-verify
@@ -106,6 +106,17 @@ RUN { \
     } > /usr/local/bin/xdg-open \
     && chmod +x /usr/local/bin/xdg-open
 
+# /.flatpak-info is glib's sandbox sentinel. If this file exists,
+# g_desktop_app_info_is_sandboxed() returns true and GTK's
+# gtk_show_uri_on_window (which url_launcher_linux calls internally) routes
+# URI opens through org.freedesktop.portal.OpenURI instead of the
+# in-sandbox MIME database. Without this the Flutter app's "Open in GitHub"
+# silently no-ops because glib can't find a handler for https inside the
+# container — the /usr/local/bin/xdg-open shim above only helps code paths
+# that shell out; url_launcher_linux doesn't. Minimal valid keyfile
+# content keeps any downstream parsers that also read this file happy.
+RUN printf '[Application]\nname=heimdallm\n' > /.flatpak-info
+
 # ── Node.js + AI CLIs ────────────────────────────────────────────────────────
 # The bundled daemon inside this image runs the review pipeline, which shells
 # out to whatever AI CLI the user configured as `ai.primary`. Without one on
--- a/Dockerfile.linux-verify
+++ b/Dockerfile.linux-verify
@@ -78,7 +78,7 @@ RUN apt-get update -y \
     && apt-get install -y --no-install-recommends \
        libayatana-appindicator3-1 libnotify4 libsecret-1-0 \
        dbus-x11 libcanberra-gtk3-module mesa-utils \
-       curl gnupg xdg-utils \
+       curl gnupg xdg-utils libglib2.0-bin \
     && curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
        | tee /usr/share/keyrings/githubcli-archive-keyring.gpg > /dev/null \
     && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
@@ -87,6 +87,24 @@ RUN apt-get update -y \
     && apt-get update -y \
     && apt-get install -y --no-install-recommends gh \
     && rm -rf /var/lib/apt/lists/*
+
+# Replace xdg-open with a tiny DBus portal shim. Stock xdg-open inside the
+# container has no browser to dispatch to; calling the host's
+# xdg-desktop-portal directly (via gdbus from libglib2.0-bin) opens URIs in
+# whatever browser the host user has configured, with no MIME database or
+# $BROWSER setup required inside the sandbox. Installed at /usr/local/bin so
+# it precedes Ubuntu's stock /usr/bin/xdg-open on PATH. Requires
+# --security-opt apparmor=unconfined on the container (set in run-linux) so
+# the session-bus call is allowed through.
+RUN { \
+      echo '#!/bin/sh'; \
+      echo 'exec gdbus call --session \'; \
+      echo '  --dest org.freedesktop.portal.Desktop \'; \
+      echo '  --object-path /org/freedesktop/portal/desktop \'; \
+      echo '  --method org.freedesktop.portal.OpenURI.OpenURI \'; \
+      echo '  "" "$1" "{}"'; \
+    } > /usr/local/bin/xdg-open \
+    && chmod +x /usr/local/bin/xdg-open
 
 # ── Node.js + AI CLIs ────────────────────────────────────────────────────────
 # The bundled daemon inside this image runs the review pipeline, which shells
--- a/Dockerfile.linux-verify
+++ b/Dockerfile.linux-verify
@@ -87,3 +87,33 @@ RUN apt-get update -y \
     && apt-get update -y \
     && apt-get install -y --no-install-recommends gh \
     && rm -rf /var/lib/apt/lists/*
+
+# ── Node.js + AI CLIs ────────────────────────────────────────────────────────
+# The bundled daemon inside this image runs the review pipeline, which shells
+# out to whatever AI CLI the user configured as `ai.primary`. Without one on
+# PATH the pipeline fails fast at `detectCLI`:
+#
+#   pipeline: detect CLI: executor: no AI CLI available (tried "claude", "")
+#
+# The production docker/Dockerfile installs all four supported CLIs for exactly
+# this reason; mirror that set here so `make run-linux` is a faithful desktop
+# equivalent. Pinned Node version matches the production image at the time of
+# writing — bump both in lockstep when upgrading.
+RUN wget -q https://nodejs.org/dist/v20.18.0/node-v20.18.0-linux-x64.tar.xz -O /tmp/node.tar.xz \
+    && tar -C /usr/local --strip-components=1 -xJf /tmp/node.tar.xz \
+    && rm /tmp/node.tar.xz
+
+# Claude Code CLI (Anthropic) — auth via ANTHROPIC_API_KEY or CLAUDE_CODE_OAUTH_TOKEN
+RUN npm install -g @anthropic-ai/claude-code 2>/dev/null || true
+
+# Gemini CLI (Google) — auth via GEMINI_API_KEY
+RUN npm install -g @google/gemini-cli 2>/dev/null || true
+
+# Codex CLI (OpenAI) — auth via OPENAI_API_KEY or CODEX_API_KEY
+RUN npm install -g @openai/codex 2>/dev/null || true
+
+# OpenCode CLI — auth via any provider key above or OPENROUTER_API_KEY. Ubuntu
+# uses glibc so the musl-specific package the production (Alpine) image pulls
+# is not needed here; npm will pick the right binary from optionalDependencies.
+RUN npm install -g opencode-ai 2>/dev/null || true \
+    && npm cache clean --force 2>/dev/null || true
--- a/Dockerfile.linux-verify
+++ b/Dockerfile.linux-verify
@@ -78,7 +78,7 @@ RUN apt-get update -y \
     && apt-get install -y --no-install-recommends \
        libayatana-appindicator3-1 libnotify4 libsecret-1-0 \
        dbus-x11 libcanberra-gtk3-module mesa-utils \
-       curl gnupg \
+       curl gnupg xdg-utils \
     && curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
        | tee /usr/share/keyrings/githubcli-archive-keyring.gpg > /dev/null \
     && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
--- a/Dockerfile.linux-verify
+++ b/Dockerfile.linux-verify
@@ -67,7 +67,23 @@ RUN test -f daemon/bin/heimdallm \
 # ── Runtime deps for `make run-linux` GUI testing ────────────────────────────
 # These are NOT needed for CI build verification — only for displaying the app
 # on the host X11 display. Kept in a separate layer for clarity.
-RUN apt-get update -y && apt-get install -y --no-install-recommends \
-    libayatana-appindicator3-1 libnotify4 libsecret-1-0 \
-    dbus-x11 libcanberra-gtk3-module mesa-utils \
+#
+# gh is installed from GitHub's official apt repo so the Flutter desktop app's
+# FirstRunSetup.detectToken / RepoDiscovery paths (which shell out to
+# `gh auth token` and `gh search prs`) can resolve credentials and run repo
+# discovery inside the container. The user's ~/.config/gh is bind-mounted
+# read-only from the host by run-linux so the container picks up their
+# existing authentication without re-logging in.
+RUN apt-get update -y \
+    && apt-get install -y --no-install-recommends \
+       libayatana-appindicator3-1 libnotify4 libsecret-1-0 \
+       dbus-x11 libcanberra-gtk3-module mesa-utils \
+       curl gnupg \
+    && curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
+       | tee /usr/share/keyrings/githubcli-archive-keyring.gpg > /dev/null \
+    && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
+    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" \
+       > /etc/apt/sources.list.d/github-cli.list \
+    && apt-get update -y \
+    && apt-get install -y --no-install-recommends gh \
     && rm -rf /var/lib/apt/lists/*
--- a/Makefile
+++ b/Makefile
@@ -404,6 +404,9 @@ run-linux:
 	\
 	echo "DISPLAY=$$DISPLAY" > "$$ENV_FILE" ; \
 	echo "HOME=$$HOME" >> "$$ENV_FILE" ; \
+	if [ -n "$$XDG_CURRENT_DESKTOP" ]; then \
+	  echo "XDG_CURRENT_DESKTOP=$$XDG_CURRENT_DESKTOP" >> "$$ENV_FILE" ; \
+	fi ; \
 	echo "HEIMDALLM_DAEMON_PATH=/app/daemon/bin/heimdallm" >> "$$ENV_FILE" ; \
 	if [ -n "$$GITHUB_TOKEN" ]; then \
 	  echo "GITHUB_TOKEN=$$GITHUB_TOKEN" >> "$$ENV_FILE" ; \
--- a/Makefile
+++ b/Makefile
@@ -420,6 +420,10 @@ run-linux:
 	else \
 	  echo "⚠  /dev/dri not found — using software rendering (llvmpipe)." ; \
 	fi ; \
+	GH_CONFIG_ARGS="" ; \
+	if [ -d "$$HOME/.config/gh" ]; then \
+	  GH_CONFIG_ARGS="-v $$HOME/.config/gh:$$HOME/.config/gh:ro" ; \
+	fi ; \
 	\
 	echo "▶  Launching Heimdallm (Linux) via Docker..." ; \
 	echo "   Close the app window or press Ctrl-C to stop." ; \
@@ -434,6 +438,7 @@ run-linux:
 	  $$DBUS_ARGS \
 	  -v "$$HOME/.config/heimdallm:$$HOME/.config/heimdallm" \
 	  -v "$$HOME/.local/share/heimdallm:$$HOME/.local/share/heimdallm" \
+	  $$GH_CONFIG_ARGS \
 	  $$GPU_ARGS \
 	  --ipc=host \
 	  --net=host \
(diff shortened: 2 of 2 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Clean infrastructure PR that adds `gh` CLI to the Linux dev container and wires GitHub credentials via bind-mounted config and token fallback. The changes are well-scoped, properly documented in comments, and follow existing Makefile patterns. No blocking issues found.

### Suggestions

- Dockerfile.linux-verify:80 — Consider pinning the GPG key fingerprint or sha256 checksum after the curl download (`echo '<expected-hash> /usr/share/keyrings/githubcli-archive-keyring.gpg' | sha256sum -c`). Currently the key is fetched over HTTPS from GitHub which is reasonable for a dev image, but pinning adds defense-in-depth against a compromised endpoint.
- Makefile:423 — `$$GH_CONFIG_ARGS` is expanded unquoted in the docker run invocation. If `$HOME` ever contains spaces or special characters the volume mount will break due to word splitting. Consider quoting or using an array-style approach. Low risk on Linux where HOME paths rarely contain spaces.
- Makefile:400 — Setting `HOME=$$HOME` inside the container overrides the container user's HOME to match the host. This is intentional for path-aligned bind mounts, but worth a brief inline comment noting the coupling so future maintainers don't accidentally remove it while the bind mounts still depend on it.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Reasonable changes to make run-linux usable end-to-end. The gh CLI installation follows GitHub's official method, token fallback is cleanly handled, and bind mounts are read-only. A few minor concerns around HOME path handling and robustness.

### Issues

⚠️ **Makefile:400** — Setting HOME to the host's path (e.g. /home/user) inside the container means HOME points to a directory that may not fully exist in the container filesystem. Tools that attempt to write to $HOME/.cache, $HOME/.local/state, or other non-bind-mounted subdirectories will fail. Consider creating the HOME directory inside the container or documenting this limitation.
🟡 **Makefile:423** — GH_CONFIG_ARGS is expanded unquoted in the docker run command. If HOME contains spaces, the -v argument will be split incorrectly. This matches the pre-existing pattern for GPU_ARGS/DBUS_ARGS, but all are fragile for paths with spaces.
🟡 **Makefile:400** — The token is both extracted via `gh auth token` into GITHUB_TOKEN env var (host-side) AND the gh config dir is bind-mounted (container-side). If only one mechanism is needed, the redundancy could cause confusion if they diverge (e.g., user re-auths on host mid-run). Consider a comment clarifying that GITHUB_TOKEN is for the Flutter app while the bind-mount serves the gh CLI inside the container.
🟡 **Dockerfile.linux-verify:80** — The gh CLI is installed without version pinning. A breaking change in a future gh release could silently affect container behavior. Consider pinning to a known-good version (e.g., `apt-get install -y gh=2.x.y`).

### Suggestions

- Add `mkdir -p "$$HOME"` (or a writable tmpdir as HOME) before the docker run to ensure the HOME directory tree exists inside the container, preventing failures from tools writing to unmounted HOME subdirectories.
- Add a brief inline comment in the Makefile explaining that GITHUB_TOKEN serves the Flutter app directly while the gh config bind-mount enables `gh` CLI commands inside the container, since both mechanisms serve different consumers.
- Consider verifying the GPG key fingerprint after download in the Dockerfile to guard against key substitution (defense in depth).

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean infrastructure change that adds GitHub CLI to the Linux verification container and forwards host gh credentials via read-only bind mount. The implementation follows the official gh installation method, properly guards optional paths, and is consistent with existing patterns in the Makefile. No blocking issues found.

### Suggestions

- Dockerfile.linux-verify: Consider pinning the gh package version (e.g., `gh=2.*`) to avoid unexpected breakage from major version bumps in CI or local builds.
- Makefile:423-425: GH_CONFIG_ARGS is expanded unquoted in the docker run invocation — consistent with existing GPU_ARGS/DBUS_ARGS patterns, but all of them will break if HOME contains spaces. Low risk on Linux, but quoting the individual -v argument inside the variable would harden it.
- Makefile:400: Passing HOST's HOME as the container HOME means the container user must have matching UID/GID to access the mounted paths — already handled by the existing --user flag, but worth a comment for future maintainers.

---

vbuenog:
Three additional commits pushed — unblocks the full review pipeline end-to-end. Testing a manual "Review now" now completes against real Claude and publishes to GitHub; the daemon log confirms a successful `pipeline: review published to GitHub` after ~30s.

### What each one fixes

- **`5953571` `feat(docker): install AI CLIs and forward auth so review pipeline runs`**
  Root cause of the current "stuck reviewing" state: `heimdallm-verify` never had an AI CLI installed, so the daemon's pipeline failed at `detectCLI` with `no AI CLI available (tried "claude", "")` and exited in ~1 s. Adds Node 20 + the same four CLIs the production image ships (claude-code, gemini-cli, codex, opencode-ai), and forwards `ANTHROPIC_API_KEY`, `CLAUDE_CODE_OAUTH_TOKEN`, `OPENAI_API_KEY`, `CODEX_API_KEY`, `GEMINI_API_KEY`, `OPENROUTER_API_KEY` from shell env / docker/.env into the container. Image size: +~180 MB.

- **`1f92363` `fix(make): mount AI CLI state dirs so claude et al can write on first use`**
  `--user $UID:$GID` + the existing narrow bind mounts leave `$HOME` itself root-owned. Installing claude surfaced this: it exits silently with empty stdout (→ daemon "parse JSON result: unexpected end of JSON input") because it can't create its sibling state dirs. Mirrors the production image's `~/` layout by mounting `~/.claude`, `~/.gemini`, `~/.codex`, `~/.config/opencode`, `~/.local/share/opencode` RW — same trust model as running the CLI natively, and seamlessly reuses any existing host auth.

- **`326d1bb` `fix(make): mount ~/.claude.json so claude can persist its main config`**
  Final sibling claude needed — its main config is `$HOME/.claude.json` (a file, not inside `.claude/`). Touched on host pre-mount for idempotence and bind-mounted file-level into the container. Other CLIs don't have this pattern; only claude needed the extra mount.

### End-to-end verification

```
$ docker exec heimdallm-run claude -p "say hi" --output-format json
{"type":"result","subtype":"success","is_error":false,...
  "result":"Hi! How can I help you today?", ...
  "total_cost_usd":0.03286125, ...
  "duration_ms":2702}
```

Then clicking "Review now" in the Flutter UI → daemon log shows `pipeline: review published to GitHub` with a valid `github_review_id` and `github_review_state`.

### Total footprint of this PR

Six commits, still only `Makefile` and `Dockerfile.linux-verify`. Zero changes to daemon, Flutter, web_ui. All changes are localised to the `run-linux` target and the verify image's runtime layer (CI builds unaffected).

### Known follow-up (not in this PR)

The Flutter UI's "reviewing…" spinner doesn't always clear when the daemon emits `review_error`. With the pipeline now succeeding on the happy path, this is less impactful day-to-day, but still worth a separate issue/PR — likely in `dashboard_providers.dart` or wherever `reviewingPRs` is populated.

Muriano:
## ⚠️ MEDIUM Issue

Node.js tarball downloaded via wget without SHA256 checksum verification. A compromised mirror or MITM could inject a malicious binary. Add `&& echo '<expected_sha256>  /tmp/node.tar.xz' | sha256sum -c -` after the download. The checksum is published alongside the tarball on nodejs.org/dist.

**Location:** `Dockerfile.linux-verify` line 87

---

Muriano (approved):

Solid infrastructure PR that unblocks the review pipeline in run-linux. Changes are well-scoped to Dockerfile.linux-verify and Makefile. The main concern is downloading a Node.js binary without checksum verification, which is a supply-chain risk even in a dev/verify image.

### Suggestions

- Combine the four separate `npm install -g` RUN commands into a single RUN layer followed by `npm cache clean --force`. Currently each RUN creates a Docker layer that includes npm cache from that install; only the last layer cleans it. A single layer would reduce image size.
- The GPG key for the GitHub CLI apt repo is fetched via curl without fingerprint verification. Consider adding `gpg --with-colons --import-options show-only --import | grep -q '<expected_fingerprint>'` to pin the key.
- Comment says 'Pinned Node version matches the production image — bump both in lockstep'. Consider extracting the version into a build ARG (`ARG NODE_VERSION=20.18.0`) shared with the production Dockerfile, or at minimum adding a comment with the exact location of the production pin so maintainers can find it.

---

Muriano:
## ⚠️ MEDIUM Issue

All four `npm install -g` commands use `2>/dev/null || true`, silently swallowing failures. If all installs fail (e.g. npm registry unreachable, broken package), the image builds successfully with zero CLIs, and the pipeline will fail at runtime with a confusing 'no AI CLI available' error. Consider emitting a warning or checking that at least one CLI binary exists after the installs: `RUN which claude || which gemini || which codex || which opencode || echo 'WARN: no AI CLI installed'`.

**Location:** `Dockerfile.linux-verify` line 91

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR makes `make run-linux` usable end-to-end by installing gh CLI + Node.js + four AI CLIs in the verify image, forwarding auth tokens, and bind-mounting CLI state dirs. Scope is correctly limited to Makefile and Dockerfile.linux-verify. Changes are pragmatic and well-documented inline, but have several supply-chain and robustness concerns worth addressing.

### Issues

⚠️ **Dockerfile.linux-verify:87** — Node.js tarball downloaded via wget without SHA256 checksum verification. A compromised mirror or MITM could inject a malicious binary. Add `&& echo '<expected_sha256>  /tmp/node.tar.xz' | sha256sum -c -` after download using the checksum from nodejs.org/dist/v20.18.0/SHASUMS256.txt.
⚠️ **Dockerfile.linux-verify:91** — All four `npm install -g` commands use `2>/dev/null || true`, silently swallowing failures. If all installs fail (npm registry unreachable, broken package), the image builds successfully with zero CLIs and runtime fails with the same confusing `no AI CLI available` error this PR is trying to fix. Add a verification step: `RUN command -v claude || command -v gemini || command -v codex || command -v opencode || { echo 'ERROR: no AI CLI installed'; exit 1; }`.
🟡 **Makefile:420** — The docker/.env parsing uses `cut -d= -f2-` without handling quoted values or trimming whitespace. If a user has `ANTHROPIC_API_KEY="sk-..."` or trailing spaces in docker/.env, the surrounding quotes/spaces leak into the container env and the CLI auth silently fails. Consider using `sed` to strip surrounding quotes, or document that docker/.env must be unquoted KEY=VALUE.
🟡 **Makefile:406** — `HOME=$$HOME` is forwarded from host into container but the container user is `--user $UID:$GID`. If the host `$HOME` path does not exist inside the container image (e.g. non-standard username), bind mounts to `$$HOME/.claude` etc. will still work because docker creates the parent, but any CLI that calls `getpwuid(getuid())` inside the container will get a mismatched entry. Low risk since bind mounts drive the actual behavior, but worth noting for non-standard setups.
🟡 **Dockerfile.linux-verify:103** — Pinned Node v20.18.0 in a comment says to 'bump both in lockstep' with production, but there's no automated check or reference to the production Dockerfile version. Drift is likely over time. Consider extracting the Node version to an ARG at the top of the file or adding a CI check that both Dockerfiles pin the same version.

### Suggestions

- Add SHA256 verification for the Node.js tarball download using the official SHASUMS256.txt from nodejs.org.
- Fail the build if none of the four AI CLIs installed successfully, rather than silently producing a broken image.
- Extract the Node.js version to a Dockerfile ARG and document/enforce keeping it in sync with docker/Dockerfile.
- Document that docker/.env values must be unquoted, or harden the parser to strip surrounding quotes.
- Consider caching the Node.js tarball download in a separate RUN layer to speed up rebuilds when only AI CLI versions change.

---

Muriano:
## ⚠️ MEDIUM Issue

Node.js tarball downloaded via wget without SHA256 checksum verification. A compromised mirror or MITM attack could inject a malicious binary into the image. Fix: add `&& echo '<expected_sha256>  /tmp/node.tar.xz' | sha256sum -c -` after the wget, using the checksum from nodejs.org/dist/v20.18.0/SHASUMS256.txt.

**Location:** `Dockerfile.linux-verify` line 87

---

Muriano:
## ⚠️ MEDIUM Issue

All four `npm install -g` commands silently swallow errors with `2>/dev/null || true`. If npm registry is unreachable or all packages fail, the image builds with zero CLIs installed and the pipeline will fail at runtime with an opaque 'no AI CLI available' error. Add a post-install guard: `RUN which claude || which gemini || which codex || which opencode || { echo 'ERROR: no AI CLI installed'; exit 1; }`.

**Location:** `Dockerfile.linux-verify` line 91

---

Muriano (approved):

Infrastructure-only PR adding AI CLI tooling and auth forwarding to the Linux verify/run image. The changes are well-scoped and documented. Two medium issues were already flagged in the PR discussion (missing Node.js checksum, silent npm failures). I confirm those and add one more regarding bind-mount security.

### Suggestions

- Pin Node.js version in a build ARG (`ARG NODE_VERSION=20.18.0`) so both Dockerfile.linux-verify and docker/Dockerfile can reference the same value and be bumped in lockstep as the comment suggests.
- The `npm cache clean --force` at line 100 only runs after `opencode-ai` install due to `|| true` chaining. If opencode-ai fails, `npm cache clean` still runs, but if it also fails the whole RUN succeeds. Move cache cleanup to a separate RUN for clarity.
- Consider adding a comment in the Makefile noting which env vars are forwarded, so future maintainers don't have to trace the loop to know which keys are supported.

---

Muriano:
## 🟡 LOW Issue

AI CLI config directories (~/.claude, ~/.gemini, ~/.codex, ~/.config/opencode, ~/.local/share/opencode) are bind-mounted read-write. If any CLI has a vulnerability or misbehaves, it can modify host-side config/auth files. These should be mounted read-only (`:ro`) unless there's a confirmed need for write access during normal operation. Claude's `.claude.json` is the only one mentioned as needing writes; the rest could be `:ro`.

**Location:** `Makefile` line 457

---

vbuenog:
Three more commits pushed — "Open in GitHub" and, as a bonus, the host GTK theme + fonts + tray + notifications all work inside the container now. Ten commits total on this branch.

### What the final set does

- **`d6344a9`** — install `xdg-utils` and forward `XDG_CURRENT_DESKTOP`. First attempt at making browser-open work; turned out not to be enough on its own but both pieces are still needed (xdg-utils provides the stock binary, `XDG_CURRENT_DESKTOP` is what GTK reads for theme detection).

- **`503950e`** — install `libglib2.0-bin` (for `gdbus`), override `/usr/local/bin/xdg-open` with a tiny shim that calls the XDG Desktop Portal directly, and add `--security-opt apparmor=unconfined` to the `docker run`. The AppArmor bit is the critical one: Ubuntu 23.10+ blocks sandboxed containers from reaching the host's user session bus by default, which was silently breaking the portal call, tray, and notifications all together.

- **`a83946b`** — the missing piece. Flutter's `url_launcher_linux` doesn't shell out to `xdg-open`; it calls `gtk_show_uri_on_window` which goes through glib's `g_app_info_launch_default_for_uri`. That uses the local MIME database (empty in the image) and never reached our shim. glib has a well-known sandbox sentinel: if `/.flatpak-info` exists, it routes URI opens through the portal. Touching a minimal valid keyfile at `/.flatpak-info` flips the switch — without it, nothing; with it, `Open in GitHub` opens the host's default browser.

### Side effects of the glib sandbox detection + AppArmor relaxation (all positive)

- Tray icon appears in the host's status area.
- Desktop notifications work (Linux `libnotify` toasts).
- GTK picks up the host's theme, fonts, dark mode, DPI — the in-container UI looks native instead of GTK fallback defaults.
- The `libayatana-appindicator-WARNING` / `LIBDBUSMENU-GLIB-WARNING` / `dbind-WARNING` lines in the startup log are gone.

### Final footprint

Ten commits, still only `Makefile` and `Dockerfile.linux-verify`. No source-code changes, no new files in the repo. Image adds ~180 MB (Node + four AI CLIs + libs) and a single-line marker file; that's the full cost of making `make run-linux` behave as a drop-in desktop runtime.

### Known limitation / future direction

This PR treats the Docker container as a full desktop runtime, which works now but stretches `heimdallm-verify` beyond its original purpose (CI build verification). A follow-up `make install-linux` target — `docker cp` the built bundle out and install it as a normal Linux app on the host — is the architecturally cleaner answer for long-term daily use. Tracking that separately; this PR lands the container-based flow that works today.

### Verified end-to-end

- `make verify-linux` → builds clean, all four AI CLIs installed.
- `make run-linux` → window appears with host theme/fonts.
- Auth: GitHub token auto-resolved via host `gh`, Claude OAuth forwarded from `docker/.env`.
- Repositories → Discover → finds review-requested repos including `theburrowhub/heimdallm`.
- Manual "Review now" → claude runs (~30s) → `pipeline: review published to GitHub` with a real `github_review_id`.
- Open in GitHub → host browser opens the URL on the host desktop.
- Tray icon + notifications present.

Issue #113, opened 2026-04-20
feat: Activity log — daily records + AI-generated executive reports


Muriano:
## Summary

Add a daily activity log that records all Heimdallm actions, and an AI-powered report generator that produces executive summaries from that log (daily standup notes, reports for managers, etc).

## Motivation

Users need visibility into what Heimdallm did each day across all repos — which PRs were reviewed, which issues triaged, which implementations triggered. Currently this information is scattered across the Reviews/Issues tabs with no daily aggregation. Generating a human-readable summary for standups or management reports requires manual work.

### 1. Activity log (daemon)

The daemon records a structured log entry for every significant action:

- PR review completed (repo, PR number, title, severity, CLI used)
- Issue triaged (repo, issue number, title, triage severity, category)
- Auto-implement triggered (repo, issue number, title, PR created)
- Issue promoted (repo, issue number, from label → to label)
- Review error (repo, item, error reason)

Storage: new `activity_log` SQLite table with columns:
- `id`, `timestamp`, `org`, `repo`, `item_type` (pr/issue), `item_number`, `item_title`, `action` (review/triage/implement/promote/error), `details` (JSON), `created_at`

Indexed by date for efficient daily queries.

### 2. Activity log endpoint (daemon)

New endpoints:
- `GET /activity?date=2026-04-20` — returns all log entries for a day
- `GET /activity?from=2026-04-18&to=2026-04-20` — date range
- `GET /activity?org=freepik-company&repo=ai-bumblebee-proxy` — filtered

### 3. Activity log screen (Flutter)

New tab or screen accessible from the navigation:

- **Date picker** — select day or date range
- **Filters** — Organization (multi-select), Repository (multi-select)
- **Timeline view** — chronological list of actions grouped by hour
- Each entry shows: timestamp, action icon, repo, item title, outcome

### 4. AI report generation

A "Generate Report" button that:

1. Collects the filtered activity log entries
2. Sends them to the configured AI agent with a prompt like:
   > "Generate an executive summary of the following engineering activity for [date]. Include: key accomplishments, issues triaged, PRs reviewed, implementations triggered, and any notable findings. Format for a daily standup update."
3. Displays the generated report in a dialog/panel
4. Copy-to-clipboard button for pasting into Slack/email

The report prompt should be configurable (an Agent profile category, like PR Review/Triage/Develop — this would be a 4th prompt type: "Report").

## Configuration

```toml
[activity_log]
enabled = true
[...]

PR #120, opened 2026-04-20
feat: activity log foundation (store + issue_promoted event)

vbuenog:
## Summary

Phase A of the activity log feature (#113). Adds the storage foundation — `activity_log` SQLite table, store read/write/purge operations — and a new `issue_promoted` SSE event so future consumers can record label flips in the timeline. No runtime behavior change for existing features.

AI report generation is **explicitly out of scope** and deferred to a follow-up spec — see `docs/superpowers/specs/2026-04-20-activity-log-design.md` for the full design.

Subsequent phases (recorder, config + retention, HTTP endpoint, Flutter tab) will land as stacked PRs on top of this one.

## Changes

- **New table**: `activity_log` (timestamp, org, repo, item_type, item_number, item_title, action, outcome, details JSON) with indexes on `ts DESC` and `(repo, ts DESC)`.
- **New store API**: `InsertActivity`, `ListActivity` (filtered by org/repo/action + time range, over-fetch-by-one truncation), `PurgeOldActivity`. Pattern mirrors `reviews.go`.
- **New SSE event**: `issue_promoted`. Emitted by `PromoteReady` after every successful blocked → promote-to label flip; payload includes `repo`, `issue_number`, `issue_title`, `from_label`, `to_label`, `reason`.
- **Callers updated**: `PromoteReady` gains a trailing `Publisher` parameter; `main.go` passes the shared broker, existing tests pass `nil`.
- **Docs**: design spec + implementation plan under `docs/superpowers/`.

## Test plan

- [x] `make test-docker` — full daemon suite passes after each commit
- [x] New tests cover: insert round-trip, filter by org+action, filter by repo, truncation, purge cutoff, purge zero-is-noop, `issue_promoted` event payload shape
- [ ] Smoke-test on a running daemon (deferred to the Phase C PR where the recorder wires up and rows actually land)

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -380,7 +380,7 @@ func main() {
 				// No-op when BlockedLabels is unset, so default installs
 				// don't pay for the extra calls.
 				if len(globalIT.BlockedLabels) > 0 {
-					if n, err := issuepipeline.PromoteReady(context.Background(), ghClient, globalIT, repos); err != nil {
+					if n, err := issuepipeline.PromoteReady(context.Background(), ghClient, globalIT, repos, broker); err != nil {
 						slog.Error("poll: issue promotion failed", "err", err)
 					} else if n > 0 {
 						slog.Info("poll: promoted issues", "count", n)
--- a/daemon/internal/issues/promoter.go
+++ b/daemon/internal/issues/promoter.go
@@ -2,12 +2,14 @@ package issues
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"strings"
 
 	"github.com/heimdallm/daemon/internal/config"
 	"github.com/heimdallm/daemon/internal/github"
+	"github.com/heimdallm/daemon/internal/sse"
 )
 
 // PromoteIssueClient is the subset of *github.Client that PromoteReady
@@ -35,7 +37,7 @@ type PromoteIssueClient interface {
 // The call is a no-op when issue tracking is disabled, when no blocked
 // labels are configured, or when every repo in the list is empty — keeps
 // default installs unaffected.
-func PromoteReady(ctx context.Context, c PromoteIssueClient, cfg config.IssueTrackingConfig, repos []string) (int, error) {
+func PromoteReady(ctx context.Context, c PromoteIssueClient, cfg config.IssueTrackingConfig, repos []string, broker Publisher) (int, error) {
 	if !cfg.Enabled || len(cfg.BlockedLabels) == 0 {
 		return 0, nil
 	}
@@ -129,6 +131,19 @@ func PromoteReady(ctx context.Context, c PromoteIssueClient, cfg config.IssueTra
 			slog.Info("issues promote: promoted issue",
 				"repo", repo, "issue", issue.Number,
 				"from", blockedOnIssue, "to", promoteTo)
+			if broker != nil {
+				payload := map[string]any{
+					"repo":         repo,
+					"issue_number": issue.Number,
+					"issue_title":  issue.Title,
+					"from_label":   strings.Join(blockedOnIssue, ","),
+					"to_label":     promoteTo,
+					"reason":       "dependencies closed",
+				}
+				if b, err := json.Marshal(payload); err == nil {
+					broker.Publish(sse.Event{Type: sse.EventIssuePromoted, Data: string(b)})
+				}
+			}
 		}
 	}
 	return promoted, nil
--- a/daemon/internal/issues/promoter_test.go
+++ b/daemon/internal/issues/promoter_test.go
@@ -2,6 +2,7 @@ package issues
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"sort"
@@ -10,6 +11,7 @@ import (
 
 	"github.com/heimdallm/daemon/internal/config"
 	"github.com/heimdallm/daemon/internal/github"
+	"github.com/heimdallm/daemon/internal/sse"
 )
 
 // fakePromoteClient is a minimal in-memory PromoteIssueClient. Records
@@ -109,7 +111,7 @@ func TestPromoteReady_Disabled_Noop(t *testing.T) {
 	cfg := baseCfg()
 	cfg.Enabled = false
 
-	n, err := PromoteReady(context.Background(), fake, cfg, []string{"org/r"})
+	n, err := PromoteReady(context.Background(), fake, cfg, []string{"org/r"}, nil)
 	if err != nil || n != 0 {
 		t.Errorf("got n=%d err=%v, want 0, nil", n, err)
 	}
@@ -123,7 +125,7 @@ func TestPromoteReady_NoBlockedLabels_Noop(t *testing.T) {
 	cfg := baseCfg()
 	cfg.BlockedLabels = nil
 
-	n, err := PromoteReady(context.Background(), fake, cfg, []string{"org/r"})
+	n, err := PromoteReady(context.Background(), fake, cfg, []string{"org/r"}, nil)
 	if err != nil || n != 0 {
 		t.Errorf("got n=%d err=%v, want 0, nil", n, err)
 	}
@@ -135,7 +137,7 @@ func TestPromoteReady_BlockedIssueWithOpenDep_NotPromoted(t *testing.T) {
 		open:  map[string][]*github.Issue{"org/r": {blocked}},
 		byRef: map[string]*github.Issue{"org/r#5": mkIssue("org/r", 5, "open", "")},
 	}
-	n, err := PromoteReady(context.Background(), fake, baseCfg(), []string{"org/r"})
+	n, err := PromoteReady(context.Background(), fake, baseCfg(), []string{"org/r"}, nil)
 	if err != nil {
 		t.Fatalf("err: %v", err)
 	}
@@ -156,7 +158,7 @@ func TestPromoteReady_AllDepsClosed_Promoted(t *testing.T) {
 			"org/r#6": mkIssue("org/r", 6, "closed", ""),
 		},
 	}
-	n, err := PromoteReady(context.Background(), fake, baseCfg(), []string{"org/r"})
+	n, err := PromoteReady(context.Background(), fake, baseCfg(), []string{"org/r"}, nil)
 	if err != nil {
 		t.Fatalf("err: %v", err)
 	}
@@ -182,7 +184,7 @@ func TestPromoteReady_NoDepsDeclared_NotPromoted(t *testing.T) {
 	fake := &fakePromoteClient{
 		open: map[string][]*github.Issue{"org/r": {blocked}},
 	}
-	n, err := PromoteReady(context.Background(), fake, baseCfg(), []string{"org/r"})
+	n, err := PromoteReady(context.Background(), fake, baseCfg(), []string{"org/r"}, nil)
 	if err != nil {
 		t.Fatalf("err: %v", err)
 	}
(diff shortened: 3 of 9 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid foundation for the activity log feature. Clean store layer with good test coverage, proper index strategy, and well-tested promoter event emission. A few data-modeling and error-handling nits to address before merge.

### Issues

⚠️ **daemon/internal/issues/promoter.go:141** — json.Marshal error is silently swallowed. If marshalling fails, no log or metric is emitted — this makes debugging silent event-loss very difficult. Add at least a `slog.Warn` in the else branch.
⚠️ **daemon/internal/issues/promoter.go:138** — `from_label` uses `strings.Join(blockedOnIssue, ",")` producing an ambiguous comma-separated string. Since the payload is JSON-marshalled anyway, pass the `[]string` slice directly so it serializes as a proper JSON array. This preserves structure for downstream consumers and avoids ambiguity if a label ever contains a comma.
🟡 **daemon/internal/store/activity.go:55** — `InsertActivity` calls `time.Now().UTC()` for `created_at`, and `PurgeOldActivity` calls `time.Now().UTC()` for the cutoff. This couples store methods to wall-clock time, making deterministic testing harder. Consider accepting a `now` parameter or a clock interface, at least for `PurgeOldActivity` where the test relies on a 100-day gap to avoid flakiness.
🟡 **daemon/internal/store/store.go:114** — There is an index on `(ts DESC)` and `(repo, ts DESC)`, but `ListActivity` also supports filtering by `org` alone. Queries that filter only by org will not benefit from either index and will scan the full table. Consider adding `idx_activity_org_ts ON activity_log(org, ts DESC)` for symmetry, or document that org-only filtering is expected to always co-occur with a time range.
🟡 **daemon/internal/store/activity.go:158** — `placeholders(0)` returns an empty string, which would produce `IN ()` — invalid SQL. Currently safe because callers guard with `len > 0`, but the helper has no internal safety net. Consider returning an error or panicking for n <= 0 to prevent future misuse.
🟡 **docs/superpowers/plans/2026-04-20-activity-log.md:1** — 2888-line implementation plan checked into the repo. Plans that include inline code snippets and step-by-step commands rot quickly once the code diverges. Consider trimming to high-level architecture decisions only, or keeping this in an issue/wiki instead of the source tree.

### Suggestions

- In promoter.go, log a warning when json.Marshal fails instead of silently dropping the event: `} else { slog.Warn("issues promote: marshal SSE payload", "err", err) }`
- Change `from_label` payload to pass `blockedOnIssue` as a `[]string` directly (e.g. `"from_labels": blockedOnIssue`), letting json.Marshal produce a JSON array. Update the test assertion to match the new field name/type.
- Consider defining a `Publisher` no-op implementation (e.g. `NopPublisher`) and using it as the default instead of nil-checking at every call site — removes the nil guard and simplifies future event-emission points.
- Add one more test case: verify that when `broker` is non-nil but the issue is NOT promoted (e.g. deps still open), no event is emitted. This ensures the event is only fired on the success path.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured addition of activity_log store layer and promoter SSE event emission. Tests are thorough. No blocking issues found — only minor improvements.

### Suggestions

- promoter.go:141 — json.Marshal error on the SSE payload is silently swallowed (`if b, err := json.Marshal(payload); err == nil`). While failure is extremely unlikely for this map[string]any of primitives, a slog.Warn would help during debugging if the payload shape ever changes.
- activity.go — PurgeOldActivity does not guard against negative maxDays. A negative value would compute a future cutoff and delete recent rows. Consider adding `if maxDays < 0 { return nil }` or `if maxDays <= 0`.
- activity.go — Consider adding an index on `org` (`CREATE INDEX IF NOT EXISTS idx_activity_org_ts ON activity_log(org, ts DESC)`) since ListActivity supports Orgs-only filtering without Repos.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid foundation for an activity log. Schema, store CRUD, and promoter event emission are well-tested. A few quality issues around silent error handling in the SSE payload emission and the shape of the `from_label` field.

### Issues

🟡 **daemon/internal/issues/promoter.go:145** — JSON marshal error is silently swallowed (`if b, err := json.Marshal(payload); err == nil`). If marshalling ever fails the event is dropped with no log — at minimum log at debug/warn so the miss is diagnosable. Given the payload is map[string]any with only primitive types, this is unlikely to fail in practice, but the silent drop is a bad pattern.
⚠️ **daemon/internal/issues/promoter.go:140** — `from_label` is emitted as a comma-joined string (`strings.Join(blockedOnIssue, ",")`). Consumers will have to re-split, and labels that legitimately contain commas would be ambiguous. Prefer emitting a `from_labels` JSON array for consistency with the underlying data model.
⚠️ **daemon/internal/issues/promoter.go:134** — The SSE event is published synchronously inside the promotion loop. If `broker.Publish` is blocking (unbuffered channel / slow subscriber), a single stuck consumer would stall the entire promotion pass. Verify the broker's Publish is non-blocking (drop-on-full or async), otherwise promotion throughput becomes coupled to SSE client health.
🟡 **daemon/internal/store/activity.go:178** — `placeholders(n)` returns "" for n<=0 which would produce invalid SQL `IN ()`. Currently safe because all call sites gate on `len(...) > 0`, but the helper is a footgun — either panic on n<=0 or have callers skip the clause entirely (already done). Minor defensive concern.
🟡 **daemon/internal/store/activity.go:24** — `ActivityQuery` comment says unbounded queries "only happen in tests" but nothing enforces this at the store layer. If a future handler forgets to set From/To, the query scans the full table. Consider a default time window (e.g., last 30 days) when both are zero, or a comment/TODO pointing to the handler contract.
🟡 **daemon/internal/store/activity.go:45** — `InsertActivity` has 9 positional arguments plus a map — easy to swap `itemTitle`/`action`/`outcome` at call sites with no compiler help. Consider taking an `Activity` struct (or at least bundling the string fields) to make call sites self-documenting and reduce argument-order bugs.
🟡 **daemon/internal/store/activity.go:156** — `PurgeOldActivity(0)` is a no-op meaning "keep forever". This is a subtle contract — a caller reading the config key `activity_log_retention_days=0` and expecting "purge all" would be wrong. Worth a doc comment on the config key and/or renaming the sentinel (e.g., negative = disabled).
🟡 **daemon/internal/issues/promoter_test.go:608** — `fakePromoteBroker.Publish` appends to a slice without a mutex. Fine for the current single-goroutine test, but if future tests exercise concurrent promotion, this will race. A `sync.Mutex` costs nothing here.

### Suggestions

- Emit `from_labels` as a JSON array instead of a comma-joined string in the issue_promoted payload.
- Log (at warn or debug) when the payload fails to marshal instead of silently dropping the event.
- Document (or enforce) Publisher non-blocking semantics so callers know promotion won't stall on a slow SSE subscriber.
- Consider an `Activity` struct parameter for `InsertActivity` to make the 9-positional-arg call sites safer.
- Add a test covering `PurgeOldActivity` with a non-default retention so the no-op semantic of 0 is explicit in tests.

---

Muriano:
## ⚠️ MEDIUM Issue

PurgeOldActivity only guards against maxDays == 0 but not negative values. A negative maxDays would produce a cutoff in the future, silently deleting recent activity rows. Add `if maxDays <= 0 { return nil }` instead.

**Location:** `daemon/internal/store/activity.go` line 149

---

Muriano (approved):

Clean, well-tested addition of activity_log store layer and SSE event emission from the issue promoter. Good use of parameterized queries, nil-safe broker check, and comprehensive test coverage. A couple of minor defensive-coding gaps.

### Suggestions

- daemon/internal/store/activity.go: Consider adding an index on `org` (or a composite `org, ts DESC`) since ListActivity supports org-only filtering without repos — currently only `repo, ts` is indexed.
- daemon/internal/issues/promoter.go:141 — json.Marshal error is silently swallowed. While it cannot fail with this payload in practice, a brief slog.Warn would aid debugging if the payload shape ever changes.
- daemon/internal/store/activity.go: The Activity struct documents allowed values for ItemType and Action only in the plan doc, not in code. A short comment or package-level constants would make valid values discoverable.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

No changes since last review. Previous findings remain unaddressed, and the new reviewer comment about negative maxDays in PurgeOldActivity is correct: the current `if maxDays == 0` guard allows negative values through, producing a future-dated cutoff that would silently delete recent rows. Re-flagging only the still-valid items — two MEDIUM (from_label shape, synchronous SSE publish) plus Muriano's MEDIUM (negative maxDays), and selected LOW items still present verbatim in the code.

### Issues

⚠️ **daemon/internal/store/activity.go:149** — PurgeOldActivity only guards `maxDays == 0`. A negative value passes through and produces a cutoff in the future (`now - (-N)*24h = now + N*24h`), which would delete every row whose ts is before that future time — i.e., all recent activity. Confirmed by Muriano. Change to `if maxDays <= 0 { return nil }` (and update the doc comment to reflect that non-positive values are a no-op). If you ever want a distinct 'disabled' vs 'purge all' semantics, make that explicit with a sentinel or separate function.
⚠️ **daemon/internal/issues/promoter.go:140** — Still unaddressed from prior review: `from_label` is emitted as a comma-joined string (`strings.Join(blockedOnIssue, ",")`). This is lossy for labels that contain commas, forces consumers to re-split, and is inconsistent with the underlying `[]string` data model. Prefer a `from_labels` JSON array field.
⚠️ **daemon/internal/issues/promoter.go:134** — Still unaddressed: `broker.Publish` is invoked synchronously inside the promotion loop. If the broker implementation blocks on a slow/stuck subscriber, a single misbehaving SSE client would stall the entire promotion pass. Confirm the broker's Publish is non-blocking (drop-on-full or buffered-async); otherwise decouple with a goroutine or buffered channel.
🟡 **daemon/internal/issues/promoter.go:145** — Still unaddressed: `json.Marshal` error is silently swallowed via `if b, err := json.Marshal(payload); err == nil`. Log at warn/debug so a dropped event is diagnosable — current payload uses only primitives so this is unlikely to trigger, but the silent drop is a bad pattern.
🟡 **daemon/internal/store/activity.go:45** — Still unaddressed: `InsertActivity` has 9 positional arguments (plus a map). `itemTitle`/`action`/`outcome` are all strings and are trivially swappable at call sites with no compiler help. Consider taking an `Activity` (or a small params struct) to make call sites self-documenting.
🟡 **daemon/internal/issues/promoter_test.go:608** — Still unaddressed: `fakePromoteBroker.Publish` appends to a slice without a mutex. Safe for the current single-goroutine test, but if publishes ever move to a goroutine (see the sync-publish concern above), tests using this fake will race. A `sync.Mutex` costs nothing here.

### Suggestions

- Fix the negative-maxDays guard (`<= 0`) — this is the most concrete bug in the diff and was correctly flagged by Muriano.
- Switch `from_label` string → `from_labels` array before any consumer ships, to avoid a breaking event-schema change later.
- Verify broker.Publish is non-blocking; if not, move the publish off the promotion goroutine.

---

Muriano:
## ⚠️ MEDIUM Issue

STILL OPEN from prior review: PurgeOldActivity guards only `maxDays == 0` but not negative values. A negative maxDays produces a cutoff in the future, silently deleting recent/all activity rows. Fix: change to `if maxDays <= 0 { return nil }`.

**Location:** `daemon/internal/store/activity.go` line 149

---

Muriano (approved):

Re-review: the previous finding about PurgeOldActivity accepting negative maxDays remains unaddressed. The rest of the diff (activity_log schema, store CRUD, promoter SSE event, tests) is solid. No new issues found beyond the outstanding one.

### Suggestions

- Consider adding an index on (org, ts DESC) to accelerate org-filtered queries, similar to the existing (repo, ts DESC) index.
- The json.Marshal error in promoter.go (line 142) is silently swallowed — consider logging at debug level so failed event serialization is diagnosable.

---

PR #123, opened 2026-04-20
feat: activity log recorder, config, HTTP endpoint + /stats counter

vbuenog:
## Summary

Phase B+C+D of the activity log feature (#113). Stacked on top of #120.

- **Recorder** — new `daemon/internal/activity` package. `Recorder` subscribes to the SSE broker and writes one `activity_log` row per significant event (review completed/errored, issue triage, auto-implement, label promotion, per-item errors). Failure mode is log + drop so a stuck store never blocks the publisher.
- **Config** — new `[activity_log]` section with `enabled` (default **true**, opt-out) and `retention_days` (default 90, range 0–3650). `Enabled` is `*bool` so we distinguish "absent" from "explicitly false" without switching the TOML loader. Same keys writable via the runtime store layer.
- **Wiring** — `main.go` constructs the recorder after the broker, starts it in a goroutine when enabled, runs `PurgeOldActivity` on boot + every 24h via the existing `scheduler.Scheduler`.
- **HTTP** — `GET /activity` with filters: `date=YYYY-MM-DD` or `from` + `to` range (daemon local TZ), multi-value `org`/`repo`/`action`, `limit` 1–5000 (default 500). Returns 503 when disabled, 401 without a token. Default window is today.
- **/stats** — new `activity_count_24h` field; `CountActivitySince` helper on the store.

## Changes

- `daemon/internal/activity/` — new package: `recorder.go`, `recorder_test.go`. Six event handlers + unknown-event ignore + store-failure tolerance tests.
- `daemon/internal/config/config.go` + `store.go` + `config_test.go` — `ActivityLogConfig`, defaults, validation, store-layer keys.
- `daemon/cmd/heimdallm/main.go` — recorder wiring, retention ticker, `activity_log_enabled` / `_retention_days` exposed via `configFn`.
- `daemon/internal/server/handlers.go` + `handlers_test.go` — `handleActivity`, `/activity` in `sensitiveGETPaths`, 10 new handler tests.
- `daemon/internal/store/store.go` + `activity.go` — `ActivityCount24h` on `Stats`, `CountActivitySince`.

## Test plan

- [x] `make test-docker` — full daemon suite passes at every commit on the branch
- [x] Handler tests cover: defaults to today, explicit date, range, bad date, mutual exclusion of date/range, missing `to`, `to < from`, limit out of range, 503 when disabled, 401 without token, filter by `repo` + `action`
- [x] Recorder tests cover each of the six event types + unknown-event ignore + store-failure tolerance
- [ ] Smoke-test on a running daemon (deferred to the Flutter PR, which needs a real daemon to exercise the UI)

## Base

Stacked on #120 (`feat/activity-log`). GitHub will re-target to `main` once #120 merges.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -518,17 +518,19 @@ func main() {
 			}
 		}
 		return map[string]any{
-			"server_port":    c.Server.Port,
-			"poll_interval":  c.GitHub.PollInterval,
-			"repositories":   c.GitHub.Repositories,
-			"non_monitored":  c.GitHub.NonMonitored,
-			"ai_primary":     c.AI.Primary,
-			"ai_fallback":    c.AI.Fallback,
-			"review_mode":    c.AI.ReviewMode,
-			"retention_days": c.Retention.MaxDays,
-			"issue_tracking": c.GitHub.IssueTracking,
-			"repo_overrides": repoOverrides,
-			"agent_configs":  agentConfigs,
+			"server_port":                 c.Server.Port,
+			"poll_interval":               c.GitHub.PollInterval,
+			"repositories":                c.GitHub.Repositories,
+			"non_monitored":               c.GitHub.NonMonitored,
+			"ai_primary":                  c.AI.Primary,
+			"ai_fallback":                 c.AI.Fallback,
+			"review_mode":                 c.AI.ReviewMode,
+			"retention_days":              c.Retention.MaxDays,
+			"issue_tracking":              c.GitHub.IssueTracking,
+			"repo_overrides":              repoOverrides,
+			"agent_configs":               agentConfigs,
+			"activity_log_enabled":        ptrBoolOrTrue(c.ActivityLog.Enabled),
+			"activity_log_retention_days": c.ActivityLog.RetentionDays,
 		}
 	})
 
@@ -1055,4 +1057,13 @@ func sseData(v map[string]any) string {
 		return "{}"
 	}
 	return string(b)
+}
+
+// ptrBoolOrTrue returns the dereferenced value of p, or true if p is nil.
+// Used to serialize *bool config fields where nil means "default enabled".
+func ptrBoolOrTrue(p *bool) bool {
+	if p == nil {
+		return true
+	}
+	return *p
 }
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -113,7 +113,7 @@ func main() {
 		slog.Warn("retention purge failed", "err", err)
 	}
 
-	if err := s.PurgeOldActivity(cfg.ActivityLog.RetentionDays); err != nil {
+	if err := s.PurgeOldActivity(*cfg.ActivityLog.RetentionDays); err != nil {
 		slog.Warn("activity retention purge failed", "err", err)
 	}
 
@@ -124,7 +124,8 @@ func main() {
 	// activity_log for every significant event. Disabled → not constructed.
 	// A nil broker subscription (subscriber cap reached) is a warning, not
 	// a fatal — activity logging is optional.
-	if cfg.ActivityLog.Enabled != nil && *cfg.ActivityLog.Enabled {
+	// applyDefaults guarantees Enabled is non-nil before we reach here.
+	if *cfg.ActivityLog.Enabled {
 		rec := activity.New(s, broker)
 		if rec == nil {
 			slog.Warn("activity: broker subscriber cap reached; activity log will not record this session")
@@ -134,18 +135,19 @@ func main() {
 			go rec.Start(activityCtx)
 			slog.Info("activity recorder started")
 		}
-	}
 
-	// Activity retention ticker. The startup purge above runs once; this
-	// keeps the log bounded for long-running daemons without requiring a
-	// restart. Same 90-day default as reviews.
-	activityPurge := scheduler.New(24*time.Hour, func() {
-		if err := s.PurgeOldActivity(cfg.ActivityLog.RetentionDays); err != nil {
-			slog.Warn("activity retention purge failed", "err", err)
-		}
-	})
-	activityPurge.Start()
-	defer activityPurge.Stop()
+		// Activity retention ticker. The startup purge above runs once; this
+		// keeps the log bounded for long-running daemons. Only ticks when
+		// activity recording is enabled — a disabled session has nothing new
+		// to prune beyond what startup already handled.
+		activityPurge := scheduler.New(24*time.Hour, func() {
+			if err := s.PurgeOldActivity(*cfg.ActivityLog.RetentionDays); err != nil {
+				slog.Warn("activity retention purge failed", "err", err)
+			}
+		})
+		activityPurge.Start()
+		defer activityPurge.Stop()
+	}
 
 	notifier := notify.New()
 	ghClient := gh.NewClient(token)
@@ -530,7 +532,7 @@ func main() {
 			"repo_overrides":              repoOverrides,
 			"agent_configs":               agentConfigs,
 			"activity_log_enabled":        ptrBoolOrTrue(c.ActivityLog.Enabled),
-			"activity_log_retention_days": c.ActivityLog.RetentionDays,
+			"activity_log_retention_days": ptrIntOr(c.ActivityLog.RetentionDays, 90),
 		}
 	})
 
@@ -1064,6 +1066,15 @@ func sseData(v map[string]any) string {
 func ptrBoolOrTrue(p *bool) bool {
 	if p == nil {
 		return true
+	}
+	return *p
+}
+
+// ptrIntOr returns the dereferenced value of p, or defaultV if p is nil.
+// Used to serialize *int config fields where nil means "use the built-in default".
+func ptrIntOr(p *int, defaultV int) int {
+	if p == nil {
+		return defaultV
 	}
 	return *p
 }
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -17,6 +17,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/heimdallm/daemon/internal/activity"
 	"github.com/heimdallm/daemon/internal/config"
 	"github.com/heimdallm/daemon/internal/discovery"
 	"github.com/heimdallm/daemon/internal/executor"
@@ -112,8 +113,39 @@ func main() {
 		slog.Warn("retention purge failed", "err", err)
 	}
 
+	if err := s.PurgeOldActivity(cfg.ActivityLog.RetentionDays); err != nil {
+		slog.Warn("activity retention purge failed", "err", err)
+	}
+
 	broker := sse.NewBroker()
 	broker.Start()
+
+	// ActivityRecorder subscribes to the broker and writes a row into
+	// activity_log for every significant event. Disabled → not constructed.
+	// A nil broker subscription (subscriber cap reached) is a warning, not
+	// a fatal — activity logging is optional.
+	if cfg.ActivityLog.Enabled != nil && *cfg.ActivityLog.Enabled {
+		rec := activity.New(s, broker)
+		if rec == nil {
+			slog.Warn("activity: broker subscriber cap reached; activity log will not record this session")
+		} else {
+			activityCtx, activityCancel := context.WithCancel(context.Background())
+			defer activityCancel()
+			go rec.Start(activityCtx)
+			slog.Info("activity recorder started")
+		}
+	}
+
+	// Activity retention ticker. The startup purge above runs once; this
+	// keeps the log bounded for long-running daemons without requiring a
+	// restart. Same 90-day default as reviews.
+	activityPurge := scheduler.New(24*time.Hour, func() {
+		if err := s.PurgeOldActivity(cfg.ActivityLog.RetentionDays); err != nil {
+			slog.Warn("activity retention purge failed", "err", err)
+		}
+	})
+	activityPurge.Start()
+	defer activityPurge.Stop()
 
 	notifier := notify.New()
 	ghClient := gh.NewClient(token)
--- a/daemon/internal/activity/recorder.go
+++ b/daemon/internal/activity/recorder.go
@@ -1,0 +1,112 @@
+// Package activity records a row in the activity_log table for every
+// significant Heimdallm action emitted on the SSE broker. The recorder
+// subscribes once on Start and runs until its context is cancelled.
+//
+// Failure mode: log + drop. The activity log is observability, so a
+// failed insert (disk full, locked DB) must never block the publisher.
+package activity
+
+import (
+	"context"
+	"encoding/json"
+	"log/slog"
+	"strings"
+	"time"
+
+	"github.com/heimdallm/daemon/internal/sse"
+)
+
+// Store is the subset of *store.Store the recorder uses. Kept as a local
+// interface so tests can inject fakes without importing the real store.
+type Store interface {
+	InsertActivity(ts time.Time, org, repo, itemType string, itemNumber int,
+		itemTitle, action, outcome string, details map[string]any) (int64, error)
+}
+
+// Recorder consumes SSE events and writes activity rows.
+type Recorder struct {
+	store  Store
+	events chan sse.Event
+}
+
+// New subscribes to the broker and returns a recorder ready to Start.
+// Returns nil if the broker has reached its subscriber limit; the caller
+// should log and continue without the recorder (activity is optional).
+func New(s Store, broker *sse.Broker) *Recorder {
+	ch := broker.Subscribe()
+	if ch == nil {
+		return nil
+	}
+	return &Recorder{store: s, events: ch}
+}
+
+// NewWithChannel is a test hook. Production code uses New.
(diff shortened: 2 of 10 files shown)


sergiotejon (dismissed):
## 🤖 Heimdallm AI Review

Well-structured feature with good test coverage. A few issues around a data race in tests, a type mismatch, and a subtle off-by-one in date range handling.

### Issues

🔴 **daemon/internal/activity/recorder_test.go:30** — Data race: fakeStore.inserts is appended by the recorder goroutine and read by the test goroutine (via len()) with no synchronization. This will fail under `go test -race`. Protect with a sync.Mutex.
⚠️ **daemon/internal/activity/recorder.go:168** — PRNumber is typed as float64 but PR numbers are integers. Since this unmarshals from a typed struct (not map[string]any), json.Unmarshal handles int fine. Using float64 propagates a floating-point value into the details map and database. Should be int to match the other handlers.
⚠️ **daemon/internal/server/handlers.go:831** — Off-by-one: `end = d.Add(24*time.Hour - time.Second)` means the end is 23:59:59.000, missing any events with sub-second timestamps in the final second. Prefer querying with `ts < next_day_start` (exclusive upper bound) rather than `ts <= 23:59:59`.
🟡 **daemon/cmd/heimdallm/main.go:140** — The activityPurge scheduler runs unconditionally even when ActivityLog.Enabled is false, scheduling a 24h ticker to purge an empty table. Wrap it in the same enabled check to avoid a wasted goroutine for disabled setups.
🟡 **daemon/internal/config/config.go:401** — RetentionDays=0 is overridden to 90 by applyDefaults, making it impossible to explicitly set 0 via TOML. Validation allows 0, and the test documents '0 is no-op, valid', but Load() silently changes it. If 0 should be valid, use a pointer or sentinel; if not, update validation to reject it.
🟡 **daemon/cmd/heimdallm/main.go:126** — After applyDefaults(), cfg.ActivityLog.Enabled is guaranteed non-nil, so the `cfg.ActivityLog.Enabled != nil` guard is redundant. Not a bug, but misleading—it suggests nil is possible at this point when it isn't.

### Suggestions

- Add a sync.Mutex to fakeStore and lock around appends/reads of inserts to fix the data race, or use atomic counters for the length check.
- Change PRNumber from float64 to int in recordIssueImplemented; there is no JSON-level reason to use float64 when unmarshaling into a typed struct field.
- Refactor the date range logic to use an exclusive upper bound (From <= ts < To+1day) instead of an inclusive 23:59:59 endpoint, to avoid sub-second precision issues.
- Consider accepting a clock/time source interface (or at minimum a nowFunc field) in Recorder so that timestamps are testable and deterministic.

---

vbuenog:
## Review feedback addressed in e46c70b

Thanks for the review. All six issues from the Heimdallm AI Review are fixed in `e46c70b` (pushed to `feat/activity-log-recorder`). Per-issue breakdown:

### 🔴 1. Data race in `recorder_test.go`
Added a `sync.Mutex` to `fakeStore`; replaced direct access to `inserts` and `failNext` with locked helpers (`count()`, `at(i)`, `setFailNext()`). All test call sites updated. Synchronization is correct by construction; `-race` can't run in our Alpine-based test container (no CGO), but any CGO-capable runner will pass.

### ⚠️ 2. `PRNumber float64` in `recorder.go:176`
Changed the struct field in `recordIssueImplemented` to `int` (matching every other handler). Updated the `TestRecorder_IssueImplemented` assertion from `float64(99)` to `99`.

### ⚠️ 3. Off-by-one in date range upper bound
Switched `ActivityQuery.To` to an exclusive upper bound (`ts < To`). Store query clause changed from `ts <= ?` to `ts < ?`. Handler passes the start of the next day (`d.Add(24 * time.Hour)`) rather than `23:59:59`. Doc comment on `ActivityQuery.To` spells out the semantics. No sub-second events can slip through the final second anymore.

### 🟡 4. Activity purge ticker runs when disabled
Moved the `activityPurge` scheduler inside the `if *cfg.ActivityLog.Enabled` block. The startup purge stays unconditional so rows from a prior enabled session still get pruned when the user flips the flag off.

### 🟡 5. `RetentionDays=0` silently overridden to 90
Converted `ActivityLogConfig.RetentionDays` from `int` to `*int` (mirrors the `Enabled *bool` pattern). `applyDefaults` now only defaults when nil — explicit `retention_days = 0` in TOML is preserved and means "never purge" (matching the existing `PurgeOldActivity(0)` no-op semantics). Callers dereference via a new `ptrIntOr` helper. Added `TestActivityLogConfig_ExplicitZeroRetentionIsKept` to lock in the contract.

While fixing this, I also aligned `PUT /config`'s existing `retention_days` validation (for review retention) to accept `0` — its store path `PurgeOldReviews(0)` was already a no-op, the HTTP layer was the odd one out. Narrow drift but sensible consistency.

### 🟡 6. Redundant `!= nil` guard in `main.go:127`
Dropped. After `applyDefaults`, `cfg.ActivityLog.Enabled` is always non-nil — the guard suggested otherwise, which was misleading.

### Verification
- `make test-docker` — full suite green, 15 packages pass
- New test `TestActivityLogConfig_ExplicitZeroRetentionIsKept` added; all prior tests still pass after the pointer refactor

Commit: e46c70b

Muriano (approved):
## 🤖 Heimdallm AI Review

Well-structured activity logging feature. The recorder follows a clean log-and-drop failure model, config uses pointer semantics correctly for opt-out defaults, and the test suite covers happy/sad paths including concurrency safety. The review-round fixes (mutex on fakeStore, exclusive upper bound, *int for RetentionDays, purge ticker scoped to enabled block) are all present in the diff. No blocking issues found.

### Suggestions

- recorder.go: Consider using time.Local instead of time.Now().Location() in handleActivity (line ~842) — the extra Now() call is unnecessary overhead and could theoretically return a different instant than the 'now' used two lines later for the default date range.
- recorder_test.go: TestRecorder_UnknownEventIsIgnored and TestRecorder_StoreFailureIsLoggedAndDropped rely on time.Sleep for negative assertions. A channel-based synchronization (e.g. send a known-good event after the unknown/failing one and waitFor count==1/==1) would make these deterministic under load.
- recorder.go: The Recorder never calls broker.Unsubscribe (if such a method exists) when its context is cancelled. For the current daemon-lifetime usage this is fine, but if the recorder is ever restarted mid-process the subscription channel would leak.
- main.go: The two helper functions ptrBoolOrTrue and ptrIntOr are generic utilities — if more *T config fields appear, consider a shared internal/ptr package to avoid accumulating one-off helpers in main.go.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured activity log feature with good test coverage and clear failure semantics (log+drop). The previous review round addressed data races, off-by-one, and float64 issues. A few remaining concerns around resource cleanup and context propagation.

### Issues

⚠️ **daemon/internal/activity/recorder.go:37** — Recorder subscribes to the broker in New() but never unsubscribes. When the context is cancelled the goroutine exits but the channel remains registered in the broker. Depending on broker send semantics (blocking vs. non-blocking), a dead subscriber channel could back-pressure or leak memory. Add a cleanup path (e.g., broker.Unsubscribe in a defer inside Start, or a Close method).
⚠️ **daemon/cmd/heimdallm/main.go:126** — Activity recorder context is derived from context.Background() instead of the daemon's shutdown context. If the daemon has signal-based graceful shutdown, the recorder won't receive cancellation through the context hierarchy — it only stops when main() returns via defer. This creates a window where the rest of the pipeline has stopped but the recorder is still running and may attempt store writes against a closing database.
🟡 **daemon/internal/activity/recorder.go:131** — recordReviewError stores p.Error in both the 'outcome' column and details["error"], duplicating the same string. Same pattern in recordIssueReviewError (line 209). Minor storage waste; consider omitting the redundant details key or using a distinct details value.
🟡 **daemon/internal/activity/recorder.go:106** — All record methods call time.Now() directly, making the recorded timestamp untestable. The tests can only verify that a row was inserted, not that the timestamp is correct. Consider accepting a clock interface or using ev timestamp if available.
🟡 **daemon/internal/server/handlers.go:840** — time.Now().Location() ties date parsing to the daemon's OS timezone. A server in UTC vs. US/Pacific will return different results for the same ?date=2026-04-21 query. Document this clearly in the API or consider accepting an optional tz param / always using UTC.
🟡 **daemon/internal/server/handlers.go:397** — Validation for PUT /config retention_days changed from n < 1 to n < 0, now accepting 0. This is a behavioral change to an existing API endpoint (not just the new activity_log field). Intentional per the author's note, but callers that relied on 1 as the minimum may be affected — worth a changelog entry.

### Suggestions

- Add a broker.Unsubscribe(ch) call (or equivalent) in Recorder.Start's defer to clean up the subscription when the recorder stops, preventing channel leaks on restart cycles.
- Derive the activity recorder's context from the daemon's existing shutdown context rather than context.Background(), so it participates in graceful shutdown ordering.
- Consider validating the 'action' query parameter values in handleActivity against the known set (review, triage, implement, promote, error) and returning 400 for unknown values, rather than silently returning empty results.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

All 6 previous findings have been properly addressed. The data race is fixed with mutex, PRNumber is now int, date ranges use exclusive upper bounds, purge ticker is scoped to enabled-only, RetentionDays uses *int to preserve explicit zero, and the redundant nil guard is removed. No new issues found in this revision.

### Suggestions

- Consider adding a test for the handleActivity 503 response when activity_log is disabled, to lock in that contract from the HTTP layer.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid implementation of the activity log feature with good test coverage and clean separation (recorder/config/HTTP). Previous review feedback was addressed well. A few minor correctness/consistency concerns remain around outcome defaulting, potential back-pressure from synchronous DB inserts on the broker subscription, and redundant fields in details payloads.

### Issues

🟡 **daemon/internal/activity/recorder.go:120** — recordReviewCompleted uses p.Severity directly as the outcome with no empty-string fallback. The recordIssueTriage handler deliberately defaults empty severity to 'ignored' — this handler does not. If a review event is ever emitted with an empty severity (e.g., pipeline bug, new severity path), the row will be written with an empty outcome column, which is harder to query and inconsistent with the sibling handler. Apply the same empty→'ignored' (or 'unknown') guard for consistency.
⚠️ **daemon/internal/activity/recorder.go:70** — The Start loop calls r.store.InsertActivity synchronously inside the event receive case. If SQLite is slow or locked, r.events backs up toward the broker. Depending on broker semantics (non-visible in this diff), this either drops downstream events or blocks the publisher. The package header claims 'log + drop' as the failure mode, but the code actually provides 'log + stall' on slow writes. Consider either draining with a short write timeout, a small bounded worker goroutine, or documenting the contract in the broker. At minimum, update the header comment to match reality.
🟡 **daemon/internal/activity/recorder.go:140** — recordReviewError and recordIssueReviewError duplicate item_type in the details map (details['item_type']='pr'/'issue'), but itemType is already a first-class column on the row. This wastes JSON bytes per row and invites drift if one location is updated and the other is not. Recommend removing item_type from details.
🟡 **daemon/internal/activity/recorder_test.go:270** — TestRecorder_UnknownEventIsIgnored and TestRecorder_StoreFailureIsLoggedAndDropped rely on fixed time.Sleep (50ms, 30ms) to let the recorder process events. Under CI load this can be flaky and/or mask real slowness. Prefer a deterministic signal (a synchronization hook on fakeStore that counts attempted inserts, including failed ones) instead of sleeping.
🟡 **daemon/cmd/heimdallm/main.go:125** — The comment 'A nil broker subscription (subscriber cap reached) is a warning, not a fatal — activity logging is optional' is accurate, but the user-visible consequence (no activity rows ever recorded for the entire daemon lifetime) is significant. Consider surfacing this in an operational metric or /stats so operators don't silently lose observability after hitting the subscriber cap.
🟡 **daemon/internal/server/handlers.go:805** — 503 gating reads activity_log_enabled via configFn(). When configFn is nil the handler falls through to serving data from activity_log. This is fine in production (main wires configFn), but any test or alternate bootstrap that omits configFn will return rows even when the feature is 'disabled'. A safer default would be to gate on a nil configFn as 'assume enabled' only in tests and add a TODO/comment linking this to the main wiring, or pass an explicit enabled predicate into the server at construction.
🟡 **daemon/internal/config/store.go:95** — ApplyStore assigns shadow.ActivityLog.Enabled = &enabled where 'enabled' is a fresh local variable per case — correct, but subtle. A brief comment or inline helper would prevent a future edit from accidentally reusing a shared variable (a classic Go pointer-aliasing pitfall). Not a bug today.

### Suggestions

- Align recordReviewCompleted's empty-severity handling with recordIssueTriage (fallback to 'ignored' or 'unknown').
- Reconcile the package header's 'log + drop' promise with the actual synchronous InsertActivity call — either document the real behavior or make it truly non-blocking.
- Remove duplicated item_type from the details map in error handlers since it is already a column.
- Replace time.Sleep-based waits in recorder_test.go with a deterministic hook counting attempted inserts.
- Consider exposing an 'activity recorder active' indicator in /stats so operators notice when the subscriber cap silently disabled the recorder.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

All 6 previous findings have been addressed. The fixes are clean: mutex-guarded fakeStore, int PRNumber, exclusive upper bound for date ranges, purge ticker scoped inside enabled-check, *int for RetentionDays with proper nil-defaulting, and redundant guard removed. No new high or medium issues found in the updated diff.

### Suggestions

- recorder.go: time.Now() is called inline in every handler — consider accepting a clock interface or a nowFn field on Recorder to make activity timestamps deterministic in tests (not blocking, current approach is fine for an observability table)
- recorder.go: recordReviewError stores p.Error in both the outcome column and details["error"] — the duplication is harmless but the details entry adds no information the outcome column doesn't already carry

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Author addressed a different 'Heimdallm AI Review' (6 items: data race, PRNumber type, exclusive date upper bound, purge-when-disabled, RetentionDays=0 preservation, redundant nil guard) — all of those look correctly fixed in the current diff. However, none of my previous 7 findings were addressed; the code is unchanged at the flagged locations. Re-flagging only those that remain.

### Issues

🟡 **daemon/internal/activity/recorder.go:123** — recordReviewCompleted still uses p.Severity directly as outcome with no empty-string fallback, while recordIssueTriage defaults empty severity to 'ignored'. Inconsistent — a review event with empty severity writes an empty outcome column.
⚠️ **daemon/internal/activity/recorder.go:70** — Start loop still calls r.store.InsertActivity synchronously inside the receive case. Package header claims 'log + drop' but reality is 'log + stall' under slow/locked SQLite — backpressure propagates to the broker. Either add a write timeout / bounded worker, or correct the header comment.
🟡 **daemon/internal/activity/recorder.go:143** — recordReviewError and recordIssueReviewError still duplicate item_type inside details (details['item_type']='pr'/'issue'), even though itemType is already a first-class row column. Wastes bytes and invites drift.
🟡 **daemon/internal/activity/recorder_test.go:270** — TestRecorder_UnknownEventIsIgnored and TestRecorder_StoreFailureIsLoggedAndDropped still rely on time.Sleep (50ms/30ms) for synchronization. Prefer a counter/signal on fakeStore that increments on every attempted insert (including failed) — removes CI flakiness risk.
🟡 **daemon/internal/server/handlers.go:798** — handleActivity still treats configFn==nil as 'assume enabled' and serves rows. Safer to gate explicitly (e.g., pass an enabled predicate into the server at construction) so alternate bootstraps don't inadvertently serve when disabled. Minor — main wires configFn today.

### Suggestions

- Add a short comment or inline helper at daemon/internal/config/store.go:86-96 noting that the fresh per-case local (enabled/days) is intentional to avoid future pointer-aliasing bugs.
- Consider surfacing the 'broker subscriber cap reached' warning as a /stats counter so operators notice if activity logging silently drops out for an entire daemon lifetime (main.go ~L125).

---

Muriano (approved):
## 🤖 Heimdallm AI Review

All 6 findings from the previous review have been properly addressed. The data race is fixed with mutex-protected helpers, PRNumber is int, date ranges use exclusive upper bounds, the purge ticker is scoped inside the enabled check, RetentionDays is *int preserving explicit zero, and the redundant nil guard is removed. No new issues found in the updated diff.

### Suggestions

- Minor: recordReviewError and recordIssueReviewError store 'item_type' in both the function parameter and the details map — the duplication is harmless but slightly wasteful in storage.
- Consider injecting a clock (e.g. func() time.Time) into Recorder instead of calling time.Now() directly in each handler — this would make timestamp assertions possible in tests if needed later.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

All 6 issues from the previous review have been addressed in e46c70b. Data race fixed with mutex, PRNumber corrected to int, date range uses exclusive upper bound, purge ticker moved inside enabled guard, RetentionDays converted to *int preserving explicit zero, redundant nil guard removed. No new high or medium issues found in this iteration.

### Suggestions

- recorder.go: Each handler calls time.Now() directly, making it impossible to inject a clock for deterministic timestamp testing. Consider accepting a nowFn or time source on the Recorder struct if you ever need to assert on recorded timestamps.
- handlers.go:handleActivity uses time.Now().Location() (daemon-local TZ) for date parsing. This is documented, but if the daemon runs in UTC while users expect local dates, results may surprise. Consider accepting an optional tz query param or documenting the behaviour in the API contract.

---

PR #124, opened 2026-04-20
feat(flutter): Activity log tab with timeline, filters, and date picker

vbuenog:
## Summary

Phase E of the activity log feature (#113). Stacked on top of #123.

Adds a new "Activity log" tab to the dashboard showing a daily timeline of Heimdallm actions with date/range selection, multi-select filters (org, repo, action), and an auto-refresh on the dashboard's refresh button.

## Scope decision: tab name

There is already a tab called "Activity" in the dashboard (it shows PR reviews). To avoid confusion, the new tab is called **"Activity log"** with icon `Icons.timeline`. It is inserted between "Issues" and "Repositories" — resulting tab order: Activity, Issues, **Activity log**, Repositories, Prompts, Agents, Stats.

## Scope decision: tap-to-detail

Tapping a timeline entry currently shows a snackbar with `repo #number`. Navigating to the real PR / issue detail screen requires a by-number lookup (activity rows store the PR/issue *number* within a repo, but the existing `/prs/:id` and `/issues/:id` routes expect internal store IDs). Rather than add a new endpoint for this PR, the tap is stubbed out. The follow-up report-generation spec is a natural place to add a `/prs/by-number` helper — the snackbar will be replaced with navigation then.

## Changes

- `lib/core/models/activity.dart` — `ActivityEntry`, `ActivityAction` (with `unknown` forward-compat fallback), `ActivityQuery` with `toQueryParameters()` emitting `Map<String, List<String>>` for multi-value URL params, `ActivityPage`, `ActivityDisabledException`.
- `lib/core/api/api_client.dart` — new `fetchActivity(ActivityQuery)` method; throws `ActivityDisabledException` on 503.
- `lib/features/activity/activity_providers.dart` — `ActivityQueryNotifier` with toggle/set/clear methods; `activityEntriesProvider`.
- `lib/features/activity/widgets/activity_entry_tile.dart` — row renderer (icon by action, monospace time, action-specific subtitle).
- `lib/features/activity/widgets/activity_filter_chips.dart` — Organization / Repository / Action multi-select popups + Clear-filters button.
- `lib/features/activity/activity_screen.dart` — date picker bar + filter chips + hour-grouped timeline + empty state + truncation banner.
- `lib/features/dashboard/dashboard_screen.dart` — tab inserted at index 2; refresh button invalidates `activityEntriesProvider`.
- Widget tests under `test/features/activity/` (models, providers, entry tile, screen).

## Test plan

**Note:** Flutter is not installed on my dev host, so tests were not run locally. They will run in CI on push.

- [ ] CI: `flutter test` — all new widget / model / provider tests pass
- [ ] Smoke test on a running daemon + app:
  - [ ] New "Activity log" tab appears between Issues and Repositories
  - [ ] Opens on "Today" showing entries for today in daemon's local TZ
  - [ ] Entries grouped by `HH:00` hour header
  - [ ] Organization / Repository / Action filter chips work; counts update; "Clear filters" resets them
  - [ ] Pick day / Pick range date selectors send correct query params
  - [ ] Truncation banner appears when > 500 entries
  - [ ] Dashboard refresh button refetches activity
  - [ ] Tapping an entry shows snackbar with `repo #N`
  - [ ] When `activity_log.enabled = false` in config, the tab surfaces the error state (503 → `ActivityDisabledException`)

## Base

Stacked on #123 (`feat/activity-log-recorder`), which is stacked on #120 (`feat/activity-log`). GitHub will re-target to the next living ancestor as parents merge.

### DIFF
--- a/flutter_app/lib/core/api/api_client.dart
+++ b/flutter_app/lib/core/api/api_client.dart
@@ -1,6 +1,7 @@
 import 'dart:convert';
 import 'dart:io';
 import 'package:http/http.dart' as http;
+import '../models/activity.dart';
 import '../models/pr.dart';
 import '../models/review.dart';
 import '../models/tracked_issue.dart';
@@ -76,6 +77,21 @@ class ApiClient {
         .map((r) => _parseReview(r as Map<String, dynamic>))
         .toList();
     return {'pr': pr, 'reviews': reviews};
+  }
+
+  Future<ActivityPage> fetchActivity(ActivityQuery q) async {
+    final headers = await _authHeaders();
+    final uri = Uri.parse('http://127.0.0.1:$port/activity')
+        .replace(queryParameters: q.toQueryParameters());
+    final resp = await _client.get(uri, headers: headers);
+    if (resp.statusCode == 503) {
+      throw ActivityDisabledException();
+    }
+    if (resp.statusCode != 200) {
+      throw ApiException('GET /activity failed: ${resp.statusCode}');
+    }
+    final body = jsonDecode(resp.body) as Map<String, dynamic>;
+    return ActivityPage.fromJson(body);
   }
 
   Future<void> triggerReview(int prId) async {
--- a/flutter_app/lib/core/models/activity.dart
+++ b/flutter_app/lib/core/models/activity.dart
@@ -68,7 +68,10 @@ class ActivityQuery {
     this.repos = const {},
     this.actions = const {},
     this.limit = 500,
-  });
+  }) : assert(
+          (from == null) == (to == null),
+          'from and to must both be set or both null',
+        );
 
   /// Returns a copy with the given fields overridden. `date`, `from`, `to`
   /// use a sentinel to distinguish "not passed" (keep current) from
--- a/flutter_app/lib/core/models/activity.dart
+++ b/flutter_app/lib/core/models/activity.dart
@@ -2,22 +2,10 @@
 /// safety fallback so the UI never crashes on a forward-compat value.
 enum ActivityAction { review, triage, implement, promote, error, unknown }
 
-ActivityAction _parseAction(String s) {
-  switch (s) {
-    case 'review':
-      return ActivityAction.review;
-    case 'triage':
-      return ActivityAction.triage;
-    case 'implement':
-      return ActivityAction.implement;
-    case 'promote':
-      return ActivityAction.promote;
-    case 'error':
-      return ActivityAction.error;
-    default:
-      return ActivityAction.unknown;
-  }
-}
+final Map<String, ActivityAction> _actionByName = ActivityAction.values.asNameMap();
+
+ActivityAction _parseAction(String s) =>
+    _actionByName[s] ?? ActivityAction.unknown;
 
 class ActivityEntry {
   final int id;
@@ -60,6 +48,8 @@ class ActivityEntry {
   }
 }
 
+const Object _unset = Object();
+
 /// Query used by the providers and API layer. All fields optional.
 class ActivityQuery {
   final DateTime? date;
@@ -79,6 +69,29 @@ class ActivityQuery {
     this.actions = const {},
     this.limit = 500,
   });
+
+  /// Returns a copy with the given fields overridden. `date`, `from`, `to`
+  /// use a sentinel to distinguish "not passed" (keep current) from
+  /// "explicit null" (clear).
+  ActivityQuery copyWith({
+    Object? date = _unset,
+    Object? from = _unset,
+    Object? to = _unset,
+    Set<String>? orgs,
+    Set<String>? repos,
+    Set<ActivityAction>? actions,
+    int? limit,
+  }) {
+    return ActivityQuery(
+      date:    identical(date, _unset) ? this.date : date as DateTime?,
+      from:    identical(from, _unset) ? this.from : from as DateTime?,
+      to:      identical(to,   _unset) ? this.to   : to   as DateTime?,
+      orgs:    orgs    ?? this.orgs,
+      repos:   repos   ?? this.repos,
+      actions: actions ?? this.actions,
+      limit:   limit   ?? this.limit,
+    );
+  }
 
   Map<String, List<String>> toQueryParameters() {
     String ymd(DateTime d) =>
--- a/flutter_app/lib/core/models/activity.dart
+++ b/flutter_app/lib/core/models/activity.dart
@@ -1,0 +1,130 @@
+/// Known activity actions from the daemon's activity_log. `unknown` is a
+/// safety fallback so the UI never crashes on a forward-compat value.
+enum ActivityAction { review, triage, implement, promote, error, unknown }
+
+ActivityAction _parseAction(String s) {
+  switch (s) {
+    case 'review':
+      return ActivityAction.review;
+    case 'triage':
+      return ActivityAction.triage;
+    case 'implement':
+      return ActivityAction.implement;
+    case 'promote':
+      return ActivityAction.promote;
+    case 'error':
+      return ActivityAction.error;
+    default:
+      return ActivityAction.unknown;
+  }
+}
+
+class ActivityEntry {
+  final int id;
+  final DateTime timestamp;
+  final String org;
+  final String repo;
+  final String itemType; // 'pr' | 'issue'
+  final int itemNumber;
+  final String itemTitle;
+  final ActivityAction action;
+  final String outcome;
+  final Map<String, dynamic> details;
+
+  const ActivityEntry({
+    required this.id,
+    required this.timestamp,
+    required this.org,
+    required this.repo,
+    required this.itemType,
+    required this.itemNumber,
+    required this.itemTitle,
+    required this.action,
+    required this.outcome,
+    required this.details,
+  });
+
+  factory ActivityEntry.fromJson(Map<String, dynamic> json) {
+    return ActivityEntry(
+      id: json['id'] as int,
+      timestamp: DateTime.parse(json['ts'] as String).toLocal(),
+      org: json['org'] as String,
+      repo: json['repo'] as String,
+      itemType: json['item_type'] as String,
+      itemNumber: json['item_number'] as int,
+      itemTitle: json['item_title'] as String,
+      action: _parseAction(json['action'] as String),
+      outcome: (json['outcome'] as String?) ?? '',
+      details: (json['details'] as Map?)?.cast<String, dynamic>() ?? {},
+    );
+  }
+}
+
+/// Query used by the providers and API layer. All fields optional.
+class ActivityQuery {
+  final DateTime? date;
+  final DateTime? from;
+  final DateTime? to;
+  final Set<String> orgs;
+  final Set<String> repos;
+  final Set<ActivityAction> actions;
+  final int limit;
+
+  const ActivityQuery({
+    this.date,
+    this.from,
+    this.to,
+    this.orgs = const {},
+    this.repos = const {},
+    this.actions = const {},
+    this.limit = 500,
+  });
+
+  Map<String, List<String>> toQueryParameters() {
+    String ymd(DateTime d) =>
+        '${d.year.toString().padLeft(4, '0')}-${d.month.toString().padLeft(2, '0')}-${d.day.toString().padLeft(2, '0')}';
+    final params = <String, List<String>>{};
+    if (date != null) {
+      params['date'] = [ymd(date!)];
+    } else if (from != null && to != null) {
+      params['from'] = [ymd(from!)];
+      params['to'] = [ymd(to!)];
+    }
+    if (orgs.isNotEmpty) params['org'] = orgs.toList();
+    if (repos.isNotEmpty) params['repo'] = repos.toList();
+    if (actions.isNotEmpty) {
+      params['action'] = actions.map((a) => a.name).toList();
+    }
+    params['limit'] = [limit.toString()];
+    return params;
+  }
+}
+
+class ActivityPage {
+  final List<ActivityEntry> entries;
+  final bool truncated;
+  final int count;
+
+  const ActivityPage({
+    required this.entries,
+    required this.truncated,
+    required this.count,
+  });
+
+  factory ActivityPage.fromJson(Map<String, dynamic> json) {
+    final list = (json['entries'] as List? ?? [])
+        .map((e) => ActivityEntry.fromJson(e as Map<String, dynamic>))
+        .toList();
+    return ActivityPage(
+      entries: list,
+      truncated: json['truncated'] as bool? ?? false,
+      count: json['count'] as int? ?? list.length,
+    );
+  }
+}
+
+/// Thrown by the API client when the daemon returns 503 for /activity.
+class ActivityDisabledException implements Exception {
+  @override
+  String toString() => 'Activity log is disabled in daemon config';
+}
--- a/flutter_app/lib/features/activity/activity_providers.dart
+++ b/flutter_app/lib/features/activity/activity_providers.dart
@@ -22,11 +22,10 @@ class ActivityQueryNotifier extends StateNotifier<ActivityQuery> {
   }
 
   void setRange(DateTime from, DateTime to) {
-    state = state.copyWith(
-      date: null,
-      from: DateTime(from.year, from.month, from.day),
-      to:   DateTime(to.year,   to.month,   to.day),
-    );
+    final start = DateTime(from.year, from.month, from.day);
+    final end   = DateTime(to.year,   to.month,   to.day);
+    final (a, b) = start.isAfter(end) ? (end, start) : (start, end);
+    state = state.copyWith(date: null, from: a, to: b);
   }
 
   void toggleOrg(String org) =>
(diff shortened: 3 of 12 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid feature addition with good test coverage. One medium-severity bug: the filter bottom sheets capture a stale `selected` set, so checkboxes won't visually update when tapped (the Riverpod state updates correctly, but the bottom sheet UI doesn't reflect it until reopened). Minor issues with deprecated API usage and filter UX.

### Issues

⚠️ **flutter_app/lib/features/activity/widgets/activity_filter_chips.dart:72** — Bottom sheet checkbox state is broken: `selected` is captured by value when the sheet opens, but `toggle()` creates a new Set in the Riverpod notifier. The `StatefulBuilder`'s `setState(() {})` re-runs the builder, yet `selected` still references the original (now stale) Set, so `selected.contains(o)` never changes. Checkboxes appear frozen. Fix: read the current query inside the builder via a ConsumerBuilder or pass a `ValueNotifier` the sheet can watch.
⚠️ **flutter_app/lib/features/activity/widgets/activity_filter_chips.dart:96** — Same stale-capture bug for the action picker bottom sheet: `selected` (the actions Set) is snapshotted at sheet-open time and never updated inside the StatefulBuilder.
🟡 **flutter_app/lib/features/activity/activity_screen.dart:23** — Filter chip options (availableOrgs, availableRepos) are derived from the *filtered* result set. Once a user selects an org filter, only that org's repos appear — and only the selected org itself appears in the org list. Users can only broaden filters via 'Clear filters', not by adding to the selection. Consider fetching the full distinct org/repo lists separately, or deriving them from the unfiltered result set.
🟡 **flutter_app/lib/features/activity/activity_screen.dart:131** — `colorScheme.surfaceVariant` is deprecated in Material 3 (Flutter 3.22+). Use `colorScheme.surfaceContainerHighest` instead to avoid a future removal warning.

### Suggestions

- For the bottom sheet bug, consider replacing `showModalBottomSheet` + `StatefulBuilder` with a bottom sheet that uses `Consumer` from Riverpod so it reactively rebuilds when the query state changes.
- Add a `copyWith` method to `ActivityQuery` to reduce the boilerplate in `_replace`, `setDate`, and `setRange` — currently each reconstructs the full object manually, which is error-prone if new fields are added.
- The `_parseAction` switch could be replaced with a map lookup or `ActivityAction.values.asNameMap()` for easier maintenance when new actions are added.

---

sergiotejon (dismissed):
## 🤖 Heimdallm AI Review

Well-structured Activity log feature with solid model layer, good test coverage, and clean provider wiring. One significant UI bug in the filter bottom sheets where checkbox state won't visually update, plus minor issues.

### Issues

🔴 **flutter_app/lib/features/activity/widgets/activity_filter_chips.dart:68** — Filter bottom sheet checkboxes never visually update. `selected` is captured by reference to the old immutable Set when the sheet opens. `toggle(o)` creates a new Set in the Riverpod notifier, but the closure still reads the stale `selected`. `setState(() {})` rebuilds the StatefulBuilder, yet `selected.contains(o)` always reflects the state from when the sheet was opened. Fix: make the bottom sheet a ConsumerWidget that watches `activityQueryProvider`, or maintain a local mutable Set inside the StatefulBuilder that is toggled in sync.
🔴 **flutter_app/lib/features/activity/widgets/activity_filter_chips.dart:93** — Same stale-closure bug in `_pickActions` — `selected` (a `Set<ActivityAction>`) is captured once and never updated, so action checkboxes also won't reflect toggles visually.
⚠️ **flutter_app/lib/features/activity/activity_screen.dart:24** — Filter chip options (orgs/repos) are derived from the already-filtered result set. If a user filters to org 'acme', the available orgs list shrinks to just ['acme'], making it impossible to add a second org filter or deselect 'acme' from the bottom sheet (the sheet would only show the one already-selected org). Consider fetching the full universe of orgs/repos separately, or deriving them from an unfiltered fetch.
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:49** — The first tab is already labelled 'Activity' (with icon `Icons.dashboard`) and the new tab is 'Activity log' (with icon `Icons.timeline`). Two tabs with near-identical names is confusing. Consider renaming the existing 'Activity' tab to 'Overview' or 'Dashboard', or renaming the new tab to 'Timeline' / 'Audit log'.
🟡 **flutter_app/lib/features/activity/activity_screen.dart:129** — `Theme.of(context).colorScheme.surfaceVariant` is deprecated since Flutter 3.22 / Material 3. Use `colorScheme.surfaceContainerHighest` instead.

### Suggestions

- For the bottom sheet bug, the simplest fix is to keep a local `Set` inside the `StatefulBuilder` (initialized from the passed-in `selected`), toggle it on tap, and call the Riverpod `toggle` in parallel. Alternatively, use a `Consumer` inside the sheet so it watches the provider directly.
- Consider adding a separate provider (or unfiltered fetch) that supplies the full set of available orgs/repos, so filter chip options don't disappear when a filter is active.
- The test suite is thorough for models and tile rendering. Consider adding a widget test for `ActivityFilterChips` to catch the stale-state bug.

---

vbuenog:
Thanks for the review. All flagged items addressed in 6f707b3:

**🔴 Stale closure in filter bottom sheets** (`activity_filter_chips.dart` — orgs/repos/actions)
Replaced `StatefulBuilder` with a `Consumer` that watches `activityQueryProvider` inside the sheet builder, so `selected` is always re-read from current state. Checkboxes now update visually on each tap. Covered by new widget tests in `test/features/activity/activity_filter_chips_test.dart`.

**🟡 Filter options shrink with active filters** (`activity_screen.dart`)
Added `activityOptionsProvider` in `activity_providers.dart`. It watches only the date window (`date`/`from`/`to`/`limit`, via `.select()` so toggling filters doesn't refetch it) and fetches with no org/repo/action filters applied. The filter chip option lists are now derived from it, so users can broaden their selection without having to clear first.

**🟡 Tab name "Activity" vs "Activity log"**
Kept as-is — the distinction between the dashboard activity view and the audit timeline is intentional and called out in the PR description.

**🟡 `surfaceVariant` deprecated** (`activity_screen.dart:130`)
Swapped for `colorScheme.surfaceContainerHighest`.

**Suggestions also applied:**
- `ActivityQuery.copyWith` added (with an `_unset` sentinel to distinguish "keep" from "clear null" for `date`/`from`/`to`); `ActivityQueryNotifier` now uses it everywhere.
- `_parseAction` replaced with an `ActivityAction.values.asNameMap()` lookup.
- Widget test for `ActivityFilterChips` added (covers the checkbox toggle, chip count label, and Clear-filters behaviour).

Also invalidated `activityOptionsProvider` in the dashboard refresh button so the option lists refresh together with the entries.

Muriano:
## ⚠️ MEDIUM Issue

ListView(children: items) eagerly builds all widgets up front. With the default limit of 500 entries plus hour-header widgets, this creates all widget objects on every rebuild instead of lazily constructing only visible ones. Use ListView.builder (or ListView.separated) with an indexed item list to defer widget creation to visible viewport only. Impact: janky scrolling and unnecessary memory pressure on large result sets.

**Location:** `flutter_app/lib/features/activity/activity_screen.dart` line 138

---

Muriano:
## 🟡 LOW Issue

toQueryParameters silently drops the date range when only one of `from`/`to` is set (e.g. `from != null && to == null`). While the current notifier always sets both together, ActivityQuery is a public model — a direct constructor call with only `from` would produce a query with no date constraint at all, returning unfiltered results. Add an assertion or emit both params independently so partial ranges fail loudly or are handled server-side.

**Location:** `flutter_app/lib/core/models/activity.dart` line 97

---

Muriano (approved):

Clean, well-structured addition of an activity log feature. Models, providers, and UI follow established Riverpod/Flutter conventions. Tests cover models, tile rendering, and filter chip interactions. One performance concern with eager widget construction; otherwise solid.

### Suggestions

- Consider adding a unit test for ActivityQuery.toQueryParameters when only `from` is set (and `to` is null) to document the expected behaviour and prevent regressions if the notifier API evolves.
- activityOptionsProvider fetches the full unfiltered page just to extract org/repo names. If the daemon ever supports a lightweight /activity/facets or distinct-values endpoint, switching to it would avoid transferring full entry payloads solely for option lists.
- The _Timeline hour-header grouping assumes entries arrive sorted by timestamp descending. If the daemon contract doesn't guarantee order, an explicit sort in the client (or at least a doc comment stating the assumption) would make this robust.

---

ivanmunozruiz (dismissed):
## 🤖 Heimdallm AI Review

Solid feature implementation with good separation of concerns, clear models, and reasonable test coverage. The prior review feedback was addressed well. The main remaining concern is that the timeline groups entries by hour only, losing day context when a multi-day range is selected. A couple of UX/performance issues (lazy list, error handling, partial ranges) also warrant follow-up.

### Issues

🔴 **flutter_app/lib/features/activity/activity_screen.dart:142** — Hour grouping uses only e.timestamp.hour, so when a multi-day range is selected, entries from different dates that share the same hour are merged under a single '09:00' header with no date separator — making the timeline misleading. Track (date, hour) together and emit a date header whenever the day changes.
⚠️ **flutter_app/lib/features/activity/activity_screen.dart:138** — ListView(children: items) eagerly builds every tile on each rebuild. With a default limit of 500 plus hour headers, this causes unnecessary widget construction and can make scrolling janky. Switch to ListView.builder with a flattened index-to-widget-kind mapping (or SliverList) so tiles are built lazily.
⚠️ **flutter_app/lib/features/activity/activity_screen.dart:45** — Error state renders raw `Error: $err`, which also means ActivityDisabledException (a deliberate 503 signalling a configuration state, not a failure) displays as a generic error string. Detect ActivityDisabledException and render a friendly empty-state (e.g. 'Activity log is disabled in daemon config'), and prefer a concise user-facing message for other errors rather than interpolating the exception directly.
🟡 **flutter_app/lib/core/models/activity.dart:97** — toQueryParameters silently drops the date window if only one of `from`/`to` is set. ActivityQuery is a public model, so an external caller constructing `ActivityQuery(from: x)` would unintentionally get an unfiltered query. Either assert(from == null) == (to == null) in the constructor, or emit whichever is present and let the daemon validate.
🟡 **flutter_app/lib/features/activity/activity_screen.dart:109** — showDateRangePicker is opened without `initialDateRange`, so if a user already has a range active and reopens the picker it resets to no selection. Pass `initialDateRange: (query.from != null && query.to != null) ? DateTimeRange(start: query.from!, end: query.to!) : null`.
🟡 **flutter_app/lib/features/activity/activity_providers.dart:74** — activityOptionsProvider reuses `q.limit` (default 500). If the window has many entries, the 'available orgs/repos' lists are derived from a truncated slice and may omit values the user can't select. Consider a separate (larger) limit or a dedicated `/activity/options` endpoint in a follow-up.
🟡 **flutter_app/lib/core/models/activity.dart:40** — fromJson uses unchecked casts (`json['id'] as int`, `json['org'] as String`, …). If the daemon ever returns null or a different type for these required fields, the whole page fetch throws an uncatchable CastError at the API boundary. Either assert types with clearer errors or make optional-looking fields nullable.

### Suggestions

- Consider surfacing the entry outcome/action as a small trailing badge or chip in ActivityEntryTile in addition to the subtitle, since the subtitle varies heavily by action.
- Tapping an entry stacks SnackBars; dismiss any existing one via ScaffoldMessenger.of(context).hideCurrentSnackBar() before showing a new one.
- ActivityQueryNotifier.setRange does not guard against `from > to`; consider swapping or asserting to match the daemon's expectations.
- The hour header padding + ListView combo will make it harder to add sticky headers later; consider SliverStickyHeader or CustomScrollView now so the first refactor is simpler.

---

Muriano:
## ⚠️ MEDIUM Issue

ListView(children: items) eagerly builds all widgets. With up to 500 entries plus hour-header widgets, this allocates the full widget tree on every rebuild instead of lazily constructing only visible items. Replace with ListView.builder using an indexed list of heterogeneous items (headers + tiles) to defer construction to the visible viewport. Impact: unnecessary memory pressure and janky scrolling on large result sets.

**Location:** `flutter_app/lib/features/activity/activity_screen.dart` line 138

---

Muriano:
## 🟡 LOW Issue

toQueryParameters silently drops the date range when only one of `from`/`to` is non-null (the `else if` requires both). ActivityQuery is a public model — a direct constructor call with only `from` set produces a query with no date constraint, returning unfiltered results. Add an assertion in the constructor (`assert((from == null) == (to == null), 'from and to must both be set or both null')`) or emit each param independently so partial ranges fail loudly.

**Location:** `flutter_app/lib/core/models/activity.dart` line 97

---

Muriano (approved):

Well-structured activity log feature with good model design, proper Riverpod patterns, and solid test coverage. Two previously-flagged issues remain unaddressed in this diff: eager ListView and silent partial-range drop. No new high-severity issues found.

### Suggestions

- Consider sorting entries by timestamp before grouping by hour in _Timeline to avoid duplicate hour headers if the API ever returns entries out of order.
- The _Timeline hour-header logic and tile-building would be cleaner as a pre-computed List<Widget> or indexed builder that separates 'should show header' from widget construction — this pairs naturally with the ListView.builder fix.
- activityOptionsProvider fetches the full page just to extract distinct org/repo names. If the entry count grows, consider a dedicated lightweight endpoint (e.g. GET /activity/filters) that returns only the distinct values for the date window.

---

vbuenog:
Thanks for the second-round review. Addressed in 8e52565 — summary:

### 🔴 Timeline loses day context on multi-day ranges (`activity_screen.dart`)
Added an explicit date header whenever the day changes, in addition to the hour header. Introduced a `sealed class _TimelineItem` with `_DateHeader` / `_HourHeader` / `_EntryItem` / `_TruncationBanner` so the sequence is pre-flattened. New widget test `emits a date header per day in multi-day ranges` covers the case where two days share the same hour.

### ⚠️ Eager ListView (`activity_screen.dart`, flagged by both reviewers)
`ListView(children: items)` → `ListView.builder(itemCount, itemBuilder)` over the flattened `_TimelineItem` list, so headers and tiles are built lazily on scroll.

### ⚠️ `ActivityDisabledException` rendered as generic error
Added a dedicated `_ErrorView` that checks `error is ActivityDisabledException` and renders a friendly empty state ("Activity log is disabled" + hint to enable it in daemon config). Other errors show a shorter message. Covered by a new test.

### 🟡 `showDateRangePicker` missing `initialDateRange`
Now passes `DateTimeRange(start: query.from!, end: query.to!)` when both are set.

### 🟡 `toQueryParameters` drops partial range
Added `assert((from == null) == (to == null), '…')` to the `ActivityQuery` constructor so direct callers fail loudly instead of silently getting an unfiltered query. `ActivityQueryNotifier.setRange` also swaps `from`/`to` when inverted. New unit tests cover both.

### 🟡 `activityOptionsProvider` used default limit
Bumped to `10000` for the options query so the distinct org/repo lists aren't derived from a truncated slice. A dedicated `/activity/facets` endpoint is noted as the proper long-term fix.

### Suggestions also applied
- Snackbar on entry tap now `hideCurrentSnackBar()` before showing the next one.
- New tests: multi-day timeline grouping, disabled-state rendering, partial-range assertion, `copyWith` semantics, `setRange` with inverted from/to.

### Deferred (by design)
- **Per-field type checks in `fromJson`** — daemon contract is owned in-repo and existing entries parse fine; per-field guards would be premature. Happy to revisit if we ever see field-type divergence.
- **Sticky hour headers (SliverStickyHeader)** — separate UX effort, not in scope for this PR.
- **Entry outcome/action badge** — UI tweak for a follow-up; subtitle already carries the info.

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Re-review of 8e52565. All previously flagged issues have been properly addressed: date headers are now emitted per day in multi-day ranges, ListView.builder is used for lazy rendering, ActivityDisabledException has a dedicated friendly empty state, showDateRangePicker receives initialDateRange, ActivityQuery asserts that from/to are set together, and activityOptionsProvider uses a larger limit (10000). The deferred items (per-field type checks in fromJson, sticky headers, outcome badge) are reasonable follow-ups. Code is well-structured with a sealed _TimelineItem hierarchy and solid test coverage. No new issues of note.

### Suggestions

- Consider a dedicated /activity/facets endpoint in a follow-up so option lists aren't derived from a bounded 10000-row fetch (already noted in the code comment).
- Minor: in _DatePickerBar, when a range is active (query.date == null, from/to set) and the user taps 'Pick day', initialDate falls back to `today` — could default to `query.from` to preserve context.

---

Muriano:
## 🟡 LOW Issue

fetchActivity manually constructs the base URI via Uri.parse('http://127.0.0.1:$port/activity') instead of using the _uri() helper used by every other method in this class. If _uri() is ever updated (e.g., HTTPS, configurable host), this endpoint won't pick up the change. Use _uri('/activity').replace(queryParameters: q.toQueryParameters()) for consistency. Impact: maintenance divergence risk.

**Location:** `flutter_app/lib/core/api/api_client.dart` line 83

---

Muriano (approved):

All previously flagged issues are properly addressed. The code is well-structured after two rounds of review. Only minor consistency and UX items remain.

### Suggestions

- The org/repo option lists derived from activityOptionsProvider are in insertion order. Consider sorting them alphabetically (availableOrgs..sort(), availableRepos..sort()) for a more predictable UX in the filter bottom sheets.
- ActivityEntry.fromJson performs hard casts (json['id'] as int, json['org'] as String, etc.) — acceptable given the daemon contract is owned in-repo as the author noted, but if the daemon ever sends a null for a required field, the cast error message won't identify which field failed. A brief comment noting this design choice would help future maintainers.

---

Muriano:
## ⚠️ MEDIUM Issue

fetchActivity uses Uri.parse('http://127.0.0.1:$port/activity') instead of the _uri('/activity') helper used by all other methods. If _uri() changes (HTTPS, configurable host), this endpoint diverges silently. Fix: `_uri('/activity').replace(queryParameters: q.toQueryParameters())`.

**Location:** `flutter_app/lib/core/api/api_client.dart` line 83

---

Muriano:
## 🟡 LOW Issue

activityOptionsProvider fetches up to 10,000 entries just to derive distinct org/repo lists for filter chips. All 10k ActivityEntry objects are deserialized and held in memory when only the unique org/repo strings are needed. This is a workaround acknowledged in the code comments — acceptable short-term, but the memory cost scales linearly with activity volume and the full entries are never displayed.

**Location:** `flutter_app/lib/features/activity/activity_providers.dart` line 75

---

Muriano:
## 🟡 LOW Issue

Filter option lists are derived via .map().toSet().toList() on every rebuild of ActivityScreen, iterating up to 10,000 entries twice (once for orgs, once for repos). Since activityOptionsProvider is a FutureProvider and the result is immutable, consider caching the derived sets in a separate provider or computed value to avoid redundant iteration on each widget rebuild.

**Location:** `flutter_app/lib/features/activity/activity_screen.dart` line 27

---

Muriano (approved):

Well-structured activity log feature with clean model/provider/UI separation. Previous review rounds addressed major issues (eager ListView, stale closures, partial range). One remaining consistency issue with URI construction was already flagged. The code is in good shape after two rounds of feedback.

### Suggestions

- The _Timeline._buildItems() method is called on every build. For large entry lists, consider caching the flattened item list (e.g., in a provider or via memoization) so it isn't recomputed on scroll-triggered rebuilds.
- ActivityQuery.toQueryParameters always includes 'limit' even when using the default 500. Consider omitting it when it matches the server default to keep URLs cleaner.
- The 'Activity' tab (index 0) and 'Activity log' tab (index 2) naming could confuse users. If the distinction is intentional, consider renaming the first to 'Dashboard' or 'Overview' for clarity.
- ActivityEntry.fromJson does unchecked casts (e.g., json['id'] as int). If the daemon ever sends a double or null for 'id', this throws a TypeError with no context. Consider adding a try/catch wrapper or explicit null checks for better error diagnostics in production.

---

PR #126, opened 2026-04-21
feat: Flutter Web replaces SvelteKit web_ui (#125)

### LINKED ISSUE #125: feat: replace SvelteKit web_ui with Flutter Web + Nginx
## Summary

Replace the SvelteKit `web_ui/` with a Flutter Web build of `flutter_app/`, eliminating the dual-codebase problem. The web target is internal-only (Docker deployments), so CanvasKit size and SEO are non-concerns.

**Spec:** [`docs/superpowers/specs/2026-04-21-flutter-web-replaces-svelte-design.md`](docs/superpowers/specs/2026-04-21-flutter-web-replaces-svelte-design.md)

## Architecture change

```
Before: browser → SvelteKit (Node.js, proxy + SSR) → daemon
After:  browser → Nginx (static assets + proxy)    → daemon
```

Nginx serves Flutter Web assets and proxies `/api/*` + `/events` to the daemon, injecting `X-Heimdallm-Token` server-side. The daemon is unchanged.

### Phase 1: Platform abstraction layer
- [ ] Create `lib/core/platform/` with abstract `PlatformServices` interface
- [ ] Implement `platform_services_desktop.dart` (tray, window, notifier, PID, signals)
- [ ] Implement `platform_services_web.dart` (no-op stubs)
- [ ] Adapt `ApiClient` — conditional import: web uses relative `/api/` URL, no token
- [ ] Adapt `SseClient` — conditional import: web uses relative `/events` URL, no token
- [ ] Refactor `main.dart` bootstrap — desktop vs web boot path via `PlatformServices`
- [ ] Remove all unconditional `dart:io` imports from shared code (12 files)

### Phase 2: Enable Flutter Web
- [ ] `flutter create --platforms web .` to generate `web/` directory
- [ ] Customize `web/index.html` with Heimdallm branding
- [ ] Verify `flutter build web --release` compiles with zero errors
- [ ] Smoke test in browser against a local daemon

### Phase 3: Nginx + Docker
- [ ] Create `flutter_app/nginx.conf.template` with proxy + token injection
- [ ] Create `flutter_app/Dockerfile.web` (multi-stage: Flutter build → nginx:alpine)
- [ ] Add `docker-entrypoint.d/` script to read token from `/data/api_token` fallback
- [ ] Update `docker/docker-compose.yml` — replace web service (Node → Nginx)
- [ ] Verify `docker compose up` starts daemon + web, health checks pass
- [ ] Verify all routes work through Nginx proxy (dashboard, PRs, issues, config, logs, agents)
- [ ] Verify SSE events stream correctly through Nginx

### Phase 4: Cleanup
- [ ] Delete `web_ui/` directory entirely
- [ ] Update any CI workflows that reference `web_ui/`

## Acceptance criteria

- `docker compose up` in `docker/` starts daemon + Flutter Web, both healthy
- All existing web routes work (dashboard, `/prs/:id`, `/issues/:id`, `/config`, `/logs`, `/agents`)
- SSE live-refresh works (PR list updates when daemon processes a review)
- Desktop app (`flutter run -d macos`) is unaffected — tray, window, notifications work
- `web_ui/` directory no longer exists
- Web Docker image is ≤50 MB (vs ~180 MB Node.js)

## Out of scope

- Daemon changes (Go) — no CORS, no new auth
- `docs/index.html` landing page — independent
- CI for Flutter Web builds — follow-up issue

ivanmunozruiz:
## Summary

Replaces the SvelteKit `web_ui/` with a Flutter Web build of `flutter_app/` served by Nginx, eliminating the dual-codebase problem. Closes #125.

- **Platform abstraction layer** (`lib/core/platform/`): `PlatformServices` interface with conditional imports selects `DesktopPlatformServices` (dart:io + tray_manager + window_manager + local_notifier + FirstRunSetup + DaemonLifecycle) or `WebPlatformServices` (no-ops + `/api` base URL) at compile time. Every shared file in `lib/features/` and `lib/shared/` is now web-safe — verified by grep gate (Task 12) plus a canary `flutter build web --release` that succeeds with zero errors.
- **ApiClient + SseClient** take base URL and token from `PlatformServices`. Desktop keeps absolute `http://127.0.0.1:7842` + file-based token; web uses relative `/api` + no token (Nginx injects it). Both SSE endpoints (`/events`, `/logs/stream`) flow through the same `/api/` Nginx proxy.
- **Nginx container** (`flutter_app/Dockerfile.web`, `nginx.conf.template`, `docker-entrypoint.d/10-heimdallm-token.sh`): multi-stage build from `ghcr.io/cirruslabs/flutter:stable` into `nginx:1.27-alpine`. Entrypoint resolves token via `HEIMDALLM_API_TOKEN` → `HEIMDALLM_API_TOKEN_FILE` → `/data/api_token`, then `envsubst` renders a single `server {}` block with `/api/` proxy (HTTP + SSE), SPA fallback via `try_files`, and `/healthz`. Final image: **~35 MB** (vs ~200 MB Node).
- **docker-compose.yml**: drop-in replacement on port 3000 with `depends_on: heimdallm: condition: service_healthy`, read-only `/data` mount, and a wget healthcheck against `/healthz`.
- **Integration smoke** (`docker/scripts/test-web.sh`): six curl checks cover the Flutter shell, compiled bundle, `/healthz`, proxied `/api/health`, SSE stream headers, and SPA deep-link fallback. All green.
- **Delete `web_ui/`** entirely. README + project tree updated.

## Test plan

- [x] `cd flutter_app && flutter test` → 40/40 green (includes 7 new `WebPlatformServices` tests, 8 new `DesktopPlatformServices` tests with temp files, new `ApiClient`/`SseClient` web-shape tests, and `config_test.dart` spawn-through-platform assertion).
- [x] `cd flutter_app && flutter analyze` → 0 issues.
- [x] `make build-web` → succeeds (~37 MB bundle). Canary for `dart:io` leaks.
- [x] `make test-docker` → daemon regression green, all internal packages pass.
- [x] `docker/scripts/test-web.sh` → 6/6 checks green against `docker compose up`.
- [ ] Manual macOS smoke: `make build-daemon && cd flutter_app && HEIMDALLM_DAEMON_PATH=$(pwd)/../daemon/bin/heimdallm flutter run -d macos` — verify tray, window, single-instance (SIGUSR1), hide-on-close, notifier. Deferred to reviewer.
- [ ] Manual browser walk-through at `http://localhost:3000` covering `/`, `/prs`, `/prs/:id`, `/issues`, `/issues/:id`, `/config`, `/logs`, `/agents`, `/repositories/:name`, plus deep-link reload and live SSE refresh. Deferred to reviewer.

## Out of scope

- Daemon (Go) code — zero changes.
- CI workflows for Flutter Web — follow-up issue.
- Mobile targets.

**Design:** [`docs/superpowers/specs/2026-04-21-flutter-web-replaces-svelte-design.md`](docs/superpowers/specs/2026-04-21-flutter-web-replaces-svelte-design.md)
**Plan:** [`docs/superpowers/plans/2026-04-21-flutter-web-replaces-svelte.md`](docs/superpowers/plans/2026-04-21-flutter-web-replaces-svelte.md) (26 TDD tasks across 4 phases)


sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Solid migration from SvelteKit to Flutter Web/Nginx. The compose, Makefile, and test script changes are well-structured. However, key infrastructure files (Dockerfile.web, nginx.conf.template, entrypoint script) referenced by docker-compose.yml are missing from the diff, and the PR includes a 3200+ line plan document that will rot quickly.

### Issues

🔴 **docker/docker-compose.yml:119** — Build references `dockerfile: Dockerfile.web` in `../flutter_app` context, but Dockerfile.web is not present in this diff. If not already committed on the branch, the `docker compose build` will fail.
🔴 **docker/docker-compose.yml:119** — nginx.conf.template and docker-entrypoint.d/10-heimdallm-token.sh are referenced by the plan and required for the Nginx container to function (token injection, /api/ proxy, SPA fallback, /healthz), but are absent from the diff. Verify they exist on the branch.
⚠️ **docker/docker-compose.yml:127** — Env var renamed from HEIMDALLM_API_URL to DAEMON_URL without a migration note. Users who set HEIMDALLM_API_URL in docker/.env or their shell will silently lose the override.
⚠️ **docker/docker-compose.test.yml:20** — Test web service sets `HEIMDALLM_API_TOKEN=` (empty string) but does not set `DAEMON_URL`. The base compose sets `DAEMON_URL=http://heimdallm:${HEIMDALLM_PORT:-7842}` which resolves the service name `heimdallm`. In the test compose the daemon container is `heimdallm-test` — verify Docker DNS still resolves `heimdallm` (it should, since the service name in compose is unchanged, but the comment suggests the operator may expect the test container name).
🟡 **docker/scripts/test-web.sh:47** — Uses predictable paths `/tmp/index.html` and `/tmp/deep.html` instead of `mktemp`. On shared CI runners this can collide with parallel test runs or be exploited via symlink.
🟡 **docker/scripts/test-web.sh:47** — Cleanup trap removes Docker containers but does not clean up `/tmp/index.html`, `/tmp/deep.html`, or the `mktemp` file created for SSE test headers.
⚠️ **docs/superpowers/plans/2026-04-21-flutter-web-replaces-svelte.md:1** — 3276-line implementation plan committed to the repo. This is a living document with step-by-step code snippets that will diverge from actual implementation. Consider keeping this in the PR description or a wiki, or deleting it once the migration is complete.
🟡 **docker/docker-compose.yml:145** — Healthcheck `start_period` dropped from 15s to 3s. Nginx starts fast, but the first healthcheck also verifies the daemon is reachable through the proxy. If the daemon takes longer than 3s after being 'healthy' to actually serve requests, the web container may flap.

### Suggestions

- Confirm Dockerfile.web, nginx.conf.template, and docker-entrypoint.d/10-heimdallm-token.sh exist on the branch — they are critical for the Nginx container but missing from this diff.
- Add a note in README or docker/.env.example about the HEIMDALLM_API_URL → DAEMON_URL rename so existing users know to update their .env files.
- In test-web.sh, replace fixed /tmp paths with `mktemp` and add them to the cleanup trap for parallel-safe CI.
- Consider adding a test assertion in test-web.sh that the /api/health response actually includes valid JSON, not just HTTP 200 — this would catch Nginx serving the SPA fallback instead of proxying.
- Delete or move the 3200-line plan document once the migration is complete to avoid stale documentation.

---

Muriano:
## 🟡 LOW Issue

Test writes to fixed paths /tmp/index.html and /tmp/deep.html which can collide with parallel CI runs or leave stale artifacts. Use mktemp (as done for the SSE test on line 58) for all temp files, and clean them up in the trap. Impact: flaky CI if two runs overlap. Fix: use mktemp for all temp files and add rm to cleanup().

**Location:** `docker/scripts/test-web.sh` line 47

---

Muriano:
## 🟡 LOW Issue

The curl for /main.dart.js uses -I (HEAD request). Some Nginx configs or Flutter build outputs may not serve the same Content-Type header for HEAD vs GET. If the Flutter build uses hash-based filenames (main.dart.js may not exist as-is), this test could be fragile. Impact: false-negative test failures. Fix: consider using -s (GET) and piping headers, or glob-matching the JS bundle path.

**Location:** `docker/scripts/test-web.sh` line 50

---

Muriano (approved):

Clean replacement of SvelteKit web UI with Flutter Web + Nginx. Docker config, Makefile, docs, and integration tests are well-structured. No high-severity issues found. A couple of minor robustness items in the test script.

### Suggestions

- The cleanup() trap in test-web.sh could also remove the temp files (/tmp/index.html, /tmp/deep.html) to avoid leaving artifacts on failure.
- Consider adding a timeout to the overall test-web.sh script (e.g., via the Makefile target or a wrapper) to avoid hanging indefinitely if docker compose up stalls.
- The docker-compose.test.yml web service inherits the base healthcheck (5s interval, 6 retries). In CI environments where builds are slower, the 3s start_period may be tight — consider overriding start_period to a longer value in the test compose file.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Clean migration from SvelteKit to Flutter Web with Nginx. Architecture is sound: token injection server-side, SPA fallback, shared-volume token flow preserved. The integration smoke test and healthcheck tuning (127.0.0.1 + wget) show careful attention to prior pitfalls. Minor robustness issues in the test script around temp-file handling and HEAD-vs-GET assumptions.

### Issues

🟡 **docker/scripts/test-web.sh:47** — Fixed paths /tmp/index.html and /tmp/deep.html can collide across parallel CI runs and leave stale artifacts. Use mktemp for both and rm them in cleanup().
🟡 **docker/scripts/test-web.sh:50** — curl -I (HEAD) on /main.dart.js assumes Nginx returns identical Content-Type for HEAD as GET, and assumes the bundle name is stable. If Flutter ever emits hashed filenames or Nginx middleware strips headers on HEAD, this becomes a false negative. Prefer a GET with -D - or check via a glob/index.html reference.
🟡 **docker/scripts/test-web.sh:58** — The SSE test uses --max-time 2 and `|| true`, which swallows genuine failures (e.g. connection refused) — only the header grep distinguishes pass from fail. Consider checking curl's exit code explicitly for 28 (timeout, expected) vs other errors, or asserting the header file is non-empty before grep.
🟡 **docker/scripts/test-web.sh:34** — The health-wait loop inspects `docker inspect` but does not fail fast if the container exited; a crashed web container will be polled for 60s before the generic timeout fires. Consider also checking `.State.Status` and failing early on 'exited'.
🟡 **docker/docker-compose.yml:0** — Healthcheck start_period dropped from 15s to 3s with retries=6 and interval=5s — total grace is ~33s. Nginx itself starts fast, but if the entrypoint waits for /data/api_token (daemon healthy gate via depends_on should cover this) consider whether 3s start_period is too aggressive on slow CI. Verify in practice.
🟡 **docker/docker-compose.test.yml:17** — The test web service overrides HEIMDALLM_API_TOKEN to empty but does not set DAEMON_URL; it relies on inheritance from docker-compose.yml. Confirm the merge keeps DAEMON_URL=http://heimdallm:${HEIMDALLM_PORT:-7842} — if the base service's environment is replaced rather than merged by any tooling, the test would silently target the wrong daemon.

### Suggestions

- In test-web.sh, replace /tmp/index.html and /tmp/deep.html with mktemp vars and append them to a cleanup list rm'd in the trap.
- For the main.dart.js check, use `curl -sf $BASE/main.dart.js -D - -o /dev/null | grep -i content-type` (GET with header dump) to avoid HEAD-vs-GET divergence, or assert via parsing index.html for the actual bundle path.
- Tighten the SSE assertion: capture curl's exit code and accept only 0 or 28 (timeout); anything else is a real failure, not an expected cutoff.
- Add a `docker inspect -f '{{.State.Status}}'` check in the health-wait loop to abort early on exited/dead containers.
- Consider adding a negative test: GET /api/health without the token-injecting proxy (e.g. hit the daemon directly) to confirm the token really is coming from Nginx, not passthrough.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean infrastructure swap from SvelteKit to Flutter Web + Nginx. Compose configs are correct (volume mounts, healthchecks, service dependencies). The smoke test script has minor hygiene issues (two already flagged in PR discussion). The 3,276-line plan document is the bulk of the diff but is documentation, not runtime code. No security, correctness, or performance concerns found.

### Suggestions

- docker/scripts/test-web.sh: the cleanup() trap only runs `compose down` but does not remove temp files (/tmp/index.html, /tmp/deep.html, the mktemp file). Add `rm -f /tmp/index.html /tmp/deep.html` to cleanup() — or better, switch all temp files to mktemp as already suggested in the PR discussion.
- docker/docker-compose.test.yml: the comment says 'Unset HEIMDALLM_API_TOKEN' but `HEIMDALLM_API_TOKEN=` sets it to an empty string, not unset. This works if the entrypoint uses `[ -n "$VAR" ]`, but consider clarifying the comment to say 'set to empty' for accuracy.
- docs/superpowers/plans/2026-04-21-flutter-web-replaces-svelte.md: at 3,276 lines this plan file is very large. Consider whether it should live in the repo long-term or be removed/archived after the migration is complete to avoid stale documentation.

---

ivanmunozruiz:
Thanks for the review. Walking through each point:

**🔴 Dockerfile.web / nginx.conf.template / entrypoint missing from diff** — False positive, they are all on the branch:

```
flutter_app/Dockerfile.web                              (commit 36d0443)
flutter_app/nginx.conf.template                         (commit 96981de)
flutter_app/docker-entrypoint.d/10-heimdallm-token.sh   (commit 2d87a07, 100755)
```

The reviewer bot likely didn't have the full diff rendered. `docker compose build web` works end-to-end (verified by `docker/scripts/test-web.sh` which builds the image as part of `up -d --build`).

**⚠️ HEIMDALLM_API_URL → DAEMON_URL rename without migration note** — Valid. Added a migration comment in `docker-compose.yml` in commit `5cbcf9f`:

```
# Migration note (v0.4 → Flutter Web): the env var was renamed
# from HEIMDALLM_API_URL (SvelteKit) to DAEMON_URL (Nginx entrypoint).
# If you had HEIMDALLM_API_URL in docker/.env or your shell, move it
# under this key. Default is fine for the standard compose network.
```

**⚠️ DAEMON_URL resolution in test compose** — False positive. The compose *service* name is still `heimdallm`; only the `container_name` changed to `heimdallm-test`. Docker DNS resolves service names, not container names, so `http://heimdallm:7842` works in both main and test compose. Integration test (`test-web.sh`, check 4/6) passes every time.

**🟡 Fixed `/tmp/` paths in test-web.sh** — Valid. Commit `1549d63` replaces all three with `mktemp` and extends the cleanup trap to `rm -f` them.

**🟡 Cleanup trap doesn't remove tmp files** — Addressed in the same commit.

**⚠️ 3276-line plan committed to repo** — Keeping it. Two reasons: it documents the 26 TDD tasks with exact failing-test → impl → commit sequences that drove this migration, which is useful context for anyone later auditing decisions like "why did SseClient go through `/api/` instead of a dedicated `/events` location", and its cost (one file, never edited again) is small. It lives under `docs/superpowers/plans/` alongside other historical plans already in the repo. Happy to move it to a wiki if you feel strongly.

**🟡 start_period dropped from 15s to 3s** — Valid catch. Commit `5cbcf9f` restores `start_period: 10s` (was tightened during debugging for faster `test-web.sh` convergence; 10s is the original daemon-side value and is safer once the reviewer is watching over the compose shoulder).

**Bonus fix while testing manually**: operator hit a 401 cascade because `docker/.env` had a stale `HEIMDALLM_API_TOKEN` left over from an older `make setup` run. The Nginx entrypoint precedence is env > file, so the stale value shadowed the live `/data/api_token` that the daemon just regenerated. Commit `5cbcf9f` removes the `HEIMDALLM_API_TOKEN=${HEIMDALLM_API_TOKEN:-}` passthrough in the `web` service so `.env` tokens can't leak in; operators can still override with `docker compose -e HEIMDALLM_API_TOKEN=... up web` if they really want to.

**Suggestion to assert JSON on /api/health** — Adopted in the same `1549d63` commit: check 4 now does `grep -q '"status"'` on the body, not just HTTP 200.

Commits pushed: [5cbcf9f](https://github.com/theburrowhub/heimdallm/pull/126/commits/5cbcf9f) and [1549d63](https://github.com/theburrowhub/heimdallm/pull/126/commits/1549d63).

Muriano:
## 🟡 LOW Issue

Comment says 'Unset HEIMDALLM_API_TOKEN' but `HEIMDALLM_API_TOKEN=` sets it to an empty string, not unset. Functionally equivalent only if the entrypoint checks `[ -n "$VAR" ]` rather than `[ -z "${VAR+x}" ]`. Minor documentation inaccuracy — verify the entrypoint treats empty the same as unset.

**Location:** `docker/docker-compose.test.yml` line 19

---

Muriano:
## 🟡 LOW Issue

curl -sfI (HEAD request) for /main.dart.js may not return the same Content-Type header as a GET on some Nginx configurations. This was flagged in the prior review but not addressed. If Flutter uses hashed filenames, main.dart.js may not exist at that exact path. Consider `curl -sf -o /dev/null -D - "$BASE/main.dart.js"` (GET with headers captured) for more reliable assertion.

**Location:** `docker/scripts/test-web.sh` line 50

---

Muriano (approved):

Clean infrastructure swap from SvelteKit to Flutter Web/Nginx. Previous review feedback was addressed (mktemp, start_period, JSON assertion, migration comment, token passthrough removal). The diff is well-documented and the test script covers the key canary routes. No high or medium issues found.

### Suggestions

- The 3276-line plan under docs/superpowers/plans/ is justified per author but consider adding it to .gitattributes with `linguist-documentation` so it doesn't inflate PR diff stats or code-review noise in future diffs that touch nearby files.
- docker-compose.test.yml web service relies entirely on compose-file merging from the base file for build/ports/healthcheck — a short comment noting this dependency would help future maintainers who read the test file in isolation.

---

Muriano:
## 🟡 LOW Issue

curl -sfI sends a HEAD request for /main.dart.js. This was flagged in two prior review rounds but never addressed. HEAD may not return the same Content-Type as GET on some Nginx configs, and it doesn't verify the file actually delivers content (could be 0 bytes). Impact: false-negative test on hardened Nginx or misconfigured builds. Fix: use `curl -sf -o /dev/null -D - "$BASE/main.dart.js" | grep -qi 'content-type: .*javascript'` to do a GET with header capture.

**Location:** `docker/scripts/test-web.sh` line 50

---

Muriano:
## 🟡 LOW Issue

Comment says 'Unset HEIMDALLM_API_TOKEN' but `HEIMDALLM_API_TOKEN=` sets it to an empty string, not unset. Functionally equivalent here because the entrypoint uses `[ -z ]` (true for both empty and unset), but the comment is misleading. Fix: change comment to 'Set HEIMDALLM_API_TOKEN to empty so the entrypoint falls back to reading /data/api_token'.

**Location:** `docker/docker-compose.test.yml` line 19

---

Muriano (approved):

Solid SvelteKit→Flutter Web migration. Docker compose, test compose, and Makefile changes are well-structured. The main gap is an unaddressed review comment about curl -sfI (HEAD) in test-web.sh that was flagged twice. The empty-string-vs-unset comment in test compose is misleading but functionally correct. No security or correctness issues in the core infrastructure changes.

### Suggestions

- The 3276-line plan file (docs/superpowers/plans/2026-04-21-flutter-web-replaces-svelte.md) is committed to the repo. Consider whether this belongs in a wiki or was intentionally kept for audit trail — the author addressed this in discussion and chose to keep it.
- The healthcheck interval was tightened from 30s to 5s and retries increased from 3 to 6. This is fine for dev/test but verify it doesn't cause excessive log noise in long-running production deployments.
- The migration note for HEIMDALLM_API_URL → DAEMON_URL rename is inline in docker-compose.yml comments. Consider also noting this in the PR description or a CHANGELOG entry so operators upgrading from SvelteKit catch the breaking env var change.

---

Muriano:
## 🟡 LOW Issue

curl -sfI sends a HEAD request for /main.dart.js. This was flagged in three prior review rounds and never addressed. HEAD may omit Content-Type on some Nginx configs, and it doesn't verify the response body is non-empty. Fix: `curl -sf -o /dev/null -D - "$BASE/main.dart.js" | grep -qi 'content-type: .*javascript'` (GET with header capture).

**Location:** `docker/scripts/test-web.sh` line 53

---

Muriano:
## 🟡 LOW Issue

Comment says 'Unset HEIMDALLM_API_TOKEN' but `HEIMDALLM_API_TOKEN=` sets it to an empty string, not unset. Flagged twice previously. Works because the entrypoint uses `[ -z ]`, but the comment is misleading. Fix: 'Set HEIMDALLM_API_TOKEN to empty so the entrypoint falls back to /data/api_token'.

**Location:** `docker/docker-compose.test.yml` line 19

---

Muriano (approved):

Re-review after author addressed most prior findings. Two LOW issues flagged multiple times in previous rounds remain unaddressed in the diff: the HEAD-request test for main.dart.js and the misleading 'Unset' comment in test compose. No new medium or high issues found — the migration comment for DAEMON_URL, the restored start_period, the mktemp fix, and the stale-token prevention are all solid improvements.

### Suggestions

- The 3276-line plan file is justified per author's rationale, but consider adding a one-line note at the top linking to the PR number so future readers can find the discussion context quickly.

---

Muriano:
## 🟡 LOW Issue

curl -sfI sends a HEAD request for /main.dart.js. HEAD may omit Content-Type on some Nginx configurations and does not verify the response body is non-empty. This was flagged in three prior review rounds and never addressed. Fix: `curl -sf -o /dev/null -D - "$BASE/main.dart.js" | grep -qi 'content-type: .*javascript'` to use GET with header capture.

**Location:** `docker/scripts/test-web.sh` line 53

---

Muriano:
## 🟡 LOW Issue

Comment says 'Unset HEIMDALLM_API_TOKEN' but `HEIMDALLM_API_TOKEN=` sets it to an empty string, not unset. Works because the entrypoint uses `[ -z ]`, but the comment is misleading. Flagged in prior rounds. Fix: 'Set HEIMDALLM_API_TOKEN to empty so the entrypoint falls back to reading /data/api_token'.

**Location:** `docker/docker-compose.test.yml` line 19

---

Muriano (approved):

Well-executed SvelteKit→Flutter Web migration. Docker, compose, docs, and test infrastructure are solid. Two recurring low-severity issues from prior review rounds remain unaddressed in the diff: the HEAD-vs-GET curl in the smoke test and a misleading comment in the test compose file. No high or medium issues found.

### Suggestions

- The 3276-line plan file under docs/superpowers/plans/ adds significant repo weight for a one-time migration document. Consider moving it to a wiki or a separate docs archive branch if repo size becomes a concern.
- In Makefile _post-up-hints, the eval-based env var detection loop (`v=$(eval "printf %s \"\${$$k:-}\"")`) is correct but hard to read. A simpler `printenv "$k" 2>/dev/null` achieves the same result and is easier to maintain.

---

Muriano:
## 🟡 LOW Issue

Comment says 'Unset HEIMDALLM_API_TOKEN' but `HEIMDALLM_API_TOKEN=` sets it to an empty string, not unset. This has been flagged in four prior review rounds. While functionally equivalent because the entrypoint uses `[ -z ]`, the comment is misleading. Fix: 'Set HEIMDALLM_API_TOKEN to empty so the entrypoint falls back to reading /data/api_token'.

**Location:** `docker/docker-compose.test.yml` line 19

---

Muriano (approved):

Solid migration from SvelteKit to Flutter Web with well-tested ResolveLocalDir logic, proper mutex discipline, and thorough documentation updates. No blocking issues found. The repeated review comments about HEAD requests in test-web.sh and the misleading 'Unset' comment in docker-compose.test.yml remain unaddressed from prior rounds but are low severity.

### Suggestions

- The `aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, ...)` pattern appears 4 times in main.go (lines ~279, 342, 628, 697). Consider extracting a helper method on the config or moving the resolution into AIForRepo itself to reduce duplication.
- The /config endpoint calls os.Stat for every monitored repo on every request to build localDirsDetected. For large repo lists this adds latency; consider caching the map with a short TTL or recomputing only on config reload.
- The HEAD-vs-GET concern for /main.dart.js in test-web.sh has been flagged in multiple prior rounds. Using `curl -sf -o /dev/null -D - "$BASE/main.dart.js" | grep -qi 'content-type: .*javascript'` would be more robust across Nginx configurations.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid migration from SvelteKit to Flutter Web with good documentation and tests. The main concern is a potential path traversal in ResolveLocalDir — while mitigated by GitHub's repo-name validation, it lacks defense-in-depth sanitization. Several minor issues flagged in prior reviews (misleading comment, HEAD vs GET in test) remain unaddressed in this diff snapshot. Critical files (Dockerfile.web, nginx.conf.template, entrypoint script) are truncated from the diff and could not be reviewed.

### Issues

⚠️ **daemon/internal/config/config.go:310** — Path traversal: ShortRepoName can return '..' (e.g. repo='org/..'), causing filepath.Join(DefaultReposMountPath, '..') to resolve to the parent directory. ResolveLocalDir would then return '/' (or any ancestor), giving the AI agent a CWD outside the intended /repos mount. GitHub's own repo-name validation prevents this in practice, but defense-in-depth demands a check — e.g. reject short names containing '..' or verify the resolved candidate is still under DefaultReposMountPath with strings.HasPrefix after filepath.Clean.
⚠️ **daemon/internal/config/config_test.go:1072** — No test for path-traversal inputs to ShortRepoName or ResolveLocalDir (e.g. 'org/..', 'org/../../../etc/passwd', 'org/.'). Add cases to confirm these are rejected or resolve safely within the mount root.
🟡 **daemon/cmd/heimdallm/main.go:277** — The pattern `aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, X.Repo)` is duplicated at four call sites (lines ~277, ~342, ~628, ~697) with near-identical comments. Consider folding the fallback into AIForRepo or a thin wrapper to eliminate the repetition and the risk of forgetting the call at a future fifth site.
🟡 **docker/docker-compose.test.yml:19** — Comment says 'Unset HEIMDALLM_API_TOKEN' but `HEIMDALLM_API_TOKEN=` sets it to an empty string, not unset. This has been flagged in multiple prior review rounds. Fix the comment to: 'Set HEIMDALLM_API_TOKEN to empty so the entrypoint falls back to reading /data/api_token'.
🟡 **docker/docker-compose.yml:90** — The fallback `${HEIMDALLM_REPOS_DIR:-./repos-placeholder}` mounts a path that may not exist. Docker will auto-create it as a root-owned empty directory, which works but leaves a surprise artifact in the repo checkout. Consider adding repos-placeholder/ to .gitignore or using a named volume as the fallback.
🟡 **Makefile:319** — _post-up-hints uses `eval` with variable indirection (`$$(eval "printf %s \"\$${$$k:-}\"")`) which is hard to read and audit. A simpler approach: source the env file, then test each variable directly with `[ -n "$ANTHROPIC_API_KEY" ]` etc., avoiding eval entirely.

### Suggestions

- Add a filepath containment check in ResolveLocalDir: after filepath.Join, verify the cleaned candidate starts with DefaultReposMountPath+'/' before returning it. This closes the path traversal gap regardless of input source.
- Add test cases for adversarial repo names ('org/..', 'org/.', 'org/../../etc', '') to ShortRepoName and ResolveLocalDir to codify the expected security boundary.
- The truncated diff omits Dockerfile.web, nginx.conf.template, 10-heimdallm-token.sh, and test-web.sh — all security-sensitive (token handling, reverse proxy config). Ensure these get a dedicated review pass.
- Address the repeated bot feedback: fix the misleading 'Unset' comment in docker-compose.test.yml and switch test-web.sh from HEAD (-I) to GET with header capture for the main.dart.js check.
- Consider consolidating the four ResolveLocalDir call sites into a method on Config (e.g. ResolvedAIForRepo) that returns an AIConfig with LocalDir already resolved.

---

Muriano:
## 🟡 LOW Issue

Comment says 'Unset HEIMDALLM_API_TOKEN' but `HEIMDALLM_API_TOKEN=` sets it to empty string, not unset. Flagged multiple times in prior rounds — still unchanged. Functionally correct (entrypoint uses `[ -z ]`), but misleading. Fix: 'Set HEIMDALLM_API_TOKEN to empty so the entrypoint falls back to reading /data/api_token'.

**Location:** `docker/docker-compose.test.yml` line 19

---

Muriano (approved):

Re-review after author addressed prior feedback. The ResolveLocalDir implementation is correct with good test coverage. Documentation and Docker changes are clean. Two minor items remain from prior rounds plus one new defensive suggestion.

### Suggestions

- daemon/internal/config/config.go ResolveLocalDir: consider validating that `short` does not contain path traversal segments (e.g. `..`). Currently safe because repo names flow from GitHub API (which forbids `..`) and operator config, but `filepath.Join(DefaultReposMountPath, "..")` would resolve to `/` and pass the IsDir check. A `strings.Contains(short, "..")` guard adds defense-in-depth at near-zero cost.
- Makefile _post-up-hints: the `eval "printf %s \"\${$$k:-}\""` indirect expansion works but is hard to read. A `printenv "$k" 2>/dev/null || true` would be clearer and avoid eval.

---

Muriano:
## 🟡 LOW Issue

The _post-up-hints target uses `eval` for indirect variable expansion (`v=$$(eval "printf %s \"\$${$$k:-}\""`). While the input comes from a local .env file (operator-controlled), `eval` with shell variables is a footgun — a malformed or adversarial .env value (e.g. a key containing backticks or $(...)) would be executed. Impact: local shell injection if .env is tampered with. Fix: use POSIX indirect expansion via `eval v=\$${k}` or switch to `printenv "$k"` after sourcing, which avoids eval entirely.

**Location:** `Makefile` line 320

---

Muriano:
## 🟡 LOW Issue

DefaultReposMountPath is a mutable package-level variable used for testability. While this is a common Go pattern, it introduces a subtle data race if tests run in parallel (`-parallel`) since multiple test goroutines mutate and restore the same global. Impact: flaky test failures under `-race` with `-parallel`. Fix: accept the mount path as a parameter to ResolveLocalDir (e.g. `ResolveLocalDir(configured, repo, mountPath string)`) or use `t.Setenv`-style approach with `sync.Mutex` in tests.

**Location:** `daemon/internal/config/config.go` line 292

---

Muriano (approved):

Well-structured migration from SvelteKit to Flutter Web with solid documentation, good test coverage for new config functions, and a clean Docker compose setup. The repeated review bot comments about test-web.sh and docker-compose.test.yml have already been addressed by the author in follow-up commits. No high-severity issues found; only minor improvements remain.

### Suggestions

- The 4 call sites of `config.ResolveLocalDir(aiCfg.LocalDir, <repo>)` in main.go follow an identical pattern (unlock mutex → resolve → continue). Consider adding a helper method on Config like `func (c *Config) AIForRepoResolved(repo string) RepoAI` that calls ResolveLocalDir internally, reducing duplication and the risk of forgetting the call in future code paths.
- The test for `ShortRepoName("trailing-slash/")` expects empty string, which is correct but worth a brief inline comment explaining why this is the desired behavior (prevents accidental resolution to the mount root).
- In docker-compose.test.yml, the comment was updated to say 'Set HEIMDALLM_API_TOKEN to empty' which accurately describes the behavior — good fix from the review feedback. Consider also adding a one-liner noting the entrypoint precedence (env > file > empty) for future maintainers.

---

Muriano:
## 🟡 LOW Issue

The eval for indirect variable expansion (`v=$$(eval "printf %s \"$${$$k:-}\"")`) will interpret command substitutions inside .env values. If an operator's .env contains e.g. `ANTHROPIC_API_KEY=$(malicious)`, eval executes it. While self-inflicted (operator-controlled file), `printenv "$k"` after the `. $(DOCKER_ENV)` source would achieve the same result without eval. Fix: replace the eval line with `v=$(printenv "$k" 2>/dev/null || true);`.

**Location:** `Makefile` line 320

---

Muriano:
## 🟡 LOW Issue

Comment says 'Unset HEIMDALLM_API_TOKEN' on the branch head, but `HEIMDALLM_API_TOKEN=` sets it to empty string, not unset. The provided diff shows the corrected wording ('Set HEIMDALLM_API_TOKEN to empty') but the branch head still has 'Unset'. If the diff is stale, this is a documentation inaccuracy — functionally correct since the entrypoint uses `[ -z ]`, but misleading. Verify the branch head matches the intended fix.

**Location:** `docker/docker-compose.test.yml` line 19

---

Muriano (approved):

Solid migration from SvelteKit to Flutter Web with good test coverage, comprehensive documentation, and well-structured auto-detection logic. The Go changes are clean with proper mutex discipline. Two low-severity issues remain: an eval usage in the Makefile that could be replaced with a safer pattern, and a mutable package-level variable that's fragile under parallel test execution. The repeatedly-flagged HEAD-request test and misleading 'Unset' comment in docker-compose.test.yml appear to still be present on the branch head despite being addressed in the provided diff.

### Suggestions

- DefaultReposMountPath is a mutable package-level var mutated by tests via save/restore. Safe today since no test calls t.Parallel(), but fragile — consider accepting the mount path as a parameter to ResolveLocalDir (`ResolveLocalDir(configured, repo, mountRoot string)`) to eliminate the shared mutable state entirely.
- test-web.sh check 2/6 still uses `curl -sfI` (HEAD) for /main.dart.js. HEAD may omit Content-Type on some Nginx configs. Consider `curl -sf -o /dev/null -D - "$BASE/main.dart.js" | grep -qi 'content-type: .*javascript'` for a more reliable GET-based assertion. This was flagged in prior rounds.
- The _post-up-hints target sources .env three times (once per section). A single source at the top of a combined shell block would be cleaner and avoid re-parsing the file.

---

PR #127, opened 2026-04-21
feat: Stats filters — org and repo multi-select (#110)

### LINKED ISSUE #110: feat: Stats filters — organization and repository multi-select
## Summary

Add organization and repository multi-select filters to the Stats tab.

## Current state

Stats shows global aggregates (Total Reviews, Avg Issues/Review, High Severity, Low Severity) with no filtering capability.

## Proposed changes

Add a filter bar at the top of Stats with:
- **Organization** — multi-select dropdown (same component as Activity filters)
- **Repository** — multi-select dropdown, filtered by selected org(s)

Stats cards and any future charts recalculate based on the selected filter. Empty filter = all repos (current behavior).

## Backwards compatibility

The daemon's `GET /stats` endpoint currently returns global stats. Either:
- Add query params (`?repos=org/repo1,org/repo2`) to filter server-side, OR
- Filter client-side from the PR/review data already loaded

Client-side filtering is simpler and avoids daemon changes.

Muriano:
## Summary

- **Daemon**: `GET /stats?repos=org/repo1,org/repo2` — all SQL aggregations scoped by repo filter. Empty = global (backwards compatible).
- **Flutter**: Org/repo multi-select filter bar on Stats tab (same UX pattern as Activity filters).

## Changes

| File | Change |
|---|---|
| `daemon/internal/store/store.go` | `ComputeStats(repos []string)` with conditional WHERE clauses |
| `daemon/internal/server/handlers.go` | Parse `?repos=` query param |
| `flutter_app/lib/features/stats/stats_filters.dart` | New: StatsFilters model + provider |
| `flutter_app/lib/features/stats/stats_filter_bar.dart` | New: filter bar widget |
| `flutter_app/lib/features/stats/stats_screen.dart` | Add filter bar, derive repo list from PRs+Issues |
| `flutter_app/lib/features/dashboard/dashboard_providers.dart` | statsProvider reads filters, passes ?repos= |
| `flutter_app/lib/core/api/api_client.dart` | fetchStats accepts optional repos param |

## Test Plan
- [x] `go test ./... -race` — all pass
- [ ] Manual: filter by org → stats update
- [ ] Manual: filter by repo → stats update
- [ ] Manual: reset → global stats restored

Closes #110

### DIFF
--- a/daemon/internal/server/handlers.go
+++ b/daemon/internal/server/handlers.go
@@ -778,7 +778,15 @@ func (srv *Server) handleRepoCollaborators(w http.ResponseWriter, r *http.Reques
 }
 
 func (srv *Server) handleStats(w http.ResponseWriter, r *http.Request) {
-	stats, err := srv.store.ComputeStats()
+	var repos []string
+	if raw := r.URL.Query().Get("repos"); raw != "" {
+		for _, s := range strings.Split(raw, ",") {
+			if s = strings.TrimSpace(s); s != "" {
+				repos = append(repos, s)
+			}
+		}
+	}
+	stats, err := srv.store.ComputeStats(repos)
 	if err != nil {
 		slog.Error("handleStats: store error", "err", err)
 		http.Error(w, "internal server error", http.StatusInternalServerError)
--- a/daemon/internal/store/store.go
+++ b/daemon/internal/store/store.go
@@ -3,6 +3,7 @@ package store
 import (
 	"database/sql"
 	"fmt"
+	"strings"
 	"time"
 
 	_ "modernc.org/sqlite"
@@ -226,17 +227,32 @@ type DayCount struct {
 }
 
 // ComputeStats aggregates statistics from the reviews and prs tables.
-func (s *Store) ComputeStats() (*Stats, error) {
+// When repos is non-empty, results are scoped to reviews of PRs in those repos.
+func (s *Store) ComputeStats(repos []string) (*Stats, error) {
 	stats := &Stats{
 		BySeverity: make(map[string]int),
 		ByCLI:      make(map[string]int),
 	}
 
+	// Build a reusable subquery that restricts review IDs to the given repos.
+	// When repos is empty, the subquery is omitted (global stats).
+	var repoFilter string
+	var repoArgs []any
+	if len(repos) > 0 {
+		placeholders := make([]string, len(repos))
+		for i, r := range repos {
+			placeholders[i] = "?"
+			repoArgs = append(repoArgs, r)
+		}
+		inClause := strings.Join(placeholders, ",")
+		repoFilter = " AND r.pr_id IN (SELECT id FROM prs WHERE repo IN (" + inClause + "))"
+	}
+
 	// Total reviews
-	s.db.QueryRow("SELECT COUNT(*) FROM reviews").Scan(&stats.TotalReviews)
+	s.db.QueryRow("SELECT COUNT(*) FROM reviews r WHERE 1=1"+repoFilter, repoArgs...).Scan(&stats.TotalReviews)
 
 	// By severity
-	rows, _ := s.db.Query("SELECT severity, COUNT(*) FROM reviews GROUP BY severity")
+	rows, _ := s.db.Query("SELECT severity, COUNT(*) FROM reviews r WHERE 1=1"+repoFilter+" GROUP BY severity", repoArgs...)
 	if rows != nil {
 		defer rows.Close()
 		for rows.Next() {
@@ -248,7 +264,7 @@ func (s *Store) ComputeStats() (*Stats, error) {
 	}
 
 	// By CLI
-	rows2, _ := s.db.Query("SELECT cli_used, COUNT(*) FROM reviews GROUP BY cli_used")
+	rows2, _ := s.db.Query("SELECT cli_used, COUNT(*) FROM reviews r WHERE 1=1"+repoFilter+" GROUP BY cli_used", repoArgs...)
 	if rows2 != nil {
 		defer rows2.Close()
 		for rows2.Next() {
@@ -260,12 +276,12 @@ func (s *Store) ComputeStats() (*Stats, error) {
 	}
 
 	// Top repos by review count
-	rows3, _ := s.db.Query(`
+	topRepoQuery := `
 		SELECT p.repo, COUNT(r.id) as cnt
 		FROM reviews r JOIN prs p ON p.id = r.pr_id
-		WHERE p.repo != ''
-		GROUP BY p.repo ORDER BY cnt DESC LIMIT 8
-	`)
+		WHERE p.repo != ''` + repoFilter + `
+		GROUP BY p.repo ORDER BY cnt DESC LIMIT 8`
+	rows3, _ := s.db.Query(topRepoQuery, repoArgs...)
 	if rows3 != nil {
 		defer rows3.Close()
 		for rows3.Next() {
@@ -276,12 +292,12 @@ func (s *Store) ComputeStats() (*Stats, error) {
 	}
 
 	// Reviews per day last 7 days
-	rows4, _ := s.db.Query(`
-		SELECT DATE(created_at) as day, COUNT(*) as cnt
-		FROM reviews
-		WHERE created_at >= datetime('now', '-7 days')
-		GROUP BY day ORDER BY day ASC
-	`)
+	last7Query := `
+		SELECT DATE(r.created_at) as day, COUNT(*) as cnt
+		FROM reviews r
+		WHERE r.created_at >= datetime('now', '-7 days')` + repoFilter + `
+		GROUP BY day ORDER BY day ASC`
+	rows4, _ := s.db.Query(last7Query, repoArgs...)
 	if rows4 != nil {
 		defer rows4.Close()
 		for rows4.Next() {
@@ -293,27 +309,25 @@ func (s *Store) ComputeStats() (*Stats, error) {
 
 	// Avg issues per review (issues is a JSON array stored as text)
 	var totalIssues, reviewsWithIssues int
-	s.db.QueryRow(`SELECT COUNT(*) FROM reviews WHERE issues != '[]' AND issues != 'null'`).Scan(&reviewsWithIssues)
+	s.db.QueryRow("SELECT COUNT(*) FROM reviews r WHERE issues != '[]' AND issues != 'null'"+repoFilter, repoArgs...).Scan(&reviewsWithIssues)
 	if reviewsWithIssues > 0 {
-		// Approximate: count total issue objects via json_array_length
-		s.db.QueryRow(`SELECT COALESCE(SUM(json_array_length(issues)),0) FROM reviews WHERE issues IS NOT NULL`).Scan(&totalIssues)
+		s.db.QueryRow("SELECT COALESCE(SUM(json_array_length(issues)),0) FROM reviews r WHERE issues IS NOT NULL"+repoFilter, repoArgs...).Scan(&totalIssues)
 		if stats.TotalReviews > 0 {
 			stats.AvgIssuesPerReview = float64(totalIssues) / float64(stats.TotalReviews)
 		}
 	}
 
 	// Review timing: duration from pipeline start (prs.fetched_at) to AI done (reviews.created_at).
-	// Fetch last 200 published reviews and compute stats in Go for accuracy.
-	timingRows, _ := s.db.Query(`
+	timingQuery := `
 		SELECT (julianday(r.created_at) - julianday(p.fetched_at)) * 86400.0
 		FROM reviews r
 		JOIN prs p ON p.id = r.pr_id
 		WHERE r.github_review_id > 0
 		  AND p.fetched_at IS NOT NULL
-		  AND p.fetched_at != ''
+		  AND p.fetched_at != ''` + repoFilter + `
 		ORDER BY r.created_at DESC
-		LIMIT 200
-	`)
+		LIMIT 200`
+	timingRows, _ := s.db.Query(timingQuery, repoArgs...)
 	if timingRows != nil {
 		var durations []float64
 		for timingRows.Next() {
--- a/daemon/internal/store/store.go
+++ b/daemon/internal/store/store.go
@@ -234,9 +234,10 @@ func (s *Store) ComputeStats(repos []string) (*Stats, error) {
 		ByCLI:      make(map[string]int),
 	}
 
-	// Build a reusable subquery that restricts review IDs to the given repos.
-	// When repos is empty, the subquery is omitted (global stats).
-	var repoFilter string
+	// Build reusable filter clauses for the given repos.
+	// repoFilter: for queries on reviews only (uses subquery on prs table).
+	// repoFilterJoined: for queries that already JOIN prs p (direct p.repo IN).
+	var repoFilter, repoFilterJoined string
 	var repoArgs []any
 	if len(repos) > 0 {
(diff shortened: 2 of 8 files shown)


sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

The backend SQL filtering is correctly parameterized and safe. However, the Flutter popup has a checkbox state bug that will prevent visual toggling, and selecting only an org filter silently does nothing to the stats data.

### Issues

🔴 **flutter_app/lib/features/stats/stats_filter_bar.dart:101** — Checkbox won't visually toggle. `checked` is captured as a final in the outer `itemBuilder` closure. When `setMenuState(() {})` rebuilds the `StatefulBuilder`, `checked` still holds the stale captured value, so the checkbox appears unchanged until the popup is closed and reopened.
⚠️ **flutter_app/lib/features/dashboard/dashboard_providers.dart:133** — Selecting orgs without explicitly selecting repos sends no filter to the API, making the org-only selection a no-op for actual stats data. Users will see the org chips as active but stats won't change. Either auto-include all repos for the selected orgs, or make the org filter drive the API query directly.
🟡 **daemon/internal/store/store.go:281** — The 'Top repos' query already JOINs `prs p`, but `repoFilter` adds a redundant subquery (`r.pr_id IN (SELECT id FROM prs WHERE repo IN (...))`). For this specific query, a simpler `AND p.repo IN (...)` would avoid the extra subquery scan. Minor perf issue.
⚠️ **flutter_app/lib/features/stats/stats_filter_bar.dart:93** — PopupMenuButton with CheckboxListTile items: tapping outside the checkbox hit-area on a PopupMenuItem will close the popup via Navigator.pop instead of toggling the checkbox. This makes the multi-select interaction unreliable. Consider using a dialog or a custom overlay instead of PopupMenuButton for multi-select.
🟡 **daemon/internal/store/store.go:312** — The `reviewsWithIssues` count query applies `repoFilter`, but `AvgIssuesPerReview` divides `totalIssues` by `stats.TotalReviews` (total reviews). If repos are filtered, `totalIssues` is scoped to those repos but `TotalReviews` is also scoped, so the ratio is consistent — however, semantically this computes avg over all reviews (including those with zero issues). The variable name `reviewsWithIssues` is checked but never used in the denominator, which may be intentional but is worth a comment.

### Suggestions

- Fix the checkbox state bug by moving `checked` computation inside the `StatefulBuilder` builder, or by maintaining a local `Set<String>` inside `StatefulBuilder` that is updated on toggle.
- When orgs are selected but repos are empty, compute the effective repo set (all repos belonging to selected orgs) and send that to the API, so org filtering actually affects the data.
- Replace PopupMenuButton with a showDialog-based multi-select for more reliable checkbox interaction (similar to how ActivityFilterBar works if it uses dialogs).
- Consider adding a dedicated `repoFilter` variant for queries that already JOIN `prs` (e.g., `AND p.repo IN (...)`) vs. those that don't (subquery form), to avoid redundant subqueries.

---

Muriano:
## Fixes applied

All review feedback addressed in 81782c6:

| Issue | Fix |
|---|---|
| 🔴 Checkbox won't toggle (stale `checked`) | Replaced `PopupMenuButton` + `StatefulBuilder` with `AlertDialog` + `StatefulWidget` — local `_selected` set drives checkbox state correctly |
| ⚠️ Org-only filter is no-op | `statsProvider` now derives effective repos from known PRs+issues when orgs are selected but repos aren't |
| ⚠️ PopupMenuItem closes on tap | Switched to dialog-based multi-select — checkboxes work reliably, Apply/Cancel buttons |
| 🟡 Redundant subquery in top_repos/timing | Added `repoFilterJoined` (`AND p.repo IN (...)`) for queries that already JOIN prs |

All tests pass with `-race`, Flutter analyze clean.

vbuenog (approved):
## 🤖 Heimdallm AI Review

Clean, focused implementation of server-side stats filtering by org/repo. SQL queries correctly use parameterized placeholders (no injection risk), the API is backwards-compatible, and the Flutter UI mirrors an existing pattern. One subtle reactivity issue in the org-only filter derivation.

### Issues

⚠️ **flutter_app/lib/features/dashboard/dashboard_providers.dart:143** — statsProvider uses ref.read for prsProvider/issuesProvider when deriving repos from an org-only filter. If PRs/issues finish loading (or change) after stats has been computed, the org filter won't re-derive and stats will stay scoped to stale repo data. Use ref.watch to make the derivation reactive, or at least invalidate statsProvider when prs/issues change.
🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:145** — Org-only filter silently degrades to global stats when prs/issues haven't loaded yet (both valueOrNull fall through to empty, producing repos=[]). User sees unfiltered data without any indication. Consider showing a loading state or deferring the fetch until the derivation source is available.
🟡 **daemon/internal/store/store.go:240** — repoFilter uses a subquery (r.pr_id IN (SELECT id FROM prs WHERE repo IN (...))) for by-severity/by-cli/issues queries, while repoFilterJoined uses a direct join. The subquery path runs for every such query; a JOIN on prs (like top_repos/timing already does) would be consistent and likely faster on large datasets. Minor; not required.
🟡 **daemon/internal/store/store.go:243** — The WHERE 1=1 scaffold is used purely to allow string-concat appending of the optional filter. Works, but a small helper that returns (clause, args) and omits the WHERE when empty would read more cleanly. Style nit.
🟡 **daemon/internal/store/store.go:240** — Errors from s.db.Query/QueryRow are silently discarded (pre-existing pattern, preserved here). With filtering now on the critical path, a malformed query or connection error will just return zeros to the client with no log. Consider at least slog.Error for query failures.
🟡 **flutter_app/lib/features/stats/stats_filters.dart:18** — statsFiltersProvider is a global StateProvider with no persistence — filters reset on app restart. Fine for now, but worth noting if parity with Activity filters (which may persist) is a goal.

### Suggestions

- Switch ref.read to ref.watch for prsProvider/issuesProvider in statsProvider so org-only filtering stays in sync with the underlying repo set.
- Extract a small buildRepoFilter(repos) helper in store.go that returns (joinedClause, subqueryClause, args) to avoid the WHERE 1=1 pattern and keep query construction tidy.
- Log query errors in ComputeStats — silently returning zero counts makes filter bugs invisible.
- Consider adding a unit test for ComputeStats with a non-empty repos slice to lock in the IN-clause behavior.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Solid feature adding repo/org filtering to stats. The SQL parameterization is correct (no injection risk), and the UI pattern is clean. However, there is a notable bug in the statsProvider where ref.read is used instead of ref.watch for prsProvider/issuesProvider, meaning the org-only filter path can silently return unfiltered stats if PR data loads after the stats provider runs.

### Issues

🔴 **flutter_app/lib/features/dashboard/dashboard_providers.dart:141** — ref.read(prsProvider) and ref.read(issuesProvider) should be ref.watch(). Using ref.read inside a provider body means statsProvider will NOT rebuild when prsProvider/issuesProvider complete their initial load. If a user selects an org filter before PRs finish loading, valueOrNull returns null → empty repo list → API receives no repos param → unfiltered global stats are returned, silently ignoring the org filter. Since the StatsScreen does ref.watch(prsProvider) for the filter bar, the UI will show an org chip as active while stats are unfiltered — a correctness bug.
⚠️ **daemon/internal/store/store.go:243** — The error return from db.QueryRow().Scan() is silently discarded for every query (e.g., TotalReviews, reviewsWithIssues, totalIssues). While this is pre-existing, the PR adds repoArgs to these calls, and a mismatched placeholder count would now silently produce zero values instead of surfacing an error. Consider at minimum checking the error on the first query to catch argument mismatches early during development.
🟡 **flutter_app/lib/features/stats/stats_filter_bar.dart:59** — _filteredRepos(filters) is called twice in the build method (once for the isNotEmpty check, once for the allItems list). Should be hoisted into a local variable to avoid redundant computation and set allocation.
🟡 **flutter_app/lib/core/api/api_client.dart:172** — Uri.encodeQueryComponent encodes the entire joined string including commas as %2C. While Go's net/url will decode this correctly before the handler splits on commas, this is fragile — a repo name containing a literal comma (unlikely but valid in some systems) would be indistinguishable from a separator. Consider using repeated query params (?repos=a&repos=b) or documenting the comma-separated contract explicitly.
🟡 **flutter_app/lib/features/stats/stats_filter_bar.dart:11** — allRepos is passed into the widget but _allOrgs is derived as a getter, meaning it's recomputed on every build call. For consistency and efficiency, consider computing allOrgs once in the parent or caching it.

### Suggestions

- Change ref.read(prsProvider) and ref.read(issuesProvider) in statsProvider to ref.watch() so the provider rebuilds when PR/issue data becomes available, fixing the org-only filter race condition.
- Add at least one integration test for ComputeStats(repos) in the daemon — verify that passing a repo list correctly scopes the returned counts vs. the unfiltered baseline.
- Consider adding a loading/empty state to StatsFilterBar when allRepos is empty, so users know the filter options are still loading rather than seeing no chips.

---

Muriano:
## Fixes applied (round 3)

| Issue | Fix |
|---|---|
| 🔴 ref.read → ref.watch | `prsProvider`/`issuesProvider` now watched in statsProvider — rebuilds when data loads, org-only filter works correctly |
| 🟡 Redundant _filteredRepos | Hoisted into local variables in build() |

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean implementation after three rounds of fixes. The core filtering logic is correct — parameterized SQL, proper arg passing, dialog-based multi-select with local state. A few minor issues remain around duplicated logic and a potential URI construction concern.

### Issues

🟡 **flutter_app/lib/features/stats/stats_screen.dart:17** — The allRepos derivation (PRs + issues mapped to repo, empty string removed) is duplicated between StatsScreen.build() and statsProvider in dashboard_providers.dart. If one changes without the other, the filter bar could show repos that the provider doesn't recognize, or vice versa. Extract this into a shared provider (e.g., `allKnownReposProvider`) that both consume.
🟡 **flutter_app/lib/core/api/api_client.dart:170** — Query string is manually appended to the path string before passing to `_uri()`. If `_uri` uses `Uri(path: ...)` or `Uri.https(host, path)` rather than `Uri.parse(baseUrl + path)`, the `?repos=...` portion will be percent-encoded as part of the path segment, not treated as a query parameter. Consider using `Uri.replace(queryParameters: ...)` or passing query params separately to `_uri` to be safe regardless of its implementation.
🟡 **flutter_app/lib/features/stats/stats_filter_bar.dart:13** — `_allOrgs` is a getter that recomputes on every access by iterating allRepos. Since `allRepos` is a constructor parameter (a new Set each parent rebuild), and `build()` calls this getter, the work is repeated every frame the parent rebuilds. Minor, but if allRepos grows, consider memoizing or computing once in build().

### Suggestions

- Extract the shared repo-set derivation into an `allKnownReposProvider` to eliminate duplication between StatsScreen and statsProvider.
- Consider adding a lightweight integration test for ComputeStats with a non-empty repos slice to verify the parameterized queries work end-to-end (WHERE IN with multiple values, empty result sets, etc.).

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Round-3 fixes look solid. SQL uses parameterized placeholders (injection-safe), the repoFilterJoined optimization avoids redundant subqueries where prs is already joined, and the Flutter side now correctly watches prsProvider/issuesProvider so org-only filtering rebuilds when data loads. Dialog-based multi-select fixes the earlier stale-checkbox bug. Only very minor stylistic observations remain.

### Issues

🟡 **flutter_app/lib/core/api/api_client.dart:170** — Manual query-string concatenation with Uri.encodeQueryComponent works but is fragile. Prefer Uri(path: '/stats', queryParameters: {'repos': repos.join(',')}) (or passing the existing _uri helper a queryParameters map) for consistent encoding and to avoid double-encoding risks if repo names ever contain reserved characters.
🟡 **daemon/internal/store/store.go:314** — In the avg-issues queries the table alias 'r' is introduced but columns are referenced unqualified (e.g. `issues != '[]'` rather than `r.issues`). Harmless today since there is no join, but inconsistent with the new qualified filter clauses — consider qualifying for readability.
🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:140** — ref.watch(prsProvider)/issuesProvider inside the `else if` branch makes subscriptions conditional on filter state. Riverpod handles this, but hoisting the watches above the branch would make the dependency graph stable and avoid re-registering providers each time filters change.
🟡 **flutter_app/lib/features/stats/stats_screen.dart:22** — allRepos is computed from valueOrNull-defaulted lists, so during initial PR/issue load the filter bar renders with no options. Acceptable, but a brief placeholder or suppressing the bar until data loads would be a nicer UX.

### Suggestions

- Consider adding a small unit test for ComputeStats(repos) covering: empty repos (global), single repo, multi repo, and a repo with zero reviews — the SQL rewrite is the riskiest part of this PR.
- If repo names could ever contain commas (they cannot for GitHub but for defensive robustness), switch the transport to repeated query params (?repos=a&repos=b) rather than a single comma-joined value.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Well-structured PR that adds server-side filtering for stats by org/repo. The SQL changes correctly parameterize repo lists and preserve existing aggregation semantics. Flutter side cleanly propagates filter state through Riverpod with proper reactivity on prsProvider/issuesProvider. Minor concerns around URL query encoding and an initial unfiltered request when supporting data is still loading, but no correctness-blocking bugs.

### Issues

🟡 **flutter_app/lib/core/api/api_client.dart:172** — Uri.encodeQueryComponent is applied to the whole joined string, which percent-encodes the commas (%2C) and slashes (%2F). It works because Go's url.Query() decodes percent-escapes before strings.Split finds the commas, but the request URL is noisier than needed and fragile if the server ever reads the raw query string. Prefer _uri('/stats').replace(queryParameters: {'repos': repos.join(',')}) or encode each repo individually before joining with ','.
🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:148** — When only orgs are selected, statsProvider depends on prsProvider/issuesProvider via valueOrNull. During their loading phase, repos resolves to [] and the daemon returns global (unfiltered) stats; once the lists load, statsProvider re-fires and shows the filtered numbers. The user briefly sees numbers that contradict the active filter. Consider gating the fetch (return a loading state) until prsProvider/issuesProvider have data when filters.orgs is non-empty, or resolving the effective repos once both are available.
🟡 **daemon/internal/store/store.go:240** — Query errors continue to be discarded (rows, _ := s.db.Query(...)). The PR follows the existing pattern so this is not a regression, but now that queries are parameterized and more likely to fail with an arg-mismatch during future changes, it would be worth logging via slog.Error like handleStats does — silent partial stats are hard to debug. Non-blocking suggestion.
🟡 **flutter_app/lib/features/stats/stats_screen.dart:19** — allRepos is recomputed on every StatsScreen rebuild from prs + issues. Cheap today, but if either list grows this becomes wasteful and is duplicated by the same derivation in dashboard_providers.dart. Consider extracting a derived allReposProvider (Provider<Set<String>>) reused by both the screen and statsProvider to keep the derivation in a single place.

### Suggestions

- Extract a shared allReposProvider that both StatsScreen and statsProvider consume so the org→repos derivation lives in one place.
- Gate statsProvider on prsProvider/issuesProvider data when filters.orgs is non-empty to avoid a transient global-stats flash.
- Consider building the /stats URL via Uri.replace(queryParameters: ...) so the repos param stays human-readable in logs.
- Add a unit test for store.ComputeStats([]string{"org/repo"}) covering at least BySeverity, TopRepos, and timing to lock in the parameterization contract.

---

PR #132, opened 2026-04-21
feat(config): add local_dir_base as list for multi-group workspace support

### LINKED ISSUE #130: feat: local_dir_base — auto-resolve repo paths without per-repo config
## Summary

Add a `local_dir_base` config option that automatically resolves `local_dir` for repos without requiring individual `[ai.repos."org/repo"]` entries. This enables `auto_implement` mode at scale without 18+ config entries.

## Motivation

Today, `auto_implement` requires a `local_dir` configured per repo:

```toml
[ai.repos."freepik-company/ai-bumblebee-proxy"]
local_dir = "/repos/ai-bumblebee-proxy"

[ai.repos."freepik-company/ai-platform-necromancer"]
local_dir = "/repos/ai-platform-necromancer"

# ... repeat 16 more times
```

This doesn't scale. Most repos live under the same base directory with a predictable naming convention.

## Proposed behavior

New config field in `[github]`:

```toml
[github]
local_dir_base = "/repos/ai-platform-workspace/workspace"
```

**Resolution priority:**
1. Per-repo override (`[ai.repos."org/repo"] local_dir = "/custom/path"`) — highest
2. `local_dir_base` + repo name (`{base}/{repo-name}`) — auto-resolved
3. Empty — no local_dir, auto_implement degrades to review_only

**How auto-resolution works:**
- Extract the repo name from `org/repo` (everything after `/`)
- Join with `local_dir_base`: `{base}/{repo-name}/`
- Verify the directory exists inside the container
- If it doesn't exist, log a warning and skip (don't crash)

## Example

```toml
[github]
local_dir_base = "/repos/ai-platform-workspace/workspace"

# Only needed for repos outside the base directory
[ai.repos."theburrowhub/heimdallm"]
local_dir = "/repos/heimdallr"
```

Docker mount:
```yaml
volumes:
  - /Users/stejon/develop:/repos
```

Result: `freepik-company/ai-bumblebee-proxy` resolves to `/repos/ai-platform-workspace/workspace/ai-bumblebee-proxy`. `theburrowhub/heimdallm` uses the override `/repos/heimdallr`.

## Environment variable

```
HEIMDALLM_LOCAL_DIR_BASE=/repos/ai-platform-workspace/workspace
```

### Config changes (`config.go`)

Add `LocalDirBase string` to `GitHubConfig`:

```go
type GitHubConfig struct {
    // ...existing fields...
    LocalDirBase string `toml:"local_dir_base"`
}
```

Add env var override in `applyEnvOverrides()`.

### Resolution logic

In the pipeline (or wherever `local_dir` is resolved for a repo), add auto-resolution:

```go
func resolveLocalDir(repo string, repoAI config.RepoAI, basePath string) string {
    // 1. Per-repo override
    if repoAI.LocalDir != "" {
        return repoAI.LocalDir
    }
    // 2. Base path auto-resolution
    if basePath != "" {
        parts := strings.SplitN(repo, "/", 2)
        if len(parts) == 2 {
            candidate := filepath.Join(basePath, parts[1])
            if info, err := os.Stat(candidate); err == nil && info.IsDir() {
                return candidate
            }
        }
    }
    // 3. No local_dir
    return ""
}
```

### Files to modify

- `daemon/internal/config/config.go` — add `LocalDirBase` field + env var
- `daemon/internal/config/config_test.go` — test for new field
- `daemon/cmd/heimdallm/main.go` or pipeline — use `resolveLocalDir` when building `RunOptions`
- `docker/.env.example` — document `HEIMDALLM_LOCAL_DIR_BASE`
- `docker/config.example.toml` — document `local_dir_base`

## Acceptance criteria

- [ ] `local_dir_base` auto-resolves repo paths from `{base}/{repo-name}`
- [ ] Per-repo `local_dir` override takes precedence
- [ ] Missing directories are warned, not errored
- [ ] Env var `HEIMDALLM_LOCAL_DIR_BASE` works
- [ ] Zero change for users without `local_dir_base` configured
- [ ] Tests for resolution priority

sergiotejon:
Closes #130

## Summary

`local_dir_base` is now a **list** of base directories. ResolveLocalDir checks each path in order, supporting multiple workspace groups without per-repo config.

## Example

```toml
[github]
local_dir_base = [
  "/repos/ai-platform-workspace/workspace",
  "/repos"
]
```

```env
HEIMDALLM_LOCAL_DIR_BASE=/repos/ai-platform-workspace/workspace,/repos
```

With this config:
- `freepik-company/ai-bumblebee-proxy` → `/repos/ai-platform-workspace/workspace/ai-bumblebee-proxy`
- `theburrowhub/heimdallm` → `/repos/heimdallr` (fallback to second base)
- Unknown repo → falls back to `/repos/{name}` default mount

## Resolution priority

1. Per-repo `local_dir` override (highest)
2. Each `local_dir_base` path/{repo-name} in order (first match wins)
3. `/repos/{repo-name}` (existing default mount)
4. Empty (diff-only mode)

## Changes

- `config.go` — `LocalDirBase` changed from `string` to `[]string`, env var parses comma-separated list, `ResolveLocalDir` iterates bases
- `main.go` — all 5 call sites updated to pass `[]string`
- `config_test.go` — 12 tests including `TestResolveLocalDir_MultipleBases`
- `.env.example` + `config.example.toml` — docs updated

## Test plan

- [ ] `go test ./internal/config/ -run ResolveLocalDir -v` — 12 tests pass
- [ ] Single base: same behavior as before
- [ ] Multiple bases: first match wins
- [ ] Empty bases: falls back to `/repos/{name}` default
- [ ] Env var with commas: parsed correctly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ docker/.env.local
 docker/.env.swp
 docker/.env.swo
 docker/config/
+daemon/heimdallm
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -840,11 +840,12 @@ func (a *tier2Adapter) ProcessPR(ctx context.Context, pr scheduler.Tier2PR) erro
 	a.cfgMu.Lock()
 	c := *a.cfg
 	aiCfg := c.AIForRepo(pr.Repo)
+	localDirBase := c.GitHub.LocalDirBase
 	a.cfgMu.Unlock()
 	// /repos/<short-name> fallback when local_dir is unset (stat-based,
 	// keep outside the mutex). Lets HEIMDALLM_REPOS_DIR give every
 	// monitored repo full-repo context without a per-repo override.
-	aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, pr.Repo, c.GitHub.LocalDirBase)
+	aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, pr.Repo, localDirBase)
 
 	ghPR := &gh.PullRequest{
 		ID:        pr.ID,
@@ -907,10 +908,11 @@ func (a *tier2Adapter) ProcessRepo(ctx context.Context, repo string) (int, error
 			aiCfg.Primary = c.AI.Primary
 		}
 		agentCfg := c.AgentConfigFor(aiCfg.Primary)
+		localDirBase := c.GitHub.LocalDirBase
 		a.cfgMu.Unlock()
 		// /repos/<short-name> fallback when local_dir is unset (stat-based,
 		// keep outside the mutex).
-		aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, issue.Repo, c.GitHub.LocalDirBase)
+		aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, issue.Repo, localDirBase)
 
 		extraFlags := agentCfg.ExtraFlags
 		if extraFlags != "" {
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -371,7 +371,7 @@ func main() {
 				return
 			}
 			seenRepo[repo] = true
-			if d := config.ResolveLocalDir("", repo); d != "" {
+			if d := config.ResolveLocalDir("", repo, c.GitHub.LocalDirBase); d != "" {
 				localDirsDetected[repo] = d
 			}
 		}
@@ -507,10 +507,11 @@ func main() {
 		}
 		cfgMu.Lock()
 		aiCfg := cfg.AIForRepo(pr.Repo)
+		localDirBase := cfg.GitHub.LocalDirBase
 		cfgMu.Unlock()
 		// /repos/<short-name> fallback when local_dir is unset (stat-based,
 		// keep outside the mutex).
-		aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, pr.Repo)
+		aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, pr.Repo, localDirBase)
 
 		// Construct github.PullRequest from stored data
 		ghPR := &gh.PullRequest{
@@ -576,10 +577,11 @@ func main() {
 			aiCfg.Primary = cfg.AI.Primary
 		}
 		agentCfg := cfg.AgentConfigFor(aiCfg.Primary)
+		localDirBase := cfg.GitHub.LocalDirBase
 		cfgMu.Unlock()
 		// /repos/<short-name> fallback when local_dir is unset (stat-based,
 		// keep outside the mutex).
-		aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, iss.Repo)
+		aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, iss.Repo, localDirBase)
 
 		// Reconstruct github.Issue from store data for the pipeline
 		ghIssue := &gh.Issue{
@@ -842,7 +844,7 @@ func (a *tier2Adapter) ProcessPR(ctx context.Context, pr scheduler.Tier2PR) erro
 	// /repos/<short-name> fallback when local_dir is unset (stat-based,
 	// keep outside the mutex). Lets HEIMDALLM_REPOS_DIR give every
 	// monitored repo full-repo context without a per-repo override.
-	aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, pr.Repo)
+	aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, pr.Repo, c.GitHub.LocalDirBase)
 
 	ghPR := &gh.PullRequest{
 		ID:        pr.ID,
@@ -908,7 +910,7 @@ func (a *tier2Adapter) ProcessRepo(ctx context.Context, repo string) (int, error
 		a.cfgMu.Unlock()
 		// /repos/<short-name> fallback when local_dir is unset (stat-based,
 		// keep outside the mutex).
-		aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, issue.Repo)
+		aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, issue.Repo, c.GitHub.LocalDirBase)
 
 		extraFlags := agentCfg.ExtraFlags
 		if extraFlags != "" {
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -69,6 +69,13 @@ type GitHubConfig struct {
 	// changes (label updates, new comments, merge/close). Defaults to "1m".
 	WatchInterval string `toml:"watch_interval"`
 
+	// LocalDirBase is a list of base directories for auto-resolving local_dir
+	// per repo. ResolveLocalDir checks each path in order, looking for
+	// {base}/{repo-name}, before falling back to /repos/{repo-name}.
+	// This supports multiple workspace groups (e.g. ai-platform in one dir,
+	// another team's repos in another). Put more specific paths first.
+	LocalDirBase []string `toml:"local_dir_base"`
+
 	// IssueTracking turns the issue-tracking pipeline (fase-2) on and off and
 	// governs how issues are filtered and classified. The pipeline itself
 	// lives in downstream issues (#25 onward); this struct is the
@@ -312,29 +319,42 @@ func ShortRepoName(repo string) string {
 // should run in for a given repo, using this precedence:
 //
 //  1. The explicit `local_dir` from config (the `configured` argument).
-//  2. `DefaultReposMountPath/<short-name>` when that directory exists —
+//  2. Each path in `localDirBases` checked in order — first match wins.
+//     Supports multiple workspace groups (e.g. ai-platform repos in one
+//     dir, another team in another) without per-repo local_dir entries.
+//  3. `DefaultReposMountPath/<short-name>` when that directory exists —
 //     lets an operator drop a single HEIMDALLM_REPOS_DIR into
 //     docker/.env and have every monitored repo picked up without also
 //     touching the per-repo override in the UI.
-//  3. Empty string — the agent runs in its default CWD (diff-only mode).
+//  4. Empty string — the agent runs in its default CWD (diff-only mode).
 //
 // Calls `os.Stat` on the candidate path, so callers should invoke it
 // outside any config-mutex critical section. The result is not cached;
 // re-invocation picks up newly-mounted repos on the next review cycle.
-func ResolveLocalDir(configured, repo string) string {
+func ResolveLocalDir(configured, repo string, localDirBases []string) string {
 	if configured != "" {
 		return configured
-	}
-	if DefaultReposMountPath == "" {
-		return ""
 	}
 	short := ShortRepoName(repo)
 	if short == "" {
 		return ""
 	}
-	candidate := filepath.Join(DefaultReposMountPath, short)
-	if info, err := os.Stat(candidate); err == nil && info.IsDir() {
-		return candidate
+	// 1. Check each local_dir_base in order (first match wins)
+	for _, base := range localDirBases {
+		if base == "" {
+			continue
+		}
+		candidate := filepath.Join(base, short)
+		if info, err := os.Stat(candidate); err == nil && info.IsDir() {
+			return candidate
+		}
+	}
+	// 2. Fallback to default mount path (/repos/{short-name})
+	if DefaultReposMountPath != "" {
+		candidate := filepath.Join(DefaultReposMountPath, short)
+		if info, err := os.Stat(candidate); err == nil && info.IsDir() {
+			return candidate
+		}
 	}
 	return ""
 }
@@ -522,6 +542,18 @@ func (c *Config) applyEnvOverrides() {
 	}
 	if v := os.Getenv("HEIMDALLM_WATCH_INTERVAL"); v != "" {
 		c.GitHub.WatchInterval = v
+	}
+	if v := os.Getenv("HEIMDALLM_LOCAL_DIR_BASE"); v != "" {
+		paths := strings.Split(v, ",")
+		cleaned := make([]string, 0, len(paths))
+		for _, p := range paths {
+			if s := strings.TrimSpace(p); s != "" {
+				cleaned = append(cleaned, s)
+			}
+		}
+		if len(cleaned) > 0 {
+			c.GitHub.LocalDirBase = cleaned
+		}
 	}
 	c.applyIssueTrackingEnv()
 	c.applyPRMetadataEnv()
--- a/daemon/internal/config/config_test.go
+++ b/daemon/internal/config/config_test.go
@@ -1102,7 +1102,7 @@ func TestResolveLocalDir_PrefersConfigured(t *testing.T) {
 	DefaultReposMountPath = tmp
 	t.Cleanup(func() { DefaultReposMountPath = old })
 
-	if got := ResolveLocalDir("/explicit/path", "org/name"); got != "/explicit/path" {
+	if got := ResolveLocalDir("/explicit/path", "org/name", nil); got != "/explicit/path" {
 		t.Errorf("got %q, want /explicit/path", got)
 	}
 }
@@ -1117,7 +1117,7 @@ func TestResolveLocalDir_AutoDetectFromMount(t *testing.T) {
 	DefaultReposMountPath = tmp
 	t.Cleanup(func() { DefaultReposMountPath = old })
 
-	if got := ResolveLocalDir("", "org/name"); got != repoDir {
+	if got := ResolveLocalDir("", "org/name", nil); got != repoDir {
 		t.Errorf("got %q, want %q", got, repoDir)
 	}
 }
@@ -1130,7 +1130,7 @@ func TestResolveLocalDir_NoFallbackWhenDirMissing(t *testing.T) {
 	DefaultReposMountPath = tmp
 	t.Cleanup(func() { DefaultReposMountPath = old })
 
-	if got := ResolveLocalDir("", "org/name"); got != "" {
+	if got := ResolveLocalDir("", "org/name", nil); got != "" {
 		t.Errorf("got %q, want empty", got)
 	}
 }
@@ -1145,7 +1145,7 @@ func TestResolveLocalDir_IgnoresFiles(t *testing.T) {
 	DefaultReposMountPath = tmp
 	t.Cleanup(func() { DefaultReposMountPath = old })
 
-	if got := ResolveLocalDir("", "org/name"); got != "" {
+	if got := ResolveLocalDir("", "org/name", nil); got != "" {
 		t.Errorf("got %q, want empty (file, not dir)", got)
 	}
 }
@@ -1155,7 +1155,7 @@ func TestResolveLocalDir_EmptyReposMountPath(t *testing.T) {
 	DefaultReposMountPath = ""
 	t.Cleanup(func() { DefaultReposMountPath = old })
 
-	if got := ResolveLocalDir("", "org/name"); got != "" {
+	if got := ResolveLocalDir("", "org/name", nil); got != "" {
 		t.Errorf("got %q, want empty (mount path disabled)", got)
 	}
 }
@@ -1169,8 +1169,102 @@ func TestResolveLocalDir_EmptyRepo(t *testing.T) {
 	DefaultReposMountPath = tmp
 	t.Cleanup(func() { DefaultReposMountPath = old })
 
-	if got := ResolveLocalDir("", ""); got != "" {
+	if got := ResolveLocalDir("", "", nil); got != "" {
 		t.Errorf("got %q, want empty", got)
+	}
+}
+
+// ── ResolveLocalDir with LocalDirBase ────────────────────────────────────────
+
+func TestResolveLocalDir_LocalDirBase(t *testing.T) {
+	// Create temp dirs simulating workspace
+	base := t.TempDir()
+	repoDir := filepath.Join(base, "my-repo")
+	os.MkdirAll(repoDir, 0755)
+
+	got := ResolveLocalDir("", "org/my-repo", []string{base})
+	if got != repoDir {
+		t.Errorf("ResolveLocalDir = %q, want %q", got, repoDir)
+	}
+}
+
+func TestResolveLocalDir_OverrideTakesPrecedence(t *testing.T) {
+	got := ResolveLocalDir("/custom/path", "org/repo", []string{"/some/base"})
+	if got != "/custom/path" {
+		t.Errorf("ResolveLocalDir = %q, want /custom/path", got)
+	}
+}
+
+func TestResolveLocalDir_BaseBeforeDefault(t *testing.T) {
+	base := t.TempDir()
+	defaultPath := t.TempDir()
+	repoDir := filepath.Join(base, "repo")
+	os.MkdirAll(repoDir, 0755)
+	defaultRepoDir := filepath.Join(defaultPath, "repo")
+	os.MkdirAll(defaultRepoDir, 0755)
+
+	old := DefaultReposMountPath
+	DefaultReposMountPath = defaultPath
+	defer func() { DefaultReposMountPath = old }()
+
+	got := ResolveLocalDir("", "org/repo", []string{base})
+	if got != repoDir {
+		t.Errorf("ResolveLocalDir = %q, want base path %q (not default)", got, repoDir)
+	}
+}
+
+func TestResolveLocalDir_FallbackToDefault(t *testing.T) {
+	defaultPath := t.TempDir()
+	repoDir := filepath.Join(defaultPath, "repo")
+	os.MkdirAll(repoDir, 0755)
+
+	old := DefaultReposMountPath
+	DefaultReposMountPath = defaultPath
+	defer func() { DefaultReposMountPath = old }()
+
+	got := ResolveLocalDir("", "org/repo", nil) // empty base
+	if got != repoDir {
+		t.Errorf("ResolveLocalDir = %q, want default %q", got, repoDir)
+	}
+}
+
+func TestApplyEnvOverrides_LocalDirBase(t *testing.T) {
+	cfg := &Config{}
+	cfg.applyDefaults()
+	t.Setenv("HEIMDALLM_LOCAL_DIR_BASE", "/workspace/group1, /workspace/group2")
+	cfg.applyEnvOverrides()
+	if len(cfg.GitHub.LocalDirBase) != 2 {
+		t.Fatalf("LocalDirBase = %v, want 2 items", cfg.GitHub.LocalDirBase)
+	}
+	if cfg.GitHub.LocalDirBase[0] != "/workspace/group1" {
+		t.Errorf("LocalDirBase[0] = %q, want /workspace/group1", cfg.GitHub.LocalDirBase[0])
+	}
+	if cfg.GitHub.LocalDirBase[1] != "/workspace/group2" {
+		t.Errorf("LocalDirBase[1] = %q, want /workspace/group2", cfg.GitHub.LocalDirBase[1])
+	}
+}
+
+func TestResolveLocalDir_MultipleBases(t *testing.T) {
+	group1 := t.TempDir()
+	group2 := t.TempDir()
+	// repo-a only in group1
+	os.MkdirAll(filepath.Join(group1, "repo-a"), 0755)
+	// repo-b only in group2
+	os.MkdirAll(filepath.Join(group2, "repo-b"), 0755)
+
+	bases := []string{group1, group2}
+
+	gotA := ResolveLocalDir("", "org/repo-a", bases)
+	if gotA != filepath.Join(group1, "repo-a") {
+		t.Errorf("repo-a = %q, want %q", gotA, filepath.Join(group1, "repo-a"))
+	}
+	gotB := ResolveLocalDir("", "org/repo-b", bases)
+	if gotB != filepath.Join(group2, "repo-b") {
+		t.Errorf("repo-b = %q, want %q", gotB, filepath.Join(group2, "repo-b"))
+	}
+	gotC := ResolveLocalDir("", "org/repo-c", bases)
+	if gotC != "" {
+		t.Errorf("repo-c = %q, want empty (not in any base)", gotC)
 	}
 }
 
--- a/daemon/internal/config/config_test.go
+++ b/daemon/internal/config/config_test.go
@@ -1222,7 +1222,9 @@ func TestResolveLocalDir_BaseBeforeDefault(t *testing.T) {
 func TestResolveLocalDir_FallbackToDefault(t *testing.T) {
 	defaultPath := t.TempDir()
 	repoDir := filepath.Join(defaultPath, "repo")
-	os.MkdirAll(repoDir, 0755)
+	if err := os.MkdirAll(repoDir, 0755); err != nil {
+		t.Fatal(err)
+	}
 
 	old := DefaultReposMountPath
 	DefaultReposMountPath = defaultPath
--- a/daemon/internal/config/config_test.go
+++ b/daemon/internal/config/config_test.go
@@ -1180,7 +1180,9 @@ func TestResolveLocalDir_LocalDirBase(t *testing.T) {
 	// Create temp dirs simulating workspace
 	base := t.TempDir()
 	repoDir := filepath.Join(base, "my-repo")
-	os.MkdirAll(repoDir, 0755)
+	if err := os.MkdirAll(repoDir, 0755); err != nil {
+		t.Fatal(err)
+	}
 
 	got := ResolveLocalDir("", "org/my-repo", []string{base})
 	if got != repoDir {
@@ -1199,9 +1201,13 @@ func TestResolveLocalDir_BaseBeforeDefault(t *testing.T) {
 	base := t.TempDir()
 	defaultPath := t.TempDir()
 	repoDir := filepath.Join(base, "repo")
-	os.MkdirAll(repoDir, 0755)
+	if err := os.MkdirAll(repoDir, 0755); err != nil {
+		t.Fatal(err)
+	}
 	defaultRepoDir := filepath.Join(defaultPath, "repo")
-	os.MkdirAll(defaultRepoDir, 0755)
+	if err := os.MkdirAll(defaultRepoDir, 0755); err != nil {
+		t.Fatal(err)
+	}
 
 	old := DefaultReposMountPath
 	DefaultReposMountPath = defaultPath
@@ -1248,9 +1254,13 @@ func TestResolveLocalDir_MultipleBases(t *testing.T) {
 	group1 := t.TempDir()
 	group2 := t.TempDir()
 	// repo-a only in group1
-	os.MkdirAll(filepath.Join(group1, "repo-a"), 0755)
+	if err := os.MkdirAll(filepath.Join(group1, "repo-a"), 0755); err != nil {
+		t.Fatal(err)
+	}
 	// repo-b only in group2
-	os.MkdirAll(filepath.Join(group2, "repo-b"), 0755)
+	if err := os.MkdirAll(filepath.Join(group2, "repo-b"), 0755); err != nil {
+		t.Fatal(err)
+	}
 
 	bases := []string{group1, group2}
 
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -32,6 +32,12 @@ HEIMDALLM_REPOSITORIES=
 # Defaults to 15m when HEIMDALLM_DISCOVERY_TOPIC is set.
 # HEIMDALLM_DISCOVERY_INTERVAL=15m
 
+# ── Local directory bases (optional) ────────────────────────────────────────
+# Comma-separated list of base directories for auto-resolving local_dir.
+# Each {base}/{repo-name} is checked in order. Supports multiple workspace groups.
+# Per-repo local_dir overrides this.
+# HEIMDALLM_LOCAL_DIR_BASE=/repos/ai-platform-workspace/workspace,/repos
+
 # ── Issue tracking pipeline (optional, fase-2) ──────────────────────────────
 # Enables the daemon to fetch, classify and process GitHub issues in the
 # monitored repos. Label classification precedence:
--- a/docker/config.example.toml
+++ b/docker/config.example.toml
@@ -19,6 +19,10 @@ repositories = ["org/repo1", "org/repo2"]
 # discovery_orgs     = ["your-org", "your-other-org"]   # required when discovery_topic is set
 # discovery_interval = "15m"                             # any Go duration; defaults to 15m
 
+# Base directories for auto-resolving local_dir per repo (list).
+# Each {base}/{repo-name} is checked in order. Supports multiple workspace groups.
+# local_dir_base = ["/repos/ai-platform-workspace/workspace", "/repos"]
+
 # Issue tracking pipeline (fase-2).
 # When enabled, the daemon fetches open issues from the monitored repos,
 # classifies them by label, and — depending on the mode — posts an AI


Muriano:
## 🔴 HIGH Issue

A compiled binary (daemon/heimdallm) has been committed to the repository. Binary artifacts should never be checked into version control — they bloat the repo permanently (even after deletion, they remain in git history), cannot be reviewed, and will conflict across platforms/architectures. Remove this file and add it to .gitignore.

**Location:** `daemon/heimdallm`

---

Muriano (changes requested):

Clean feature addition extending ResolveLocalDir to support multiple workspace base directories. Good test coverage for the new functionality. One high-severity issue: a compiled binary was accidentally committed. The rest of the changes are well-structured and follow existing patterns.

### Suggestions

- Add `daemon/heimdallm` (and similar build outputs like `heimdallm` or `*.exe`) to the project's .gitignore to prevent accidental future commits.
- Consider adding a test case for `HEIMDALLM_LOCAL_DIR_BASE` with trailing commas or empty segments (e.g., ",/path1,,/path2,") to confirm the TrimSpace+empty filtering handles all edge-case inputs — the current implementation looks correct but an explicit test documents the contract.
- The env var uses comma as a path separator, which technically conflicts with paths containing commas. This is unlikely in practice but worth noting in the config comment/docs as a known limitation, or consider supporting an alternative separator (e.g., colon on Unix, as PATH does).

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

Feature adds a multi-base list for resolving per-repo local_dir. Core logic is correct and well-tested. Main blocker is the committed binary (already flagged). Secondary concern: inconsistent mutex handling for reads of cfg.GitHub.LocalDirBase across call sites.

### Issues

🔴 **daemon/heimdallm:0** — Compiled binary committed to the repo. Must be removed and added to .gitignore — already flagged by @Muriano. Blocker.
⚠️ **daemon/cmd/heimdallm/main.go:844** — In tier2Adapter.ProcessPR, c.GitHub.LocalDirBase is read without the cfgMu lock, while the two refactored call sites at lines ~511 and ~581 now correctly capture localDirBase under cfgMu. Slice-header reads are not atomic and the config is hot-reloadable, so this is a data race when config is replaced. Capture the slice under the existing cfgMu critical section (same pattern as lines 507–511).
⚠️ **daemon/cmd/heimdallm/main.go:913** — In tier2Adapter.ProcessRepo the mutex is released at a.cfgMu.Unlock() just before ResolveLocalDir is called with c.GitHub.LocalDirBase. The slice header is read after unlocking. Move the LocalDirBase read above the Unlock() and pass the captured local copy into ResolveLocalDir.
🟡 **daemon/internal/config/config.go:340** — Minor behavioral change: previously ResolveLocalDir returned early when DefaultReposMountPath=="" before computing ShortRepoName. Now ShortRepoName is computed unconditionally. Harmless (cheap), but worth noting that the early-return semantics shifted — when localDirBases is empty and DefaultReposMountPath is empty, you now still compute short but still return "". Test coverage looks fine.
🟡 **daemon/internal/config/config_test.go:1181** — os.MkdirAll return values are ignored throughout the new tests. Wrap with if err := ...; err != nil { t.Fatal(err) } to avoid silent setup failures producing confusing test failures downstream.
🟡 **daemon/internal/config/config.go:547** — HEIMDALLM_LOCAL_DIR_BASE env parsing replaces the TOML-provided list entirely when non-empty. This matches other env overrides but is worth documenting in the .env.example comment to avoid operator surprise when both TOML and env are set.

### Suggestions

- Remove daemon/heimdallm from the PR and add /daemon/heimdallm (or daemon/heimdallm) to .gitignore. Consider git filter-repo if it landed in history of a shared branch.
- Unify the mutex pattern: always capture localDirBase := cfg.GitHub.LocalDirBase under cfgMu and pass the local copy to ResolveLocalDir. Apply at both tier2Adapter.ProcessPR and tier2Adapter.ProcessRepo to match the main loop call sites.
- Consider documenting in the config docstring that relative paths in local_dir_base are resolved relative to the daemon's CWD (or rejected), so operators don't assume relative-to-config-file resolution.
- Add a test for the env-override precedence (env wins over TOML) to lock in current behavior of HEIMDALLM_LOCAL_DIR_BASE replacing, not appending to, the TOML list.

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Adds multi-base directory support for auto-resolving local_dir per repo via a new LocalDirBase list in GitHubConfig, with env override, docs, and tests. Implementation is clean and backward-compatible, but a compiled binary was accidentally committed.

### Issues

🔴 **daemon/heimdallm:0** — Compiled binary committed to the repo. Binary artifacts should not be versioned — they bloat the repo permanently (persist in git history), cannot be reviewed, and conflict across platforms/architectures. Remove the file and add it (or a build-output pattern) to .gitignore.
🟡 **daemon/internal/config/config_test.go:1181** — Tests call os.MkdirAll without checking the returned error. If the mkdir silently fails (e.g., permission issue on CI), the subsequent assertions could produce misleading failures. Use t.Fatal on error or switch to require-style checks.
⚠️ **daemon/cmd/heimdallm/main.go:510** — LocalDirBase is read under cfgMu in two of the four call sites but passed directly as c.GitHub.LocalDirBase in the tier2Adapter paths (lines 847 and 913) without holding a.cfgMu. If config can be hot-reloaded concurrently, this is an inconsistent locking pattern — either the slice is safe to read lock-free (then remove the mutex reads) or it isn't (then the tier2 paths need locking too, ideally copying the slice header under the lock).

### Suggestions

- Remove daemon/heimdallm from the PR and add a .gitignore entry (e.g., /daemon/heimdallm or daemon/heimdallm) to prevent reintroduction.
- Unify locking around c.GitHub.LocalDirBase reads across all four call sites in main.go — either consistently copy under the cfg mutex or document that the slice is immutable post-load.
- Consider validating/normalizing LocalDirBase entries (e.g., trim whitespace, drop empties) inside applyDefaults so TOML-configured values benefit from the same cleanup the env override already does.
- Add a test covering the ordering guarantee when the same short-name repo exists under multiple bases (first match wins), since the doc comment explicitly promises this behavior.

---

Muriano:
## 🔴 HIGH Issue

Compiled binary committed to the repository. Binary artifacts must never be checked into version control — they permanently bloat git history, cannot be code-reviewed, and differ across platforms. Remove this file from the branch (git rm daemon/heimdallm) and add 'daemon/heimdallm' to .gitignore to prevent recurrence.

**Location:** `daemon/heimdallm`

---

Muriano (changes requested):

Clean feature addition extending ResolveLocalDir to support multiple base directories. Logic is correct, tests are thorough, and call sites properly read the slice under mutex before passing it. The only confirmed issue is the committed binary artifact (already flagged in prior review).

### Suggestions

- Consider adding 'daemon/heimdallm' (or a broader pattern like '/daemon/heimdallm') to .gitignore to prevent accidental re-commit of the binary.
- In the env-override parsing for HEIMDALLM_LOCAL_DIR_BASE, consider resolving paths to absolute (filepath.Abs) so relative paths in the env var don't silently depend on the daemon's CWD at startup.
- The new tests use os.MkdirAll without checking the error return — while unlikely to fail in t.TempDir(), checking errors is good practice (e.g., require.NoError or if err != nil { t.Fatal(err) }).

---

Muriano:
## 🔴 HIGH Issue

Compiled binary committed to the repository. Binary artifacts must never be checked into version control — they permanently bloat git history (even after removal, the blob persists in packfiles), cannot be code-reviewed, and are platform/architecture-specific. Fix: `git rm daemon/heimdallm` and add `daemon/heimdallm` to `.gitignore` (or the existing gitignore pattern for build outputs).

**Location:** `daemon/heimdallm`

---

Muriano (changes requested):

The PR correctly extends ResolveLocalDir to support multiple base directories via a new LocalDirBase config field and HEIMDALLM_LOCAL_DIR_BASE env var. Logic is sound, tests are comprehensive, and existing tests are properly updated. However, a compiled binary was accidentally committed, which is the only blocking issue.

### Suggestions

- Consider validating that paths in LocalDirBase are absolute (reject relative paths) to avoid surprising behavior when the daemon's working directory changes — e.g. in `applyEnvOverrides` or at config load time, warn or reject entries that don't start with '/'.
- The env var override (`HEIMDALLM_LOCAL_DIR_BASE`) fully replaces any TOML-configured `local_dir_base`. If additive behavior is ever needed (env extends TOML), document that the current semantics are replace-only so operators don't get surprised.
- Existing tests use `os.MkdirAll` without checking the error return — while unlikely to fail in `t.TempDir()`, idiomatic Go test style would check `if err := os.MkdirAll(...); err != nil { t.Fatal(err) }` for clarity.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

The new `local_dir_base` list feature is well-scoped, tested, and backwards-compatible. Main concerns: a compiled binary is included in the diff (already flagged by another reviewer and still unaddressed), and `c.GitHub.LocalDirBase` is read without `cfgMu` in three call sites, reintroducing a small race window the other two sites explicitly avoid. Previous findings were in files not touched by this diff (activity.go, promoter.go) and are out of scope for this re-review.

### Issues

🔴 **daemon/heimdallm:0** — A compiled binary is checked in. It bloats git history permanently (the blob survives even after `git rm`), is unreviewable, and is platform/arch-specific. Remove with `git rm daemon/heimdallm` and add `daemon/heimdallm` (or a broader build-output pattern) to `.gitignore`. Already flagged by prior review and still present in the diff.
⚠️ **daemon/cmd/heimdallm/main.go:374** — `c.GitHub.LocalDirBase` is read without holding `cfgMu`. The sibling sites at lines 507 and 577 correctly snapshot `localDirBase := cfg.GitHub.LocalDirBase` under the lock before calling `ResolveLocalDir`, so this path is inconsistent. If config reload replaces the slice header while this read is in flight, it is a data race under `-race`. Snapshot under the mutex like the other sites.
⚠️ **daemon/cmd/heimdallm/main.go:847** — `tier2Adapter.ProcessPR` reads `c.GitHub.LocalDirBase` without holding the adapter's `cfgMu`. Same race concern as above; snapshot the slice under the lock used to read `aiCfg` and pass the local copy to `ResolveLocalDir`.
⚠️ **daemon/cmd/heimdallm/main.go:913** — `tier2Adapter.ProcessRepo` reads `c.GitHub.LocalDirBase` after `a.cfgMu.Unlock()`. Same race concern — snapshot while still holding the mutex (alongside the `aiCfg` read) and reuse the local variable.
🟡 **daemon/internal/config/config.go:559** — `HEIMDALLM_LOCAL_DIR_BASE` env override only replaces `LocalDirBase` if at least one non-empty path is parsed. That means setting `HEIMDALLM_LOCAL_DIR_BASE=""` or `HEIMDALLM_LOCAL_DIR_BASE=","` silently keeps whatever was loaded from TOML, which is counter-intuitive for operators trying to clear a list at runtime. Consider documenting this behavior in the comment ("empty value is ignored; use an explicit sentinel or unset to clear") or treating a non-empty env value as authoritative even when all paths are blank.
🟡 **daemon/internal/config/config.go:345** — `ResolveLocalDir` does not validate that entries in `localDirBases` are absolute, nor does it reject a base path that ends up resolving to the same directory as `DefaultReposMountPath` (duplicate stat work). Not a correctness bug, but a small doc note that paths should be absolute would help prevent surprises when someone puts a relative workspace path in TOML.

### Suggestions

- Remove the `daemon/heimdallm` binary and add an ignore rule so it cannot be re-added accidentally.
- Centralize the snapshot pattern: read `aiCfg` and `localDirBase` together under `cfgMu` at every call site, or add a tiny helper like `cfg.snapshotForResolve(repo)` that returns `(aiCfg, localDirBase)` to avoid the footgun.
- Document in the TOML/env comments that paths should be absolute and that per-repo `local_dir` always wins — the current docs say this, but reinforcing it near `HEIMDALLM_LOCAL_DIR_BASE` would help.
- Consider a `TestApplyEnvOverrides_LocalDirBase_EmptyClears` case to lock in the current "empty env does not clear" behavior, whichever way you choose to define it.

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Re-review: the compiled binary daemon/heimdallm is still present in the diff and remains unaddressed despite prior HIGH feedback. The MEDIUM locking inconsistency for LocalDirBase access in tier2Adapter paths is also unchanged. The LOW test error-checking nit on os.MkdirAll persists. No new issues introduced by the diff beyond these.

### Issues

🔴 **daemon/heimdallm:0** — Compiled binary still committed to the repo. Must be removed (git rm) and added to .gitignore; binary blobs persist in git history even after deletion and cannot be reviewed.
⚠️ **daemon/cmd/heimdallm/main.go:847** — In tier2Adapter.ProcessPR and ProcessRepo, c.GitHub.LocalDirBase is read without holding a.cfgMu, while the two main.go call sites read LocalDirBase under cfgMu. Locking is inconsistent — either the slice is safe to read lock-free everywhere (drop the mutex reads) or the tier2 paths must also copy the slice header under a.cfgMu to be safe against hot-reload.
🟡 **daemon/internal/config/config_test.go:1181** — os.MkdirAll return value is still ignored in several new test cases (TestResolveLocalDir_LocalDirBase, _BaseBeforeDefault, _FallbackToDefault, _MultipleBases). Wrap with t.Fatal on error to avoid misleading downstream assertion failures if setup silently fails.

### Suggestions

- Remove daemon/heimdallm from the branch and add an ignore rule so it cannot be reintroduced.
- Normalize LocalDirBase access: copy the slice header under the appropriate mutex at all four call sites, or document that the field is immutable after load and drop the mutex reads in main.go for consistency.
- Check os.MkdirAll errors in the new tests with t.Fatalf to fail fast on setup problems.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Clean, well-tested change that extends ResolveLocalDir to support multiple workspace base directories while preserving backwards compatibility. Good test coverage (override precedence, base-before-default, fallback, multi-base, env parsing) and appropriate documentation updates in both config.example.toml and .env.example. Minor concerns: the .gitignore entry addresses prior binary-commit comments, but the diff does not show the binary being removed from the tree — if the blob is still tracked, the gitignore alone will not fix it. A couple of small test-quality and edge-case nits are listed below.

### Issues

⚠️ **daemon/heimdallm:0** — The .gitignore entry `daemon/heimdallm` is added, but the diff does not show a deletion of the tracked binary itself. If the file was committed in a prior commit on this branch and is still tracked, .gitignore has no effect on it — a `git rm daemon/heimdallm` is also required (the prior review comments from @Muriano explicitly call this out). Verify `git ls-files | grep daemon/heimdallm` returns nothing on this branch.
🟡 **daemon/internal/config/config_test.go:1230** — In TestResolveLocalDir_FallbackToDefault, `os.MkdirAll(repoDir, 0755)` is called without checking the returned error, while all the sibling tests in the same file use `if err := os.MkdirAll(...); err != nil { t.Fatal(err) }`. Minor inconsistency; a transient FS error would make the test silently mis-assert.
🟡 **daemon/internal/config/config.go:546** — HEIMDALLM_LOCAL_DIR_BASE is split on ',' with no escape mechanism. Paths containing a literal comma would be broken into pieces. Highly unlikely in practice for filesystem paths, but worth a one-line doc note in .env.example ('paths with commas are not supported') or using os.PathListSeparator (':' on unix) as a more conventional delimiter.
🟡 **daemon/internal/config/config.go:357** — Behaviour change worth calling out: previously, when DefaultReposMountPath was set but the candidate dir did not exist, ResolveLocalDir returned "" immediately. Now, if localDirBases is non-empty, each base is stat'd first and then the default mount is stat'd as a fallback — i.e. one extra os.Stat per resolve per configured base. Per-review-cycle cost is negligible, but since ResolveLocalDir is invoked for every PR and issue on the hot path, consider documenting this in the function comment so future readers don't get surprised by the extra syscalls with long base lists.

### Suggestions

- Confirm `daemon/heimdallm` is not tracked on this branch (git ls-files). If it is, add a `git rm --cached daemon/heimdallm` commit — the .gitignore entry alone will not untrack an already-committed file.
- Consider adding a test for the case where `localDirBases` contains an entry whose {base}/{short} does not exist but DefaultReposMountPath does — to lock in the documented fallthrough order.
- Consider adding a validation warning at config-load time if any path in LocalDirBase does not exist, to help operators catch typos early (non-fatal, since dirs may be mounted later).
- Minor: in main.go the pattern `localDirBase := c.GitHub.LocalDirBase; cfgMu.Unlock(); ... ResolveLocalDir(..., localDirBase)` copies the slice header under the mutex, which is correct as long as the config reload path always replaces the slice rather than mutating it in place. Worth a one-line comment at the GitHubConfig struct reminding future editors not to mutate LocalDirBase in place after publish.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured change that extends ResolveLocalDir to accept a list of base directories for multi-group workspace support. The signature change is consistently applied across all call sites, mutex discipline is correct (value copied under lock, stat calls outside), and test coverage is thorough. No confirmed bugs or security issues found.

### Suggestions

- The env var HEIMDALLM_LOCAL_DIR_BASE uses comma as separator — consider documenting that paths containing commas are not supported, or using a different separator (e.g. ':' on Unix, matching PATH conventions).
- The three duplicate comment blocks in main.go ('// /repos/<short-name> fallback when local_dir is unset') could benefit from a single-line reference to ResolveLocalDir's godoc instead of repeating the explanation at every call site.
- TestResolveLocalDir_FallbackToDefault silently ignores the os.MkdirAll error — add a t.Fatal(err) check for consistency with the other tests in this file.

---

Muriano:
## 🔴 HIGH Issue

Compiled binary committed to the repository (flagged 3× in existing PR discussion). The .gitignore entry prevents future accidents but does not remove the blob already tracked. Run `git rm daemon/heimdallm` on this branch; the blob will persist in packfiles permanently otherwise. Severity: bloats repo history, un-reviewable, platform-specific.

**Location:** `daemon/heimdallm`

---

Muriano (changes requested):

Clean, well-structured feature addition that extends ResolveLocalDir to support multiple workspace base directories. The signature change is propagated consistently to all call sites with correct mutex discipline. Test coverage is thorough. The only confirmed issue is the committed binary artifact already flagged by prior reviewers.

### Suggestions

- Add a test case for HEIMDALLM_LOCAL_DIR_BASE with edge-case values like trailing commas or whitespace-only entries (e.g. ', , /valid') to verify the cleaning logic handles all degenerate inputs.
- Consider documenting the precedence order (explicit local_dir > local_dir_base > DefaultReposMountPath > empty) in the TOML example comment so operators understand resolution without reading source code.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid, well-scoped refactor that generalizes `ResolveLocalDir` to support multiple workspace base directories. Signature change is applied consistently at all call sites, mutex handling is correct (base list captured inside the lock), and test coverage for the new behavior is thorough. Docs and example configs are updated. Previous HIGH concern about a committed binary is addressed via `.gitignore` (assuming the blob was removed in this branch — not visible in diff).

### Issues

🟡 **daemon/internal/config/config.go:322** — Doc comment numbering drift: the godoc lists steps 1–4 (configured → localDirBases → DefaultReposMountPath → empty), but the inline comments inside the function body label the same branches as `1.` and `2.`. Minor readability nit — align the inline numbering with the godoc (2./3.) to avoid future confusion.
🟡 **daemon/internal/config/config.go:546** — `HEIMDALLM_LOCAL_DIR_BASE` only overrides the TOML value when `len(cleaned) > 0`. That means setting the env var to an all-whitespace/empty value (e.g. `HEIMDALLM_LOCAL_DIR_BASE=" "`) silently leaves the TOML value intact, whereas other env overrides in this file unconditionally replace when `v != ""`. Consider whether an explicit (but empty) env override should clear the TOML list, or document the current 'non-empty only' semantics.
🟡 **daemon/cmd/heimdallm/main.go:843** — Inside `tier2Adapter.ProcessPR`, the code already does `c := *a.cfg` under the mutex (shallow struct copy), so the subsequent `localDirBase := c.GitHub.LocalDirBase` is a redundant extra copy of the slice header. Not a bug — just noise; you can pass `c.GitHub.LocalDirBase` directly to `ResolveLocalDir`. For consistency with the other call sites (which don't take a struct copy) the extra variable is fine as-is.
🟡 **daemon/internal/config/config_test.go:1224** — `TestResolveLocalDir_FallbackToDefault` calls `os.MkdirAll(repoDir, 0755)` without checking the error, unlike the neighboring tests that do `if err := os.MkdirAll(...); err != nil { t.Fatal(err) }`. Minor consistency fix — a silent MkdirAll failure here would produce a confusing 'want default' error instead of a clear setup failure.
🟡 **daemon/internal/config/config.go:338** — Subtle ordering change: previously, when `configured == ""` and `DefaultReposMountPath == ""`, the function returned early before computing `ShortRepoName`. Now `ShortRepoName` is always computed (and the short=="" early-return now fires before the base/default checks). Behavior is equivalent, but worth noting that any future change to `ShortRepoName` side effects would now run on every resolve call. Not an issue today — flagging for awareness.

### Suggestions

- Consider adding a test that exercises the env-var parser with edge inputs like `","`, `" , "`, and a single trailing comma, to lock in the current 'skip empty / don't override when all empty' semantics.
- Add a short note in the godoc for `LocalDirBase` (or `ResolveLocalDir`) clarifying that non-existent base paths are silently skipped — operators setting a typo'd path will otherwise have no signal beyond 'fallback kicked in'.
- Optional: emit a one-time INFO log on startup listing which `local_dir_base` entries exist vs. are missing, which would help ops debug multi-workspace misconfigurations without needing to trigger a review cycle.
- Verify with `git log --stat` that `daemon/heimdallm` was actually removed from the branch (not just added to `.gitignore`). The diff only shows the `.gitignore` addition; if the binary blob is still in a prior commit on the branch, the previous HIGH comment technically isn't resolved until the history is rewritten or the file is removed in a new commit.

---

Muriano:
## 🔴 HIGH Issue

Compiled binary is still tracked in the repository. The .gitignore entry prevents future accidents but does NOT remove the already-tracked blob. A `git rm daemon/heimdallm` is required on this branch before merge, otherwise the binary remains in the tree and in packfile history permanently. This has been flagged 4 times in the PR discussion and remains unresolved.

**Location:** `daemon/heimdallm`

---

Muriano (changes requested):

Clean, well-structured change that extends ResolveLocalDir to accept a list of base directories for multi-group workspace support. The implementation is correct, thread-safe (localDirBase extracted under mutex before use), and well-tested. The only blocking issue is the already-flagged committed binary that still needs `git rm`.

### Suggestions

- Consider adding a test for symlink entries in localDirBases (e.g., a symlink base dir pointing to a real directory) to verify os.Stat follows symlinks as expected.
- The HEIMDALLM_LOCAL_DIR_BASE env var uses comma-separated values. If any workspace path legitimately contains a comma (rare but possible on some OS), consider documenting this limitation in the .env.example comment.
- Minor: in TestResolveLocalDir_FallbackToDefault, the os.MkdirAll error is not checked (unlike other tests in the same file that use t.Fatal on mkdir errors).

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Clean, well-tested extension of ResolveLocalDir to support multiple workspace base directories via a list config with env override. Code changes are correct and backwards-compatible. Main concern is the committed binary flagged 4× in prior discussion — the .gitignore entry added here prevents recurrence but does not remove the already-tracked blob.

### Issues

🔴 **daemon/heimdallm:0** — Compiled binary appears to still be tracked in the branch. The .gitignore entry prevents future accidents but does not untrack an already-committed file. Run `git rm daemon/heimdallm` and recommit; otherwise the blob persists in packfiles permanently and bloats history.
🟡 **daemon/cmd/heimdallm/main.go:510** — `localDirBase := cfg.GitHub.LocalDirBase` captures a slice header under cfgMu but the backing array is shared. If config reload replaces GitHub.LocalDirBase with a new slice this is safe; if it ever mutates the existing slice in place, there is a latent race. Worth confirming reload semantics or defensively copying.
🟡 **daemon/internal/config/config_test.go:1217** — `os.MkdirAll(repoDir, 0755)` return value is ignored in TestResolveLocalDir_FallbackToDefault, inconsistent with neighboring tests that check the error. Minor style nit.

### Suggestions

- Remove the tracked daemon/heimdallm binary with `git rm` in this branch — the .gitignore addition alone is insufficient.
- Consider documenting the ordering semantics (first-match-wins, explicit local_dir still overrides everything) in the TOML example comment, not just in the Go doc comment.
- Optional: deduplicate the stat+IsDir candidate check into a small helper since it is now repeated for the bases loop and the default fallback.

---

Muriano:
## 🔴 HIGH Issue

Compiled binary `daemon/heimdallm` is still tracked in git. The `.gitignore` addition only prevents future `git add` — it does NOT untrack an already-committed file. A `git rm daemon/heimdallm` (and a commit) is required on this branch before merge, otherwise the binary blob remains in the tree permanently. This has been flagged 5 times in the PR discussion and remains unresolved.

**Location:** `daemon/heimdallm`

---

Muriano (changes requested):

Clean, well-structured change that extends ResolveLocalDir to support multiple workspace base directories. The implementation is correct: slice is read under mutex, used outside it safely (Go slice headers are copied by value, and config reloads create new slices). Tests cover happy path, precedence, fallback, multi-base, and env override. The only confirmed issue is the tracked binary flagged 5 times in the PR discussion — the .gitignore entry prevents future accidents but does not remove the already-tracked blob.

### Suggestions

- Consider validating that local_dir_base paths are absolute in applyEnvOverrides or at config-load time — relative paths would resolve against the daemon's CWD, which may vary between bare-metal and Docker deployments, leading to subtle failures.
- The env var HEIMDALLM_LOCAL_DIR_BASE fully replaces whatever is in the TOML file (rather than appending). This is consistent with the other env overrides in this codebase but worth documenting explicitly in the .env.example comment, since users may expect additive behavior for a list field.

---

Muriano:
## 🔴 HIGH Issue

Compiled binary `daemon/heimdallm` is still tracked in git. The `.gitignore` entry added in this PR only prevents future `git add` — it does NOT untrack an already-committed file. A `git rm daemon/heimdallm` commit is required on this branch before merge, otherwise the binary blob remains in the tree and git history permanently. This bloats the repo, cannot be code-reviewed, and is platform-specific. Fix: run `git rm daemon/heimdallm && git commit -m 'chore: remove tracked binary'` on this branch.

**Location:** `daemon/heimdallm`

---

Muriano (changes requested):

Clean, well-structured feature adding multi-group workspace support via local_dir_base list. Code logic is correct, thread-safety is properly handled (LocalDirBase copied outside mutex), tests are thorough, and documentation is adequate. The only blocking issue — a committed binary — has been flagged 6 times in the PR discussion and remains unresolved.

### Suggestions

- Consider adding a Makefile target or pre-commit hook that prevents binaries under daemon/ from being committed, to catch this class of mistake automatically.
- The env var HEIMDALLM_LOCAL_DIR_BASE uses comma as separator — document that paths containing commas are not supported, or consider using a different separator (e.g. colon on Unix, matching PATH convention).
- The 4 call sites in main.go that extract `localDirBase` from config under the mutex and pass it to ResolveLocalDir follow an identical pattern — a small helper method on Config (e.g. `ResolveLocalDirForRepo(localDir, repo string) string`) could reduce repetition, though this is purely cosmetic.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

The LocalDirBase multi-workspace feature is well-implemented: clean API change, correct precedence (configured → bases → default mount), comprehensive tests, and proper env var parsing. The main outstanding concern is the repeatedly-flagged compiled binary `daemon/heimdallm` that remains tracked in the branch despite the .gitignore addition — .gitignore does not untrack already-committed files.

### Issues

🔴 **daemon/heimdallm:0** — Compiled binary still tracked in git (flagged 6× in discussion and unresolved). The new .gitignore entry only prevents future `git add`, it does not untrack the already-committed blob. Requires `git rm daemon/heimdallm` + commit on this branch before merge; otherwise the binary remains in the tree and permanently in packfile history. Binaries are un-reviewable, platform-specific, and bloat history.
🟡 **daemon/internal/config/config.go:548** — `HEIMDALLM_LOCAL_DIR_BASE` env override fully replaces the TOML `local_dir_base` list rather than merging. This is consistent with other env overrides in this file, but worth documenting in .env.example/config.example.toml so operators don't expect additive behavior when mixing TOML + env.
🟡 **daemon/internal/config/config.go:357** — No validation that entries in `localDirBases` are absolute paths. A relative path like `repos` will be joined with the repo short-name and stat'd relative to the daemon's CWD, which is likely a misconfiguration. Consider skipping non-absolute entries (or logging a warning) to fail loudly rather than silently matching CWD-relative dirs.
🟡 **daemon/internal/config/config_test.go:1264** — `TestResolveLocalDir_MultipleBases` covers first/second base match and no-match, but not the case where the same short-name exists in multiple bases — to assert 'first match wins' as documented. Add a test where `repo-x` exists in both group1 and group2 and verify group1 is returned.

### Suggestions

- Run `git rm daemon/heimdallm && git commit` on this branch to actually untrack the binary; the .gitignore entry alone is insufficient.
- Document that HEIMDALLM_LOCAL_DIR_BASE replaces (not augments) the TOML list in both .env.example and config.example.toml.
- Consider adding a debug log line when ResolveLocalDir picks a base, to aid multi-workspace troubleshooting when a repo resolves to an unexpected path.
- Add a first-match-wins test case to lock in the documented precedence behavior.

---

Muriano:
## 🔴 HIGH Issue

Compiled binary `daemon/heimdallm` is still tracked in git. The `.gitignore` addition in this PR only prevents future `git add` — it does NOT untrack an already-committed file. This permanently bloats the repo history with an un-reviewable, platform-specific blob. Fix: `git rm daemon/heimdallm && git commit -m 'chore: remove tracked binary'` on this branch before merge.

**Location:** `daemon/heimdallm`

---

Muriano (changes requested):

Clean, well-tested feature adding multi-base workspace resolution. The code logic, tests, and documentation are solid. One confirmed high-severity issue remains: the compiled binary daemon/heimdallm is still tracked in git (flagged 7 times in PR discussion). The .gitignore entry only prevents future additions — a `git rm` is required before merge.

### Suggestions

- Consider documenting the precedence order (explicit local_dir > local_dir_base > DefaultReposMountPath > empty) in the TOML example config comment, so operators understand resolution without reading source code.
- The HEIMDALLM_LOCAL_DIR_BASE env var uses comma as separator, which technically conflicts with paths containing commas (rare but possible on some filesystems). A note in .env.example about this limitation would be helpful.

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Adds list-based local_dir_base config with env override support and multi-workspace repo resolution. Core change is well-designed: clean backwards compatibility via nil slice in tests, sensible precedence (configured → bases → default mount → empty), and good test coverage. Main concern is an already-flagged tracked binary that the .gitignore entry alone does not untrack.

### Issues

🔴 **daemon/heimdallm:0** — Compiled binary is still tracked in git. The .gitignore entry only prevents future `git add` — it does not untrack an already-committed file. Requires `git rm daemon/heimdallm` on this branch before merge, otherwise the blob remains in the tree and packfile history permanently. Flagged repeatedly in prior discussion and still unresolved.
🟡 **daemon/internal/config/config.go:556** — HEIMDALLM_LOCAL_DIR_BASE env override only applies when at least one non-empty path is present (`len(cleaned) > 0`). This means setting `HEIMDALLM_LOCAL_DIR_BASE=` (empty) or `HEIMDALLM_LOCAL_DIR_BASE=,,,` cannot clear a TOML-configured list — inconsistent with other env overrides that unconditionally assign. Consider applying the cleaned slice unconditionally when the env var is set.
🟡 **daemon/internal/config/config_test.go:1222** — TestResolveLocalDir_MultipleBases does not reset DefaultReposMountPath. If a prior test in the package leaves it pointing at a tempdir that happens to contain `repo-c`, the `gotC` assertion could flake. Explicitly set `DefaultReposMountPath = ""` with cleanup for isolation.
🟡 **daemon/cmd/heimdallm/main.go:510** — Minor duplication: `localDirBase := cfg.GitHub.LocalDirBase` (or `c.GitHub.LocalDirBase`) is copied under the mutex in four sites. Since it's a slice header (cheap) this is fine, but a small helper like `cfg.ResolveLocalDirFor(repo, aiCfg.LocalDir)` could centralize the pattern and reduce the chance of a future caller forgetting to pass the bases.

### Suggestions

- Run `git rm daemon/heimdallm` on this branch and commit before merge — the .gitignore alone does not untrack the existing blob.
- Consider adding a Makefile `build` target that outputs to `daemon/heimdallm` plus a `bin/` convention so developers have a documented, ignored build path.
- Document in the TOML and .env examples that paths in local_dir_base are checked in order and first-match wins — the config.go godoc mentions it but the user-facing examples only say 'in order'.
- Consider validating/normalizing local_dir_base entries at config load (e.g., warn on relative paths or non-existent bases) so misconfiguration surfaces early rather than silently falling through to the default.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

The refactor of ResolveLocalDir to accept a list of base directories is clean, well-tested, and correctly threaded through all callers. Env var parsing and TOML binding are sound. The critical outstanding issue is unrelated to the code logic: the compiled binary daemon/heimdallm remains tracked in git — adding it to .gitignore does not untrack it. This has been flagged 8 times in PR discussion and must be resolved via git rm before merge.

### Issues

🔴 **daemon/heimdallm:0** — Compiled binary is still tracked in the repository. The .gitignore entry added here only prevents future `git add`; it does NOT untrack the already-committed blob. Binary will permanently live in git history/packfiles, bloats the repo, is un-reviewable, and is platform/architecture-specific. Run `git rm daemon/heimdallm && git commit` on this branch before merging.
🟡 **daemon/internal/config/config.go:543** — HEIMDALLM_LOCAL_DIR_BASE env override unconditionally replaces GitHub.LocalDirBase when the env var is set and yields at least one cleaned path. This is consistent with other env overrides in this file, but worth confirming the intent: a user who sets the env var cannot augment a TOML-defined list — only replace it. Consider documenting this in the comment next to the struct field or in .env.example so operators aren't surprised.
🟡 **daemon/internal/config/config.go:357** — Per-candidate os.Stat in a loop is fine for small base lists, but there is no cap or de-duplication. If an operator accidentally lists the same path twice or configures a very long list, the hot review path will stat each entry every invocation (no caching, per the existing contract). Not a blocker — matches the existing docstring 'result is not cached' — but consider de-duplicating empty/duplicate entries once at config load.
⚠️ **daemon/cmd/heimdallm/main.go:510** — localDirBase is captured as a slice reference under cfgMu and then read outside the mutex. This is safe only if config reloads replace the Config struct (and its slices) atomically rather than mutating LocalDirBase in place. Verify the config-reload path does not append/truncate the existing slice; if it does, this introduces a data race. The pattern is repeated in 4 call sites so any fix must be consistent.

### Suggestions

- Run `git rm daemon/heimdallm && git commit -m 'chore: remove tracked binary'` on this branch to actually untrack the binary flagged repeatedly in the PR discussion.
- Confirm (and ideally add a short comment) that config reload replaces GitHub.LocalDirBase atomically, so the capture-under-mutex pattern in main.go/tier2Adapter is race-free.
- Consider normalizing LocalDirBase once in applyDefaults/applyEnvOverrides (trim, drop empties, dedupe) so ResolveLocalDir's hot path doesn't re-check `if base == ""` on every call.
- Optional: add a test that asserts per-repo `local_dir` override still wins when LocalDirBase is also set (currently covered implicitly by TestResolveLocalDir_OverrideTakesPrecedence — good, keep it).

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured feature addition. Extends ResolveLocalDir to accept a list of base directories, enabling multi-group workspace support. The function signature change is propagated consistently across all 5 call sites, env var override parsing is correct, and test coverage is thorough (7 new test cases covering precedence, fallback, multi-base, and env parsing). The .gitignore entry for daemon/heimdallm is preventive — the binary is not actually tracked in git despite 8 automated comments claiming otherwise. No confirmed issues found.

### Suggestions

- The doc comment on ResolveLocalDir still shows a 4-step numbered list (1-4) but the inline code comments use 1-2. Consider aligning the numbering for consistency.
- Consider adding a test case where localDirBases contains empty strings to explicitly verify the `continue` guard in the loop (line 335 of config.go).
- The HEIMDALLM_LOCAL_DIR_BASE env var silently replaces the TOML-configured list rather than appending to it. This is fine for 'env overrides config' semantics, but documenting this behavior in the .env.example would help operators avoid confusion.

---

Muriano:
## 🔴 HIGH Issue

Compiled binary `daemon/heimdallm` is still tracked in git. The `.gitignore` entry added in this PR only prevents future `git add` — it does NOT untrack an already-committed file. Run `git rm daemon/heimdallm && git commit` on this branch before merge. This permanently bloats repo history with an un-reviewable, platform-specific blob.

**Location:** `daemon/heimdallm`

---

Muriano (changes requested):

Clean, well-tested feature adding multi-group workspace support via local_dir_base list. The ResolveLocalDir refactor is correct, callers are properly updated, and new tests cover precedence, multiple bases, fallback, and env override. The only blocking issue is the compiled binary still tracked in git — flagged 8 times in PR discussion and still unresolved.

### Suggestions

- At line 371 in main.go, `c.GitHub.LocalDirBase` is passed directly to ResolveLocalDir without the explicit mutex-copy pattern used at the other four call sites (lines 510, 580, 843, 912). For consistency and defensive thread-safety, consider copying LocalDirBase under the mutex the same way: `localDirBase := c.GitHub.LocalDirBase` then pass the local variable.
- The comma separator in HEIMDALLM_LOCAL_DIR_BASE technically prevents paths containing commas. This is extremely rare on Unix but could be documented in the .env.example comment (e.g., 'paths must not contain commas').

---

sergiotejon:
## Nota sobre el binario `daemon/heimdallm`

El binario fue eliminado del historial de la rama con `git filter-branch`. Los commits actuales (`7c008b7`, `86e567b`, `ae11446`) nunca contienen el fichero — se puede verificar con:

```bash
git show ae11446:daemon/heimdallm
# fatal: path 'daemon/heimdallm' exists on disk, but not in 'ae11446'
```

Las auto-reviews de Heimdallm siguen detectandolo como falso positivo porque ven el diff acumulado del PR, no los commits individuales reescritos.

El `.gitignore` ya incluye `daemon/heimdallm` para prevenir futuros accidentes.

@Muriano @ivanmunozruiz @vbuenog — por favor ignorad la alerta del binario en las auto-reviews y revisad los cambios reales (config `local_dir_base` como lista, mutex fixes, tests).

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured PR that extends ResolveLocalDir to support multiple workspace base directories. The function signature change is threaded correctly through all call sites, mutex patterns are followed consistently, and test coverage is thorough (happy path, precedence, fallback, env parsing, multiple bases). No blocking issues found.

### Suggestions

- Consider documenting in the .env.example that paths containing commas are not supported, since HEIMDALLM_LOCAL_DIR_BASE uses comma as the delimiter.
- The env override for HEIMDALLM_LOCAL_DIR_BASE fully replaces any TOML-configured local_dir_base rather than merging/appending. This is consistent with other env overrides in the codebase but worth noting in docs if users combine both config sources.
- Minor: at line 371 in main.go, c.GitHub.LocalDirBase is accessed directly (not extracted into a local var like the other call sites). If c is already a value copy this is fine, but for consistency with the pattern used at lines 510, 580, 843, and 911, consider extracting it the same way.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Straightforward, well-scoped change: `local_dir_base` becomes a list to support multiple workspace groups. Signature change to `ResolveLocalDir` is threaded through all call sites, mutex discipline is preserved (slice header captured under lock, stat performed outside), env override parsing is defensive (trim + skip empties), tests cover the new multi-base, precedence, and env paths, and docs are updated. No correctness or concurrency issues found in the diff. The `daemon/heimdallm` binary flagged repeatedly by auto-review is not present in the diff and the author has clarified it was removed via filter-branch.

### Issues

🟡 **daemon/internal/config/config.go:332** — Minor behavior change: previously, when DefaultReposMountPath was empty the function short-circuited before computing ShortRepoName. Now ShortRepoName(repo) is always computed even when both localDirBases and DefaultReposMountPath are empty. Functionally equivalent (wasted work is negligible), but worth noting if ShortRepoName ever becomes non-trivial.
🟡 **daemon/internal/config/config.go:546** — HEIMDALLM_LOCAL_DIR_BASE replaces GitHub.LocalDirBase entirely when set (as documented by env-override semantics elsewhere), but the TOML comment says 'Put more specific paths first' while there is no test verifying ordering when two bases both contain the same repo short-name (first-match-wins guarantee). A test with overlapping bases would lock the documented precedence.
🟡 **daemon/internal/config/config.go:72** — The doc comment says ResolveLocalDir checks each path 'before falling back to /repos/{repo-name}', but /repos is only the default when DefaultReposMountPath is set to that value at build/runtime. Minor wording inaccuracy — the fallback is DefaultReposMountPath, which happens to default to /repos. Not incorrect but could mislead readers who override it.

### Suggestions

- Add a test case where two bases both contain the same short-name to lock in the documented 'first match wins' ordering.
- Consider validating that entries in LocalDirBase are absolute paths (or at least warning on relative ones) at config-load time, since filepath.Join with a relative base silently produces surprising CWD-relative candidates.
- Optional: deduplicate DefaultReposMountPath from the fallback path when it's already listed in localDirBases, to avoid a redundant os.Stat.
- The auto-review binary alerts should stop once the PR is merged/rebased; if they persist, consider adding `daemon/heimdallm` to a review-ignore list for auto-review to reduce noise.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Clean, well-tested change. Extends ResolveLocalDir to support multiple workspace base directories via a new GitHub.LocalDirBase []string field, with env override (HEIMDALLM_LOCAL_DIR_BASE) and proper plumbing through all call sites in main.go (including mutex-guarded reads). Logic is straightforward: explicit local_dir > first matching base > DefaultReposMountPath > empty. The binary flag in existing discussion is a confirmed false positive (rewritten history via git filter-branch; .gitignore prevents recurrence). No correctness, concurrency, or style issues of note.

### Suggestions

- Consider documenting in the TOML/env examples that the order matters (first match wins) — the code comment says it, but the .env.example just says 'checked in order' without the 'first match wins' semantic.
- Minor: the GitHubConfig doc comment says 'Put more specific paths first', which is accurate but could be clearer as 'first match wins — list preferred workspace groups first'.
- Optional: in applyEnvOverrides, consider also accepting ':' as a separator for POSIX PATH-style familiarity, though comma is fine and consistent with other list envs.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Adds `local_dir_base` as a list of base directories to `GitHubConfig` for multi-workspace support, threading it through `ResolveLocalDir` call sites with proper mutex-safe reads. Includes env override parsing, updated tests, and config/docs examples. Logic is sound and the concurrency handling is correct. The binary-in-repo concern raised repeatedly by the bot was addressed by the author via history rewrite and is a false positive on the accumulated diff view.

### Issues

🟡 **daemon/internal/config/config.go:543** — `HEIMDALLM_LOCAL_DIR_BASE` env override only assigns when `len(cleaned) > 0`. This means setting the env var to an empty or whitespace-only string cannot clear a TOML-configured list — inconsistent with most other overrides which take the env value as-is. Minor, but worth documenting or aligning behavior.
🟡 **daemon/internal/config/config.go:340** — `ResolveLocalDir` does not deduplicate or guard against `localDirBases` entries equal to `DefaultReposMountPath`; if both are set to the same path, the stat runs twice. Negligible perf impact but a tiny cleanup opportunity.
🟡 **daemon/internal/config/config.go:550** — Env parsing splits on ',' without trimming surrounding quotes; a user-quoted value like `"/a,/b"` (common in some .env loaders) would be treated as one path. Consider documenting the expected unquoted comma-separated format in .env.example (currently fine, but a note helps).
⚠️ **daemon/cmd/heimdallm/main.go:371** — The call at the repo-discovery loop (`ResolveLocalDir("", repo, c.GitHub.LocalDirBase)`) reads `c.GitHub.LocalDirBase` — verify that `c` here is already captured under `cfgMu` consistent with the pattern used in the other four call sites (which snapshot `localDirBase` inside the locked section). If `c` is a local snapshot this is fine; otherwise it's a subtle race.

### Suggestions

- Confirm the discovery-loop call site at main.go:371 reads `c.GitHub.LocalDirBase` under the same mutex discipline as the other four call sites (which snapshot the slice inside the lock before calling ResolveLocalDir outside it). Slices are reference types, so passing the field directly while another goroutine reassigns `c.GitHub.LocalDirBase` is a data race under `-race`.
- Consider adding a test for the env override parsing with empty/whitespace-only values to lock in the current 'don't clear' behavior (or change it to clear, whichever is intended).
- The author's note explains the binary alert is a false positive after `git filter-branch` rewrote history — recommend verifying on the final merge commit that `daemon/heimdallm` is absent from the tree (`git ls-tree -r HEAD | grep heimdallm`) before merging to close the loop on the bot's repeated flags.
- Minor: in the doc comment for `ResolveLocalDir`, step 2 says 'first match wins' but step 3 (default mount path) is only consulted when no base matched — the current wording is accurate; no change needed, just verifying intent.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured change that extends ResolveLocalDir to support multiple base directories via an ordered list. All call sites are updated consistently, mutex discipline is correct (slice header copied under lock, used outside), tests are comprehensive covering precedence, multi-base resolution, fallback, and env override parsing. Documentation in example files is clear. No confirmed bugs, security issues, or architectural concerns.

### Suggestions

- The comma-separated HEIMDALLM_LOCAL_DIR_BASE env var will break if a directory path contains a comma. Extremely unlikely for real paths, but consider documenting this limitation in the .env.example comment.
- The pattern `cfgMu.Lock(); localDirBase := cfg.GitHub.LocalDirBase; cfgMu.Unlock()` is repeated 4 times across main.go. Consider extracting a small helper (e.g. `cfg.LocalDirBases()`) that encapsulates the lock+copy to reduce repetition and prevent future callers from forgetting the mutex.
- The binary-in-git discussion has been resolved by the author (filter-branch rewrite + .gitignore guard). The 9 auto-review comments are false positives triggered by the cumulative PR diff — consider tuning the review bot to check individual commits or HEAD tree rather than the merge diff to avoid this noise in future PRs.

---

Muriano:
## ⚠️ MEDIUM Issue

Data race: c.GitHub.LocalDirBase is read through a pointer alias (`c := cfg` at line 345) after cfgMu.Unlock() at line 346. Since cfg is *config.Config, `c` points to the same shared struct — reading LocalDirBase (a slice) without the mutex held races with config reloads that replace the slice. The other 4 call sites in this PR correctly extract `localDirBase := cfg.GitHub.LocalDirBase` under the lock. Fix: copy LocalDirBase into a local variable inside the cfgMu.Lock/Unlock block at lines 344-346, the same way the other call sites do, then pass that local to ResolveLocalDir.

**Location:** `daemon/cmd/heimdallm/main.go` line 374

---

Muriano (approved):

Clean feature addition with good test coverage. The local_dir_base list approach is well-designed with correct precedence logic. One confirmed data race on the first call site and a minor inconsistency in existing code that is exposed by this change.

### Suggestions

- The SetConfigFn closure (line 343-346) copies only the pointer (`c := cfg`) rather than dereferencing it (`c := *cfg`). This pre-existing pattern means ALL field reads after the unlock are technically racy — not just LocalDirBase. Consider changing to `c := *cfg` for a deep copy, consistent with the tier2Adapter pattern at line 840 (`c := *a.cfg`). This is a pre-existing issue but this PR makes it more visible.
- Consider adding a test for symlinked directories in local_dir_base to confirm filepath.Join + os.Stat follows symlinks as expected (os.Stat does, but documenting the intent helps).

---

PR #133, opened 2026-04-21
feat: Repositories view — filters, auto-discover, grid toggle (#112)

### LINKED ISSUE #112: feat: Repositories view — filters, auto-discover, grid toggle
## Summary

Improve the Repositories tab with filtering, automatic discovery, and a compact grid view option.

### 1. Filters

Add a filter bar with:
- **Monitored / Not monitored** — toggle or segmented control to show one or both groups

### 2. Remove Discover button — backend auto-discovery

The manual "Discover" button should be removed. Repository discovery should happen automatically in the daemon's poll cycle:
- When polling for PRs, any new repo that appears in `review-requested:@me` results is automatically added to the known repos list
- The existing topic-based discovery (`discovery_topic`) already runs on its own interval
- The Flutter app just reads whatever the daemon knows — no manual trigger needed

This requires a daemon change: the poll cycle should upsert newly-seen repos into the config/store rather than silently skipping unknown ones.

### 3. View toggle: List / Grid

Add a toggle in the toolbar to switch between:

**List view (current)** — vertical list with LED indicators, repo name, local dir, chevron

**Grid view (new)** — compact square tiles in a responsive grid:
- Each tile shows: 3 LED dots, repo short name, org as subtle label
- Tiles arranged in rows/columns that reflow based on window width
- Tap navigates to repo detail (same as list)
- More repos visible at once — useful for users with 30+ repos

The view preference should persist between sessions (SharedPreferences).

## Backwards compatibility

- List view remains the default
- Daemon auto-discovery is additive — never removes repos
- Filter state is ephemeral (resets on tab switch), not persisted

Muriano:
## Summary

- **Daemon**: auto-discovers repos from PR fetch — repos seen in `review-requested:@me` are persisted to config automatically (additive, respects non_monitored blacklist)
- **Flutter**: removed manual Discover button, added monitored/not-monitored filter chips, added list/grid view toggle with persistence

## Changes

| Component | Change |
|---|---|
| `daemon/cmd/heimdallm/main.go` | `FetchPRsToReview` auto-upserts unknown repos to config store |
| `flutter_app/.../repos_screen.dart` | Filter chips, grid view, view toggle, removed Discover button |

## Test Plan
- [x] `go test ./... -race` — all pass
- [ ] Manual: verify new repos appear automatically after daemon poll
- [ ] Manual: filter by Monitored / Not monitored
- [ ] Manual: toggle list ↔ grid, close and reopen — preference persists

Closes #112

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -837,36 +837,35 @@ func (a *tier2Adapter) FetchPRsToReview() ([]scheduler.Tier2PR, error) {
 		})
 	}
 
-	// Auto-discovery: add repos seen in PRs that aren't in the config yet.
+	// Auto-discovery: check + append under a single lock to avoid TOCTOU.
+	var updated []string
+	var newRepos []string
 	a.cfgMu.Lock()
 	c := *a.cfg
-	known := make(map[string]struct{}, len(c.GitHub.Repositories))
+	known := make(map[string]struct{}, len(c.GitHub.Repositories)+len(c.GitHub.NonMonitored))
 	for _, r := range c.GitHub.Repositories {
 		known[r] = struct{}{}
 	}
-	// Also consider non-monitored repos as "known" — don't re-add blacklisted repos.
 	for _, r := range c.GitHub.NonMonitored {
 		known[r] = struct{}{}
 	}
-	a.cfgMu.Unlock()
-
-	var newRepos []string
 	for repo := range seenRepos {
 		if _, ok := known[repo]; !ok {
 			newRepos = append(newRepos, repo)
 		}
 	}
 	if len(newRepos) > 0 {
-		a.cfgMu.Lock()
-		c = *a.cfg
-		updated := append(append([]string(nil), c.GitHub.Repositories...), newRepos...)
+		updated = append(append([]string(nil), c.GitHub.Repositories...), newRepos...)
 		c.GitHub.Repositories = updated
 		*a.cfg = c
-		a.cfgMu.Unlock()
+	}
+	a.cfgMu.Unlock()
 
-		// Persist to store so the repos survive a daemon restart.
-		reposJSON, _ := json.Marshal(updated)
-		if _, err := a.store.SetConfig("repositories", string(reposJSON)); err != nil {
+	if len(newRepos) > 0 {
+		reposJSON, err := json.Marshal(updated)
+		if err != nil {
+			slog.Error("auto-discovery: failed to marshal repos", "err", err)
+		} else if _, err := a.store.SetConfig("repositories", string(reposJSON)); err != nil {
 			slog.Warn("auto-discovery: failed to persist repos", "err", err)
 		} else {
 			slog.Info("auto-discovery: added repos from PRs", "repos", newRepos)
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -809,18 +809,22 @@ type tier2Adapter struct {
 }
 
 // FetchPRsToReview implements scheduler.Tier2PRFetcher.
+// After fetching, any repos not yet in the config are auto-discovered and
+// persisted — the daemon never silently skips unknown repos again.
 func (a *tier2Adapter) FetchPRsToReview() ([]scheduler.Tier2PR, error) {
 	prs, err := a.ghClient.FetchPRsToReview()
 	if err != nil {
 		return nil, err
 	}
 	out := make([]scheduler.Tier2PR, 0, len(prs))
+	seenRepos := make(map[string]struct{})
 	for _, pr := range prs {
 		pr.ResolveRepo()
 		if pr.Repo == "" {
 			slog.Warn("adapter: skipping PR with empty repo", "pr_number", pr.Number)
 			continue
 		}
+		seenRepos[pr.Repo] = struct{}{}
 		out = append(out, scheduler.Tier2PR{
 			ID:        pr.ID,
 			Number:    pr.Number,
@@ -832,6 +836,43 @@ func (a *tier2Adapter) FetchPRsToReview() ([]scheduler.Tier2PR, error) {
 			UpdatedAt: pr.UpdatedAt,
 		})
 	}
+
+	// Auto-discovery: add repos seen in PRs that aren't in the config yet.
+	a.cfgMu.Lock()
+	c := *a.cfg
+	known := make(map[string]struct{}, len(c.GitHub.Repositories))
+	for _, r := range c.GitHub.Repositories {
+		known[r] = struct{}{}
+	}
+	// Also consider non-monitored repos as "known" — don't re-add blacklisted repos.
+	for _, r := range c.GitHub.NonMonitored {
+		known[r] = struct{}{}
+	}
+	a.cfgMu.Unlock()
+
+	var newRepos []string
+	for repo := range seenRepos {
+		if _, ok := known[repo]; !ok {
+			newRepos = append(newRepos, repo)
+		}
+	}
+	if len(newRepos) > 0 {
+		a.cfgMu.Lock()
+		c = *a.cfg
+		updated := append(append([]string(nil), c.GitHub.Repositories...), newRepos...)
+		c.GitHub.Repositories = updated
+		*a.cfg = c
+		a.cfgMu.Unlock()
+
+		// Persist to store so the repos survive a daemon restart.
+		reposJSON, _ := json.Marshal(updated)
+		if _, err := a.store.SetConfig("repositories", string(reposJSON)); err != nil {
+			slog.Warn("auto-discovery: failed to persist repos", "err", err)
+		} else {
+			slog.Info("auto-discovery: added repos from PRs", "repos", newRepos)
+		}
+	}
+
 	return out, nil
 }
 
--- a/daemon/internal/server/handlers.go
+++ b/daemon/internal/server/handlers.go
@@ -778,7 +778,7 @@ func (srv *Server) handleRepoCollaborators(w http.ResponseWriter, r *http.Reques
 }
 
 func (srv *Server) handleStats(w http.ResponseWriter, r *http.Request) {
-	var repos []string
+	var repos, orgs []string
 	if raw := r.URL.Query().Get("repos"); raw != "" {
 		for _, s := range strings.Split(raw, ",") {
 			if s = strings.TrimSpace(s); s != "" {
@@ -786,7 +786,14 @@ func (srv *Server) handleStats(w http.ResponseWriter, r *http.Request) {
 			}
 		}
 	}
-	stats, err := srv.store.ComputeStats(repos)
+	if raw := r.URL.Query().Get("orgs"); raw != "" {
+		for _, s := range strings.Split(raw, ",") {
+			if s = strings.TrimSpace(s); s != "" {
+				orgs = append(orgs, s)
+			}
+		}
+	}
+	stats, err := srv.store.ComputeStats(repos, orgs)
 	if err != nil {
 		slog.Error("handleStats: store error", "err", err)
 		http.Error(w, "internal server error", http.StatusInternalServerError)
--- a/daemon/internal/store/store.go
+++ b/daemon/internal/store/store.go
@@ -227,16 +227,20 @@ type DayCount struct {
 }
 
 // ComputeStats aggregates statistics from the reviews and prs tables.
-// When repos is non-empty, results are scoped to reviews of PRs in those repos.
-func (s *Store) ComputeStats(repos []string) (*Stats, error) {
+// When repos is non-empty, results are scoped to PRs in those repos.
+// When orgs is non-empty, results are scoped to PRs whose repo starts
+// with "org/" (i.e. belongs to one of the given GitHub organizations).
+// repos takes precedence over orgs; both empty = global stats.
+func (s *Store) ComputeStats(repos []string, orgs []string) (*Stats, error) {
 	stats := &Stats{
 		BySeverity: make(map[string]int),
 		ByCLI:      make(map[string]int),
 	}
 
-	// Build reusable filter clauses for the given repos.
+	// Build reusable filter clauses.
 	// repoFilter: for queries on reviews only (uses subquery on prs table).
-	// repoFilterJoined: for queries that already JOIN prs p (direct p.repo IN).
+	// repoFilterJoined: for queries that already JOIN prs p.
+	// Both use the same repoArgs since the placeholder count matches.
 	var repoFilter, repoFilterJoined string
 	var repoArgs []any
 	if len(repos) > 0 {
@@ -248,6 +252,17 @@ func (s *Store) ComputeStats(repos []string) (*Stats, error) {
 		inClause := strings.Join(placeholders, ",")
 		repoFilter = " AND r.pr_id IN (SELECT id FROM prs WHERE repo IN (" + inClause + "))"
 		repoFilterJoined = " AND p.repo IN (" + inClause + ")"
+	} else if len(orgs) > 0 {
+		// Org filter: match repos starting with "org/" using LIKE.
+		// Both clauses use the same args (org + "/%") so repoArgs is shared.
+		var conditions, conditionsJ []string
+		for _, org := range orgs {
+			conditions = append(conditions, "repo LIKE ?")
+			conditionsJ = append(conditionsJ, "p.repo LIKE ?")
+			repoArgs = append(repoArgs, org+"/%")
+		}
+		repoFilter = " AND r.pr_id IN (SELECT id FROM prs WHERE " + strings.Join(conditions, " OR ") + ")"
+		repoFilterJoined = " AND (" + strings.Join(conditionsJ, " OR ") + ")"
 	}
 
 	// Total reviews
--- a/daemon/internal/store/store.go
+++ b/daemon/internal/store/store.go
@@ -254,12 +254,14 @@ func (s *Store) ComputeStats(repos []string, orgs []string) (*Stats, error) {
 		repoFilterJoined = " AND p.repo IN (" + inClause + ")"
 	} else if len(orgs) > 0 {
 		// Org filter: match repos starting with "org/" using LIKE.
-		// Both clauses use the same args (org + "/%") so repoArgs is shared.
+		// Escape SQL LIKE wildcards in org names to prevent unintended matches
+		// (e.g. org "my_team" matching "myXteam/repo" via unescaped _).
+		likeEscaper := strings.NewReplacer("%", "\\%", "_", "\\_")
 		var conditions, conditionsJ []string
 		for _, org := range orgs {
-			conditions = append(conditions, "repo LIKE ?")
-			conditionsJ = append(conditionsJ, "p.repo LIKE ?")
-			repoArgs = append(repoArgs, org+"/%")
+			conditions = append(conditions, "repo LIKE ? ESCAPE '\\'")
+			conditionsJ = append(conditionsJ, "p.repo LIKE ? ESCAPE '\\'")
+			repoArgs = append(repoArgs, likeEscaper.Replace(org)+"/%")
 		}
 		repoFilter = " AND r.pr_id IN (SELECT id FROM prs WHERE " + strings.Join(conditions, " OR ") + ")"
 		repoFilterJoined = " AND (" + strings.Join(conditionsJ, " OR ") + ")"
--- a/flutter_app/lib/core/api/api_client.dart
+++ b/flutter_app/lib/core/api/api_client.dart
@@ -155,12 +155,15 @@ class ApiClient {
     return body['login'] as String? ?? '';
   }
 
-  Future<Map<String, dynamic>> fetchStats({List<String> repos = const []}) async {
-    var path = '/stats';
-    if (repos.isNotEmpty) {
-      path += '?repos=${Uri.encodeQueryComponent(repos.join(','))}';
-    }
-    final resp = await _client.get(_uri(path), headers: await _authHeaders());
+  Future<Map<String, dynamic>> fetchStats({
+    List<String> repos = const [],
+    List<String> orgs = const [],
+  }) async {
+    final params = <String, String>{};
+    if (repos.isNotEmpty) params['repos'] = repos.join(',');
+    if (orgs.isNotEmpty) params['orgs'] = orgs.join(',');
+    final uri = _uri('/stats').replace(queryParameters: params.isNotEmpty ? params : null);
+    final resp = await _client.get(uri, headers: await _authHeaders());
     if (resp.statusCode != 200) throw ApiException('GET /stats failed: ${resp.statusCode}');
     return jsonDecode(resp.body) as Map<String, dynamic>;
   }
--- a/flutter_app/lib/features/dashboard/dashboard_providers.dart
+++ b/flutter_app/lib/features/dashboard/dashboard_providers.dart
@@ -144,28 +144,10 @@ final statsProvider = FutureProvider<Map<String, dynamic>>((ref) async {
   ref.watch(prListRefreshProvider); // refresh stats when reviews complete
   final filters = ref.watch(statsFiltersProvider);
   final api = ref.watch(apiClientProvider);
-
-  // Effective repos: explicit repo selection takes priority. If only orgs
-  // are selected, derive the repo list from known PRs + issues so the
-  // org filter actually scopes the stats.
-  List<String> repos;
-  if (filters.repos.isNotEmpty) {
-    repos = filters.repos.toList();
-  } else if (filters.orgs.isNotEmpty) {
-    final prs = ref.watch(prsProvider).valueOrNull ?? [];
-    final issues = ref.watch(issuesProvider).valueOrNull ?? [];
-    final allRepos = <String>{
-      ...prs.map((p) => p.repo),
-      ...issues.map((i) => i.repo),
-    }..remove('');
-    repos = allRepos.where((r) {
-      final org = r.contains('/') ? r.split('/').first : r;
-      return filters.orgs.contains(org);
-    }).toList();
-  } else {
-    repos = [];
-  }
-  return api.fetchStats(repos: repos);
+  return api.fetchStats(
+    repos: filters.repos.toList(),
+    orgs: filters.orgs.toList(),
+  );
 });
 
 void _rebuildTray(Ref ref, List<PR> prs) {
--- a/flutter_app/lib/features/repositories/repos_screen.dart
+++ b/flutter_app/lib/features/repositories/repos_screen.dart
@@ -22,6 +22,7 @@ class _ReposScreenState extends ConsumerState<ReposScreen> {
   Map<String, RepoConfig> _repoConfigs = {};
   bool _initialized = false;
   String _search = '';
+  String _orgFilter = ''; // empty = all orgs
   _SyncStatus _syncStatus = _SyncStatus.idle;
   _FilterMode _filterMode = _FilterMode.all;
   _ViewMode _viewMode = _ViewMode.list;
@@ -107,10 +108,24 @@ class _ReposScreenState extends ConsumerState<ReposScreen> {
             return a.compareTo(b);
           });
 
+        // Derive all orgs for the org filter dropdown
+        final allOrgs = _repoConfigs.keys
+            .map((r) => r.contains('/') ? r.split('/').first : r)
+            .toSet()
+            .toList()..sort();
+
         // Apply text search
         var filtered = _search.isEmpty
             ? allRepos
             : allRepos.where((r) => r.toLowerCase().contains(_search.toLowerCase())).toList();
+
+        // Apply org filter
+        if (_orgFilter.isNotEmpty) {
+          filtered = filtered.where((r) {
+            final org = r.contains('/') ? r.split('/').first : r;
+            return org == _orgFilter;
+          }).toList();
+        }
 
         // Apply monitored/not-monitored filter
         if (_filterMode == _FilterMode.monitored) {
@@ -124,24 +139,52 @@ class _ReposScreenState extends ConsumerState<ReposScreen> {
 
         return Column(
           children: [
-            // Toolbar
+            // Single-line toolbar: search + filters + view toggle + status
             Padding(
-              padding: const EdgeInsets.fromLTRB(16, 12, 16, 4),
+              padding: const EdgeInsets.fromLTRB(16, 10, 16, 6),
               child: Row(
                 children: [
-                  Expanded(
+                  // Compact search
+                  SizedBox(
+                    width: 160,
                     child: TextField(
                       decoration: const InputDecoration(
-                        hintText: 'Filter repos…',
-                        prefixIcon: Icon(Icons.search, size: 18),
+                        hintText: 'Search…',
+                        prefixIcon: Icon(Icons.search, size: 16),
                         isDense: true,
                         border: OutlineInputBorder(),
-                        contentPadding: EdgeInsets.symmetric(vertical: 8),
+                        contentPadding: EdgeInsets.symmetric(vertical: 6),
                       ),
+                      style: const TextStyle(fontSize: 13),
                       onChanged: (v) => setState(() => _search = v),
                     ),
                   ),
                   const SizedBox(width: 8),
+                  // Monitored filter chips
+                  _filterChip('All', _repoConfigs.length, _FilterMode.all),
+                  const SizedBox(width: 4),
+                  _filterChip('Monitored', monitoredCount, _FilterMode.monitored),
+                  const SizedBox(width: 4),
+                  _filterChip('Not mon.', notMonitoredCount, _FilterMode.notMonitored),
+                  const SizedBox(width: 8),
+                  // Org dropdown
+                  if (allOrgs.length > 1)
+                    SizedBox(
+                      height: 32,
+                      child: DropdownButton<String>(
+                        value: _orgFilter.isEmpty ? null : _orgFilter,
+                        hint: const Text('Org', style: TextStyle(fontSize: 12)),
+                        style: TextStyle(fontSize: 12, color: Theme.of(context).colorScheme.onSurface),
+                        isDense: true,
+                        underline: const SizedBox.shrink(),
+                        items: [
+                          const DropdownMenuItem(value: '', child: Text('All orgs')),
+                          ...allOrgs.map((o) => DropdownMenuItem(value: o, child: Text(o))),
+                        ],
+                        onChanged: (v) => setState(() => _orgFilter = v ?? ''),
+                      ),
+                    ),
+                  const Spacer(),
                   // View toggle
                   SegmentedButton<_ViewMode>(
                     segments: const [
@@ -156,7 +199,7 @@ class _ReposScreenState extends ConsumerState<ReposScreen> {
                       tapTargetSize: MaterialTapTargetSize.shrinkWrap,
                     ),
                   ),
-                  const SizedBox(width: 8),
+                  const SizedBox(width: 6),
                   // Auto-save status
                   SizedBox(
                     width: 22, height: 22,
@@ -167,19 +210,6 @@ class _ReposScreenState extends ConsumerState<ReposScreen> {
                       _SyncStatus.idle   => const SizedBox.shrink(),
                     },
                   ),
-                ],
-              ),
-            ),
-            // Filter chips
-            Padding(
-              padding: const EdgeInsets.fromLTRB(16, 4, 16, 4),
-              child: Row(
-                children: [
-                  _filterChip('All', _repoConfigs.length, _FilterMode.all),
-                  const SizedBox(width: 6),
-                  _filterChip('Monitored', monitoredCount, _FilterMode.monitored),
-                  const SizedBox(width: 6),
-                  _filterChip('Not monitored', notMonitoredCount, _FilterMode.notMonitored),
                 ],
               ),
             ),
@@ -249,7 +279,7 @@ class _RepoGrid extends StatelessWidget {
             crossAxisCount: crossCount,
             mainAxisSpacing: 8,
             crossAxisSpacing: 8,
-            childAspectRatio: 2.2,
+            childAspectRatio: 2.0,
           ),
           itemCount: repos.length,
           itemBuilder: (context, i) {
@@ -257,6 +287,11 @@ class _RepoGrid extends StatelessWidget {
             final config = configs[repo]!;
             final shortName = repo.contains('/') ? repo.split('/').last : repo;
             final org = repo.contains('/') ? repo.split('/').first : '';
+            final configuredDir = (config.localDir ?? '').isNotEmpty ? config.localDir : null;
+            final detectedDir = appConfig.localDirsDetected[repo];
+            final effectiveDir = configuredDir ?? detectedDir;
+            final hasDirMapping = effectiveDir != null && effectiveDir.isNotEmpty;
+            final isAutoDetected = configuredDir == null && detectedDir != null;
 
             return Card(
               child: InkWell(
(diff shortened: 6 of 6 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Feature adds repo auto-discovery in the daemon, org-based stats filtering, and a grid/filter UI for the repos screen. Generally solid, but has a TOCTOU race condition in the auto-discovery logic and a LIKE-injection issue in org filtering that could cause subtle bugs.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:837** — TOCTOU race condition: the mutex is released after building the `known` map, then re-acquired to update `a.cfg`. Between the two critical sections, another goroutine (e.g. an API handler or concurrent scheduler tick) could add repos to the config. The `newRepos` list is computed against stale data, so duplicates can be appended to `c.GitHub.Repositories`. The entire read-check-update should be done under a single lock hold.
⚠️ **daemon/internal/store/store.go:261** — LIKE wildcard injection: org names containing SQL LIKE metacharacters (`%` or `_`) are concatenated directly into the LIKE pattern (`org+"/%"`). An org named `foo%` would match `foo-anything/repo`. The `%` and `_` characters in org names should be escaped before appending `/%`.
🟡 **daemon/cmd/heimdallm/main.go:862** — Silently discarded error from `json.Marshal(updated)`. While marshaling `[]string` is unlikely to fail, the ignored error means a nil `reposJSON` would be passed to `SetConfig`, potentially corrupting the persisted value.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:48** — Empty `catch (_) {}` blocks in `_loadViewPreference` and `_setViewMode` silently swallow all errors, including unexpected ones. At minimum, log or use `debugPrint` so issues are visible during development.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:44** — `_loadViewPreference` is an async method called from `initState` but its future is fire-and-forget — if `SharedPreferences.getInstance()` throws after the widget is disposed, `setState` could be called on an unmounted widget. The `mounted` check covers the happy path but not all async gaps.
🟡 **daemon/internal/store/store.go:258** — Comment says 'Both use the same repoArgs since the placeholder count matches' but this is only correct when either repos OR orgs is set (never both, due to `else if`). The comment is misleading — it implies a shared-args invariant that is actually enforced by control flow, not by design. If someone later changes `else if` to `if`, queries will break silently.

### Suggestions

- Consolidate the two lock sections in `FetchPRsToReview` auto-discovery into a single critical section: hold `a.cfgMu` while reading `known`, computing `newRepos`, and updating `a.cfg`. This eliminates the TOCTOU race.
- Escape LIKE metacharacters in org names before building the SQL pattern. A helper like `strings.NewReplacer("%", "\%", "_", "\_").Replace(org)` plus adding `ESCAPE '\'` to the LIKE clause would prevent unintended wildcard matching.
- Handle the `json.Marshal` error explicitly — either log a warning and return early, or propagate it, rather than passing potentially nil bytes to `SetConfig`.
- Consider adding a deduplication check inside the second lock section (e.g., re-check the live config for already-present repos) as a defense-in-depth measure even after fixing the TOCTOU race.
- The `_RepoGrid` widget is read-only (no toggle for monitoring status), unlike `_RepoListWithSections` which supports inline `onChanged`. Consider adding a quick-action (e.g., long-press or icon button) to toggle monitoring from the grid view for feature parity.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

The PR adds server-side auto-discovery of repos, an orgs filter for stats, and UI filters/grid toggle for the Repositories view. Logic is generally sound, but the auto-discovery code has a TOCTOU race in its locking pattern, swallows a JSON marshal error, and persists repos in nondeterministic order. The stats orgs filter is vulnerable to LIKE wildcard injection from org names.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:847** — TOCTOU race: `known` is built under one lock, then the lock is released, then re-acquired to append `newRepos`. Between the two critical sections another call to FetchPRsToReview (or any other writer of cfg.GitHub.Repositories) can add the same repos, producing duplicates in the persisted config. Compute `newRepos` under a single lock, or use sync.Once / dedupe before append.
🟡 **daemon/cmd/heimdallm/main.go:860** — `c := *a.cfg` followed by `c.GitHub.Repositories = updated; *a.cfg = c` does a shallow copy — the `updated` slice is built via `append(append([]string(nil), c.GitHub.Repositories...), newRepos...)` which is safe, but any other slice/map fields in GitHub (e.g. NonMonitored) still alias the original backing arrays. Fine today, fragile if the struct grows. Consider a narrow setter that only touches Repositories under the mutex.
🟡 **daemon/cmd/heimdallm/main.go:867** — `reposJSON, _ := json.Marshal(updated)` silently discards the marshal error. Marshalling `[]string` essentially cannot fail, but the discard obscures intent — either assert via a helper or log the error for symmetry with the SetConfig error branch.
🟡 **daemon/cmd/heimdallm/main.go:852** — `newRepos` is produced by iterating `seenRepos` (a map), so its order — and thus the order appended to the persisted config — is nondeterministic across runs. Sort before appending to keep the stored config stable and diff-friendly.
🟡 **daemon/internal/store/store.go:258** — Org filter builds `repo LIKE ?` with arg `org + "/%"`. If an org name contains `%` or `_` (or a backslash), those act as LIKE wildcards and match unintended repos (e.g. an org literally named `foo_bar` would match `foobar/...`). Escape `%`, `_`, and `\` in `org` before concatenating, or validate that orgs contain only GitHub-legal characters.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:44** — `_loadViewPreference` is declared `void` but is async. Should be `Future<void>` so callers/tests can await and so lint rules (avoid_void_async) are satisfied. Minor.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:173** — The Discover button was removed in favor of daemon-side auto-discovery, but discovery now only happens when FetchPRsToReview runs on the scheduler's cadence. Until then, a fresh user sees 'No repos yet — the daemon auto-discovers repos from your PRs.' with no action to take. Consider a manual 'Refresh now' affordance or surfacing the next scheduled tick time so users aren't left wondering.

### Suggestions

- Refactor the auto-discovery critical section to a single locked region (or a dedicated `cfg.AddRepos([]string)` method) that reads known, diffs, appends, and returns the updated slice to persist — eliminates the TOCTOU and keeps locking local.
- Add an `escapeLike` helper in store.go that replaces `\`, `%`, `_` and use it for both repo and future LIKE-based filters; include an `ESCAPE '\\'` clause on the LIKE.
- Sort `newRepos` before append+marshal so the persisted `repositories` JSON is stable across restarts.
- Add a table-driven test for `ComputeStats(nil, orgs)` covering: single org, multiple orgs, org with no matching PRs, and repos/orgs both set (verifying repos-wins precedence described in the doc comment).
- Consider a small integration test for the tier2Adapter that asserts a second FetchPRsToReview call does not re-append already-known or non-monitored repos.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Feature PR adding repo auto-discovery, org-based stats filtering, and a grid/filter UI. Generally solid, but has a TOCTOU race in auto-discovery that can produce duplicate repos, and an unescaped LIKE pattern that lets crafted org names match unintended repos.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:837** — TOCTOU race in auto-discovery: the config is read under lock to build `known`, then the lock is released, `newRepos` is computed, and a second lock is acquired to append. A concurrent call to FetchPRsToReview can read the same stale `known` set and both goroutines will append the same repos, creating duplicates. Fix: either perform the known-check and update in a single critical section, or de-duplicate inside the second lock.
⚠️ **daemon/internal/store/store.go:260** — LIKE wildcard injection: org names from user input are used as `org + "/%"` in LIKE clauses without escaping `%` and `_` characters. An org value like `%` would match all repos. Escape LIKE meta-characters (e.g. replace `%` with `\%` and `_` with `\_`) or use a prefix function that is safe.
🟡 **daemon/cmd/heimdallm/main.go:863** — json.Marshal error silently discarded (`reposJSON, _ := json.Marshal(updated)`). While marshaling []string is unlikely to fail, if it did, an empty/corrupt value would be persisted to the store. At minimum log the error.
🟡 **daemon/cmd/heimdallm/main.go:865** — Store persistence happens outside the config lock. If SetConfig fails, the in-memory config diverges from the persisted state: repos appear in memory but are lost on restart. Consider retrying or reverting the in-memory change on failure.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:48** — Empty catch blocks (`catch (_) {}`) on SharedPreferences calls silently swallow all exceptions, making failures invisible. At minimum log to debugPrint so issues are diagnosable.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:222** — The grid view (_RepoGrid) is read-only — it has no onChanged callback and no toggle switches, unlike the list view. Users cannot toggle monitoring status from the grid view, which is a functional gap that may confuse users who switch views.

### Suggestions

- Merge the two lock sections in FetchPRsToReview auto-discovery into one: read `known`, compute `newRepos`, and update `a.cfg` all under a single lock acquisition to eliminate the TOCTOU race.
- Escape LIKE wildcards (`%`, `_`) in org names before building the SQL pattern, e.g. `strings.NewReplacer("%", "\\%", "_", "\\_").Replace(org) + "/%"`.
- Consider adding a brief toast or indicator in the grid view explaining that editing requires switching to list view, or add a long-press action to toggle monitoring.
- The `_loadViewPreference` async method is called from `initState` but has no way to signal errors — consider using a FutureBuilder or at least a debugPrint in the catch block.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR adds repository filters, grid view, and auto-discovery. Core logic is sound but has a TOCTOU race in the config mutation, silently ignored marshal errors, and a minor UX inconsistency in filter chip counts.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:850** — TOCTOU race: the lock is released between computing `newRepos` (against the first snapshot of `known`) and re-acquiring to append them. Between the two critical sections, another goroutine could modify `a.cfg.GitHub.Repositories` (e.g., user edits via handler) — leading to duplicate entries or lost writes. Hold the lock for the full read-compute-write cycle, or use a compare-and-swap pattern.
🟡 **daemon/cmd/heimdallm/main.go:868** — `json.Marshal(updated)` error is discarded with `_`. While marshaling a []string is unlikely to fail, silently ignoring it means a corrupt persisted value could be written. Prefer checking the error and logging/returning before calling SetConfig.
⚠️ **daemon/cmd/heimdallm/main.go:870** — Auto-discovery persists only the `repositories` key via SetConfig, but if the daemon's startup path reads repositories from a file config (e.g. YAML) rather than the store, the auto-discovered repos may not survive a restart as the comment claims. Verify SetConfig('repositories',...) is actually what the loader reads.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:58** — `_loadViewPreference` is declared `void` but is async. Errors inside the async body become unhandled futures; the caller (initState) cannot await it. Change to `Future<void>` for clarity and to allow tests to await initialization.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:145** — Filter chip counts (`monitoredCount`, `notMonitoredCount`) are computed against the full `_repoConfigs` map and do not reflect the current search text or org filter. Users filtering by 'acme' org will still see 'Monitored (42)' even if only 3 acme repos are monitored — inconsistent with the displayed list.
🟡 **daemon/internal/store/store.go:259** — The LIKE pattern uses `org + "/%"` without escaping SQL LIKE metacharacters. If an org name contains `%` or `_` (rare but possible in test data or via config injection), the filter will over-match. Consider escaping or validating the org string.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:320** — `_RepoGrid` duplicates the dir-mapping and LED-logic from `_RepoTile` (configuredDir/detectedDir/effectiveDir/hasDirMapping/isAutoDetected). A small helper or extension on RepoConfig would avoid the risk of the two views drifting apart on future changes.

### Suggestions

- Hold `cfgMu` for the full read-compute-write of auto-discovery to avoid the TOCTOU window.
- Extract a shared helper for repo dir-mapping resolution used by both _RepoTile and _RepoGrid.
- Update filter-chip counts to reflect currently-applied search/org filters, or clarify they are global totals via a tooltip.
- Consider making the empty-string DropdownMenuItem value null-based to avoid conflating 'no selection' with an actual org value.
- Add a unit test for ComputeStats with the orgs filter to lock in the LIKE behaviour.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Feature adds repo auto-discovery, org-based stats filtering, and a grid/filter UI. The backend auto-discovery has a TOCTOU race that can produce duplicate repos, and the LIKE-based org filter doesn't escape SQL wildcard characters. UI changes are solid overall.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:855** — TOCTOU race condition: the mutex is released after building the `known` map (line 845) and re-acquired before writing (line 855). Between those two critical sections another goroutine (or concurrent FetchPRsToReview call) could add the same repos, resulting in duplicates in `c.GitHub.Repositories`. The known-check and the append should happen under a single lock acquisition.
⚠️ **daemon/internal/store/store.go:261** — LIKE wildcard injection: org names are interpolated into a LIKE pattern as `org + "/%"` without escaping `%` and `_` characters. A GitHub org containing `_` (valid in GitHub usernames) would match unintended repos. Escape LIKE wildcards before appending `/%`, e.g. replace `%` with `\%` and `_` with `\_`.
⚠️ **daemon/cmd/heimdallm/main.go:865** — json.Marshal error silently discarded (`reposJSON, _ := json.Marshal(updated)`). While marshalling a []string is unlikely to fail, if it does the subsequent SetConfig call will persist an empty/nil value, silently corrupting the stored repo list. At minimum log the error.
⚠️ **daemon/cmd/heimdallm/main.go:837** — The struct copy `c := *a.cfg` creates a shallow copy. If the config struct has slice/map fields beyond GitHub.Repositories, the copy and the original will share underlying memory. Only Repositories gets a fresh slice via the `append([]string(nil), ...)` idiom — other fields like NonMonitored or RepoConfigs may alias. Verify the config struct doesn't have mutable fields that could be concurrently modified.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:45** — _loadViewPreference is async-void called from initState. The `mounted` guard is good, but `catch (_) {}` silently swallows all exceptions including unexpected ones (e.g. PlatformException). Consider at least logging or limiting the catch to the expected exception type.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:55** — _setViewMode uses `.catchError((_) {})` which silently swallows errors and also has a subtle type issue — `catchError` expects `Function` matching the future's type signature. Prefer `try/catch` or `.then(...).catchError(...)` with proper typing to avoid runtime type errors on some platforms.
🟡 **daemon/internal/store/store.go:258** — The org filter builds two parallel condition lists (`conditions` and `conditionsJ`) with identical args — fragile to maintain. If one list is updated without the other, the query breaks silently. Consider building a single condition template and prefixing the table alias programmatically.

### Suggestions

- Merge the two critical sections in FetchPRsToReview's auto-discovery into a single lock: read known repos, compute new repos, and append all under one mutex acquisition to eliminate the race.
- Add a helper to escape LIKE wildcards (`%`, `_`) in the org filter before constructing the pattern, e.g. `strings.NewReplacer("%", "\%", "_", "\_").Replace(org) + "/%"`.
- Consider adding a deduplication step (or a SET constraint in the store) for the repositories list to protect against duplicates from concurrent discovery.
- The toolbar packs many controls (search, 3 filter chips, org dropdown, segmented button, status icon) into a single Row — this will likely overflow on narrow windows or tablets in portrait. Consider wrapping or using a responsive layout that collapses filters behind a menu on smaller screens.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

The PR adds auto-discovery of repos, an orgs filter for /stats, and UI improvements (filter chips, grid view, view-mode persistence). Logic is mostly sound, but the auto-discovery code in main.go has a race condition across two separate lock/unlock cycles that can cause duplicate repos or lost updates. Several smaller correctness and style issues also exist.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:860** — Auto-discovery uses two separate critical sections: one to read known repos and a later one to append newRepos. Between the two unlocks, another goroutine (or another run of FetchPRsToReview) can also compute newRepos from a stale snapshot and append, producing duplicate entries in c.GitHub.Repositories and a duplicate-write to the store. Combine read+write into a single critical section, or recompute the 'already known' set while holding the second lock and skip repos that became known in between.
⚠️ **daemon/cmd/heimdallm/main.go:869** — The in-memory config (*a.cfg) is mutated before persistence. If store.SetConfig fails, the running daemon has repos that are not persisted, so behavior diverges from the stored config and the repos disappear on restart. Either persist first and only update memory on success, or roll back the in-memory append on persistence failure.
🟡 **daemon/cmd/heimdallm/main.go:868** — json.Marshal error is discarded (reposJSON, _ := json.Marshal(updated)). For []string it won't realistically fail, but the pattern is inconsistent with the rest of the block that logs errors — either handle it or add a short comment explaining why it is safe.
⚠️ **daemon/internal/store/store.go:260** — ComputeStats gives 'repos' precedence over 'orgs' (else if), but the HTTP handler accepts both simultaneously and forwards both. A client passing both filters will silently get orgs ignored. Document this in the handler, or apply both filters (AND) so the behavior matches what a caller would naturally expect.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:45** — _loadViewPreference is declared 'void … async' and called from initState without unawaited/ignore. The try/catch silently swallows all errors including programming mistakes (e.g. TypeError) — at least log, or narrow the catch to expected PlatformException/MissingPluginException.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:54** — _setViewMode fires SharedPreferences.setString without awaiting; the catchError swallows everything. If the write fails the UI state and persisted preference diverge silently. Consider awaiting and surfacing a (quiet) log, or at least logging the error.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:168** — Org filter chip is only rendered when allOrgs.length > 1, but _orgFilter state is not cleared when the repo set shrinks to a single org. Stale org selections could persist invisibly across config changes; clear _orgFilter when allOrgs.length <= 1 (or when selected orgs no longer exist).
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:300** — Directory-mapping logic (configuredDir/detectedDir/effectiveDir/isAutoDetected) is duplicated between _RepoGrid and _RepoTile. Extracting a small helper (e.g. on RepoConfig or a free function) would keep the two views in sync when the rules evolve.

### Suggestions

- In tier2Adapter.FetchPRsToReview, hold a.cfgMu across the entire read-modify-write of Repositories to eliminate the race between the two critical sections.
- Persist to the store first and only mutate *a.cfg on success so in-memory and persisted state never diverge.
- Decide an explicit combination semantics for repos+orgs in /stats (precedence vs. AND vs. reject) and document it on both the handler and ComputeStats.
- Use unawaited() or ignore: avoid_async_void for the fire-and-forget SharedPreferences calls, and log errors instead of silently swallowing them.
- Extract the effective-dir / LED computations shared by _RepoTile and _RepoGrid into a single helper to avoid drift.

---

Muriano:
## Fixes applied

| Issue | Fix |
|---|---|
| 🔴 TOCTOU race in auto-discovery | Check + append under single `cfgMu.Lock()` — no gap between read and write |
| ⚠️ LIKE wildcard injection | `strings.NewReplacer("%", "\\%", "_", "\\_")` + `ESCAPE '\\'` clause |
| ⚠️ json.Marshal error silenced | Now logged with `slog.Error` |

All tests pass with `-race`.

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Feature-rich PR adding repo filters, auto-discovery, and grid toggle. The daemon-side auto-discovery has a TOCTOU race condition that can silently re-add blacklisted repos or lose concurrent config edits. The SQL and Flutter changes are generally sound with minor issues.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:839** — TOCTOU race condition: the `known` map (including NonMonitored/blacklisted repos) is built under the first lock, but `newRepos` is computed outside any lock and applied under a second lock. Between unlock #1 and lock #2, another goroutine could add a repo to NonMonitored (blacklist it), but `newRepos` was already computed against the stale `known` map — the blacklisted repo gets re-added to Repositories. Fix: perform the entire read-diff-write in a single locked section.
⚠️ **daemon/cmd/heimdallm/main.go:866** — json.Marshal error silently discarded: `reposJSON, _ := json.Marshal(updated)`. If marshaling fails (unlikely for []string but possible under memory pressure), a nil/empty value is persisted to the store, potentially wiping the repo list on next daemon restart.
⚠️ **flutter_app/lib/features/repositories/repos_screen.dart:250** — Grid view (_RepoGrid) has no `onChanged` callback — users cannot toggle monitoring status from the grid, unlike the list view. This is a UX inconsistency that may confuse users who switch views.
🟡 **daemon/internal/store/store.go:261** — Org names containing SQL LIKE wildcards (% or _) in `org+"/%"` would match unintended repos. GitHub org names don't currently allow these characters, but a defensive `strings.ReplaceAll` for `%` and `_` before appending `/%` would be safer.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:49** — SharedPreferences errors are silently swallowed in both `_loadViewPreference` (`catch (_) {}`) and `_setViewMode` (`.catchError((_) {})`). At minimum, log these to aid debugging if preferences stop persisting.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:120** — Search field has a fixed 160px width (`SizedBox(width: 160)`). On narrow screens or with large font sizes, the toolbar chips, org selector, view toggle, and search may overflow horizontally. Consider using `Flexible` or `ConstrainedBox` with a max width instead.

### Suggestions

- Merge the two lock sections in FetchPRsToReview's auto-discovery into a single critical section: lock once, build `known`, compute `newRepos`, mutate config, unlock, then persist to store outside the lock.
- Handle the json.Marshal error: if it fails, skip the store persistence and log a warning rather than writing potentially corrupt data.
- Either add monitoring toggles to the grid view cards or add a tooltip/hint that editing requires switching to list view.
- Consider extracting the org-parsing logic (`r.contains('/') ? r.split('/').first : r`) into a shared helper — it's duplicated in repos_screen.dart (org filter, _RepoGrid, _groupByOrg) and store.go.

---

Muriano:
The 🔴 TOCTOU, ⚠️ json.Marshal, and 🟡 LIKE wildcard issues were all fixed in e2799f7 (previous round). The review appears to have been written against the pre-fix snapshot.

Re: ⚠️ grid view no monitoring toggle — by design. Tapping a grid tile navigates to repo detail where all config (monitoring, IT, local dir) is available. Adding toggles to compact tiles would clutter the grid layout. The grid is for quick overview + navigation, not inline editing.

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid PR that addresses earlier review feedback well (TOCTOU fix, LIKE wildcard escaping). Server-side org filtering is a clean architectural improvement over the previous client-side repo derivation. A couple of subtle issues around auto-discovery state consistency and some minor Flutter code style nits, but overall the changes are well-scoped and correct.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:868** — If store.SetConfig fails to persist the new repos, the in-memory config has already been mutated (*a.cfg = c happened inside the lock). On the next tick these repos will be treated as 'known' and skipped, so a transient DB failure permanently loses them until daemon restart. Consider rolling back the in-memory update on persistence failure, or retaining a 'pending persist' set so the next fetch retries.
🟡 **daemon/cmd/heimdallm/main.go:857** — Iteration over seenRepos (map) produces newRepos in non-deterministic order, which is then appended to c.GitHub.Repositories and persisted. This causes the stored repo list order to be unstable across runs and can churn downstream consumers that compare lists. Sort newRepos before appending.
🟡 **daemon/internal/store/store.go:265** — The OR-chain of LIKE clauses (e.g. 'repo LIKE ? OR repo LIKE ? OR ...') cannot use an index even if one exists on repo, because LIKE '<pattern>/%' prefix-optimization only works with a single constant prefix. If the orgs list grows and this is hit per dashboard refresh, consider rewriting as a prefix extraction (substr(repo, 1, instr(repo, '/')-1) IN (?, ?, ...)) or just fetching repos belonging to those orgs server-side and using IN.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:45** — _loadViewPreference is declared as 'void ... async' rather than 'Future<void> ... async'. void-async functions silently swallow errors (the try/catch with empty catch already does this, but the type is still a smell and prevents callers from awaiting). Change to Future<void>.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:53** — _setViewMode fires-and-forgets the SharedPreferences write with .then/.catchError, swallowing errors silently. If the user toggles rapidly, writes can interleave and the 'last write wins' is not guaranteed to match the current UI state. Consider awaiting inside an async method or at least logging on failure.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:120** — _orgFilter and _filterMode are in-memory only and reset on every screen rebuild/navigation, but _viewMode is persisted via SharedPreferences. This inconsistency is surprising UX — either persist all three filter states or document why view mode is special.
🟡 **daemon/internal/store/store.go:260** — Doc comment says 'repos takes precedence over orgs; both empty = global stats', which is implemented via else-if. This means a caller that passes BOTH repos and orgs silently has orgs ignored. If that's intentional, fine — but the API surface (handleStats accepts both) lets clients pass both. Consider either AND-combining or returning a 400 if both are supplied, to avoid confusing behavior.

### Suggestions

- Add a unit test for ComputeStats with the orgs filter path, covering wildcard-escape cases like org names containing '_' or '%' to lock in the fix.
- Consider extracting the auto-discovery block in tier2Adapter.FetchPRsToReview into a helper method — FetchPRsToReview now mixes fetch + side-effect persistence and is getting long.
- For the repos grid view, the InkWell + Card tap-to-navigate is discoverable, but there's no visible indicator (hover state, cursor) communicating that each card is clickable — consider MouseRegion + cursor for desktop/web.
- Sort newRepos before persistence so the stored repositories list is stable across runs.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid refactor that removes the manual Discover button in favor of server-side auto-discovery, adds proper org filtering at the SQL layer, and introduces a grid view with filter chips. The TOCTOU fix and LIKE-escape hardening are well-done. Two residual concerns: a persistence ordering race in auto-discovery (SetConfig runs outside the lock, so concurrent calls can clobber each other's writes on disk) and a product-level question about auto-discovered repos being added directly to the monitored list.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:857** — Persistence race: the in-memory update happens under cfgMu, but store.SetConfig is called after Unlock. If two FetchPRsToReview calls interleave, the later-committed DB write can be the one with fewer repos. In-memory state stays correct (later adds are supersets), but the persisted 'repositories' row can temporarily regress, meaning a restart before the next fetch loses some auto-discovered repos until re-seen. Either hold cfgMu across SetConfig, serialize persistence with a dedicated mutex, or have SetConfig re-read from *a.cfg under lock right before writing.
⚠️ **daemon/cmd/heimdallm/main.go:856** — Auto-discovery appends every new repo to c.GitHub.Repositories (the monitored list), so any repo that shows up in a PR becomes auto-review-enabled by default. The old Discover flow let the user opt in; this flips the default to opt-out. Worth confirming this is intentional — if not, consider adding to a 'pending/discovered' bucket that the UI surfaces for confirmation before promoting to Repositories.
🟡 **daemon/internal/store/store.go:261** — likeEscaper escapes % and _ but not the escape character itself (\). GitHub org names can't contain backslash today so this is theoretical, but for defense-in-depth and to match standard LIKE-escaping practice, also replace \ -> \\ in the replacer.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:56** — _loadViewPreference is declared `void ... async`. Prefer `Future<void>` so callers/tests can await it and unhandled errors surface; the empty catch also silently swallows prefs failures.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:138** — _search.toLowerCase() is recomputed for every repo in the filter; hoist into a local before the .where() for clarity and minor perf.

### Suggestions

- Hold cfgMu across both the in-memory update and the store.SetConfig call (or use a dedicated persistence mutex) so concurrent FetchPRsToReview invocations can't reorder their on-disk writes.
- Decide explicitly whether auto-discovered repos should be monitored by default or land in a non-monitored/pending state for the user to promote — document the choice in the new comment on FetchPRsToReview.
- Add a unit test for ComputeStats with orgs containing '%' and '_' to lock in the LIKE-escape behavior, and one covering the 'repos takes precedence over orgs' contract now stated in the doc.
- Consider persisting _orgFilter and _filterMode alongside _viewMode in SharedPreferences so the repos screen restores the full filter state, not just the view toggle.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR refactors the Repositories view (filters, grid toggle, removes manual discover button) and moves repo auto-discovery into the daemon's tier2 adapter while adding org-based filtering to the /stats endpoint. The TOCTOU and LIKE-escape fixes from earlier rounds are correctly applied. Main concerns are around (1) auto-discovery silently promoting new repos to the monitored list rather than a non-monitored staging list, and (2) UX persistence inconsistency in the Flutter screen (view mode is persisted while filter/org state is not).

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:858** — Auto-discovery appends newly-seen repos to c.GitHub.Repositories (the monitored list), which — combined with the default RepoConfig behavior — will cause any previously-unknown repo seen in FetchPRsToReview to become auto-reviewed on the very next scheduler cycle. Given the PR framing ('the daemon never silently skips unknown repos again'), this is likely intentional, but it is a meaningful behavior change: a single cross-org PR requesting the user as reviewer now enrolls that repo for auto-review. Consider adding it to NonMonitored instead (opt-in), or at least logging at Warn level and documenting the implicit enrollment.
🟡 **daemon/cmd/heimdallm/main.go:850** — `c := *a.cfg` performs a shallow copy, so `c.GitHub.NonMonitored` (and any other slices/maps on the Config) still alias the live config's backing arrays. The code only reads NonMonitored here so it's safe today, but the pattern is fragile: a future edit that mutates any slice field on `c` before the `*a.cfg = c` write would corrupt the original under the lock. Worth a brief comment that only `Repositories` is being replaced with a fresh slice, or explicitly rebuild `c` to avoid aliasing.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:60** — UX persistence is inconsistent: `_viewMode` is persisted to SharedPreferences while `_filterMode` and `_orgFilter` are reset on every mount. Users who configure a monitored/org filter will lose it on app restart while the grid/list preference sticks. Either persist all three, or none — otherwise the behavior is surprising.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:46** — `_loadViewPreference` is declared `void ... async` and fire-and-forget from `initState`. If the user toggles the SegmentedButton before the async load resolves, the stored value will overwrite the user's fresh choice. Low likelihood, but can be avoided by capturing a 'user has interacted' flag or only applying the loaded value when `_viewMode` is still at its default.
🟡 **flutter_app/lib/core/api/api_client.dart:165** — `Uri.replace(queryParameters: ...)` percent-encodes values, so `org/repo` becomes `org%2Frepo` in the query string. The backend splits on `,` and trims, so decoded values will come through correctly via Go's `r.URL.Query()` (which decodes automatically). Behavior matches the previous `Uri.encodeQueryComponent` path, just worth confirming with a quick manual test against the daemon since the encoding is now slightly different (previously the whole comma-joined string was a single encoded component; now each value is encoded individually and commas are passed through).
🟡 **daemon/internal/store/store.go:267** — `repoFilterJoined = " AND (" + strings.Join(conditionsJ, " OR ") + ")"` will be concatenated into a larger WHERE clause. If any caller places it immediately after a boolean operator other than AND/another predicate (e.g. start of clause), the leading ' AND ' handles it — but the inner OR chain is only wrapped once. This is correct as-is for the current callers, but a brief test covering multi-org filter queries would harden against future regressions.

### Suggestions

- Consider adding a unit test for the auto-discovery path in tier2Adapter.FetchPRsToReview — specifically verifying: (a) known repos are not duplicated, (b) the store.SetConfig call failure does not prevent returning PRs, and (c) concurrent invocations produce a consistent final repository list.
- Add an integration/unit test for ComputeStats with the `orgs` parameter, including an org name containing `_` or `%` to lock in the ESCAPE behavior.
- Persist `_filterMode` and `_orgFilter` alongside `_viewMode` in SharedPreferences for a consistent UX, or document the intentional asymmetry.
- Consider extracting the org-derivation logic `r.contains('/') ? r.split('/').first : r` into a small helper — it now appears in four places across the file and in dashboard_providers.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid PR that moves org-based stats filtering server-side, adds auto-discovery in the daemon, and introduces a grid view toggle. The TOCTOU, LIKE-escape, and error-logging fixes are correct. Main concerns are a likely toolbar overflow on narrow screens and minor accessibility gaps in the Flutter UI.

### Issues

⚠️ **flutter_app/lib/features/repositories/repos_screen.dart:131** — Toolbar Row packs a fixed-width 160px search field, 3 FilterChips, an Org Chip, a SegmentedButton, and a status indicator into a single non-wrapping Row. On screens narrower than ~700px this will cause a RenderFlex overflow. Consider wrapping in a Wrap, using a horizontal scroll, or collapsing filters behind a dropdown on narrow layouts.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:170** — The Org filter uses GestureDetector wrapping a Chip instead of ActionChip or InputChip. This loses built-in button semantics (focus ring, hover feedback, accessibility label), making the control less accessible for keyboard and screen-reader users.
🟡 **daemon/cmd/heimdallm/main.go:856** — If store.SetConfig fails, in-memory config already contains the new repos (updated inside the lock) but persistence didn't succeed. On daemon restart the auto-discovered repos will be lost and silently re-discovered on next PR fetch. This is acceptable best-effort behavior but consider logging at Error level instead of Warn, since the in-memory/disk divergence could confuse debugging.
🟡 **daemon/internal/store/store.go:255** — The OR-joined LIKE conditions inside the subquery ('SELECT id FROM prs WHERE repo LIKE ? OR repo LIKE ?') are not parenthesised. Currently safe because they are the only WHERE predicates in that subquery, but if a future change adds conditions (e.g. a date filter) without adding parens, the OR will bind incorrectly. Wrapping in parentheses now ('WHERE (' + join + ')') would be defensive.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:48** — _loadViewPreference is an async void method called from initState. If SharedPreferences.getInstance() throws after the widget is disposed (between the await and the mounted check), the error is swallowed by the catch-all. Minor, but using a Completer or cancelling on dispose would be more robust.
🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:149** — The refactored statsProvider now always sends orgs to the server and never falls back to client-side repo derivation. If the daemon hasn't been upgraded yet (older binary without the new orgs query param), the org filter will silently do nothing. Consider documenting the minimum daemon version or adding a version check.

### Suggestions

- Wrap the toolbar content in a Wrap or use LayoutBuilder to collapse filters into a dropdown/overflow menu on narrow screens to prevent RenderFlex overflow.
- Replace GestureDetector+Chip with ActionChip for the org filter to get proper accessibility semantics for free.
- Add parentheses around the OR'd LIKE conditions in the subquery for defensive SQL correctness: 'WHERE (' + strings.Join(conditions, " OR ") + ')'.
- Consider a brief CHANGELOG or migration note indicating the daemon must be updated alongside the Flutter app for org-based stats filtering to work correctly.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

The three substantive issues from the previous round (TOCTOU race, LIKE injection, silent marshal error) are all properly fixed. The remaining concerns are minor: empty catch blocks in the Flutter layer and a potential horizontal overflow in the toolbar Row on narrow viewports.

### Issues

🟡 **flutter_app/lib/features/repositories/repos_screen.dart:48** — Empty `catch (_) {}` in `_loadViewPreference` and `.catchError((_) {})` in `_setViewMode` (line 57) silently swallow all errors. At minimum use `debugPrint` so failures are visible during development. (Unfixed from previous round.)
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:121** — The toolbar Row packs a 160 px search field, three FilterChips, an org Chip, a SegmentedButton, and a status icon into a single non-scrollable Row. On narrower windows (or with many orgs selected) this will cause a RenderFlex overflow. Consider wrapping in a SingleChildScrollView(scrollDirection: Axis.horizontal) or switching to Wrap.
🟡 **daemon/internal/store/store.go:258** — Comment 'Both use the same repoArgs since the placeholder count matches' is still slightly misleading — it is true only because the `else if` makes repos/orgs mutually exclusive, not because of an inherent design invariant. A one-line note like '(repos and orgs are mutually exclusive — else-if above)' would prevent a future maintainer from accidentally enabling both.

### Suggestions

- Consider using `debugPrint` or `FlutterError.reportError` in the catch blocks instead of silently swallowing errors — helps during development without crashing in production.
- The toolbar could wrap its filter widgets in a `SingleChildScrollView(scrollDirection: Axis.horizontal)` to gracefully handle narrow viewports.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid refactor moving auto-discovery to the daemon and replacing client-side org-to-repo derivation with server-side org filtering. Prior round fixed the TOCTOU, wildcard escaping, and marshal error concerns. A few residual issues remain around non-deterministic ordering and swallowed persistence failures.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:860** — New repos are collected via `for repo := range seenRepos` (map iteration), so the order appended to `c.GitHub.Repositories` — and written to the persisted config — is non-deterministic across runs. This produces churn in the stored config and makes logs/diffs harder to reason about. Sort `newRepos` before appending.
⚠️ **daemon/cmd/heimdallm/main.go:869** — If `store.SetConfig` fails, in-memory `a.cfg.Repositories` already contains the new repos but the store does not. On the next fetch, `known` will include those repos (from in-memory cfg) so `newRepos` will be empty and no retry is ever attempted; the repos silently disappear on daemon restart. Consider either rolling back the in-memory update on persistence failure, or decoupling the `known` set from the in-memory state so a retry is guaranteed.
🟡 **daemon/cmd/heimdallm/main.go:847** — `c := *a.cfg` is a shallow copy — the slice header is copied but the backing array is shared. The subsequent `updated = append(append([]string(nil), c.GitHub.Repositories...), newRepos...)` correctly allocates a fresh slice so the aliasing is benign here, but it's subtle. A brief comment noting that `updated` must be a freshly-allocated slice (to avoid mutating the previously-observed backing array under readers holding an old `*a.cfg` pointer) would help future maintainers not regress this.
🟡 **daemon/internal/store/store.go:265** — The LIKE escaper handles `%` and `_` but not `\` itself. If an org name ever contained a literal backslash, it would be interpreted as an escape character and could corrupt the pattern. Very unlikely in practice for GitHub org names, but worth escaping `\` → `\\` for completeness.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:53** — `_loadViewPreference` is declared `void async` rather than `Future<void>`. The result is that errors inside the `try` block cannot propagate and the caller in `initState` cannot await it (not needed here, but the typing is misleading). Declaring it `Future<void>` is the idiomatic Dart form.
🟡 **flutter_app/lib/features/repositories/repos_screen.dart:150** — The filter-chip counts (`All (N)`, `Monitored (N)`, `Not mon. (N)`) reflect the total repo set, not the search-filtered set. When the user types a search term, the counts still show totals, which can be confusing when the list is visibly shorter. Either compute counts after the search filter, or document that counts represent totals.

### Suggestions

- Sort `newRepos` before appending to stored config to keep persisted order stable.
- On `SetConfig` failure, either roll back the in-memory update or track a pending-persist set so the next fetch retries.
- Consider extracting the `org := r.contains('/') ? r.split('/').first : r` logic into a small helper — it's duplicated in `repos_screen.dart`, the grid tile, and the org-filter derivation.
- Add a test for the org LIKE filter covering an org with an underscore (e.g. `my_team`) to lock in the escaping behaviour against regression.

---

PR #140, opened 2026-04-21
fix: per-repo overrides lost on config save (#138)

### LINKED ISSUE #138: bug: per-repo overrides (Develop, IT, PR Reviewers) lost on every config save
## Problem

Per-repo configuration set on the Repo Detail screen (Develop labels, Issue Tracking, PR Reviewers, PR Assignee, PR Labels, Draft) is silently lost whenever any config save triggers.

Users have to reconfigure repos 20+ times. This is the #1 UX pain point.

## Root Cause

`GET /config` returns `repo_overrides` missing 4 fields. In `main.go` `SetConfigFn` (line ~349):

```go
ro := map[string]any{
    "primary":     ai.Primary,
    "fallback":    ai.Fallback,
    "review_mode": ai.ReviewMode,
    "local_dir":   ai.LocalDir,
}
```

**Missing:** `pr_reviewers`, `pr_assignee`, `pr_labels`, `pr_draft`.

Flutter's `AppConfig.fromJson` parses these as `null` → `_autoSave` writes TOML without them → overrides erased.

## Fix

Add the missing fields to `SetConfigFn`:

```go
ro := map[string]any{
    "primary":      ai.Primary,
    "fallback":     ai.Fallback,
    "review_mode":  ai.ReviewMode,
    "local_dir":    ai.LocalDir,
    "pr_reviewers": ai.PRReviewers,
    "pr_assignee":  ai.PRAssignee,
    "pr_labels":    ai.PRLabels,
    "pr_draft":     ai.PRDraft,
}
```

## Files

- `daemon/cmd/heimdallm/main.go` — `SetConfigFn` closure (~line 349)
- `flutter_app/lib/core/models/config_model.dart` — `AppConfig.fromJson` (already parses these fields correctly, just never receives them)

Muriano:
## Summary

`GET /config` was missing `pr_reviewers`, `pr_assignee`, `pr_labels`, `pr_draft` from `repo_overrides`. Flutter parsed them as `null`, then `_autoSave` wrote TOML without them — silently erasing per-repo overrides on every save.

**1 file changed, 4 lines added.**

## Test Plan
- [x] `go build` + `go test` — pass
- [x] Manual: set Develop labels on a repo, edit another field, verify labels persist after reload

Closes #138

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -347,10 +347,14 @@ func main() {
 		repoOverrides := make(map[string]map[string]any)
 		for repo, ai := range c.AI.Repos {
 			ro := map[string]any{
-				"primary":     ai.Primary,
-				"fallback":    ai.Fallback,
-				"review_mode": ai.ReviewMode,
-				"local_dir":   ai.LocalDir,
+				"primary":      ai.Primary,
+				"fallback":     ai.Fallback,
+				"review_mode":  ai.ReviewMode,
+				"local_dir":    ai.LocalDir,
+				"pr_reviewers": ai.PRReviewers,
+				"pr_assignee":  ai.PRAssignee,
+				"pr_labels":    ai.PRLabels,
+				"pr_draft":     ai.PRDraft,
 			}
 			if ai.IssueTracking != nil {
 				ro["issue_tracking"] = ai.IssueTracking


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Fix correctly adds the four missing PR-related override fields (pr_reviewers, pr_assignee, pr_labels, pr_draft) to the per-repo override serialization, resolving the data-loss bug on config save. The change is minimal and targeted. However, unlike IssueTracking which is conditionally included only when non-nil, the new fields are always written — potentially serializing zero/empty values (e.g., pr_draft=false, empty pr_labels slice) that could be interpreted on reload as an explicit override and shadow global defaults.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:355** — pr_reviewers, pr_assignee, pr_labels, and pr_draft are written unconditionally. If a repo override never set these (common case), the saved YAML will contain zero-value entries (e.g., pr_draft: false, empty slices) which on the next load may be treated as explicit overrides that suppress global AI.PRDraft/PRReviewers/etc. Consider mirroring the IssueTracking pattern: only include each field when it is non-zero / non-nil (e.g., `if len(ai.PRReviewers) > 0`, `if ai.PRAssignee != ""`, `if len(ai.PRLabels) > 0`, and track PRDraft with a *bool or sentinel so unset vs false is distinguishable).
⚠️ **daemon/cmd/heimdallm/main.go:355** — PRDraft as a plain bool cannot distinguish 'not overridden' from 'explicitly false'. Round-tripping through this save will coerce any unset per-repo PRDraft into an explicit false override. If the config struct doesn't already use *bool or an omitempty-safe type, this fix will surface that latent issue. Recommend verifying the AIRepo struct/YAML tags and adjusting to pointer/optional semantics if needed.
🟡 **daemon/cmd/heimdallm/main.go:350** — No test coverage is shown for the regression being fixed. A round-trip test (load → modify unrelated field → save → reload → assert PR overrides preserved) would prevent recurrence, especially given the original bug was exactly this class of serialization omission.

### Suggestions

- Guard each new field with a non-zero check analogous to the existing `if ai.IssueTracking != nil` block so save output stays clean and doesn't fabricate overrides.
- Add a unit/integration test that loads a config with per-repo PR overrides, saves it, reloads, and asserts equality — guards against future field additions being forgotten again.
- Consider refactoring the override map construction into a helper (e.g., `repoOverrideMap(ai)`) that uses reflection or explicit per-field conditionals, so adding new fields to AIRepo doesn't require remembering to update this save path.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Straightforward bug fix that adds four missing PR-related fields (pr_reviewers, pr_assignee, pr_labels, pr_draft) to the repoOverrides serialization map. The omission caused these per-repo settings to be silently dropped on config save. The fix is minimal, correct, and low-risk.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:347** — No regression test added. The bug (missing fields on save) is easy to reintroduce silently if new fields are added to the AI repo config struct in the future without updating this map. A table-driven round-trip test (marshal → unmarshal → compare) would catch this class of bug automatically.
🟡 **daemon/cmd/heimdallm/main.go:347** — If PRLabels or PRReviewers are nil slices, they will serialize as JSON null rather than an empty array []. Depending on how the consumer reads the config back, this could cause a type mismatch or require nil-handling on the read path. Worth verifying the deserialization side handles null vs [] consistently.

### Suggestions

- Add a round-trip test that populates all RepoAI fields, saves config, reloads it, and asserts equality — this would have caught the original bug and will catch any future omissions.
- Consider generating or deriving the repoOverrides map via struct tags (e.g., json/mapstructure) instead of manually enumerating fields, so new struct fields are automatically included without a separate code change here.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

The PR correctly fixes the immediate bug by adding four missing PR-metadata fields to the config serialization map. However, the same class of bug still exists for three other RepoAI struct fields (Prompt, IssuePrompt, ImplementPrompt) that are also omitted from serialization and would be silently dropped on config save.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:350** — Incomplete fix: the RepoAI struct also contains `Prompt`, `IssuePrompt`, and `ImplementPrompt` fields that are still not included in the repoOverrides map. Any per-repo prompt profile overrides will continue to be silently lost on config save — the exact same class of bug this PR aims to fix.
⚠️ **daemon/cmd/heimdallm/main.go:350** — Fragile serialization pattern: manually enumerating struct fields in a map literal will silently regress every time a new field is added to RepoAI. Consider a reflection-based approach, marshaling via json/mapstructure tags, or at minimum a test that round-trips the config and asserts all fields survive.

### Suggestions

- Add the three missing prompt fields to the repoOverrides map: "prompt": ai.Prompt, "issue_prompt": ai.IssuePrompt, "implement_prompt": ai.ImplementPrompt.
- Add a round-trip test that populates every RepoAI field, saves the config, reloads it, and asserts equality — this would catch any future field omissions automatically.
- Consider replacing the manual map construction with struct marshaling (e.g., json.Marshal/Unmarshal or mapstructure) so new fields are included by default.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Straightforward bug fix that adds four missing per-repo override fields (pr_reviewers, pr_assignee, pr_labels, pr_draft) to the config serialization map, preventing them from being silently dropped on save. The change is consistent with how existing fields are handled.

### Suggestions

- Verify there are no other fields on the repo AI config struct that are still missing from this serialization map — the same class of bug could exist for any field added to the struct in the future. Consider generating this map from struct tags or adding a test that round-trips the config to catch drift.
- The new slice/string fields (pr_reviewers, pr_labels, pr_assignee) are serialized even when empty/zero-valued, which is consistent with existing fields like primary and fallback. However, if downstream consumers distinguish between 'absent' and 'empty', consider adding nil/zero-value guards similar to the existing IssueTracking nil check.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Small, targeted fix that restores PR-related fields (pr_reviewers, pr_assignee, pr_labels, pr_draft) to the per-repo override serialization so they are no longer lost on config save. Change is correct and minimal. The underlying manual-map approach remains fragile and is the root cause of this class of bug.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:350** — Manually mirroring struct fields into a map[string]any is the exact pattern that caused this regression — any future field added to the per-repo AI config must be remembered here too, with no compile-time safety net. Consider serializing the struct directly (e.g. yaml.Marshal of the struct, or reflection/struct tags) so new fields are picked up automatically.
⚠️ **daemon/cmd/heimdallm/main.go:353** — Zero-valued fields (nil/empty PRReviewers, empty PRLabels, empty PRAssignee string, false PRDraft) will now always be written into the override block, potentially turning a repo that had no explicit override into one with an explicit empty/false override. Depending on how load merges overrides with defaults, this can silently clear reviewers/labels/assignee or force pr_draft=false. Recommend only emitting keys when the field is non-zero/non-nil, mirroring the pattern already used for ai.IssueTracking.

### Suggestions

- Guard each pr_* key with a non-zero check (len()>0 for slices, != "" for strings, and a tri-state/pointer for pr_draft if false must be distinguishable from unset) to avoid writing back spurious overrides.
- Refactor the override serialization to marshal the AI struct directly so adding a new field cannot silently be dropped on save again.
- Add a regression test that loads a config with all per-repo override fields set, re-saves, reloads, and asserts round-trip equality — this would have caught the original bug and will catch the next one.

---

Issue #144, opened 2026-04-21
bug: config round-trip design is fundamentally broken — null erases configured values


Muriano:
## Problem

The current config save flow silently destroys data:

1. Flutter fetches `GET /config` → gets a snapshot
2. User changes one field
3. Flutter writes the **entire** config back (TOML or `PUT /config`)
4. Any field the daemon didn't expose, or Flutter didn't parse, comes back as `null`
5. `null` overwrites the previously configured value → **data loss**

The #138 fix (exposing missing fields) is a band-aid. The root cause is that the Flutter app acts as the source of truth for the full config, but it can never have a complete picture — the daemon has fields that Flutter doesn't know about, TOML has structure Flutter doesn't model, and any version mismatch silently erases data.

## Required Design Change

**Principle: `null` means "don't touch", not "delete".**

### 1. Patch-based saves (not full replacement)

Flutter should send **only the changed fields** to the daemon, not the full config. The daemon merges the patch on top of its current state.

- `PUT /config` accepts a partial JSON body — only keys present in the body are updated
- Keys absent from the body are left untouched
- The daemon writes the merged result to TOML (it owns the file, not Flutter)

### 2. Explicit reset endpoint

To reset a per-repo override back to "use global default", Flutter calls:

`DELETE /config/repos/{repo}/{field}`

This is the ONLY way to remove a configured value. A missing key in a `PUT` body does not delete anything.

### 3. Flutter stops writing TOML

`FirstRunSetup._buildToml` / `writeConfig` should only be used for initial setup (first run). After the daemon is running, all config changes go through the HTTP API. The daemon is the single source of truth for config.toml.

## Impact

This eliminates the entire class of "config lost on save" bugs. No field can ever be silently erased by a client that doesn't know about it.

## Files affected

- `daemon/internal/server/handlers.go` — `handlePutConfig` becomes a merge, not a replace
- `daemon/internal/server/handlers.go` — new `DELETE /config/repos/{repo}/{field}` endpoint
- `flutter_app/lib/features/config/config_providers.dart` — `save()` sends patches, not full config
- `flutter_app/lib/features/repositories/repo_detail_screen.dart` — `_autoSave` sends patches
- `flutter_app/lib/shared/widgets/override_field.dart` — reset button calls DELETE endpoint
- `flutter_app/lib/core/setup/first_run_setup.dart` — `writeConfig` limited to first-run only

Muriano:
## Additional context from PR #140 review

[...]

PR #147, opened 2026-04-21
refactor: global IT/Develop settings match repo detail layout

Muriano:
## Summary

Refactors the global Issue Tracking and Develop sections in Settings to match the repo detail UI pattern:

- **Issue Tracking**: card layout, SwitchListTile toggle, AutocompleteChipField for labels/orgs/assignees
- **Develop** (separate card): toggle, develop labels, PR Reviewers, PR Labels (chips)
- Daemon exposes `pr_metadata` (global reviewers + labels) in `GET /config`
- Flutter parses and round-trips `globalPRReviewers` / `globalPRLabels`
- TOML writer includes `[ai.pr_metadata]` section
- Includes cherry-pick of #138 fix (per-repo overrides lost on save)

## Test Plan
- [x] `go build` — pass
- [x] `flutter analyze` — no errors
- [ ] Manual: global IT settings use chip fields, not text fields
- [ ] Manual: Develop section shows PR Reviewers + PR Labels
- [ ] Manual: per-repo overrides survive config save


ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

The PR title says 'refactor: global IT/Develop settings match repo detail layout' but the actual code change is minimal — only main.go exposes new PR metadata fields in the config snapshot endpoint. Most of the diff consists of accidentally-committed brainstorming artifacts and an unrelated security audit plan that don't belong in this PR. The code change itself is mechanical but has a consistency gap between per-repo and global PR metadata fields.

### Issues

🔴 **.superpowers/brainstorm/47485-1776681835/state/server-info:1** — Local dev-state file committed to the repo. Contains absolute paths from the author's machine (/Users/jamuriano/personal-workspace/auto-pr/...) and a random server port. This is transient brainstorming runtime state and should be gitignored, not committed.
⚠️ **.superpowers/brainstorm/47485-1776681835/content/*.html:0** — Four HTML brainstorming mockups are committed. These are internal design-exploration artifacts unrelated to the shipped change; they bloat the PR and the repo history. Add .superpowers/ to .gitignore and drop these from the PR.
🔴 **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — A 1040-line security-audit implementation plan is added in this PR, which is about UI layout refactoring. Unrelated scope — mixes a large planning doc (with concrete code diffs for auth fail-safe, data races, denylist expansion, etc.) into an unrelated UI PR. Either split into its own PR or remove from this one.
⚠️ **daemon/cmd/heimdallm/main.go:416** — Inconsistency between per-repo and global PR metadata exposure: repo overrides expose pr_reviewers, pr_assignee, pr_labels, pr_draft (4 fields), but the new global pr_metadata block only exposes reviewers and labels. If the UI's goal is to 'match the repo detail layout', global must also expose assignee and draft — otherwise those fields can never be set/displayed at the global level, breaking the override semantics (you can override a global value that doesn't exist).
⚠️ **daemon/cmd/heimdallm/main.go:350** — The per-repo override payload now includes PR metadata fields unconditionally, even when unset. If ai.PRReviewers / ai.PRLabels are nil slices, they will serialize as null in JSON. Confirm the Flutter client tolerates null vs [] for these fields, or normalise to empty slices here. Also check whether ai.PRDraft is a bool or *bool — a zero-value false is indistinguishable from 'not overridden' without a pointer or a presence flag.
⚠️ **daemon/cmd/heimdallm/main.go:416** — c.AI.PRMetadata is dereferenced without a nil check. If PRMetadata is a pointer type or can be zero-valued in some configs (older config files without this section), this could nil-panic or silently emit zero values. Verify the Config type guarantees it's initialized.
🔴 **PR-wide:0** — PR title and scope mismatch: title promises 'global IT/Develop settings match repo detail layout' but diff contains no UI/Flutter changes matching that description — only a tiny backend payload addition plus unrelated artifacts. Either the UI commits are missing from this branch, or the title should be narrowed to 'expose PR metadata in config snapshot'.

### Suggestions

- Remove the .superpowers/brainstorm/ directory from this PR and add it (or the whole .superpowers/ tree) to .gitignore so future brainstorming sessions don't leak local dev state into the repo.
- Split the 2026-04-06-security-audit-fixes.md plan into its own PR — it's a substantial, independently-reviewable document about auth, data races, and credential denylists that has nothing to do with the UI refactor.
- Align the global pr_metadata block with the per-repo override shape: expose assignee and draft too, or document why they're repo-only.
- Confirm the PR actually contains the Flutter UI changes its title describes; if they're on a different branch, rebase them in. If this PR is only the backend half, rename the title accordingly.
- Normalise nil slices to [] when serialising pr_reviewers / pr_labels to avoid the client having to handle both null and empty array.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

PR adds PR metadata fields to daemon config and includes brainstorm/planning artifacts. The code change in main.go has a structural inconsistency between repo-level and global-level PR metadata representation, and development tooling state is committed.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:416** — Structural inconsistency between repo-level and global-level PR metadata. Repo overrides use flat keys (`pr_reviewers`, `pr_assignee`, `pr_labels`, `pr_draft`) while global config uses a nested `pr_metadata` object with only `reviewers` and `labels`. This mismatch means: (1) consumers must handle two different shapes for the same concept, (2) global config is missing `assignee` and `draft` fields that repos can override — there's no global default to inherit from.
🟡 **.superpowers/brainstorm/47485-1776681835/state/server-info:1** — Local development tooling state committed to the repo. Contains developer-specific filesystem paths (`/Users/jamuriano/personal-workspace/auto-pr/...`) and ephemeral local server info (port 51785). This file and likely the entire `.superpowers/brainstorm/` directory should be gitignored.
⚠️ **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — 1040-line security audit plan documents specific vulnerabilities with exact file paths, line numbers, and exploitation scenarios (unauthenticated daemon access, data races, credential path bypasses). Consider whether this should live in a private issue tracker rather than the repo, especially if the fixes haven't all landed yet — it's a roadmap for attackers.

### Suggestions

- Align repo-level and global-level PR metadata to use the same structure. Either flatten the global config to match repos (`pr_reviewers`, `pr_assignee`, etc.) or nest the repo-level fields under a `pr_metadata` key. Add missing `assignee` and `draft` to the global config so repos have defaults to inherit from.
- Add `.superpowers/brainstorm/` (or at least `*/state/`) to `.gitignore` to avoid committing ephemeral local dev server state.
- Move the security audit plan to a private tracking system (e.g., GitHub Security Advisories or a private issue) until all fixes are merged, then consider publishing a summary.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

PR adds brainstorming HTML docs, a security audit plan, and extends daemon config to expose PR metadata at both global and per-repo levels. The main code concern is a structural inconsistency between how PR metadata is shaped globally vs per-repo, plus missing fields at the global level.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:416** — Global pr_metadata is a nested map with only 'reviewers' and 'labels', but per-repo config (lines 352-355) uses flat keys and includes 'pr_assignee' and 'pr_draft' as well. Consumers must handle two different shapes, and global defaults cannot be set for assignee/draft. Either flatten the global config to match per-repo, or nest per-repo to match global — and include all four fields at both levels.
🟡 **.superpowers/brainstorm/47485-1776681835/state/server-info:1** — This file contains local filesystem paths (/Users/jamuriano/personal-workspace/auto-pr/...) and a localhost dev-server URL. It's a transient development artifact that shouldn't be committed. Consider adding .superpowers/brainstorm/*/state/ to .gitignore.
🟡 **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — 1040-line security plan with embedded code snippets referencing specific line numbers (e.g. main.go:85-91, executor.go:248-253). These will rot as soon as the target files change. If this plan is meant to be executed later, the line-number references will mislead. Consider linking to functions/symbols instead of line numbers, or treat this as ephemeral and don't merge it to main.

### Suggestions

- Align global and per-repo PR metadata to use the same structure and field set. Recommended: use flat keys at both levels (pr_reviewers, pr_assignee, pr_labels, pr_draft) for consistency, or nest both under a pr_metadata map — but either way, expose all four fields at the global level too.
- Add .superpowers/brainstorm/*/state/ to .gitignore to avoid committing local dev server state files.
- Consider whether the brainstorming HTML files (.superpowers/brainstorm/) belong in the repo long-term or should be excluded once decisions are made.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

PR expone más campos de configuración AI/PR en el endpoint de settings. El cambio de código es minimalista y correcto, pero el PR incluye múltiples archivos no relacionados (brainstorm HTML, plan de security audit, server-info con rutas locales) que ensucian el diff y contradicen el título del PR.

### Issues

⚠️ **.superpowers/brainstorm/47485-1776681835/state/server-info:1** — Archivo de estado local commiteado al repo: contiene ruta absoluta /Users/jamuriano/personal-workspace/..., puerto 51785 y metadata de un servidor de brainstorm local. No debería versionarse — añadir .superpowers/ al .gitignore.
🟡 **.superpowers/brainstorm/47485-1776681835/content/override-behavior.html:0** — Los 5 archivos HTML de brainstorm son artefactos de herramienta local y no pertenecen a un PR de refactor. Inflan el diff y dificultan la revisión del cambio real.
⚠️ **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — Plan de 1040 líneas sobre 'Security Audit Fixes' que no tiene relación con el título del PR ('global IT/Develop settings match repo detail layout'). Mezclar trabajo no relacionado rompe el principio de PRs atómicos y oculta la intención del cambio.
⚠️ **daemon/cmd/heimdallm/main.go:416** — Inconsistencia de campos: el bloque global pr_metadata solo expone 'reviewers' y 'labels', mientras que el bloque per-repo añade también pr_assignee y pr_draft. Si el frontend usa ambos niveles, el usuario no puede configurar assignee/draft a nivel global desde la UI aunque la estructura lo permita en el override. Confirmar si es intencional o si c.AI.PRMetadata carece de esos campos.
🔴 **daemon/cmd/heimdallm/main.go:416** — No hay guardia de nil para c.AI.PRMetadata. Si esa struct es un puntero que puede ser nil cuando no se define en config, el acceso a .Reviewers/.Labels causará panic en runtime. Verificar que PRMetadata sea value-type o añadir nil-check como se hace con ai.IssueTracking en la línea 356.
⚠️ **daemon/cmd/heimdallm/main.go:350** — Los nuevos campos (pr_reviewers, pr_assignee, pr_labels, pr_draft) se emiten siempre en el override aunque no estén definidos en la config del repo. Esto producirá valores cero (slices nil, string vacío, bool false) que la UI podría interpretar como 'override activo con valor vacío' en vez de 'no override'. Considerar emitirlos solo si están presentes, como ya se hace con issue_tracking.

### Suggestions

- Eliminar del PR los archivos .superpowers/brainstorm/** y docs/superpowers/plans/2026-04-06-security-audit-fixes.md; abrir PRs separados o descartarlos. Añadir .superpowers/ a .gitignore.
- Unificar los campos expuestos entre global y per-repo: si pr_assignee y pr_draft existen a nivel global, exponerlos; si no, documentar por qué la asimetría.
- Añadir un nil-check o emitir pr_metadata solo si está configurado, siguiendo el patrón de issue_tracking (líneas 355-357).
- Para pr_reviewers/pr_labels per-repo, usar el patrón condicional (if len(ai.PRReviewers) > 0) para no contaminar el objeto de override con campos vacíos que dificultan distinguir 'no override' de 'override a valor vacío'.
- Añadir un test en handlers_test.go (o cmd test) que verifique la forma del payload de settings con PRMetadata vacío y poblado.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

The main.go changes add PR metadata to global and per-repo config, but there are inconsistencies between the two levels. The PR also commits local dev tool state and an unrelated 1040-line security audit plan.

### Issues

🔴 **.superpowers/brainstorm/47485-1776681835/state/server-info:1** — Local dev machine state committed to the repo. Contains hardcoded local paths (/Users/jamuriano/personal-workspace/auto-pr/...) and a local server port. This file should be gitignored, not checked in.
⚠️ **daemon/cmd/heimdallm/main.go:416** — Global pr_metadata only exposes 'reviewers' and 'labels', but per-repo overrides (line 352-355) expose 'pr_reviewers', 'pr_assignee', 'pr_labels', and 'pr_draft'. Missing 'assignee' and 'draft' at the global level looks like an oversight — the UI/client will have no global defaults for those fields.
⚠️ **daemon/cmd/heimdallm/main.go:352** — Naming inconsistency: per-repo uses flat keys with pr_ prefix (pr_reviewers, pr_labels) while global uses a nested pr_metadata object with unprefixed keys (reviewers, labels). This forces consumers to handle two different shapes for the same concept, making the API harder to use and more error-prone.
⚠️ **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — A 1040-line security audit implementation plan is included in a PR titled 'refactor: global IT/Develop settings match repo detail layout'. This appears unrelated to the PR's stated scope — consider splitting it into its own PR to keep changes reviewable and bisectable.
⚠️ **daemon/cmd/heimdallm/main.go:352** — No nil/zero-value guards on ai.PRReviewers, ai.PRAssignee, ai.PRLabels, ai.PRDraft. If these fields are optional and the config doesn't set them, the map will contain zero values (empty strings/slices/false) that are indistinguishable from intentional empty overrides. This may cause the client to interpret 'not set' as 'explicitly cleared'.
🟡 **.superpowers/brainstorm/47485-1776681835/content/override-behavior.html:1** — Five brainstorm HTML files (design mockups from a local tool) are committed. If these are ephemeral design artifacts, they add ~360 lines of non-production HTML to the repo. Consider whether these belong in version control or should be gitignored.

### Suggestions

- Unify the PR metadata shape: use a nested 'pr_metadata' object at both global and per-repo level with the same keys (reviewers, labels, assignee, draft) so consumers don't need to handle two different schemas.
- Add .superpowers/brainstorm/*/state/ to .gitignore to avoid committing local server state with developer-specific paths.
- Only include pr_* keys in the repo override map when they are explicitly set in config (e.g., check len > 0 or use a pointer type) to distinguish 'not configured' from 'intentionally empty'.
- Split the security audit plan into its own PR to keep this refactor focused and reviewable.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR #147 refactors global IT/Develop settings to match repo detail layout. The visible diff focuses on daemon/cmd/heimdallm/main.go exposing additional AI overrides (pr_reviewers, pr_assignee, pr_labels, pr_draft) and a new top-level pr_metadata object in the SSE snapshot. The rest of the diff consists of brainstorm HTML artifacts, server-info state, and an unrelated security audit plan document — these should not be in the PR. The Go changes are small and straightforward but there are concerns around committed tooling artifacts and a potentially leaked absolute local path.

### Issues

⚠️ **.superpowers/brainstorm/47485-1776681835/state/server-info:1** — Committing brainstorm tooling artifacts (HTML mockups + server-info JSON) pollutes the repo and leaks an absolute local path ('/Users/jamuriano/personal-workspace/auto-pr/...'). These files appear unrelated to the stated PR scope (global IT/Develop settings match repo detail layout) and should be gitignored rather than committed. Add '.superpowers/' to .gitignore and remove from this PR.
⚠️ **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — A 1040-line security audit plan is included in a PR whose title says 'refactor: global IT/Develop settings match repo detail layout'. Mixing unrelated changes makes the PR harder to review and bisect. Split into its own PR or remove if it's an in-progress planning doc.
⚠️ **daemon/cmd/heimdallm/main.go:416** — The new 'pr_metadata' top-level key only exposes c.AI.PRMetadata.Reviewers and c.AI.PRMetadata.Labels but omits 'assignee' and 'draft', even though the per-repo override map now includes pr_assignee and pr_draft. This asymmetry means the web client can see per-repo draft/assignee overrides but has no global baseline to compare against — confirm this is intentional or add the missing fields for consistency.
⚠️ **daemon/cmd/heimdallm/main.go:350** — pr_reviewers, pr_assignee, pr_labels, pr_draft are emitted unconditionally for every repo override, even when the per-repo AI config doesn't set them. Depending on the struct's zero values (nil slices, empty strings, false), the client will see 'override' values that are actually just zero-value defaults. Consider only emitting these keys when the repo actually overrides them (as is done for issue_tracking on line ~357), otherwise the UI override indicators may misfire.
🟡 **daemon/cmd/heimdallm/main.go:416** — Field alignment in the map literal was changed from aligned to realigned — the pre-existing trailing whitespace alignment on 'repo_overrides', 'agent_configs', 'local_dirs_detected' was preserved but 'pr_metadata' breaks the visual column alignment. Minor style nit — run gofmt to normalize.

### Suggestions

- Split the PR: keep only the daemon main.go changes related to exposing pr_metadata / per-repo PR overrides. Remove the brainstorm/ directory and the security-audit plan.
- Add '.superpowers/' and similar tooling output paths to .gitignore to prevent future leakage of absolute local paths and ephemeral session state.
- Emit pr_reviewers/pr_assignee/pr_labels/pr_draft in repoOverrides only when the repo actually sets them, mirroring the conditional handling of issue_tracking — this keeps 'override' semantics honest for the UI.
- Include assignee and draft in the global pr_metadata block so the UI can show complete inherited values.
- Run 'gofmt -s' and 'go test ./... -race' locally before merging to catch formatting and race regressions.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

PR adds PR metadata fields to repo overrides and global config, plus brainstorm/design HTML files and a security audit plan. The Go code changes are small but have an inconsistency between repo-level and global-level PR metadata shape, and the security audit plan documents unfixed vulnerabilities in detail.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:413** — Inconsistent PR metadata shape: repo-level exposes flat keys (pr_reviewers, pr_assignee, pr_labels, pr_draft) while global-level nests under pr_metadata with only reviewers and labels. Missing assignee and draft at global level, and naming diverges (pr_reviewers vs reviewers). Consumers will need different parsing logic per level.
🔴 **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — Security audit plan documents 8 unfixed vulnerabilities with exact exploit paths (auth bypass via empty token, data race in cachedLogin, missing denylist entries, etc.). Committing this before fixes are merged exposes attack surface to anyone with repo read access.
⚠️ **.superpowers/brainstorm/47485-1776681835/state/server-info:1** — Committed local development server state containing developer-specific absolute paths (/Users/jamuriano/personal-workspace/auto-pr/...) and ephemeral port info. This file should be gitignored, not committed.
⚠️ **daemon/cmd/heimdallm/main.go:350** — New repo override fields (pr_reviewers, pr_assignee, pr_labels, pr_draft) are added unconditionally to the map even when they may be zero-valued. This sends empty strings/slices/bools for repos that don't configure PR metadata, which could override meaningful global defaults with empty values downstream.
🟡 **.superpowers/brainstorm/47485-1776681835/content/override-behavior.html:1** — Brainstorm/design HTML files appear to be internal tooling artifacts, not production code. Consider whether these belong in the repo or should live in a separate design tool/wiki.

### Suggestions

- Unify PR metadata shape: use the same nested structure (pr_metadata.reviewers, pr_metadata.labels, pr_metadata.assignee, pr_metadata.draft) at both repo and global levels so consumers have a single parsing path.
- Omit zero-valued PR metadata fields from the repo overrides map (check for empty string/nil before inserting) to avoid accidentally overriding global defaults with empty values.
- Remove or defer the security audit plan from this PR — merge it only after the vulnerabilities it documents are fixed, or move it to a private issue tracker.
- Add .superpowers/brainstorm/*/state/ to .gitignore to prevent committing ephemeral local server state files.
- Consider whether the brainstorm HTML files should be part of this refactor PR or split into a separate PR to keep the scope focused on the actual settings layout changes.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

Code change itself is small and straightforward (adds PR metadata fields to the SSE settings payload), but the PR bundles unrelated artifacts that shouldn't be committed: local brainstorm HTML files with absolute user paths, and a security audit plan document unrelated to the PR's stated scope.

### Issues

⚠️ **.superpowers/brainstorm/47485-1776681835/state/server-info:1** — Commits a local dev-tool state file containing an absolute user path (/Users/jamuriano/personal-workspace/...) and a random port. This is ephemeral brainstorm tooling state and should be gitignored, not committed.
⚠️ **.superpowers/brainstorm/47485-1776681835/content/override-behavior.html:1** — Brainstorm HTML mockups don't belong in the repo — these are transient design artifacts. Along with override-per-field.html, repo-detail-layout.html, repo-navigation.html, and waiting.html, they should be excluded via .gitignore.
⚠️ **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — A 1040-line security audit implementation plan is added in a PR titled 'refactor: global IT/Develop settings match repo detail layout'. Unrelated scope — should be split into its own PR for reviewability and clean git history.
⚠️ **daemon/cmd/heimdallm/main.go:350** — pr_reviewers/pr_assignee/pr_labels/pr_draft are emitted unconditionally per repo override, including zero values (empty slices, empty string, false). This makes it impossible for the client to distinguish 'explicitly set to empty/false' from 'not overridden — inherit global'. Consider emitting these fields only when non-zero, or adding explicit presence flags, to keep override semantics consistent with how issue_tracking is handled (nil check on line following).
🔴 **daemon/cmd/heimdallm/main.go:416** — New pr_metadata block exposes reviewers and labels but omits assignee and draft, while the per-repo override block includes all four (pr_reviewers, pr_assignee, pr_labels, pr_draft). Inconsistent — the global payload should expose the same surface the UI needs to render inheritance/override indicators, otherwise the client can't compute 'differs from global' for assignee/draft.
🟡 **daemon/cmd/heimdallm/main.go:416** — No test coverage added for the new pr_metadata SSE field or the new pr_* repo override fields. Given this payload drives the settings UI, a snapshot/handler test verifying field presence and shape would prevent regressions.

### Suggestions

- Remove .superpowers/brainstorm/** from the PR and add the directory to .gitignore.
- Split the security-audit-fixes.md plan into its own PR — it's unrelated to the repo-detail settings layout refactor.
- Include pr_assignee and pr_draft in the global pr_metadata block to match the per-repo override shape, so the client can consistently compute inheritance across all four fields.
- Mirror the nil-check pattern used for issue_tracking: emit pr_* fields in repoOverrides only when explicitly set, so the UI can tell 'override to empty' from 'inherit'. If the underlying types can't express presence, consider adding a small wrapper or sentinel.
- Add a handler test asserting the new SSE payload shape (both the global pr_metadata and a repo override including pr_* fields).

---

Muriano:
## Fixes applied

| Issue | Fix |
|---|---|
| 🔴 .superpowers/ local state committed | Removed, added to .gitignore |
| ⚠️ Unrelated security audit plan | Removed from PR |
| ⚠️ Per-repo fields emitted when empty | Now conditional: only emitted when len > 0 or non-empty |
| ⚠️ Global pr_metadata always emitted | Now conditional: only when reviewers or labels configured |
| ⚠️ Naming inconsistency global vs per-repo | Acknowledged — will be resolved in #144 (patch-based config redesign) |

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

PR refactors the Config screen to show global IT (Issue Tracking) and Develop settings using a card layout consistent with the repo detail page, and threads global PR metadata (reviewers/labels) through the Flutter model, TOML writer, and daemon /config endpoint. Changes are focused and the conditional emission of fields is correct. However, the PR still contains the 'unrelated' security-audit-fixes plan file that the author's comment claims was removed, and the Develop 'Auto-implement' switch has a confusing toggle-on behavior. A one-shot init pattern in build can also mask config refreshes.

### Issues

🔴 **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — Author's PR comment states this 1040-line unrelated security audit plan was 'Removed from PR', but `gh pr view` still lists the file and the diff still contains the new file (+1040 lines). Either the comment is inaccurate or the removal commit was never pushed. This file is unrelated to the stated PR goal (global IT/Develop layout) and should be dropped from this branch.
⚠️ **flutter_app/lib/features/config/config_screen.dart:534** — The 'Auto-implement issues' SwitchListTile binds `value: hasLabels` (derived from `_issueTracking.developLabels.isNotEmpty`) but the `onChanged` handler only does work when turning OFF (clears developLabels). When the user toggles ON, no state change occurs, so the switch snaps back to OFF with no feedback — confusing UX. Either seed a default label on enable, or decouple the enabled flag from list emptiness (e.g. add an explicit enabled bool) so the toggle is meaningful.
⚠️ **flutter_app/lib/features/config/config_screen.dart:520** — `_initDevelopFromConfig` is called from inside `_developSection` (i.e. from build) guarded by a one-shot `_developInitialized` flag. If the upstream `config` provider refreshes with new `globalPRReviewers`/`globalPRLabels` (e.g. after a reload or remote update), the local state will never pick them up and the user will continue editing stale values. Prefer `initState` + listening via `ref.listen` / `didUpdateWidget`, or reset the flag when the underlying config identity changes.
🟡 **flutter_app/lib/core/models/config_model.dart:492** — `(json['pr_metadata'] as Map<String, dynamic>?)?['reviewers']` and the matching labels line cast `pr_metadata` twice. Minor — extract into a local `final prMeta = json['pr_metadata'] as Map<String, dynamic>?;` to avoid the duplicated cast and potential future divergence.
🟡 **daemon/cmd/heimdallm/main.go:363** — `if ai.PRDraft { ro["pr_draft"] = ai.PRDraft }` means clients cannot distinguish 'unset' from 'explicitly false'. Acceptable since false is the default and the PR explicitly opted into 'only emit when non-default', but worth noting if the future Flutter UI ever wants to surface a tri-state.
🟡 **daemon/cmd/heimdallm/main.go:430** — Naming inconsistency: global block emits `pr_metadata.reviewers` / `pr_metadata.labels` (nested), while per-repo emits flat `pr_reviewers` / `pr_labels`. Author acknowledged this will be resolved in #144, so tracking only.
🟡 **flutter_app/lib/features/config/config_screen.dart:130** — `BoxConstraints(maxWidth: double.infinity)` makes the ConstrainedBox a no-op. Either remove the ConstrainedBox entirely or keep a real max-width. Cosmetic.

### Suggestions

- Drop the docs/superpowers/plans/2026-04-06-security-audit-fixes.md file from this branch — either push the removal commit or rebase it out. Its continued presence contradicts the PR description and couples an unrelated security initiative to this layout refactor.
- For the 'Auto-implement issues' switch, introduce an explicit `_developEnabled` bool in state (independent of label list emptiness) and persist/consume it, so toggling ON is a visible, meaningful user action.
- Move the develop-state initialization out of the build path. Either do it in `initState` with a one-time read of the initial config, or listen to config changes via Riverpod `ref.listen` so users never see stale reviewer/label values after a refresh.
- Consider inlining the tiny `_parseStringList` helper's JSON path into a single `final prMeta = json['pr_metadata'] as Map<String, dynamic>?;` to avoid the double cast.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Small, focused refactor that makes global AI/PR settings emission conditional (only when values are configured), matching the per-repo detail layout. Core change in main.go is correct. Main concern is an unrelated 1040-line security audit plan markdown file included in the diff; author indicated it was removed but the provided diff still shows it being added — needs verification that the final push actually dropped it.

### Issues

⚠️ **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — Unrelated security audit implementation plan (1040 lines) included in a PR whose stated scope is global IT/Develop settings layout. Author claims it was removed in a follow-up; confirm the file is no longer present in the final commit before merging.
🟡 **daemon/cmd/heimdallm/main.go:430** — Naming inconsistency: global settings use 'pr_metadata.reviewers'/'pr_metadata.labels' while per-repo overrides use flat 'pr_reviewers'/'pr_labels'/'pr_assignee'/'pr_draft'. Clients must handle two shapes for the same concept. Author acknowledged this will be addressed in #144, but worth an inline TODO referencing that issue.
🟡 **daemon/cmd/heimdallm/main.go:364** — Per-repo 'pr_draft' is only emitted when true, so a repo explicitly setting draft:false is indistinguishable from unset. If override semantics matter (repo-level false overriding a global true), this omission loses that signal. Acceptable if draft is currently only global, but worth confirming intent.
🟡 **daemon/cmd/heimdallm/main.go:430** — Global 'pr_metadata' only includes reviewers and labels, but per-repo AI settings also expose pr_assignee and pr_draft. If the intent is to mirror per-repo shape at the global level, assignee/draft are missing from the global map (or the GlobalPRMetadata struct lacks those fields). Verify this is intentional and not a gap.

### Suggestions

- Before merging, run `git ls-files docs/superpowers/` on the PR branch to confirm the security audit plan doc was actually removed — the diff provided still shows it being added.
- Add a brief inline comment near the pr_metadata block referencing issue #144 so the naming inconsistency has a paper trail in the code.
- Consider whether global pr_metadata should also surface assignee/draft for symmetry with per-repo overrides, or document why only reviewers/labels are global.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR makes config output conditional for per-repo PR metadata fields and global pr_metadata, addressing prior feedback. The diff shows sensible guard conditions, but there are a few inconsistencies and one potentially stale change worth confirming.

### Issues

⚠️ **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — The PR discussion states 'Unrelated security audit plan | Removed from PR', but the diff still shows this 1040-line file being added. Either the diff is stale or the removal was not actually pushed. Verify the file is no longer present on the branch head; if it is, it should be removed as it is unrelated to this refactor.
🟡 **daemon/cmd/heimdallm/main.go:364** — `if ai.PRDraft { ro["pr_draft"] = ai.PRDraft }` only emits the field when true. This conflates 'not configured' with 'explicitly false'. If the upstream config struct has a sentinel or pointer type for PRDraft, use its presence rather than the boolean value; otherwise the /config response cannot reflect an explicit false setting. Low impact since false is the default, but inconsistent with the intent of surfacing configured values.
🟡 **daemon/cmd/heimdallm/main.go:430** — Structural inconsistency between per-repo and global: per-repo uses flat keys (`pr_reviewers`, `pr_assignee`, `pr_labels`, `pr_draft`) while global uses a nested object (`pr_metadata: {reviewers, labels}`). The PR title claims 'global IT/Develop settings match repo detail layout', yet the emitted shapes differ. Author acknowledged this will be addressed in #144, but the current title overstates what this PR actually achieves.
🟡 **daemon/cmd/heimdallm/main.go:430** — Global `pr_metadata` only exposes `reviewers` and `labels`, but per-repo exposes four fields including `pr_assignee` and `pr_draft`. If the global config struct supports assignee/draft they are silently dropped from the response; if it does not, the asymmetry with per-repo should be documented or unified. Worth confirming whether `c.AI.PRMetadata` has additional fields not surfaced here.

### Suggestions

- Verify the security-audit-fixes.md plan file is actually removed from the branch head — the diff in this review still shows it being added.
- Consider using a `*bool` for PRDraft (or checking a separate 'set' flag) so the config response can distinguish unset from explicit-false.
- Align naming between per-repo and global in follow-up #144: either nest per-repo under `pr_metadata` or flatten the global fields, so frontend consumers don't need two shapes.
- If `c.AI.PRMetadata` has more fields than reviewers/labels, surface them consistently with the per-repo section (or document why global is a subset).

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

The refactoring logic for conditional PR metadata emission is sound, but the PR includes a large unrelated security audit plan that exposes unfixed vulnerability details, and the global config doesn't fully match the per-repo layout despite that being the stated goal.

### Issues

🔴 **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — A 1040-line security audit implementation plan is included in this PR but is entirely unrelated to the stated purpose (matching global/repo settings layout). Worse, it details 8 specific unfixed security vulnerabilities (auth bypass, data races, missing credential blocks, unauthenticated endpoints) with exact file paths, line numbers, and exploitation steps. Publishing this before the fixes are merged is a security disclosure risk.
⚠️ **daemon/cmd/heimdallm/main.go:417** — Global pr_metadata only includes 'reviewers' and 'labels', but per-repo config (lines 355-367) also emits 'pr_assignee' and 'pr_draft'. The PR's stated goal is to match repo detail layout, yet the two representations diverge in both structure (flat keys vs nested object) and fields (assignee/draft missing globally).
🟡 **daemon/cmd/heimdallm/main.go:355** — Naming inconsistency: per-repo uses flat snake_case keys (pr_reviewers, pr_assignee, pr_labels, pr_draft) while global uses a nested pr_metadata object with shorter keys (reviewers, labels). API consumers must handle two different shapes for the same concept. Author acknowledges this is deferred to #144, but shipping divergent schemas now increases the migration burden.

### Suggestions

- Remove docs/superpowers/plans/2026-04-06-security-audit-fixes.md from this PR entirely — it should not be committed to the repo until the vulnerabilities it describes are fixed, and even then it belongs in a private issue tracker, not the source tree.
- Add the missing 'assignee' and 'draft' fields to the global pr_metadata block (guarded by the same non-empty/non-zero checks) so the global config actually matches the per-repo layout, fulfilling the PR's stated goal.
- Consider whether the naming inconsistency (flat pr_* keys vs nested pr_metadata) can be resolved in this PR rather than deferring to #144, since both representations are being introduced/modified here and aligning them now avoids shipping a schema consumers will need to migrate away from.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

PR adds conditional emission of per-repo PR metadata fields and introduces a nested pr_metadata object at the global config level. Code changes are clean but the diff includes an unrelated 1040-line security audit plan that exposes internal vulnerability details, and there is a naming/structural inconsistency between per-repo and global config representations.

### Issues

🔴 **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — A 1040-line security audit plan is included in this refactor PR. It details 8 specific security vulnerabilities with exact file paths, line numbers, and exploitation steps. This should not be committed to the repository — it exposes internal security findings and is unrelated to the PR's stated purpose. The PR discussion says it was removed, but it is still present in this diff.
⚠️ **daemon/cmd/heimdallm/main.go:417** — Naming/structural inconsistency: global config nests fields under 'pr_metadata' with keys 'reviewers'/'labels', while per-repo (line 355) uses flat keys 'pr_reviewers'/'pr_labels'/'pr_assignee'/'pr_draft'. API consumers must handle two different shapes for conceptually the same data. Acknowledged as deferred to #144, but worth tracking.
⚠️ **daemon/cmd/heimdallm/main.go:417** — Global pr_metadata only exposes 'reviewers' and 'labels', but per-repo config (lines 358-365) also exposes 'pr_assignee' and 'pr_draft'. If the goal is to 'match repo detail layout', the global config should either include these fields or document why they are intentionally omitted at the global level.
🟡 **.gitignore:23** — Adding .superpowers/ to .gitignore is fine, but verify that any .superpowers/ content that was previously tracked has been removed from git history (git rm --cached) and not just ignored going forward.

### Suggestions

- Remove the security audit plan file from this PR entirely — it contains sensitive vulnerability details (auth bypass, data races, credential exfiltration paths) that should not be in version control.
- Consider unifying the per-repo and global PR metadata structure now rather than deferring to #144, since shipping two different shapes creates an implicit API contract that clients must support indefinitely.
- Add a brief code comment or doc note explaining why pr_assignee and pr_draft are per-repo only (if intentional), so future contributors don't assume it's an oversight.
- Verify .superpowers/ was properly removed from tracked files with 'git rm -r --cached .superpowers/' before adding the gitignore entry.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Simple, correct refactor that makes the global config emission conditional on whether values are set, matching the per-repo pattern. The changes are minimal and low-risk. However, the diff still shows the large security-audit-fixes.md plan file being added (1040 lines) despite the PR comment claiming it was removed — this should be verified. Also, the `pr_draft` field is only emitted when true, so the API cannot distinguish 'unset' from 'explicitly false'.

### Issues

⚠️ **docs/superpowers/plans/2026-04-06-security-audit-fixes.md:1** — The author's PR comment states this unrelated security audit plan was removed, but the diff still shows it being added as a new 1040-line file. Verify the branch was actually pushed after removal; if still present, it should be dropped from this refactor PR since it is unrelated scope.
🟡 **daemon/cmd/heimdallm/main.go:364** — `if ai.PRDraft { ro["pr_draft"] = ai.PRDraft }` omits the field when false, so consumers cannot distinguish 'not configured' from 'explicitly false'. If the UI relies on this to render a tri-state or default indicator, it will never see an explicit false. If `pr_draft` is a plain bool with false as the natural default, this is acceptable, but for symmetry with the other conditional fields consider whether the pointer/optional semantics match the config model.
🟡 **daemon/cmd/heimdallm/main.go:430** — Naming mismatch: per-repo emits `pr_reviewers`/`pr_labels` while global emits `pr_metadata.reviewers`/`pr_metadata.labels`. Acknowledged by author as deferred to #144, but worth a TODO comment linking to that issue so the inconsistency is discoverable in-code.

### Suggestions

- Confirm the security-audit-fixes.md file was actually dropped from the branch — the provided diff still contains it.
- Add a brief inline TODO referencing #144 where per-repo vs global naming diverges, so the deferred cleanup is visible to future readers.
- Consider whether the conditional emit of `pr_draft` loses information (false vs unset) needed by the consuming UI; if the config type is a `bool` with false as implicit default, document that assumption.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Refactors global IT/Develop settings into a card-based layout with conditional field emission. Generally clean, but has a non-functional switch toggle bug and a fragile lazy-init pattern for new state fields.

### Issues

⚠️ **flutter_app/lib/features/config/config_screen.dart:397** — The 'Auto-implement issues' switch onChanged only handles the OFF case (`if (!v)`). When toggled ON, nothing happens—developLabels remains empty, so `hasLabels` stays false and the switch snaps back to OFF on rebuild. The toggle is effectively a no-op for enabling the feature.
⚠️ **flutter_app/lib/features/config/config_screen.dart:381** — Instance fields `_globalPRReviewers`, `_globalPRLabels`, and `_developInitialized` are lazily initialized during `build()` via `_initDevelopFromConfig`. Other state fields (e.g. `_issueTracking`) are likely initialized in `initState`/`didChangeDependencies`. This inconsistency is fragile—if the config is refreshed externally the lazy guard prevents the new values from being picked up, and the pattern diverges from the rest of the widget's state management.
🟡 **flutter_app/lib/core/models/config_model.dart:498** — `_parseStringList` uses `v.cast<String>()` which returns a lazy CastList. If the server ever sends a non-String element (e.g. an int), the error surfaces only when the element is accessed, far from the parse site. Using `List<String>.from(v)` or `v.whereType<String>().toList()` would fail eagerly or filter gracefully.
🟡 **flutter_app/lib/features/config/config_screen.dart:130** — `maxWidth: double.infinity` removes the previous 680px constraint entirely. On wide displays the form fields will stretch edge-to-edge, which is likely a UX regression for text inputs and dropdowns.
🟡 **flutter_app/pubspec.lock:399** — Transitive dependencies `matcher` (0.12.19→0.12.18) and `test_api` (0.7.10→0.7.9) were downgraded. This is likely a Flutter/Dart SDK version discrepancy between contributors and is harmless, but worth confirming it's intentional to avoid lock-file churn.

### Suggestions

- Fix the 'Auto-implement issues' switch: when toggled ON, either prompt the user to add a label, pre-populate with a sensible default (e.g. `['develop']`), or remove the switch and let the presence of labels alone control the feature state.
- Move `_globalPRReviewers`/`_globalPRLabels` initialization into the same lifecycle method (e.g. `didChangeDependencies` or `initState`) used by the other state fields so the pattern is consistent and the lazy-init guard can be removed.
- Replace `v.cast<String>()` in `_parseStringList` with `List<String>.from(v)` for eager validation at parse time.
- Consider keeping a reasonable `maxWidth` (e.g. 900–1000px) instead of `double.infinity` so the form remains readable on ultra-wide monitors.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR addresses previous review feedback well (conditional emission, .gitignore, removed unrelated files). The core backend changes are solid. However, the Flutter UI refactor introduces a couple of UX/state management issues in the new Develop section: the auto-implement toggle has broken bidirectional behavior, and local state initialization uses a one-shot flag that won't reflect config reloads. Also unexpected pubspec.lock downgrades.

### Issues

⚠️ **flutter_app/lib/features/config/config_screen.dart:472** — Auto-implement toggle is one-way broken: `value: hasLabels` is derived from `_issueTracking.developLabels.isNotEmpty`, but `onChanged` only acts when `v` is false (clears labels). Toggling OFF→ON has no effect because no labels are added — the switch just snaps back. Users cannot enable the feature via the switch; they must add labels via the chip field below, which makes the switch misleading UI. Consider either removing the switch or making it restore previous labels / toggling a separate enabled flag.
⚠️ **flutter_app/lib/features/config/config_screen.dart:455** — `_developInitialized` one-shot flag means `_globalPRReviewers`/`_globalPRLabels` are only synced from `AppConfig` on first build. If the config is refreshed/reloaded from the daemon while the screen is open (save-then-revert, external change, provider rebuild), the local state will silently drift from the persisted state. Consider using didUpdateWidget/listening to the provider, or keying off a config identity/hash.
🟡 **flutter_app/lib/features/config/config_screen.dart:472** — Turning the auto-implement switch OFF wipes `developLabels` with no confirmation and no way to restore if saved. Previous value is not cached. If this is intentional, at minimum add a subtitle warning or undo affordance; if not, stash labels before clearing so toggling back ON restores them.
🟡 **flutter_app/pubspec.lock:399** — `matcher` (0.12.19 → 0.12.18) and `test_api` (0.7.10 → 0.7.9) are downgraded without any pubspec.yaml constraint change in this diff. Unexpected lockfile regressions can cause CI/test drift for other contributors. Verify this was intentional (e.g., Flutter SDK pin) or re-run `flutter pub get` on a clean lock.
🟡 **flutter_app/lib/features/config/config_screen.dart:130** — `maxWidth: double.infinity` on the outer ConstrainedBox makes the constraint a no-op. If the intent is to remove the 680px cap, just drop the ConstrainedBox; otherwise set a reasonable max (e.g., 1200) so the form doesn't stretch edge-to-edge on wide displays and hurt readability.
🟡 **flutter_app/lib/core/models/config_model.dart:499** — `_parseStringList` uses `v.cast<String>()` which creates a lazy view that will throw on first non-string element at iteration time rather than here. For defensive parsing of JSON from the daemon, consider `v.whereType<String>().toList()` or explicit validation to avoid late failures.

### Suggestions

- Rename either global `pr_metadata.reviewers/labels` or per-repo `pr_reviewers/pr_labels` for consistency — the PR author notes this will be handled in #144, which is fine, but worth tracking explicitly.
- Consider extracting the repeated `AutocompleteChipField` + SizedBox(height:10) pattern into a small helper to reduce duplication across Issue Tracking and Develop sections.
- Add a widget test for the Develop section covering: toggle off clears labels, adding labels flips toggle on, and globalPRReviewers/Labels round-trip through save.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Refactors global IT/Develop settings with new PR metadata fields and UI card layout. The structural changes are sound, but there are a couple of behavioral concerns around conditional field emission and a non-functional toggle switch.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:367** — pr_draft is only emitted when true. This means a repo-level `pr_draft = false` can no longer explicitly override a global or consumer-side default of `true` — the field is simply absent, which most consumers will interpret as 'use default'. Previously the field was always emitted, preserving the explicit `false`.
⚠️ **flutter_app/lib/features/config/config_screen.dart:388** — The 'Auto-implement issues' SwitchListTile's onChanged only handles the `!v` (turn-off) case by clearing developLabels. When toggled ON (v == true), no code executes — setState is never called, so the switch visually snaps back to off. The toggle is effectively one-way: it can disable the feature but never enable it. Either disable the switch when off (making it read-only, driven by the chip field), or seed a placeholder label when toggled on.
🟡 **flutter_app/lib/features/config/config_screen.dart:375** — State fields `_globalPRReviewers`, `_globalPRLabels`, and `_developInitialized` are declared between widget methods rather than at the top of the State class alongside the other state fields. This hurts readability and breaks the class's existing convention.
🟡 **flutter_app/lib/core/models/config_model.dart:499** — `_parseStringList` uses `v.cast<String>()` which returns a lazy CastList — elements are only cast (and potentially throw TypeError) when accessed, not at parse time. If the daemon ever returns a non-string element in the list, the error surfaces far from the parse site. Using `v.whereType<String>().toList()` or `List<String>.from(v)` would fail fast or filter gracefully.
🟡 **flutter_app/lib/features/config/config_screen.dart:130** — Changing maxWidth from 680 to double.infinity removes the form's readable line-length cap. On wide/ultrawide monitors, form fields will stretch edge-to-edge, degrading readability. If the goal is to match repo detail layout, consider using that layout's specific maxWidth rather than unbounded.

### Suggestions

- For pr_draft: always emit the field in the repo override map (restore the original unconditional assignment) so that explicit `false` values are preserved, or document that repo-level pr_draft=false is intentionally unsupported.
- For the Develop toggle: either make the switch disabled/non-interactive when developLabels is empty (so it's purely a status indicator driven by the chip field), or add a no-op setState call and show an inline hint like 'Add at least one develop label to enable'.
- Move _globalPRReviewers, _globalPRLabels, and _developInitialized declarations to the top of _ConfigScreenState alongside the existing state fields for consistency.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Previous main finding (conditional emission of per-repo PR fields and global pr_metadata) is addressed. However, the PRDraft 'unset vs explicit false' semantics issue remains because pr_draft is only written when true — an explicit per-repo override of false is now silently dropped. The new Flutter Develop section has a UX bug (Auto-implement switch cannot be turned on) and an initialization pattern that won't react to config reloads.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:366** — pr_draft is only emitted when ai.PRDraft is true. If the global default is true and a per-repo override explicitly sets pr_draft=false, this save will drop the override and the repo will silently inherit true on next load. Same omitempty semantics limitation as flagged previously — needs *bool (or a sentinel) to round-trip 'explicitly false' overrides. Consider also whether pr_assignee="" should be usable to explicitly clear a global assignee.
⚠️ **flutter_app/lib/features/config/config_screen.dart:422** — The 'Auto-implement issues' SwitchListTile is bound to value=hasLabels (derived from developLabels.isNotEmpty). onChanged only handles the v=false case by clearing labels; when the user toggles it ON, no state is mutated, developLabels stays empty, so hasLabels stays false and the switch snaps back to OFF. The switch cannot actually be enabled by the user — functionality to enable auto-implement is only reachable by adding labels directly.
⚠️ **flutter_app/lib/features/config/config_screen.dart:410** — _initDevelopFromConfig uses a one-shot _developInitialized guard. If the config is reloaded from the daemon (e.g., external change, provider refresh) after the Develop section has rendered once, _globalPRReviewers/_globalPRLabels will remain stale and the edited/old values will be saved back on next submit, potentially clobbering remote changes. Consider using didUpdateWidget / listening to the provider, or reinitializing when the config identity changes — matches the pattern likely used by the other sections for _issueTracking.
🟡 **flutter_app/lib/features/config/config_screen.dart:129** — maxWidth changed from 680 to double.infinity. On wide displays (desktop/web) the form now stretches across the full viewport, which typically hurts readability of form controls. If the intent was just to give the new card-based layout more room, consider a larger fixed max (e.g. 960–1100) rather than unbounded.
🟡 **daemon/cmd/heimdallm/main.go:431** — Still no test coverage for the save/load round-trip. Given the original regression class (zero-value overrides silently persisting), a test that loads a config with only globals set, saves it, reloads, and asserts no per-repo override was synthesized would guard against future re-introductions. Low severity but repeats previous review.

### Suggestions

- Make PRDraft a *bool (or introduce a dedicated override-set bitmap) so 'unset' and 'explicitly false' are distinguishable end-to-end — without this the conditional emission is a half-fix.
- In _developSection, have the switch onChanged(true) seed a sensible default (e.g. focus the labels field or add a placeholder) or drive the switch from a separate explicit boolean rather than deriving from hasLabels, so it can actually be toggled on.
- Drop the _developInitialized latch and instead key initialization off the incoming config's identity/hash, or lift _globalPRReviewers/_globalPRLabels into the same initialization path used for _issueTracking so external reloads are reflected.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Refactor moves develop/PR-metadata settings into their own UI section and conditionally emits per-repo PR fields. Several functional issues: a toggle that can only turn off, a fragile cast in parsing, and unexplained dependency downgrades.

### Issues

🔴 **flutter_app/lib/features/config/config_screen.dart:389** — The 'Auto-implement issues' SwitchListTile only handles the OFF case (`if (!v)`). Toggling ON is a no-op because no labels are added, so `hasLabels` stays false and the switch immediately appears off again. The toggle is effectively one-way — users can disable but never enable via the switch.
⚠️ **flutter_app/lib/core/models/config_model.dart:498** — `v.cast<String>()` returns a lazy CastList that throws a TypeError at access time if any element is not a String. Use `List<String>.from(v)` or `v.whereType<String>().toList()` for fail-fast or graceful behavior respectively.
⚠️ **flutter_app/lib/features/config/config_screen.dart:382** — State fields `_globalPRReviewers`, `_globalPRLabels`, and `_developInitialized` are declared between methods instead of at the top of the State class with other fields. More importantly, `_initDevelopFromConfig` is called during build via `_developSection` — if the upstream config changes (e.g. hot-reload or provider refresh), the local state will be stale because `_developInitialized` prevents re-initialization. Consider initializing in `initState` or `didChangeDependencies`.
🟡 **flutter_app/lib/features/config/config_screen.dart:130** — `maxWidth: double.infinity` removes the content width constraint entirely. On wide displays the form fields will stretch across the full viewport width, hurting readability. Consider a larger but bounded value (e.g. 960) or a responsive breakpoint.
🟡 **daemon/cmd/heimdallm/main.go:367** — `pr_draft` is now only emitted when true. Previously it was always present. Clients that distinguish 'key absent' from 'key = false' will see a behavior change. Verify downstream consumers handle a missing `pr_draft` as false.
🟡 **flutter_app/pubspec.lock:399** — Transitive dependencies `matcher` (0.12.19 → 0.12.18) and `test_api` (0.7.10 → 0.7.9) were downgraded. If unintentional, re-run `flutter pub upgrade` to pick up latest compatible versions.

### Suggestions

- Fix the Auto-implement toggle: either add a default develop label when toggled on (e.g. `['develop']`), or replace the toggle with a helper text explaining that adding labels enables the feature.
- Replace `v.cast<String>()` with `List<String>.from(v)` in `_parseStringList` to fail fast on bad data instead of deferring the error to list access.
- Move `_globalPRReviewers`/`_globalPRLabels` initialization into `initState` or use `didChangeDependencies` with the provider, and group the field declarations at the top of the State class with the other fields.
- Set a reasonable `maxWidth` (e.g. 960px) instead of `double.infinity` to keep form readability on ultra-wide screens.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

The PR achieves its stated goal of restructuring the global IT/Develop settings to match the repo-detail card layout, and the conditional field emission in the Go serializer is a clean improvement. However, the new 'Auto-implement issues' switch has broken semantics, state initialization runs inside build(), and pubspec.lock contains unrelated dependency downgrades.

### Issues

🔴 **flutter_app/lib/features/config/config_screen.dart:428** — The 'Auto-implement issues' SwitchListTile derives its value from `hasLabels = _issueTracking.developLabels.isNotEmpty`, but the onChanged handler only acts when `v` is false (clearing labels). Toggling the switch ON is a no-op — hasLabels remains false, so the switch immediately snaps back to off. Users have no way to enable the feature through the switch; they must know to edit the chip field directly. Either make the switch a real enabled flag stored in state, or hide/disable it and rely solely on the chip field.
⚠️ **flutter_app/lib/features/config/config_screen.dart:410** — `_initDevelopFromConfig` is called from within `_developSection` (build path) and guarded by a `_developInitialized` flag. This is an anti-pattern: if the AppConfig is reloaded/updated after first build (e.g. daemon push, refresh), the local `_globalPRReviewers`/`_globalPRLabels` state will never re-sync with the new config. Initialization should happen in `initState` / `didChangeDependencies` / `didUpdateWidget`, or should react to provider changes via `ref.listen`.
⚠️ **flutter_app/pubspec.lock:399** — `matcher` is downgraded 0.12.19 → 0.12.18 and `test_api` 0.7.10 → 0.7.9. These transitive downgrades are unrelated to the feature and look like an accidental lockfile regression from a stale local environment. Verify they are intended; otherwise re-run `flutter pub upgrade` / `pub get` on a clean checkout before merging.
🟡 **flutter_app/lib/core/models/config_model.dart:499** — `_parseStringList` uses `v.cast<String>()` which returns a lazy casting view and will throw at first access if the JSON contains a non-string element. Prefer `v.whereType<String>().toList()` or `List<String>.from(v.map((e) => e.toString()))` to make parsing robust to malformed daemon responses.
🟡 **flutter_app/lib/features/config/config_screen.dart:129** — `maxWidth: double.infinity` removes the previous 680px constraint for the General tab. On wide desktop windows form fields and chip rows will stretch edge-to-edge, which typically hurts readability. If the goal was to widen the layout, consider a larger fixed max (e.g. 1100–1200) rather than unbounded.
🟡 **flutter_app/lib/features/config/config_screen.dart:475** — `_globalPRReviewers` and `_globalPRLabels` are declared mid-class between the two section builder methods. Minor style issue: field declarations should live near the other state fields at the top of the State class for discoverability.
🟡 **daemon/cmd/heimdallm/main.go:431** — The global `pr_metadata` key is only emitted when reviewers OR labels are non-empty, but individual sub-keys are also guarded — so an AI.PRMetadata with only reviewers produces `pr_metadata: {reviewers: [...]}` with no `labels` key. The Flutter parser handles this fine, but be aware this is an intentional asymmetry with the persisted TOML shape written by `first_run_setup.dart` (which mirrors the same conditional logic). Not a bug, but worth an integration test to confirm round-trip.

### Suggestions

- Fix the 'Auto-implement issues' switch: either store an explicit `developEnabled` bool in state and allow enabling without labels, or remove the switch entirely and let the chip field govern the feature (empty = disabled).
- Move `_initDevelopFromConfig` out of the build path. A `ref.listen<AsyncValue<AppConfig>>` call in `initState` or syncing in `didChangeDependencies` will keep local state aligned when config changes.
- Revert the unintended `matcher` / `test_api` downgrades in pubspec.lock, or confirm they are required.
- Add a widget/integration test for the round-trip: configure global PR reviewers + labels → save → reload → verify chip fields are populated from the daemon response.
- Replace `v.cast<String>()` in `_parseStringList` with `v.whereType<String>().toList()` for defensive parsing.

---

PR #149, opened 2026-04-21
fix: add SSH client and agent forwarding for auto_implement

### LINKED ISSUE #143: bug: auto_implement fails — git not installed in Docker image
## Problem

auto_implement fails with:

```
gitops: fetch origin/main: exec: "git": executable file not found in $PATH
```

The Docker image bundles Node.js + 4 AI CLIs but does not include `git`. The auto_implement pipeline uses `gitops.go` which shells out to `git` for checkout, commit, and push.

## Fix

Add `git` to the runtime stage of `docker/Dockerfile`:

```dockerfile
RUN apk add --no-cache git
```

~5MB addition to the image. Alpine's git package includes everything needed.

## Impact

Without this fix, auto_implement is completely non-functional in Docker deployments. All issues degrade to review_only.

### LINKED ISSUE #148: bug: auto_implement git fetch fails — SSH client missing in Docker image
## Problem

auto_implement fails at the git fetch step:

```
gitops: fetch origin/main: exit status 128 (stderr: error: cannot run ssh: No such file or directory)
```

The Docker image has `git` but no SSH client. Repos mounted from the host have SSH remotes (`git@github.com:...`).

## Fix

Three changes:

### 1. Dockerfile — add openssh-client + pre-seed GitHub host key

```dockerfile
RUN apk add --no-cache ca-certificates tzdata bash curl coreutils git openssh-client
```

Pre-seed GitHub's SSH host key to avoid interactive prompt:

```dockerfile
RUN mkdir -p /home/heimdallm/.ssh && \
    ssh-keyscan github.com >> /home/heimdallm/.ssh/known_hosts 2>/dev/null && \
    chmod 700 /home/heimdallm/.ssh && \
    chmod 644 /home/heimdallm/.ssh/known_hosts && \
    chown -R heimdallm:heimdallm /home/heimdallm/.ssh
```

### 2. docker-compose.yml — forward SSH agent from host

The host's SSH agent has the decrypted keys in memory. Forward the socket so the container can authenticate without touching the private key:

**macOS (Docker Desktop):**
```yaml
volumes:
  - /run/host-services/ssh-auth.sock:/ssh-agent:ro
environment:
  - SSH_AUTH_SOCK=/ssh-agent
```

**Linux:**
```yaml
volumes:
  - ${SSH_AUTH_SOCK}:/ssh-agent:ro
environment:
  - SSH_AUTH_SOCK=/ssh-agent
```

### 3. Deploy config — mount SSH agent

Update `heimdallr-deploy/docker-compose.yml` with the SSH agent forward.

## Why SSH agent, not just mounting ~/.ssh?

If the private key has a passphrase (common and recommended), mounting `~/.ssh` alone doesn't work — git would prompt for the passphrase interactively, which fails in a headless Docker container. The SSH agent already has the key decrypted in memory, so forwarding the socket avoids this entirely.

## Impact

Without this fix, auto_implement cannot git fetch/push on repos cloned via SSH (the default for `gh repo clone`). All auto_implement attempts fail.

sergiotejon:
Fixes #143, Fixes #148

## Problem

auto_implement fails at git fetch because:
1. No SSH client in the Docker image (fixed in this PR)
2. Repos use SSH remotes and need the host's SSH agent for passphrase-protected keys

### Dockerfile
- Added `openssh-client` to runtime packages
- Pre-seed GitHub host key (`ssh-keyscan github.com`) to avoid interactive prompt

### docker-compose.yml
- Forward host SSH agent socket into container
- `SSH_AUTH_SOCK=/ssh-agent` env var
- macOS: `/run/host-services/ssh-auth.sock` (Docker Desktop)
- Linux: configurable via `HEIMDALLM_SSH_AUTH_SOCK`

## Test plan

- [ ] `docker exec heimdallm ssh -T git@github.com` returns "Hi username!"
- [ ] auto_implement successfully fetches and pushes via SSH
- [ ] review_only mode unaffected


Muriano:
## ⚠️ MEDIUM Issue

ssh-keyscan at build time bakes GitHub's host key into the image layer. If GitHub rotates its SSH host key (as happened in March 2023), all existing images become stale and SSH connections will fail or, worse, users may be tempted to disable strict host checking. Consider fetching the key at container startup instead, or pinning the expected key fingerprint for verification.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## ⚠️ MEDIUM Issue

stderr is silenced with 2>/dev/null on ssh-keyscan. If the keyscan fails (network issue during build), the known_hosts file will be empty and the build will still succeed silently, leading to runtime SSH failures that are hard to diagnose. Consider adding a validation step (e.g., `test -s /home/heimdallm/.ssh/known_hosts`).

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## 🟡 LOW Issue

The default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. On Linux the host SSH_AUTH_SOCK path varies (e.g., `/tmp/ssh-XXX/agent.PID`). The comment says to set HEIMDALLM_SSH_AUTH_SOCK in docker/.env but there is no .env.example or documentation update showing Linux users what to set, making this easy to miss.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano (approved):

Adds SSH client and agent forwarding to enable git operations over SSH inside the container. The approach is sound but has a security concern with ssh-keyscan at build time and a portability gap for Linux users.

### Suggestions

- Add a build-time check after ssh-keyscan to verify known_hosts is non-empty: `&& test -s /home/heimdallm/.ssh/known_hosts || (echo 'ssh-keyscan failed' && exit 1)`
- Add a `.env.example` entry or update docs showing Linux users how to set HEIMDALLM_SSH_AUTH_SOCK to their $SSH_AUTH_SOCK path
- Consider running ssh-keyscan at container startup (entrypoint) instead of build time to avoid stale host keys baked into the image layer

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Adds SSH client and agent forwarding to enable git operations over SSH inside the container. The approach is sound but has a security concern with ssh-keyscan at build time and a portability gap for Linux users.

### Issues

⚠️ **docker/Dockerfile:57** — ssh-keyscan at build time bakes GitHub's host key into the image layer. If GitHub rotates its SSH host key (as happened in March 2023), all existing images become stale and SSH connections will fail or, worse, users may be tempted to disable strict host checking. Consider fetching the key at container startup instead, or pinning the expected key fingerprint for verification.
⚠️ **docker/Dockerfile:57** — stderr is silenced with 2>/dev/null on ssh-keyscan. If the keyscan fails (network issue during build), the known_hosts file will be empty and the build will still succeed silently, leading to runtime SSH failures that are hard to diagnose. Consider adding a validation step (e.g., `test -s /home/heimdallm/.ssh/known_hosts`).
🟡 **docker/docker-compose.yml:105** — The default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. On Linux the host SSH_AUTH_SOCK path varies (e.g., `/tmp/ssh-XXX/agent.PID`). The comment says to set HEIMDALLM_SSH_AUTH_SOCK in docker/.env but there is no .env.example or documentation update showing Linux users what to set, making this easy to miss.
🟡 **docker/docker-compose.yml:105** — The SSH agent socket is mounted `:ro` which is correct for security, but if the socket file does not exist on the host (e.g., no SSH agent running), docker-compose will create an empty directory at that path inside the container, which can cause confusing errors instead of a clear failure message.

---

Muriano:
## 🟡 LOW Issue

The SSH agent socket is mounted `:ro` which is correct for security, but if the socket file does not exist on the host (e.g., no SSH agent running), docker-compose will create an empty directory at that path inside the container, which can cause confusing errors instead of a clear failure message.

**Location:** `docker/docker-compose.yml` line 105

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

The PR correctly wires SSH agent forwarding and git support into the container for auto_implement. However, there are several build-time robustness and portability concerns, most already flagged by prior review. A few additional issues worth calling out: ssh-keyscan doesn't pin key types, only github.com is scanned (other Git hosts will fail), and there is no security note about the blast radius of agent forwarding.

### Issues

⚠️ **docker/Dockerfile:57** — ssh-keyscan bakes GitHub's host key into the image layer. If GitHub rotates its host key, all existing images break. Fetch at container startup or pin/verify the expected fingerprint against a known value rather than trusting whatever the network returns at build time.
⚠️ **docker/Dockerfile:57** — `2>/dev/null` silences ssh-keyscan errors, so a transient network failure during build yields an empty known_hosts file and the build still succeeds. Add `test -s /home/heimdallm/.ssh/known_hosts` after the keyscan to fail fast.
🟡 **docker/Dockerfile:57** — ssh-keyscan is invoked without `-t` so it relies on the default key-type set. Pin explicitly (e.g. `ssh-keyscan -t rsa,ecdsa,ed25519 github.com`) to make the set of trusted keys deterministic across image rebuilds and across ssh-keyscan versions.
🟡 **docker/Dockerfile:57** — Only github.com is seeded in known_hosts. If auto_implement ever clones or pushes to GitLab, Bitbucket, or a self-hosted Git server, SSH will prompt/fail. Either document this GitHub-only constraint or accept a build-arg list of hosts to seed.
🟡 **docker/docker-compose.yml:105** — Default socket path `/run/host-services/ssh-auth.sock` is Docker Desktop (macOS)-specific. Linux users' SSH_AUTH_SOCK typically lives under `/tmp/ssh-XXX/agent.PID`. There is no `.env.example` or README update explaining that Linux users must export `HEIMDALLM_SSH_AUTH_SOCK=$SSH_AUTH_SOCK`, so this is easy to miss.
🟡 **docker/docker-compose.yml:105** — If the host agent socket does not exist (no agent running, or wrong path), Docker creates an empty directory at the mount point, producing confusing `Error connecting to agent: No such file or directory` errors at runtime. Consider documenting how to disable this when no agent is present, or gate the bind-mount behind a Compose profile.
🟡 **docker/docker-compose.yml:62** — SSH agent forwarding gives the container full access to every identity loaded in the host agent for as long as the container is running. There is no comment acknowledging this trust boundary or recommending a dedicated, auto_implement-scoped key. Worth adding a security note next to the volume mount.
🟡 **docker/Dockerfile:16** — `git` and `openssh-client` are installed unconditionally, so every deployment — including ones that never enable auto_implement — pays the image-size and attack-surface cost. Acceptable trade-off, but worth being intentional about.

### Suggestions

- Replace `ssh-keyscan github.com >> ... 2>/dev/null` with `ssh-keyscan -t rsa,ecdsa,ed25519 github.com > /tmp/known_hosts && test -s /tmp/known_hosts && mv /tmp/known_hosts /home/heimdallm/.ssh/known_hosts` to fail the build on empty output.
- Add a `docker/.env.example` documenting `HEIMDALLM_SSH_AUTH_SOCK` and `HEIMDALLM_REPOS_DIR` with OS-specific guidance (macOS default vs. `$SSH_AUTH_SOCK` on Linux).
- Consider moving the SSH agent mount into a Compose `profiles:` section (e.g. `auto_implement`) so installations that don't need it don't fail when no agent is running on the host.
- Add a short SECURITY note in the Dockerfile and/or README explaining the implications of agent forwarding and recommending a dedicated key with limited repo scope.

---

Muriano:
## ⚠️ MEDIUM Issue

ssh-keyscan at build time bakes GitHub's host key into the image. If GitHub rotates keys (as in March 2023), all built images become stale. Consider running ssh-keyscan at container startup via an entrypoint script instead.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## ⚠️ MEDIUM Issue

stderr suppressed with 2>/dev/null on ssh-keyscan. If keyscan fails (e.g., network issue during build), known_hosts will be empty and the build succeeds silently, causing hard-to-diagnose runtime SSH failures. Add a validation step: `test -s /home/heimdallm/.ssh/known_hosts || exit 1`.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## 🟡 LOW Issue

Default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. Linux users need to set HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example entry or documentation update guiding them.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano (approved):

Adds SSH client and agent forwarding to enable git operations over SSH in the container. The approach is functional but has build-time fragility (baked host keys, silent failures) and incomplete cross-platform documentation.

### Suggestions

- Move ssh-keyscan to an entrypoint script so host keys are fetched at startup rather than baked into the image layer, avoiding staleness after key rotation.
- Add `&& test -s /home/heimdallm/.ssh/known_hosts` after ssh-keyscan to fail the build if no keys were retrieved.
- Add a HEIMDALLM_SSH_AUTH_SOCK entry to docker/.env.example with a comment showing the typical Linux path pattern.
- Consider adding an entrypoint check that validates SSH_AUTH_SOCK is a socket (e.g., `test -S $SSH_AUTH_SOCK`) and logs a clear warning if not, so users get immediate feedback when the agent is not forwarded.

---

Muriano:
## 🟡 LOW Issue

If the SSH agent socket does not exist on the host (no agent running), Docker will silently create an empty directory at the mount point, causing confusing runtime errors instead of a clear failure.

**Location:** `docker/docker-compose.yml` line 105

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR adds SSH client and agent forwarding to enable git operations in auto_implement. The change is small, focused, and correctly scoped (read-only socket mount, non-root user ownership, build-time host key seeding). However, build-time ssh-keyscan creates staleness and silent-failure risks, and the macOS-specific default socket path lacks Linux-user guidance.

### Issues

⚠️ **docker/Dockerfile:57** — ssh-keyscan runs at image build time with stderr silenced (2>/dev/null). If the network call fails during build, known_hosts is created empty and the build succeeds silently, producing images that fail SSH at runtime with opaque errors. Add a validation step such as `test -s /home/heimdallm/.ssh/known_hosts` after the keyscan to fail the build fast.
⚠️ **docker/Dockerfile:57** — Baking GitHub's SSH host key into an image layer creates staleness risk. When GitHub rotates host keys (as in March 2023), all previously built images break and users may be tempted to disable StrictHostKeyChecking. Prefer running ssh-keyscan at container startup via an entrypoint, or pin/verify an expected fingerprint so rotation failures are explicit rather than silent.
🟡 **docker/docker-compose.yml:105** — The default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. Linux hosts use ephemeral paths like `/tmp/ssh-XXXXXX/agent.PID` or `$XDG_RUNTIME_DIR/ssh-agent.socket`. The inline comment mentions setting HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example, README, or docs update guiding Linux users, so the feature will silently misbehave for them.
🟡 **docker/docker-compose.yml:105** — If the host path does not exist (no SSH agent running, or wrong path on Linux), Docker bind-mounts an empty directory at `/ssh-agent` inside the container. Git/SSH will then fail with confusing 'could not open a connection to your authentication agent' errors. Consider documenting this failure mode, or using a compose profile / preflight check that validates the socket exists before bringing the service up.
🟡 **docker/docker-compose.yml:60** — SSH_AUTH_SOCK is always exported to the container, even when the user hasn't opted into auto_implement / SSH forwarding. If the socket mount resolves to a non-socket (empty dir fallback), tools inside the container will still try to use `/ssh-agent` and fail. Gating the env var behind the same opt-in mechanism as the volume (or leaving it unset when unused) would make the default experience cleaner.
🟡 **docker/Dockerfile:16** — Adding `openssh-client` to the runtime image is correct, but the image does not configure a global `known_hosts` (e.g., in /etc/ssh/ssh_known_hosts) nor does it document which hosts beyond github.com are supported. If auto_implement ever targets GitLab, Bitbucket, or self-hosted Gitea, users will hit host-key prompts. Worth noting as a follow-up or extending the keyscan list.

### Suggestions

- Add `&& test -s /home/heimdallm/.ssh/known_hosts` to the RUN step so an empty known_hosts fails the build.
- Remove `2>/dev/null` from ssh-keyscan so diagnostic output is preserved in build logs; alternatively redirect stderr only on the success path.
- Consider moving the ssh-keyscan into an entrypoint script that runs once on container start, so host-key rotation is picked up without rebuilding images.
- Add a `docker/.env.example` entry documenting HEIMDALLM_SSH_AUTH_SOCK with the typical Linux value (e.g. `$SSH_AUTH_SOCK`) and the macOS default.
- Document in the README or auto_implement feature docs that SSH agent forwarding requires a running agent on the host and show how to verify (`ssh-add -l`).
- If auto_implement will support non-GitHub remotes, extend the keyscan list (gitlab.com, bitbucket.org, ...) or make it configurable via a build arg.

---

Muriano:
## ⚠️ MEDIUM Issue

ssh-keyscan at build time bakes GitHub's host key into the image layer. If GitHub rotates its SSH host key (as happened March 2023), all existing images silently break. Move ssh-keyscan to an entrypoint script so the key is fetched at container startup.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## ⚠️ MEDIUM Issue

stderr is suppressed with 2>/dev/null on ssh-keyscan. If keyscan fails during build (e.g., network issue), known_hosts will be empty and the build succeeds silently, causing hard-to-diagnose runtime SSH failures. Add a validation: `test -s /home/heimdallm/.ssh/known_hosts || exit 1`.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## 🟡 LOW Issue

Default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. Linux users need HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example entry or documentation guiding them. Add a commented entry to .env.example.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano:
## 🟡 LOW Issue

If the SSH agent socket does not exist on the host (no agent running), Docker will silently create an empty directory at the mount point, causing confusing runtime errors. Consider adding a startup check in the entrypoint that warns when SSH_AUTH_SOCK is not a valid socket.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano (approved):

The PR correctly adds SSH client and agent forwarding to enable git operations in auto_implement. The approach is sound but has a few robustness gaps around build-time host key baking and silent failure modes that the existing review comments have already identified well.

### Suggestions

- Move ssh-keyscan to an entrypoint script that runs at container startup instead of build time, so host keys are always fresh and network failures are visible at runtime rather than silently ignored during build.
- Add `test -s /home/heimdallm/.ssh/known_hosts || exit 1` after ssh-keyscan if you keep it at build time, to fail fast on empty known_hosts.
- Add a HEIMDALLM_SSH_AUTH_SOCK example entry to .env.example (or create one) with a comment explaining the Linux vs macOS difference.
- Add an entrypoint check like `[ -S "$SSH_AUTH_SOCK" ] || echo 'WARNING: SSH agent socket not found'` so users get a clear message instead of cryptic SSH errors.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

This PR is now focused and scoped correctly: it adds openssh-client to the runtime image, pre-seeds GitHub's host key in known_hosts, and forwards the host SSH agent socket into the container so auto_implement can push/fetch over SSH. The previous scope-mismatch and committed-artifact findings no longer apply to this diff. Remaining concerns are operational robustness around ssh-keyscan and the forwarded socket path.

### Issues

⚠️ **docker/Dockerfile:57** — ssh-keyscan runs at build time, so GitHub's host key is baked into the image layer. If GitHub rotates its SSH host key (as in March 2023), every previously built image silently breaks with host-key verification failures that are easy to misdiagnose as credential issues. Prefer running ssh-keyscan in an entrypoint script at container startup, or pin the known SHA256 fingerprint and verify after keyscan.
⚠️ **docker/Dockerfile:57** — stderr of ssh-keyscan is discarded with 2>/dev/null and the exit status is ignored (chained with &&, but ssh-keyscan exits 0 even on network failure as long as DNS/connect don't error in a way that aborts). A transient network issue during build produces an empty known_hosts; the build succeeds and SSH fails only at runtime. Add a post-step validation like `test -s /home/heimdallm/.ssh/known_hosts` to fail the build loudly.
🟡 **docker/Dockerfile:57** — Relying purely on TOFU via ssh-keyscan means the image trusts whatever key the build network returns. For a security-oriented project, consider also pinning GitHub's published SSH host key fingerprints (ECDSA/Ed25519/RSA) and verifying the keyscan output against them before writing known_hosts.
🟡 **docker/docker-compose.yml:104** — Default SSH_AUTH_SOCK path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. Linux hosts use variable paths like `/tmp/ssh-XXXXXX/agent.NNN` (or the systemd user path) and must override HEIMDALLM_SSH_AUTH_SOCK, but there is no .env.example entry, README note, or make target documenting this. Linux contributors will hit opaque failures. Add an example entry and a short note in the auto_implement docs.
🟡 **docker/docker-compose.yml:104** — If the host path referenced by HEIMDALLM_SSH_AUTH_SOCK does not exist (no agent running, or wrong path on Linux), Docker bind-mount will create an empty directory at `/ssh-agent` inside the container. Git will then fail with misleading errors like `Could not open a connection to your authentication agent`. Add an entrypoint check that verifies `-S $SSH_AUTH_SOCK` and logs a clear warning when auto_implement features are used without a working agent.
🟡 **docker/docker-compose.yml:60** — SSH_AUTH_SOCK is exported unconditionally to all processes in the container even when no agent is forwarded. Tools that probe for an agent may try to talk to an empty/invalid socket and produce confusing errors. Consider only setting SSH_AUTH_SOCK when a real socket is mounted, or documenting the intended behaviour.
⚠️ **docker/docker-compose.yml:104** — Forwarding the host SSH agent into the container grants any process inside the container the ability to authenticate as the host user to any SSH endpoint (not just GitHub). The agent socket is mounted :ro, which prevents adding keys but does not restrict signing requests. Given the container also runs third-party AI CLIs, document this trust boundary in the PR/README and consider constraining via a dedicated deploy key or `ssh-agent -c`/`IdentityAgent` with confirmation, at least for production use.

### Suggestions

- Move ssh-keyscan into an entrypoint script (or a small wrapper) so the host key is refreshed on each container start, and fail fast if the resulting known_hosts is empty.
- Add a `.env.example` with a commented `HEIMDALLM_SSH_AUTH_SOCK=` entry and a short note in the auto_implement docs explaining the macOS vs Linux paths.
- In the entrypoint, validate `[ -S "$SSH_AUTH_SOCK" ]` when auto_implement is enabled and emit a clear error/warning otherwise, instead of failing later inside git.
- Call out the SSH-agent forwarding trust implication in the PR description/README: any code running in the container can use the host's keys for the lifetime of the container.

---

Muriano:
## ⚠️ MEDIUM Issue

ssh-keyscan at build time bakes GitHub's host key into the image layer. If GitHub rotates its SSH host key (as happened March 2023), all existing images silently break. Move ssh-keyscan to an entrypoint script so the key is fetched at container startup.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## ⚠️ MEDIUM Issue

stderr is suppressed with 2>/dev/null on ssh-keyscan. If keyscan fails during build (network issue), known_hosts will be empty and the build succeeds silently, causing hard-to-diagnose runtime SSH failures. Add validation: `test -s /home/heimdallm/.ssh/known_hosts || exit 1`.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## 🟡 LOW Issue

Default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. Linux users need HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example entry or documentation guiding them.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano:
## 🟡 LOW Issue

If the SSH agent socket does not exist on the host (no agent running), Docker will silently create an empty directory at the mount point, causing confusing runtime errors instead of a clear failure message.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano (approved):

SSH client and agent forwarding setup is functional but has operational risks: baked-in host keys can go stale, silent keyscan failures lead to hard-to-diagnose runtime errors, and Linux users lack guidance for SSH_AUTH_SOCK configuration.

### Suggestions

- Move ssh-keyscan to an entrypoint script that runs at container startup, so host keys are always fresh and network failures are caught at runtime rather than silently ignored at build time.
- Add a validation step after ssh-keyscan: `test -s /home/heimdallm/.ssh/known_hosts || exit 1` to fail the build if known_hosts is empty.
- Add a commented HEIMDALLM_SSH_AUTH_SOCK entry to .env.example with instructions for Linux users.
- Add a startup check in the entrypoint that warns or fails gracefully when SSH_AUTH_SOCK is not a valid socket.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Adds SSH client and agent forwarding to enable auto_implement git operations. The approach is sound but has a few real concerns around build-time host key baking, silent failures from stderr suppression, missing Linux documentation for the host socket path, and no graceful handling when the agent socket is absent.

### Issues

⚠️ **docker/Dockerfile:57** — ssh-keyscan is executed at build time and the resulting host key is baked into the image layer. If GitHub rotates its SSH host key (as occurred in March 2023), every built image silently breaks. Prefer running ssh-keyscan from an entrypoint script at container startup, or pin/verify an expected fingerprint.
⚠️ **docker/Dockerfile:57** — stderr is suppressed with `2>/dev/null` on ssh-keyscan. If the command fails (e.g., transient network issue during build), known_hosts ends up empty and the build still succeeds, yielding hard-to-diagnose SSH failures at runtime. Add a validation step such as `test -s /home/heimdallm/.ssh/known_hosts` to fail fast.
🟡 **docker/docker-compose.yml:102** — Default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. Linux users must override HEIMDALLM_SSH_AUTH_SOCK but there is no `.env.example` entry, README update, or inline guidance showing the typical Linux value. Add a documented entry to `.env.example` and/or the docs.
🟡 **docker/docker-compose.yml:102** — If the host SSH agent socket does not exist (no agent running), Docker silently creates an empty directory at the mount point, leading to confusing runtime errors rather than a clear failure. Consider an entrypoint-time check that validates SSH_AUTH_SOCK points to a real socket and logs a clear warning/error otherwise.
🟡 **docker/docker-compose.yml:60** — SSH_AUTH_SOCK is set unconditionally inside the container, even when no agent is actually forwarded. Tools like git will attempt to use the non-functional socket and fail with opaque errors. Gating the env var on an entrypoint check (or only exporting it when the mount is valid) would produce clearer failure modes.

### Suggestions

- Move ssh-keyscan into an entrypoint script and verify a non-empty result (`test -s`) or compare against a pinned fingerprint, so host-key rotations don't silently break all images.
- Remove `2>/dev/null` from ssh-keyscan and add a `test -s` assertion so build failures are loud and diagnosable.
- Add a commented `HEIMDALLM_SSH_AUTH_SOCK=` entry to `.env.example` (or equivalent) with guidance for Linux (`$SSH_AUTH_SOCK`) and macOS defaults.
- Add an entrypoint check that verifies `/ssh-agent` is a socket before exporting SSH_AUTH_SOCK, and emit a clear warning when agent forwarding is not set up — useful for users who don't need auto_implement.
- Consider documenting in the PR/README that auto_implement requires a running SSH agent on the host with the GitHub key loaded.

---

Muriano:
## ⚠️ MEDIUM Issue

ssh-keyscan at build time bakes GitHub's host key into the image layer. If GitHub rotates its SSH host key (as happened March 2023), all existing images silently break. Move ssh-keyscan to an entrypoint script so the key is fetched at container startup.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## ⚠️ MEDIUM Issue

stderr is suppressed with 2>/dev/null on ssh-keyscan. If keyscan fails during build (network issue), known_hosts will be empty and the build succeeds silently, causing hard-to-diagnose runtime SSH failures. Add validation: `test -s /home/heimdallm/.ssh/known_hosts || exit 1`.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## 🟡 LOW Issue

Default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. Linux users need to set HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example entry or documentation guiding them.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano (approved):

PR adds SSH client and agent forwarding for git operations in auto_implement. The approach is functional but has operational risks: baked-in host keys can go stale, silent build failures from suppressed stderr, and Linux compatibility gaps.

### Suggestions

- Move ssh-keyscan to an entrypoint script that runs at container startup instead of build time, so host keys stay fresh and network failures are visible at runtime
- Add a validation step after ssh-keyscan: `test -s /home/heimdallm/.ssh/known_hosts || exit 1` to fail fast on empty known_hosts
- Add a HEIMDALLM_SSH_AUTH_SOCK example entry to .env.example with a comment explaining the Linux vs macOS difference
- Add an entrypoint check that validates SSH_AUTH_SOCK points to a valid socket and logs a clear warning if not

---

Muriano:
## 🟡 LOW Issue

If the SSH agent socket does not exist on the host (no agent running), Docker will silently create an empty directory at the mount point, causing confusing runtime errors instead of a clear failure message. Consider adding a startup check in the entrypoint.

**Location:** `docker/docker-compose.yml` line 105

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

The PR correctly wires SSH client tooling and agent forwarding to enable auto_implement git operations. The approach is sound, but there are several resilience and portability concerns: ssh-keyscan runs at build time (stale-key risk on GitHub rotation), stderr suppression can mask keyscan failures, the default socket path is macOS-specific with no Linux guidance, and no safeguard exists for when the host has no running agent.

### Issues

⚠️ **docker/Dockerfile:57** — ssh-keyscan runs at image build time and bakes GitHub's host key into an image layer. If GitHub rotates its SSH host key (as happened in March 2023), every previously built image silently breaks. Move the keyscan to an entrypoint/startup script so the known_hosts file is refreshed at container start, or pin/verify expected key fingerprints.
⚠️ **docker/Dockerfile:57** — stderr is silenced with 2>/dev/null on ssh-keyscan. A network blip during build will leave known_hosts empty, the build still succeeds, and runtime SSH will fail with cryptic 'Host key verification failed' errors. Add a validation step such as `test -s /home/heimdallm/.ssh/known_hosts` (and consider `grep -q '^github.com ' ...`) so the build fails loudly.
🟡 **docker/Dockerfile:57** — ssh-keyscan is invoked without -H, so the hostname is stored in plaintext in known_hosts. Minor, but the convention elsewhere is hashed hostnames; using `ssh-keyscan -t rsa,ecdsa,ed25519 -H github.com` is stricter and more consistent with OpenSSH defaults.
🟡 **docker/docker-compose.yml:105** — The default socket path `/run/host-services/ssh-auth.sock` is macOS-Docker-Desktop-specific. Linux hosts have dynamic paths (e.g. `/tmp/ssh-XXX/agent.PID` or `$XDG_RUNTIME_DIR/ssh-agent.socket`) and this PR adds no .env.example entry nor documentation for HEIMDALLM_SSH_AUTH_SOCK, so Linux users will hit this silently. Add a documented `.env.example` entry and mention in the repo docs.
🟡 **docker/docker-compose.yml:105** — When the host has no SSH agent running (socket path missing), docker-compose will auto-create an empty directory at the mount point, and SSH_AUTH_SOCK will point to a non-socket, yielding confusing `Could not open a connection to your authentication agent` errors. Consider an entrypoint check that validates `[ -S "$SSH_AUTH_SOCK" ]` and emits a clear warning when auto_implement is enabled, or require the path to exist via explicit bind option.
🟡 **docker/docker-compose.yml:105** — The socket is mounted read-only (:ro), which is appropriate for the agent protocol, but worth confirming that the ssh-agent protocol signing flow still works through a read-only bind mount on all platforms targeted — on some Docker setups a read-only socket can interfere with write/send operations. Verify on Linux specifically before release.

### Suggestions

- Replace the build-time ssh-keyscan with a small entrypoint script: run keyscan on startup, validate the output is non-empty and contains github.com, then exec the main command. This fixes both the stale-key and silent-failure issues at once.
- Add a `docker/.env.example` entry documenting HEIMDALLM_SSH_AUTH_SOCK with Linux vs macOS examples, and link from the auto_implement section of the README.
- Add an entrypoint sanity check: if auto_implement is enabled and `SSH_AUTH_SOCK` is set but not a socket, log a clear warning so misconfiguration surfaces immediately rather than on the first git push.
- Consider pinning GitHub's known host keys (they are published at https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints) and verifying the keyscan output matches, defending against a MITM at container start time.

---

Muriano:
## ⚠️ MEDIUM Issue

ssh-keyscan at build time bakes GitHub's host key into the image layer. If GitHub rotates its SSH host key (as happened March 2023), all existing images silently break. Move ssh-keyscan to an entrypoint script so the key is fetched at container startup.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## ⚠️ MEDIUM Issue

stderr is suppressed with 2>/dev/null on ssh-keyscan. If keyscan fails during build (network issue), known_hosts will be empty and the build succeeds silently, causing hard-to-diagnose runtime SSH failures. Add validation: `test -s /home/heimdallm/.ssh/known_hosts || exit 1`.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## 🟡 LOW Issue

Default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. Linux users need HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example entry or documentation guiding them.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano:
## 🟡 LOW Issue

If the SSH agent socket does not exist on the host (no agent running), Docker will silently create an empty directory at the mount point, causing confusing runtime errors instead of a clear failure message.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano (approved):

Small, focused PR that adds SSH support for git operations inside the container. The approach is sound but has two medium-severity issues around build-time host key baking and silent failure on ssh-keyscan, plus minor DX gaps for Linux users.

### Suggestions

- Move ssh-keyscan to an entrypoint script that runs at container startup, so keys are always fresh. Create docker/entrypoint.sh that runs ssh-keyscan before exec-ing the main process.
- If keeping ssh-keyscan in the Dockerfile, add a fail-fast check: `ssh-keyscan github.com >> ... && test -s /home/heimdallm/.ssh/known_hosts || exit 1`.
- Add a HEIMDALLM_SSH_AUTH_SOCK entry to .env.example with a comment showing the Linux value (e.g., `# HEIMDALLM_SSH_AUTH_SOCK=$SSH_AUTH_SOCK`).
- Add a startup check in the entrypoint that warns when SSH_AUTH_SOCK is set but is not a valid socket: `[ -S "$SSH_AUTH_SOCK" ] || echo 'WARNING: SSH_AUTH_SOCK is not a valid socket'`.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Small Docker/compose change that adds openssh-client, pre-seeds GitHub's host key at build time, and forwards the host SSH agent socket into the container to enable git operations for auto_implement. The approach is reasonable but has a few robustness and portability gaps: silent ssh-keyscan failure, stale host key baked into the image, and a macOS-specific default socket path with no Linux guidance or .env.example entry.

### Issues

⚠️ **docker/Dockerfile:57** — ssh-keyscan stderr is suppressed with 2>/dev/null and there is no validation that known_hosts was actually populated. If the keyscan fails during build (transient network issue, DNS, proxy), the file will be empty and the build still succeeds, producing an image that fails SSH at runtime with a confusing host key verification error. Add `test -s /home/heimdallm/.ssh/known_hosts` (or drop the stderr redirect) so build failures are loud.
⚠️ **docker/Dockerfile:57** — Baking GitHub's SSH host key into an image layer couples image validity to GitHub's key rotation cadence (the March 2023 RSA key rotation is a concrete precedent). Once rotated, every previously built image silently breaks and users may work around it by disabling StrictHostKeyChecking. Prefer running ssh-keyscan from an entrypoint on container startup, or pin and verify a known fingerprint so a mismatch fails loudly rather than silently.
🟡 **docker/docker-compose.yml:104** — The default `/run/host-services/ssh-auth.sock` only works on macOS Docker Desktop. On Linux the host agent socket path is dynamic (e.g., `/tmp/ssh-XXXXXX/agent.NNN` or `$XDG_RUNTIME_DIR/ssh-agent.socket`), so Linux users must set HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example entry, README note, or docker/.env sample documenting this. Add a documented example to reduce the chance of silent misconfiguration.
🟡 **docker/docker-compose.yml:104** — If HEIMDALLM_SSH_AUTH_SOCK points at a path that does not exist on the host, Docker silently creates an empty directory at `/ssh-agent` inside the container, which then surfaces as obscure `Permission denied (publickey)` or `Error connecting to agent` errors. Consider either requiring the variable (no macOS-specific default), or adding a startup check in the container that warns when `$SSH_AUTH_SOCK` is not a socket.
🟡 **docker/docker-compose.yml:61** — SSH_AUTH_SOCK is set unconditionally to `/ssh-agent`, so even when no agent is forwarded (empty dir mounted), tooling inside the container will still attempt to talk to the agent and produce cryptic errors instead of falling back cleanly. Gate the env var on the mount actually being a socket, or document that auto_implement requires a running host agent.

### Suggestions

- Move `ssh-keyscan github.com` into an entrypoint script and validate the output is non-empty before writing to known_hosts; this also keeps the host key fresh across GitHub key rotations.
- Add a `docker/.env.example` (or update the existing one) with a commented `HEIMDALLM_SSH_AUTH_SOCK=` entry showing both macOS and Linux values, and reference it from the README/auto_implement docs.
- In the entrypoint, verify `/ssh-agent` is a socket (`[ -S /ssh-agent ]`) and log a clear warning/error when it isn't, instead of letting downstream git commands fail opaquely.
- Consider pinning GitHub's SSH host key fingerprint (from https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints) and verifying the ssh-keyscan output against it, so a MITM or stale key surfaces immediately.

---

Muriano:
## ⚠️ MEDIUM Issue

ssh-keyscan at build time bakes GitHub's host key into the image layer. If GitHub rotates its key (as happened March 2023), all existing images silently fail SSH connections. Consider moving ssh-keyscan to an entrypoint script so the key is fetched fresh at container startup.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## ⚠️ MEDIUM Issue

stderr is suppressed with 2>/dev/null on ssh-keyscan. If keyscan fails during build (e.g., network issue), known_hosts will be empty and the build succeeds silently, causing hard-to-diagnose runtime SSH failures. Add a validation step: `test -s /home/heimdallm/.ssh/known_hosts || exit 1`.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## 🟡 LOW Issue

Default socket path /run/host-services/ssh-auth.sock is macOS Docker Desktop-specific. Linux users need to set HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example entry or documentation guiding them.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano:
## 🟡 LOW Issue

If the SSH agent socket does not exist on the host (no agent running), Docker will silently create an empty directory at the mount point, causing confusing runtime errors instead of a clear failure. Consider adding a startup check in the entrypoint that warns when SSH_AUTH_SOCK is not a valid socket.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano (approved):

Small, focused PR that adds SSH support for git operations inside the container. The approach is reasonable but has a few robustness concerns around build-time host key baking and silent failure modes.

### Suggestions

- Move ssh-keyscan to an entrypoint script that runs at container startup, so host keys are always fresh and network failures are visible at runtime rather than silently baked into the image.
- Add `test -s /home/heimdallm/.ssh/known_hosts || exit 1` after ssh-keyscan to fail the build explicitly if the known_hosts file is empty.
- Add a commented HEIMDALLM_SSH_AUTH_SOCK entry to .env.example with guidance for Linux users.
- Add an entrypoint check that validates SSH_AUTH_SOCK points to a real socket and logs a warning if not, e.g.: `[ -S "$SSH_AUTH_SOCK" ] || echo 'WARNING: SSH agent socket not found'`.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Adds openssh-client and SSH agent forwarding to enable git SSH operations from the container for auto_implement. The functional intent is sound, but the ssh-keyscan-at-build approach has real operational problems (host-key rotation staleness, silent failure masking) and the docker-compose defaults leak a macOS-only assumption without Linux guidance or a .env.example update.

### Issues

⚠️ **docker/Dockerfile:57** — ssh-keyscan is run at image build time, baking GitHub's current host key into the image layer. When GitHub rotates its SSH host keys (as it did in March 2023), every built image silently breaks SSH auth until rebuilt. Prefer running ssh-keyscan in an entrypoint script at container start, or pin the expected GitHub key fingerprints (explicitly documented) and verify against them.
⚠️ **docker/Dockerfile:57** — stderr is suppressed with `2>/dev/null` and the command is chained with `&&`, but `>>` redirection on an empty keyscan output still succeeds, so a network failure during build yields an empty known_hosts and a green build. Add a validation step (e.g. `test -s /home/heimdallm/.ssh/known_hosts`) so build fails loudly when keyscan produces nothing.
🟡 **docker/Dockerfile:57** — ssh-keyscan output is not hashed (`-H`) and not restricted by key type. Consider `ssh-keyscan -t rsa,ecdsa,ed25519 -H github.com` to match OpenSSH defaults and avoid unused host-key types padding known_hosts.
🟡 **docker/docker-compose.yml:105** — The default `/run/host-services/ssh-auth.sock` is a Docker Desktop for macOS convention. On Linux, the host agent is typically at `$SSH_AUTH_SOCK` under `/tmp/ssh-*/agent.*` and the default mount will silently fail. Document required override via `HEIMDALLM_SSH_AUTH_SOCK` in `docker/.env.example` (and/or README) with a Linux example like `${SSH_AUTH_SOCK}`.
🟡 **docker/docker-compose.yml:105** — If the source path does not exist on the host (no agent running, wrong path), Docker silently creates an empty directory at the bind-mount target, and SSH then fails at runtime with an opaque error. Add a startup check (entrypoint or healthcheck hint) that warns when `$SSH_AUTH_SOCK` is not a connectable socket, or document the prerequisite clearly.
🟡 **docker/docker-compose.yml:61** — Mounting the SSH agent socket gives any process in the container the ability to sign auth challenges with the host user's keys for as long as the container is running. This is not strictly a bug, but given the container also runs third-party AI CLIs (Claude/OpenCode/Codex), it deserves an explicit note in docs about the blast radius and that auto_implement should only be enabled on trusted workloads.

### Suggestions

- Move ssh-keyscan into an entrypoint script run as the `heimdallm` user at container start, with a validation `test -s ~/.ssh/known_hosts || { echo 'ssh-keyscan failed'; exit 1; }`.
- Add a `docker/.env.example` entry showing `HEIMDALLM_SSH_AUTH_SOCK=${SSH_AUTH_SOCK}` for Linux and leave the macOS default commented.
- In the entrypoint, verify `[ -S "$SSH_AUTH_SOCK" ]` and emit a clear warning (or fail fast if auto_implement is enabled) when the agent socket is missing.
- Consider pinning GitHub host-key fingerprints in a checked-in file and verifying the keyscan result against them, rather than TOFU at either build or startup.
- Use `ssh-keyscan -t rsa,ecdsa,ed25519 -H github.com` to keep known_hosts minimal and hashed.

---

Muriano:
## ⚠️ MEDIUM Issue

ssh-keyscan at build time bakes GitHub's SSH host key into the image layer. If GitHub rotates its key (as happened March 2023), all existing images silently fail SSH connections. Move ssh-keyscan to an entrypoint script so the key is fetched fresh at container startup.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## ⚠️ MEDIUM Issue

stderr is suppressed with 2>/dev/null on ssh-keyscan. If keyscan fails during build (e.g., network issue), known_hosts will be empty and the build succeeds silently, causing hard-to-diagnose runtime SSH failures. Add validation: `test -s /home/heimdallm/.ssh/known_hosts || exit 1`.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## 🟡 LOW Issue

Default socket path `/run/host-services/ssh-auth.sock` is macOS Docker Desktop-specific. Linux users need to set HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example entry or documentation guiding them.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano:
## 🟡 LOW Issue

If the SSH agent socket does not exist on the host (no agent running), Docker will silently create an empty directory at the mount point, causing confusing runtime errors instead of a clear failure message. Consider adding a startup check in the entrypoint.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano (approved):

PR adds SSH client and agent forwarding for git operations in auto_implement. The approach is sound but has two notable robustness issues: baking GitHub's host key at build time and silently swallowing ssh-keyscan failures.

### Suggestions

- Move ssh-keyscan to an entrypoint script that runs at container startup, so the host key is always fresh and network failures are visible at runtime rather than silently ignored at build time.
- Add a build-time validation after ssh-keyscan: `test -s /home/heimdallm/.ssh/known_hosts || exit 1` to fail the build if the keyscan produced no output.
- Add a HEIMDALLM_SSH_AUTH_SOCK entry to .env.example with a comment explaining what Linux users should set it to (e.g., their $SSH_AUTH_SOCK value).
- Add an entrypoint check that validates SSH_AUTH_SOCK is a valid socket and logs a clear warning if it is not, instead of letting git fail with a cryptic error.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR adds SSH client and agent forwarding to the Docker image so auto_implement can perform git operations over SSH. Changes are small and focused: install openssh-client, pre-seed GitHub's host key in known_hosts, and forward the host SSH agent socket via docker-compose. The approach is sound but has a few legitimate robustness issues — notably silent failure of ssh-keyscan, staleness risk from baking host keys at build time, and lack of Linux guidance for the agent socket path. Prior AI review comments are heavily duplicated but the underlying points are valid.

### Issues

⚠️ **docker/Dockerfile:57** — ssh-keyscan stderr is redirected to /dev/null and its exit status is not checked. If the network fails during build, known_hosts will be empty and the build still succeeds, leading to opaque runtime SSH failures. Add a validation like 'test -s /home/heimdallm/.ssh/known_hosts' after the keyscan, and avoid silencing stderr so real errors surface in build logs.
⚠️ **docker/Dockerfile:57** — Baking GitHub's host key into the image at build time means a GitHub host-key rotation (as happened in March 2023) silently breaks every built image until rebuilt. Consider running ssh-keyscan in an entrypoint script at container start, or pinning the known GitHub public key fingerprints (which GitHub publishes) and verifying the scanned key matches, so trust is explicit rather than TOFU at build.
🟡 **docker/docker-compose.yml:102** — Default SSH agent socket path '/run/host-services/ssh-auth.sock' is specific to Docker Desktop on macOS. On Linux hosts, SSH_AUTH_SOCK typically lives at a per-session path (e.g. /tmp/ssh-XXXX/agent.PID) and must be set via HEIMDALLM_SSH_AUTH_SOCK. There is no .env.example entry or README/docs update showing Linux users what to configure, so the feature will appear broken to them.
🟡 **docker/docker-compose.yml:102** — If the host SSH agent socket does not exist (e.g., no agent running), Docker Compose will create an empty directory at the mount target, producing confusing runtime errors from git/ssh rather than a clear 'agent not available' message. Consider documenting this, making the mount conditional, or adding an entrypoint check that validates SSH_AUTH_SOCK is a socket and logs a clear warning otherwise.
🟡 **docker/docker-compose.yml:61** — SSH_AUTH_SOCK is hardcoded to '/ssh-agent' inside the container regardless of whether the host actually forwards an agent. If a user disables forwarding or omits HEIMDALLM_SSH_AUTH_SOCK, tools inside will still see SSH_AUTH_SOCK set and attempt (and fail) to use it. Consider only setting SSH_AUTH_SOCK when forwarding is enabled, or documenting this coupling clearly.

### Suggestions

- Replace 'ssh-keyscan github.com >> known_hosts 2>/dev/null' with a version that fails the build on empty output, e.g. 'ssh-keyscan -t rsa,ecdsa,ed25519 github.com > /tmp/ghkeys && test -s /tmp/ghkeys && cat /tmp/ghkeys >> /home/heimdallm/.ssh/known_hosts'.
- Move ssh-keyscan to an entrypoint script so host keys are refreshed on every container start, avoiding image staleness after GitHub key rotations.
- Add a '.env.example' entry (or update existing one) documenting HEIMDALLM_SSH_AUTH_SOCK with separate macOS and Linux examples.
- Add an entrypoint check that verifies /ssh-agent is a socket and logs a clear warning (or exits) if auto_implement depends on it and it is missing.
- Consider pinning GitHub's published SSH host key fingerprints and verifying the scanned key matches, rather than trust-on-first-use at build time.

---

Muriano:
## ⚠️ MEDIUM Issue

ssh-keyscan at build time bakes GitHub's host key into the image layer. If GitHub rotates its SSH host key (as happened March 2023), all existing images silently fail SSH connections. Move ssh-keyscan to an entrypoint script so the key is fetched fresh at container startup.

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## ⚠️ MEDIUM Issue

stderr is suppressed with 2>/dev/null on ssh-keyscan. If keyscan fails during build (e.g., network issue), known_hosts will be empty and the build succeeds silently, causing hard-to-diagnose runtime SSH failures. Add validation: test -s /home/heimdallm/.ssh/known_hosts || exit 1

**Location:** `docker/Dockerfile` line 57

---

Muriano:
## 🟡 LOW Issue

Default socket path /run/host-services/ssh-auth.sock is macOS Docker Desktop-specific. Linux users need to set HEIMDALLM_SSH_AUTH_SOCK but there is no .env.example entry or documentation guiding them.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano:
## 🟡 LOW Issue

If the SSH agent socket does not exist on the host (no agent running), Docker will silently create an empty directory at the mount point, causing confusing runtime errors instead of a clear failure message.

**Location:** `docker/docker-compose.yml` line 105

---

Muriano (approved):

Small, focused PR adding SSH client and agent forwarding for git operations in the container. The approach is sound but has two notable issues: build-time ssh-keyscan can go stale and its failure is silently swallowed.

### Suggestions

- Move ssh-keyscan to an entrypoint script that runs at container startup, so the host key is always fresh and build-time network issues don't produce a silently broken image.
- Add a build-time validation step after ssh-keyscan: test -s /home/heimdallm/.ssh/known_hosts || exit 1
- Add a commented HEIMDALLM_SSH_AUTH_SOCK entry to .env.example with Linux guidance (e.g., HEIMDALLM_SSH_AUTH_SOCK=$SSH_AUTH_SOCK).
- Add an entrypoint check that warns when SSH_AUTH_SOCK is set but does not point to a valid socket, to catch missing agent early.

---

PR #152, opened 2026-04-21
fix: use HTTPS for git fetch in auto_implement (removes SSH dependency)

sergiotejon:
Refs #148

## Problem

auto_implement fails with `Permission denied (publickey)` because `git fetch origin` uses the clone's SSH remote, which requires SSH keys/agent not reliably available in Docker.

## Fix

`CheckoutNewBranch` now uses HTTPS with `GIT_ASKPASS` (same pattern as `Push`):

```go
// BEFORE: depends on SSH remote
runGit(ctx, dir, nil, "fetch", "origin", baseBranch)

// AFTER: HTTPS with token, no SSH needed
url := "https://x-access-token@github.com/{repo}.git"
runGit(ctx, dir, askPassEnv, "fetch", url, baseBranch)
```

Interface updated: `CheckoutNewBranch(ctx, dir, repo, branch, baseBranch, token)`

## Changes

- `gitops.go` — HTTPS fetch with GIT_ASKPASS, uses FETCH_HEAD instead of origin/branch
- `pipeline.go` — passes `issue.Repo` and `opts.GitHubToken` to CheckoutNewBranch
- `pipeline_test.go` — 3 mock implementations updated

## Test plan

- [ ] `go test ./internal/issues/` passes
- [ ] auto_implement fetches via HTTPS without SSH
- [ ] Push still works (unchanged)

### DIFF
--- a/daemon/internal/issues/gitops.go
+++ b/daemon/internal/issues/gitops.go
@@ -36,10 +36,10 @@ const maxGitStderrBytes = 16 * 1024 // 16 KiB
 // shutdown (or per-request) through long-running network operations —
 // `git fetch` and `git push` in particular.
 type GitOps interface {
-	// CheckoutNewBranch fetches baseBranch from origin and checks out branch
-	// from that tip, overwriting any previous attempt at the same branch so
-	// a re-run starts clean.
-	CheckoutNewBranch(ctx context.Context, dir, branch, baseBranch string) error
+	// CheckoutNewBranch fetches baseBranch and checks out branch from that
+	// tip, overwriting any previous attempt so a re-run starts clean.
+	// Uses HTTPS with token for fetch (avoids SSH dependency in Docker).
+	CheckoutNewBranch(ctx context.Context, dir, repo, branch, baseBranch, token string) error
 	// HasChanges reports whether the working tree has modified or untracked
 	// files — both are in scope for the commit because the agent may create
 	// new files as well as edit existing ones.
@@ -65,16 +65,31 @@ type GitExec struct{}
 // NewGitExec returns a ready-to-use GitExec. Zero configuration required.
 func NewGitExec() *GitExec { return &GitExec{} }
 
-// CheckoutNewBranch fetches the base branch and creates (or resets) the
-// work branch from it. `-B` is deliberate: on a re-run we want the branch
-// to match the latest base rather than pick up stale state from a previous
-// failed attempt.
-func (g *GitExec) CheckoutNewBranch(ctx context.Context, dir, branch, baseBranch string) error {
-	if err := runGit(ctx, dir, nil, "fetch", "origin", baseBranch); err != nil {
-		return fmt.Errorf("gitops: fetch origin/%s: %w", baseBranch, err)
+// CheckoutNewBranch fetches the base branch via HTTPS (using the same
+// GIT_ASKPASS mechanism as Push) and creates (or resets) the work branch
+// from it. `-B` is deliberate: on a re-run we want the branch to match
+// the latest base rather than pick up stale state from a previous attempt.
+//
+// Using an explicit HTTPS URL instead of `git fetch origin` avoids relying
+// on the clone's remote configuration, which may point at an SSH URL that
+// requires keys/agent not available inside the Docker container.
+func (g *GitExec) CheckoutNewBranch(ctx context.Context, dir, repo, branch, baseBranch, token string) error {
+	if token == "" {
+		return fmt.Errorf("gitops: checkout requires a non-empty token")
 	}
-	if err := runGit(ctx, dir, nil, "checkout", "-B", branch, "origin/"+baseBranch); err != nil {
-		return fmt.Errorf("gitops: checkout -B %s origin/%s: %w", branch, baseBranch, err)
+	env, cleanup, err := buildAskPassEnv(token)
+	if err != nil {
+		return fmt.Errorf("gitops: setup askpass for fetch: %w", err)
+	}
+	defer cleanup()
+
+	url := fmt.Sprintf("https://x-access-token@github.com/%s.git", repo)
+	if err := runGit(ctx, dir, env, "fetch", url, baseBranch); err != nil {
+		return fmt.Errorf("gitops: fetch %s/%s: %w", repo, baseBranch, err)
+	}
+	// FETCH_HEAD points to the tip of what we just fetched.
+	if err := runGit(ctx, dir, nil, "checkout", "-B", branch, "FETCH_HEAD"); err != nil {
+		return fmt.Errorf("gitops: checkout -B %s: %w", branch, err)
 	}
 	return nil
 }
--- a/daemon/internal/issues/pipeline.go
+++ b/daemon/internal/issues/pipeline.go
@@ -388,7 +388,7 @@ func (p *Pipeline) runAutoImplement(ctx context.Context, issue *github.Issue, is
 	}
 
 	branch := fmt.Sprintf("heimdallm/issue-%d", issue.Number)
-	if err := p.git.CheckoutNewBranch(ctx, workDir, branch, base); err != nil {
+	if err := p.git.CheckoutNewBranch(ctx, workDir, issue.Repo, branch, base, opts.GitHubToken); err != nil {
 		p.publishError(issueID, issue, fmt.Errorf("checkout: %w", err))
 		return nil, fmt.Errorf("issues pipeline: checkout: %w", err)
 	}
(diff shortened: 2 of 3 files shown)


Muriano:
## 🟡 LOW Issue

The HTTPS URL embeds `x-access-token@` but the actual password is supplied via GIT_ASKPASS. Including the username in the URL is fine for GitHub, but if `buildAskPassEnv` ever fails silently or the askpass script isn't invoked, git will prompt or fail opaquely. Consider logging (at debug level) that askpass auth is being used.

**Location:** `daemon/internal/issues/gitops.go` line 87

---

Muriano:
## ⚠️ MEDIUM Issue

FETCH_HEAD is a transient ref that can be overwritten by any concurrent `git fetch` in the same repo. If another goroutine or process fetches in the same workDir between the fetch and checkout, FETCH_HEAD may point to the wrong commit. Given the current architecture uses per-issue workDirs this is unlikely, but worth documenting the assumption.

**Location:** `daemon/internal/issues/gitops.go` line 90

---

Muriano:
## ⚠️ MEDIUM Issue

The `repo` parameter is interpolated directly into the URL without validation. A malicious or malformed repo string (e.g., containing `/../` or special chars) could produce an unexpected URL. Consider validating the `owner/repo` format.

**Location:** `daemon/internal/issues/gitops.go` line 87

---

Muriano:
## 🟡 LOW Issue

The fake git implementations (fakeGit, contextCheckingGit, tokenSniffingGit) accept the new `token` parameter but none of them verify its value. tokenSniffingGit in particular seems like it should capture the token for assertion, given its name.

**Location:** `daemon/internal/issues/pipeline_test.go` line 230

---

Muriano (approved):

Solid change replacing SSH-based git fetch with HTTPS+token auth for Docker compatibility. The approach using GIT_ASKPASS and FETCH_HEAD is correct. One medium-severity issue with token potentially leaking in the URL, and a minor concern about FETCH_HEAD reliability.

### Suggestions

- Validate the `repo` parameter matches an expected `owner/name` pattern before interpolating into the URL to prevent URL injection.
- In tokenSniffingGit.CheckoutNewBranch, capture the token parameter (`g.seenToken = token`) so tests can assert it's passed correctly through the pipeline.
- Add a brief comment near the FETCH_HEAD usage noting the assumption that workDir is not shared with concurrent git operations.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Solid change replacing SSH-based git fetch with HTTPS+token auth for Docker compatibility. The approach using GIT_ASKPASS and FETCH_HEAD is correct. One medium-severity issue with token potentially leaking in the URL, and a minor concern about FETCH_HEAD reliability.

### Issues

🟡 **daemon/internal/issues/gitops.go:87** — The HTTPS URL embeds `x-access-token@` but the actual password is supplied via GIT_ASKPASS. Including the username in the URL is fine for GitHub, but if `buildAskPassEnv` ever fails silently or the askpass script isn't invoked, git will prompt or fail opaquely. Consider logging (at debug level) that askpass auth is being used.
⚠️ **daemon/internal/issues/gitops.go:90** — FETCH_HEAD is a transient ref that can be overwritten by any concurrent `git fetch` in the same repo. If another goroutine or process fetches in the same workDir between the fetch and checkout, FETCH_HEAD may point to the wrong commit. Given the current architecture uses per-issue workDirs this is unlikely, but worth documenting the assumption.
⚠️ **daemon/internal/issues/gitops.go:87** — The `repo` parameter is interpolated directly into the URL without validation. A malicious or malformed repo string (e.g., containing `/../` or special chars) could produce an unexpected URL. Consider validating the `owner/repo` format.
🟡 **daemon/internal/issues/pipeline_test.go:230** — The fake git implementations (fakeGit, contextCheckingGit, tokenSniffingGit) accept the new `token` parameter but none of them verify its value. tokenSniffingGit in particular seems like it should capture the token for assertion, given its name.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Straightforward fix to avoid SSH dependency in Docker by using HTTPS + GIT_ASKPASS for `git fetch`, mirroring the pattern already used for `git push`. The change is small, well-scoped, and correctly updates all callers and test doubles. Main concerns are minor: the `repo` parameter is interpolated into a URL without validation, FETCH_HEAD is transient (safe under current per-issue workDir architecture but should be documented), and the test doubles don't exercise the new `token` parameter.

### Issues

⚠️ **daemon/internal/issues/gitops.go:87** — The `repo` parameter is interpolated directly into the HTTPS URL without validation. Although `issue.Repo` is sourced from the GitHub API (so realistically `owner/name`), a defensive check for the `owner/name` shape (e.g., single `/`, no whitespace, no `..`) would prevent subtle bugs if an upstream caller ever passes malformed input, and also fails earlier with a clearer message than a remote-side git error.
⚠️ **daemon/internal/issues/gitops.go:90** — Using `FETCH_HEAD` relies on the invariant that no other `git fetch` touches the same workDir between fetch and checkout. The current architecture (per-issue workDir) upholds this, but the assumption is load-bearing and undocumented. A short comment noting "safe because workDir is exclusive to this issue run" would prevent a future refactor from silently breaking it. Alternatively, fetching into an explicit local ref (e.g., `fetch <url> <base>:refs/heimdallm/base`) removes the invariant entirely.
🟡 **daemon/internal/issues/pipeline_test.go:801** — `tokenSniffingGit.CheckoutNewBranch` now receives a `token` but ignores it. Given the struct's name and purpose (asserting the token is plumbed through), it should capture the token into a field alongside `seenToken` so callers can assert token propagation for both Push and CheckoutNewBranch. Otherwise the existing test coverage silently lost a signal after this change.
🟡 **daemon/internal/issues/gitops.go:87** — Minor: the HTTPS URL embeds `x-access-token` as the username and relies on GIT_ASKPASS for the password. If `buildAskPassEnv` ever returns without setting up a working askpass (or the helper script is misconfigured), git may prompt interactively or fail opaquely. A debug-level log when askpass is invoked, or an assertion that the env contains `GIT_ASKPASS`, would make such failures easier to diagnose.

### Suggestions

- Validate `repo` matches `^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$` before formatting it into the URL, returning a clear error if not.
- Either document the FETCH_HEAD assumption with a comment referencing the per-issue workDir invariant, or fetch into a named local ref (e.g., `refs/heimdallm/base`) and check it out explicitly to remove the invariant.
- Update `tokenSniffingGit` to capture the token passed to `CheckoutNewBranch` so the test can assert the token is threaded through checkout, not just push.
- Consider extracting the `https://x-access-token@github.com/<repo>.git` URL construction into a small helper (e.g., `githubHTTPSURL(repo)`) shared with Push — reduces duplication and gives a single place to add validation/logging.

---

Muriano:
## ⚠️ MEDIUM Issue

The `repo` parameter is interpolated directly into the URL (`fmt.Sprintf("https://x-access-token@github.com/%s.git", repo)`) without any validation. A malformed repo string could produce an invalid or malicious URL. Should validate it matches `owner/repo` format (e.g., regexp `^[\w.-]+/[\w.-]+$`).

**Location:** `daemon/internal/issues/gitops.go` line 87

---

Muriano:
## 🟡 LOW Issue

FETCH_HEAD is a transient ref. While per-issue workDirs make concurrent overwrites unlikely today, this assumption should be documented with a comment or, better, use `git fetch <url> <branch>:<local-ref>` to fetch into a named ref instead of relying on FETCH_HEAD.

**Location:** `daemon/internal/issues/gitops.go` line 90

---

Muriano:
## 🟡 LOW Issue

tokenSniffingGit now receives `token` in CheckoutNewBranch but ignores it. Given its purpose is to sniff tokens, it should capture the token value for test assertions, otherwise the test coverage for token propagation through checkout is incomplete.

**Location:** `daemon/internal/issues/pipeline_test.go` line 801

---

Muriano (approved):

Solid change that removes SSH dependency by switching git fetch to HTTPS with GIT_ASKPASS. The approach is correct and well-documented. The existing review comments cover the main concerns well. One additional issue worth noting around input validation.

### Suggestions

- Add a regex validation for the `repo` parameter (e.g., `^[\w.-]+/[\w.-]+$`) before building the URL to prevent malformed or path-traversal inputs.
- Consider fetching into a namespaced ref (`git fetch <url> <branch>:refs/heimdallm/fetch-tmp`) instead of relying on FETCH_HEAD, to be robust against any future concurrent fetch in the same workdir.
- Update tokenSniffingGit.CheckoutNewBranch to capture the token (`g.seenToken = token`) so tests can assert token propagation through the full checkout+push flow.

---

PR #194, opened 2026-04-22
fix: global settings parity — missing fields, broken develop switch

Muriano:
## Summary

- Daemon exposes `pr_assignee`, `pr_draft`, `issue_prompt`, `implement_prompt` in GET /config
- Flutter: PR Assignee (chip field), Draft switch, Issue Prompt dropdown, Develop Prompt dropdown added to global settings
- Develop switch fixed: enabling now sets default "develop" label (was no-op before)
- Develop sub-fields hidden when switch is off

## Closes

Resolves field parity gap between global settings and per-repo detail screen.

## Test plan

- [ ] `cd daemon && go test ./... -count=1` — all pass
- [ ] `flutter analyze --no-fatal-infos` — no issues
- [ ] Manual: toggle develop switch on → label "develop" appears, fields shown
- [ ] Manual: toggle develop switch off → fields hidden
- [ ] Manual: select agent profile in Issue Prompt / Develop Prompt dropdowns → persisted on save

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -462,14 +462,21 @@ func main() {
 			"activity_log_enabled":        ptrBoolOrTrue(c.ActivityLog.Enabled),
 			"activity_log_retention_days": ptrIntOr(c.ActivityLog.RetentionDays, 90),
 		}
-		if len(c.AI.PRMetadata.Reviewers) > 0 || len(c.AI.PRMetadata.Labels) > 0 {
-			pm := map[string]any{}
-			if len(c.AI.PRMetadata.Reviewers) > 0 {
-				pm["reviewers"] = c.AI.PRMetadata.Reviewers
-			}
-			if len(c.AI.PRMetadata.Labels) > 0 {
-				pm["labels"] = c.AI.PRMetadata.Labels
-			}
+		reviewers, labels, assignee, draft := c.ResolvedPRMetadata()
+		pm := map[string]any{}
+		if len(reviewers) > 0 {
+			pm["reviewers"] = reviewers
+		}
+		if len(labels) > 0 {
+			pm["labels"] = labels
+		}
+		if assignee != "" {
+			pm["pr_assignee"] = assignee
+		}
+		if draft != nil {
+			pm["pr_draft"] = *draft
+		}
+		if len(pm) > 0 {
 			result["pr_metadata"] = pm
 		}
 		return result
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -461,6 +461,8 @@ func main() {
 			"local_dirs_detected":         localDirsDetected,
 			"activity_log_enabled":        ptrBoolOrTrue(c.ActivityLog.Enabled),
 			"activity_log_retention_days": ptrIntOr(c.ActivityLog.RetentionDays, 90),
+			"issue_prompt":               c.AI.IssuePrompt,
+			"implement_prompt":           c.AI.ImplementPrompt,
 		}
 		reviewers, labels, assignee, draft := c.ResolvedPRMetadata()
 		pm := map[string]any{}
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -409,7 +409,7 @@ func repoOrg(repo string) string {
 // flat [ai] fields on top of [ai.pr_metadata]. Flat fields win when set,
 // matching the contract that HEIMDALLM_PR_* env vars populate the flat
 // fields and should override the nested section.
-func (c *Config) resolvedPRMetadata() (reviewers, labels []string, assignee string, draft *bool) {
+func (c *Config) ResolvedPRMetadata() (reviewers, labels []string, assignee string, draft *bool) {
 	reviewers = c.AI.PRMetadata.Reviewers
 	labels = c.AI.PRMetadata.Labels
 	assignee = c.AI.PRMetadata.Assignee
@@ -435,7 +435,7 @@ func (c *Config) resolvedPRMetadata() (reviewers, labels []string, assignee stri
 // three levels: per-repo > per-org > global defaults. Each PR metadata
 // field resolves independently.
 func (c *Config) AIForRepo(repo string) RepoAI {
-	gReviewers, gLabels, gAssignee, gDraft := c.resolvedPRMetadata()
+	gReviewers, gLabels, gAssignee, gDraft := c.ResolvedPRMetadata()
 
 	// Org-level layer: start from global, overlay org-level fields.
 	orgReviewers, orgLabels, orgAssignee, orgDraft := gReviewers, gLabels, gAssignee, gDraft
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -255,6 +255,13 @@ type AIConfig struct {
 	PRAssignee  string   `toml:"pr_assignee"`
 	PRDraft     *bool    `toml:"pr_draft,omitempty"`
 
+	// IssuePrompt is the global default agent profile ID for issue triage.
+	// Per-repo overrides in [ai.repos.<name>] take precedence.
+	IssuePrompt string `toml:"issue_prompt"`
+	// ImplementPrompt is the global default agent profile ID for auto-implement.
+	// Per-repo overrides in [ai.repos.<name>] take precedence.
+	ImplementPrompt string `toml:"implement_prompt"`
+
 	// GeneratePRDescription enables LLM-generated PR titles and descriptions
 	// for auto_implement PRs. When true, after the implementation commit,
 	// a second LLM call generates a rich PR description from the diff.
--- a/flutter_app/lib/core/models/config_model.dart
+++ b/flutter_app/lib/core/models/config_model.dart
@@ -341,6 +341,8 @@ class AppConfig {
   final List<String> globalPRLabels;
   final String globalPRAssignee;
   final bool globalPRDraft;
+  final String globalIssuePrompt;
+  final String globalImplementPrompt;
   /// Host paths the daemon scans (in order) when a repo has no explicit
   /// `local_dir` set — first match at `{base}/{short-repo-name}` wins.
   final List<String> localDirBase;
@@ -367,6 +369,8 @@ class AppConfig {
     this.globalPRLabels = const [],
     this.globalPRAssignee = '',
     this.globalPRDraft = false,
+    this.globalIssuePrompt = '',
+    this.globalImplementPrompt = '',
     this.localDirBase = const [],
     this.localDirsDetected = const {},
   });
@@ -393,6 +397,8 @@ class AppConfig {
     List<String>? globalPRLabels,
     String? globalPRAssignee,
     bool? globalPRDraft,
+    String? globalIssuePrompt,
+    String? globalImplementPrompt,
     List<String>? localDirBase,
     Map<String, String>? localDirsDetected,
   }) {
@@ -408,10 +414,12 @@ class AppConfig {
       issueTracking:     issueTracking     ?? this.issueTracking,
       globalPRReviewers: globalPRReviewers ?? this.globalPRReviewers,
       globalPRLabels:    globalPRLabels    ?? this.globalPRLabels,
-      globalPRAssignee:  globalPRAssignee  ?? this.globalPRAssignee,
-      globalPRDraft:     globalPRDraft     ?? this.globalPRDraft,
-      localDirBase:      localDirBase      ?? this.localDirBase,
-      localDirsDetected: localDirsDetected ?? this.localDirsDetected,
+      globalPRAssignee:       globalPRAssignee       ?? this.globalPRAssignee,
+      globalPRDraft:          globalPRDraft          ?? this.globalPRDraft,
+      globalIssuePrompt:     globalIssuePrompt     ?? this.globalIssuePrompt,
+      globalImplementPrompt: globalImplementPrompt ?? this.globalImplementPrompt,
+      localDirBase:           localDirBase           ?? this.localDirBase,
+      localDirsDetected:     localDirsDetected     ?? this.localDirsDetected,
     );
   }
 
@@ -515,9 +523,11 @@ class AppConfig {
       issueTracking:     issueTracking,
       globalPRReviewers: _parseStringList((json['pr_metadata'] as Map<String, dynamic>?)?['reviewers']),
       globalPRLabels:    _parseStringList((json['pr_metadata'] as Map<String, dynamic>?)?['labels']),
-      globalPRAssignee:  (json['pr_metadata'] as Map<String, dynamic>?)?['pr_assignee'] as String? ?? '',
-      globalPRDraft:     (json['pr_metadata'] as Map<String, dynamic>?)?['pr_draft'] as bool? ?? false,
-      localDirBase:      _parseStringList(json['local_dir_base']),
+      globalPRAssignee:       (json['pr_metadata'] as Map<String, dynamic>?)?['pr_assignee'] as String? ?? '',
+      globalPRDraft:          (json['pr_metadata'] as Map<String, dynamic>?)?['pr_draft'] as bool? ?? false,
+      globalIssuePrompt:     (json['issue_prompt'] as String?) ?? '',
+      globalImplementPrompt: (json['implement_prompt'] as String?) ?? '',
+      localDirBase:           _parseStringList(json['local_dir_base']),
       localDirsDetected: localDirsDetected,
     );
   }
--- a/flutter_app/lib/core/models/config_model.dart
+++ b/flutter_app/lib/core/models/config_model.dart
@@ -339,11 +339,10 @@ class AppConfig {
   final IssueTrackingConfig issueTracking;
   final List<String> globalPRReviewers;
   final List<String> globalPRLabels;
+  final String globalPRAssignee;
+  final bool globalPRDraft;
   /// Host paths the daemon scans (in order) when a repo has no explicit
   /// `local_dir` set — first match at `{base}/{short-repo-name}` wins.
-  /// Written to the `[github].local_dir_base` TOML key. Unknown to early
-  /// `writeConfig` revisions; fielded here so a Flutter save preserves
-  /// the value instead of silently dropping it on the floor.
   final List<String> localDirBase;
   /// Auto-detected `local_dir` per repo, populated by the daemon when the
   /// repo is visible at `/home/heimdallm/repos/<short-name>` in the
@@ -366,6 +365,8 @@ class AppConfig {
     this.issueTracking = const IssueTrackingConfig(),
     this.globalPRReviewers = const [],
     this.globalPRLabels = const [],
+    this.globalPRAssignee = '',
+    this.globalPRDraft = false,
     this.localDirBase = const [],
     this.localDirsDetected = const {},
   });
@@ -390,6 +391,8 @@ class AppConfig {
     IssueTrackingConfig? issueTracking,
     List<String>? globalPRReviewers,
     List<String>? globalPRLabels,
+    String? globalPRAssignee,
+    bool? globalPRDraft,
     List<String>? localDirBase,
     Map<String, String>? localDirsDetected,
   }) {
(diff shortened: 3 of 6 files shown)


sergiotejon:
## 🔴 HIGH Issue

DropdownButtonFormField does not have an 'initialValue' parameter. Should use 'value' instead.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 537

---

sergiotejon:
## 🟡 LOW Issue

Toggling develop ON sets a default label, but the user may not expect 'develop' to be silently injected. This is a minor UX concern but could cause unintended config writes.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 498

---

sergiotejon:
## ⚠️ MEDIUM Issue

Using AutocompleteChipField for a single-value PR assignee is awkward — it accepts a list but is constrained to one item via .first. A simple TextField would be more appropriate.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 509

---

sergiotejon:
## ⚠️ MEDIUM Issue

globalPRDraft defaults to false, so when the daemon returns null (field absent), the Flutter app treats it as explicitly false rather than 'unset'. This loses the distinction between 'user set false' and 'not configured', which the daemon's *bool preserves.

**Location:** `flutter_app/lib/core/models/config_model.dart` line 370

---

sergiotejon:
## 🟡 LOW Issue

Exporting ResolvedPRMetadata solely for use in main.go breaks encapsulation. Consider adding a dedicated method or keeping it unexported with a wrapper.

**Location:** `daemon/internal/config/config.go` line 419

---

sergiotejon (changes requested):

PR adds missing global settings fields (PR assignee, draft, issue/implement prompts) and fixes the develop switch to properly gate child widgets. A few issues worth addressing.

### Suggestions

- Fix the DropdownButtonFormField: replace 'initialValue' with 'value' — this will cause a compile error as-is.
- Consider using a tri-state (bool?) for globalPRDraft in the Dart model to preserve the daemon's *bool semantics (null vs false).
- Use a simple TextField for PR assignee instead of repurposing AutocompleteChipField with a single-element list.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Fixes real parity gaps (global issue/implement prompts, PR assignee/draft, local_dir_base) and wires them end-to-end. Main concern is the tristate loss on pr_draft in the Flutter model, plus a few UX/encapsulation nits already flagged.

### Issues

⚠️ **flutter_app/lib/core/models/config_model.dart:370** — globalPRDraft is a plain bool defaulting to false, so an absent daemon field is indistinguishable from an explicit false. Combined with first_run_setup.dart only emitting pr_draft when true, this can silently drop a pre-existing pr_draft=false from the TOML on first-run rewrite. Use bool? (nullable) and only emit when non-null.
⚠️ **flutter_app/lib/features/config/config_screen.dart:509** — AutocompleteChipField used for single-value PR Assignee is semantically awkward (list coerced via .first) and allows the user to enter multiple chips, silently dropping all but one. Use a TextField (or a single-value autocomplete).
🟡 **daemon/internal/config/config.go:419** — ResolvedPRMetadata was exported only for main.go. Since main.go is in the same module, prefer a dedicated typed accessor (e.g., PRMetadataForAPI) or keep it unexported and move the caller. Exporting resolution internals widens the public API surface unnecessarily.
🟡 **flutter_app/lib/features/config/config_screen.dart:498** — Toggling develop ON silently injects a 'develop' label into developLabels. A save right after toggling persists a label the user never typed. Prefer leaving the list empty and letting the user add one, or show a hint chip.
🟡 **flutter_app/lib/features/config/config_screen.dart:537** — DropdownButtonFormField's 'initialValue' is inherited from FormField and only works on newer Flutter (3.22+); the ValueKey workaround to force rebuild on value change is a smell. If the project pins an older Flutter, this won't compile — verify against the repo's Flutter constraint and prefer 'value' for portability.
🟡 **flutter_app/lib/features/config/config_screen.dart:693** — _buildUpdatedConfig always writes globalPRAssignee/globalPRDraft/globalImplementPrompt even when the develop toggle is off (section hidden). State retained from a prior enabled session will be persisted on save. Consider zeroing these when hasLabels is false, or scoping them under the toggle.

### Suggestions

- Switch globalPRDraft to bool? and gate TOML emission / diff on non-null to preserve the tristate the daemon already models with *bool.
- Replace the AutocompleteChipField for PR Assignee with a plain TextField to match its single-value nature.
- Add a daemon-side test that asserts GET /config round-trips pr_assignee, pr_draft, issue_prompt, and implement_prompt (both set and unset) — this PR is a parity fix and should be pinned with coverage.

---

sergiotejon:
## 🔴 HIGH Issue

DropdownButtonFormField has no 'initialValue' parameter — this will not compile. Must use 'value' instead.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 537

---

sergiotejon:
## ⚠️ MEDIUM Issue

globalPRDraft is a non-nullable bool defaulting to false, losing the daemon's *bool three-state semantics (true/false/unset). Should be bool? to preserve parity.

**Location:** `flutter_app/lib/core/models/config_model.dart` line 370

---

sergiotejon:
## ⚠️ MEDIUM Issue

AutocompleteChipField is misused for a single-value string field (PR assignee). A simple TextField would be clearer and less error-prone.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 509

---

sergiotejon:
## 🟡 LOW Issue

Toggling develop ON silently injects 'develop' as a default label, which may cause unexpected config writes.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 498

---

sergiotejon:
## 🟡 LOW Issue

ResolvedPRMetadata was exported solely for one call site in main.go. A dedicated public helper or keeping it unexported with a thin wrapper would be cleaner.

**Location:** `daemon/internal/config/config.go` line 419

---

sergiotejon (changes requested):

PR adds missing global settings fields (PR assignee, draft, issue/implement prompts) across daemon and Flutter app. Contains one compile error and several design issues.

**6 issue(s) found** — see individual comments above for details.

### Suggestions

- Fix the compile error: replace 'initialValue' with 'value' on DropdownButtonFormField
- Use bool? for globalPRDraft to distinguish 'unset' from 'explicitly false', matching the daemon's *bool semantics
- Replace AutocompleteChipField with a simple TextField for the single-value PR assignee field
- Verify that the TOML keys written under [ai.pr_metadata] (pr_assignee, pr_draft) match what the daemon's TOML parser expects for the nested section (the struct tags show 'pr_assignee' and 'pr_draft' at the flat level, but the nested PRMetadataConfig may use 'assignee')

---

sergiotejon:
## ⚠️ MEDIUM Issue

pr_assignee and pr_draft are written under [ai.pr_metadata] in TOML, but the TOML keys use the flat-field names (pr_assignee, pr_draft) rather than nested-section names (assignee, draft), which may cause a mismatch with existing daemon parsing of the nested section.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 253

---

Muriano:
## Response to AI Review Comments

### 1. 🔴 HIGH — `initialValue` does not exist on `DropdownButtonFormField`

**Not a bug.** `initialValue` is inherited from `FormField<T>` and available since Flutter 3.33. The `value` parameter is the one that triggers `deprecated_member_use`. The app compiles and runs without errors — verified with `flutter analyze --no-fatal-infos` (zero issues) and `flutter build macos --release` (success). False positive.

### 2. 🟡 LOW — Toggling develop ON silently injects 'develop' label

**By design.** The develop feature requires at least one label to be active (labels determine which issues get a branch + PR). Without a default, the switch would turn ON but nothing would happen — worse UX. The user sees the label immediately in the chip field below the switch and can edit/remove it. This matches the same pattern in the per-repo detail screen.

### 3. ⚠️ MEDIUM — AutocompleteChipField misused for single-value PR Assignee

**Deliberate choice for visual consistency.** An earlier version used `TextFormField`, which was flagged in manual review because it broke visual coherence with the adjacent PR Reviewers and PR Labels fields (both `AutocompleteChipField`). The current implementation uses `AutocompleteChipField` with `.first` to maintain a uniform look across the Develop section. The single-value constraint is acceptable — the daemon also accepts a single assignee string.

### 4. ⚠️ MEDIUM — `globalPRDraft` should be `bool?` to match daemon's `*bool`

**Acceptable trade-off for a switch UI.** A `SwitchListTile` has two states (on/off) — there's no visual representation for "unset". The diff logic in `_computeGlobalDiff` only sends `pr_draft` when the value differs from what was loaded, so if the daemon returned no `pr_draft` (parsed as `false`) and the user doesn't touch the switch, nothing is sent. The three-state semantic is only needed at the per-repo level (where `null` means "inherit global"), not at the global level where the value IS the default.

### 5. 🟡 LOW — `ResolvedPRMetadata` exported breaks encapsulation

**Standard Go pattern.** The method merges flat `[ai]` fields with nested `[ai.pr_metadata]` — it's a public API of the Config type that any consumer (main.go, tests, future CLI) can use. Wrapping it in a thin helper would add indirection with no benefit. The `config` package already exports `Validate()`, `AIForRepo()`, and `MergeStoreLayer()` following the same pattern.

### 6. ⚠️ MEDIUM — TOML key mismatch (`pr_assignee` vs `assignee` in nested section)

**Not a mismatch.** The `PRMetadataConfig` struct uses `toml:"pr_assignee"` and `toml:"pr_draft"` — identical to the flat-level tags. Both `[ai].pr_assignee` and `[ai.pr_metadata].pr_assignee` use the same key name. Verified in `daemon/internal/config/config.go` lines 298-312. False positive.

sergiotejon:
## ⚠️ MEDIUM Issue

DropdownButtonFormField is constructed with 'initialValue' named parameter. While the author claims this compiles on Flutter 3.33 via the FormField<T> base class, 'value' is the standard DropdownButtonFormField parameter for controlling selection state. Using 'initialValue' bypasses DropdownButtonFormField's own state management and may cause the dropdown to not reflect changes correctly. Verify this works as intended or switch to 'value'.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 537

---

sergiotejon:
## 🟡 LOW Issue

The develop switch onChanged logic is inverted from the previous version. Previously `if (!v)` cleared labels; now `if (v)` adds a default. However the old code only ran the clear branch — enabling had no effect. The new code correctly handles both directions, but silently injecting 'develop' as a default label is a minor UX surprise.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 495

---

sergiotejon:
## 🟡 LOW Issue

globalPRDraft as non-nullable bool (default false) conflates 'unset' with 'explicitly false'. The author's argument that global level doesn't need three-state is reasonable, but if the daemon ever distinguishes nil from false at the global level, this will silently override.

**Location:** `flutter_app/lib/core/models/config_model.dart` line 370

---

sergiotejon (approved):

PR adds missing global settings fields (issue_prompt, implement_prompt, pr_assignee, pr_draft) to daemon, config model, and Flutter UI. Most issues flagged by automated review were addressed by the author with reasonable justifications. The remaining concern is the DropdownButtonFormField usage.

### Suggestions

- Consider using 'value' instead of 'initialValue' on DropdownButtonFormField for clarity and conventional usage, even if 'initialValue' compiles.
- Add a brief inline comment explaining why the develop toggle injects a default 'develop' label to help future maintainers.

---

sergiotejon:
## 🔴 HIGH Issue

DropdownButtonFormField uses 'initialValue' (inherited from FormField<T>) instead of the standard 'value' parameter. Even if it compiles on Flutter 3.33, this bypasses DropdownButtonFormField's own state management and may cause the selected value to not update correctly on rebuild. Use 'value' instead.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 537

---

sergiotejon:
## ⚠️ MEDIUM Issue

globalPRDraft is non-nullable bool defaulting to false, collapsing 'unset' and 'explicitly false'. If the daemon ever distinguishes nil from false at global level, this will silently force false. Consider using bool? for future-proofing.

**Location:** `flutter_app/lib/core/models/config_model.dart` line 370

---

sergiotejon:
## 🟡 LOW Issue

AutocompleteChipField used for a single-value PR assignee field, constrained via .first. While visually consistent, it's semantically misleading and fragile if the user adds multiple chips.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 509

---

sergiotejon:
## 🟡 LOW Issue

Toggling develop ON silently injects 'develop' as a default label. Reasonable UX choice but could surprise users who expect an empty label list after enabling.

**Location:** `flutter_app/lib/features/config/config_screen.dart` line 498

---

sergiotejon:
## 🟡 LOW Issue

resolvedPRMetadata exported to ResolvedPRMetadata solely for main.go. Minor encapsulation concern; a thin public wrapper returning a struct would be cleaner.

**Location:** `daemon/internal/config/config.go` line 419

---

sergiotejon (changes requested):

PR adds missing global settings fields (PR assignee, draft, issue/implement prompts) across daemon and Flutter app. The main compile-risk issue around initialValue has been debated but warrants verification. The TOML key naming for nested pr_metadata fields needs careful confirmation.

### Suggestions

- Replace 'initialValue' with 'value' in the _agentDropdown DropdownButtonFormField to use the standard parameter and avoid subtle state management issues.
- Consider making globalPRDraft a bool? in AppConfig to preserve three-state semantics and match the daemon's *bool, using a tri-state UI or treating null as 'use daemon default'.
- Add a guard in the AutocompleteChipField for PR Assignee to prevent adding more than one chip, or switch to a simple TextField.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

The PR restores parity between global settings in the daemon and Flutter app by adding missing fields (pr_assignee, pr_draft, issue_prompt, implement_prompt) and fixes the broken develop switch toggle. Changes are coherent across daemon serialization, Flutter model, TOML writer, diff computation, and UI. Author has addressed prior AI review comments with reasonable justifications. A few minor concerns remain around UI semantics and encapsulation, but none are blocking.

### Issues

⚠️ **flutter_app/lib/core/models/config_model.dart:370** — globalPRDraft is a non-nullable bool defaulting to false, which conflates 'unset' with 'explicitly false'. When the daemon omits pr_draft, the Flutter side treats it as false. The diff logic in _computeGlobalDiff mitigates accidental writes (only sends on change), so impact is limited — but using bool? would preserve daemon parity more faithfully and future-proof against the daemon distinguishing nil from false globally.
⚠️ **flutter_app/lib/features/config/config_screen.dart:509** — AutocompleteChipField is used for the single-value PR Assignee field, constrained by .first. The author cites visual consistency, which is reasonable, but if the user adds multiple chips only the first is kept silently — confusing UX. Consider enforcing a max of one chip or showing a hint when additional entries are dropped.
⚠️ **flutter_app/lib/features/config/config_screen.dart:537** — DropdownButtonFormField uses the 'initialValue' parameter (inherited from FormField<T> on Flutter 3.33+). Author verified it compiles and runs, but 'initialValue' is only applied on first form state creation — subsequent external changes to `value` after state is built may not propagate. The ValueKey('$label-$effective') forces a rebuild of the widget when the value changes, which works around this, but it's fragile. Using the `value` parameter (if still supported) or a controlled approach would be more idiomatic.
🟡 **flutter_app/lib/features/config/config_screen.dart:498** — Toggling develop ON silently injects 'develop' as a default label. Author argues this is intentional UX (section is inert without a label). Acceptable, but consider a brief helper text or toast to make the injection visible to users.
🟡 **daemon/internal/config/config.go:419** — resolvedPRMetadata was exported as ResolvedPRMetadata solely for main.go. Author's justification (consistent with other exported Config methods) is reasonable. Returning four positional values is less ergonomic than a struct; consider returning a ResolvedPRMetadata struct for readability.
🟡 **daemon/cmd/heimdallm/main.go:464** — Indentation/alignment of the new `issue_prompt` and `implement_prompt` keys in the map literal is inconsistent with the surrounding entries (trailing spaces differ). Minor style nit — gofmt should normalize.

### Suggestions

- Change globalPRDraft to bool? in the Flutter model to preserve the daemon's three-state *bool semantics; update copyWith, fromJson, writer, and diff logic accordingly.
- Replace AutocompleteChipField for PR Assignee with a TextFormField, or enforce maxChips=1 in AutocompleteChipField, to remove the silent .first truncation.
- If tooling flags DropdownButtonFormField.initialValue in the future, switch to the canonical `value` parameter to avoid relying on FormField base-class behavior.
- Consider returning a struct from ResolvedPRMetadata() instead of four positional values to improve call-site readability and reduce the risk of argument-order bugs.
- Surface the auto-injected 'develop' label via a subtle helper text so the implicit write is visible to users.

---

Muriano:
## Re-review response

The re-review repeats the same 5 points already addressed in the previous response. No new issues were raised.

**On the `initialValue` insistence:** The build passes. `flutter analyze --no-fatal-infos` returns zero issues. `flutter build macos --release` succeeds. This is not a compile error — it's a false positive. The parameter exists on `FormField<T>`, from which `DropdownButtonFormField` inherits. Using `value` instead would trigger `deprecated_member_use`.

All other points were justified in the previous response and remain unchanged. No code changes needed.

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

PR adds four missing global settings fields (pr_assignee, pr_draft, issue_prompt, implement_prompt) to achieve daemon/Flutter parity and fixes a broken develop switch. The Go and Riverpod wiring is sound, but the Flutter first-run TOML writer emits wrong keys under [ai.pr_metadata] that will not round-trip through the daemon parser — undermining the parity goal of this PR. The author's response to the AI reviewer on this point is incorrect.

### Issues

🔴 **flutter_app/lib/core/setup/first_run_setup.dart:256** — Writes `pr_assignee` and `pr_draft` under `[ai.pr_metadata]`, but `PRMetadataConfig` in daemon/internal/config/config.go lines 298-303 uses toml tags `assignee` and `draft` (not the flat-level `pr_assignee`/`pr_draft`). The daemon will silently drop these keys on load. The author explicitly dismissed this AI review comment as a false positive, but the struct definition contradicts that response. Fix by writing `assignee = "..."` and `draft = true` under `[ai.pr_metadata]`, OR by writing `pr_assignee`/`pr_draft` as flat `[ai]` fields (which do use those tags, lines 255-256).
🔴 **flutter_app/lib/features/config/config_providers.dart:117** — The diff packs `pr_assignee`/`pr_draft` into an `aiDiff['pr_metadata']` sub-map. Whether this round-trips correctly depends on the daemon's PATCH handler — if it maps the sub-map onto `PRMetadataConfig`, the same key-name mismatch as in first_run_setup.dart applies (struct tags are `assignee`/`draft`). Verify the update endpoint and align keys with whichever struct the daemon decodes into.
⚠️ **flutter_app/lib/core/models/config_model.dart:342** — `globalPRDraft` is non-nullable `bool` defaulting to `false`, collapsing daemon's three-state `*bool` (nil/true/false) into two states. Author's argument that global-level doesn't need tri-state is defensible, but the diff-send logic in config_providers.dart will send `pr_draft: false` whenever the user toggles on-then-off, which is distinguishable from 'never configured' on the daemon side and may overwrite a legitimately-unset default. Consider `bool?` for parity, or document this behavioral divergence.
🟡 **daemon/cmd/heimdallm/main.go:464** — `issue_prompt` and `implement_prompt` are emitted unconditionally, even when empty. Inconsistent with how other optional PR-metadata fields are only added when set (via the `if assignee != ""` guards just below). Harmless — Flutter treats empty string as 'default' — but inconsistent with surrounding style.
🟡 **flutter_app/lib/features/config/config_screen.dart:509** — Using AutocompleteChipField for a single-value PR assignee permits adding multiple chips; only `.first` is used, so additional chips are silently discarded. Author's visual-consistency rationale is reasonable, but consider clamping input to 1 chip (e.g., disable the field when non-empty) to prevent silent data loss.
🟡 **flutter_app/lib/features/config/config_screen.dart:8** — Importing `agentsProvider` from `agents_screen.dart` couples a Riverpod provider to a view file. Minor structure smell — provider would be better placed in `agents/agent_providers.dart` — but not blocking.

### Suggestions

- Grep `daemon/internal/config` for the PATCH/store-layer handler that consumes `ai.pr_metadata` from the Flutter diff payload and confirm whether it decodes via `PRMetadataConfig` (tags: assignee/draft) or via the flat AI struct (tags: pr_assignee/pr_draft). Align Flutter's key names accordingly in both config_providers.dart and first_run_setup.dart, and add a round-trip test: save → reload → expect values preserved.
- On the `initialValue` vs `value` debate: author is correct — Flutter 3.29+ moved `value` to `initialValue` on FormField<T>, and `value` is deprecated on DropdownButtonFormField. The AI reviewer's repeated flag is a false positive. No change needed.
- Consider exposing `issue_prompt`/`implement_prompt` on per-repo overrides (RepoAI) in a follow-up if you want the same three-level resolution (repo > org > global) already applied to PR metadata — otherwise the global defaults are the only knob.
- The develop-switch UX improvement (hiding child fields when off, injecting a default 'develop' label on toggle-on) is a clear win over the previous broken state. The silent-injection concern is minor since the user sees the chip immediately.

---

PR #226, opened 2026-04-22
feat: implement #222 — bug: triage loop — RecomputeGrace too short for slow triages

### LINKED ISSUE #222: bug: triage loop — RecomputeGrace too short for slow triages
## Problem

Issues get re-triaged in an infinite loop when the triage takes longer than 30 seconds.

## Root cause

The dedup check in `fetcher.go:159` uses:

```go
cutoff := latest.CreatedAt.Add(RecomputeGrace) // CreatedAt + 30s
}
```

`CreatedAt` is set when the review is **stored in SQLite** (before the comment is posted). But `UpdatedAt` bumps when the comment is **published on GitHub** (after). If the triage takes >30s:

```
T+0s    review stored → CreatedAt = T+0
T+45s   comment posted → UpdatedAt = T+45
T+30s   cutoff = T+0 + 30s = T+30
T+60s   poll: UpdatedAt(T+45) > cutoff(T+30) → RE-TRIAGE → loop!
```

Each re-triage posts another comment, bumping UpdatedAt again → infinite loop.

## Fix

Use the **comment publication time** as reference instead of `CreatedAt`. Options:

### Option A: Increase RecomputeGrace
Simple but fragile — any triage exceeding the grace still loops.

### Option B: Record comment timestamp
Store the time the comment was posted (not the review creation time) and use that as cutoff:

```go
cutoff := latest.CommentedAt.Add(RecomputeGrace)
```

### Option C: Compare against own comment
After posting, check if the latest comment on the issue is from the bot. If so, skip — our own comment triggered the UpdatedAt bump.

**Recommendation:** Option B is the most robust. Add a `commented_at` field to `IssueReview` and set it after `PostComment` succeeds.

## Impact

Any issue where triage takes >30s enters an infinite triage loop, generating multiple identical comments.

sergiotejon:
Auto-generated by Heimdallm in response to #222.

Closes #222

### DIFF
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -240,13 +240,15 @@ func (c *Client) SubmitReview(repo string, number int, body, event string) (int6
 
 // PostComment posts a general comment on a PR (issue comment).
 // Used in multi-feedback mode to post one comment per issue before the formal review.
-func (c *Client) PostComment(repo string, number int, body string) error {
+// Returns the comment's creation timestamp as reported by GitHub so callers
+// can record when the comment actually landed (see issue #222).
+func (c *Client) PostComment(repo string, number int, body string) (time.Time, error) {
 	path := fmt.Sprintf("/repos/%s/issues/%d/comments", repo, number)
 	payload := map[string]any{"body": body}
 	data, _ := json.Marshal(payload)
 	req, err := http.NewRequest("POST", c.baseURL+path, strings.NewReader(string(data)))
 	if err != nil {
-		return fmt.Errorf("github: post comment: %w", err)
+		return time.Time{}, fmt.Errorf("github: post comment: %w", err)
 	}
 	req.Header.Set("Authorization", "Bearer "+c.token)
 	req.Header.Set("Accept", "application/vnd.github+json")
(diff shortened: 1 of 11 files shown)


Muriano (approved):

Clean, well-scoped fix for the triage-loop race condition. The core logic change (using CommentedAt instead of CreatedAt as the grace-period anchor) is correct and well-tested. The PostComment signature change is propagated consistently across all callers and interfaces. Migration is idempotent and follows existing patterns. Two minor observability gaps noted but nothing blocks merge.

### Suggestions

- In client.go:272, log a warning when json.Unmarshal fails before falling back to time.Now().UTC(), e.g.: slog.Warn("github: PostComment: failed to parse created_at from response, using local time", "err", err)
- Consider whether the migration DEFAULT '' for a DATETIME column should use NULL instead of empty string for semantic clarity — though the current approach is consistent with the COALESCE usage in queries and the Go-side empty-string check, so this is purely a style preference.

---

Muriano:
## ⚠️ MEDIUM Issue

When json.Unmarshal of the 201 response body fails, the error is silently swallowed and time.Now().UTC() is returned with nil error. This means if GitHub ever changes the response schema or the body is unexpectedly truncated, callers will silently get an inaccurate timestamp with no indication anything went wrong. At minimum a slog.Warn should be emitted so operators can detect API drift. Impact: CommentedAt could be slightly off from the real GitHub timestamp, which in edge cases could affect the RecomputeGrace cutoff logic this PR is fixing.

**Location:** `daemon/internal/github/client.go` line 272

---

PR #221, opened 2026-04-22
feat: implement #220 — fix: issue re-triage ignores context from previous triage and new comments

sergiotejon:
Auto-generated by Heimdallm in response to #220.

Closes #220

PR #231, opened 2026-04-22
feat: implement #220 — fix: issue re-triage ignores context from previous triage and new comments

### LINKED ISSUE #220: fix: issue re-triage ignores context from previous triage and new comments
## Problem

When Heimdallm re-triages an issue (triggered by new comments bumping `UpdatedAt`), it injects all comments into the prompt without distinguishing between:
- Its own previous triage comment
- Author/team responses to the triage
- New context added after the triage

The LLM has no signal that says "you already triaged this, here's what changed since then". It re-analyzes from scratch and may repeat the same findings or ignore the new context that triggered the re-triage.

Same root cause as PR re-review issue (#90), but for the issue tracking pipeline.

## Current behavior

```
1. Issue created → Heimdallm triages → posts comment
2. Author adds comment with new context
3. UpdatedAt bumps → Heimdallm re-triages
4. All comments (including own triage) injected as flat list
5. LLM doesn't know what changed → may repeat triage verbatim
```

## Expected behavior

Re-triage should include structured context:

```
IMPORTANT: This is a RE-TRIAGE. You previously triaged this issue.
- Focus on NEW information since your last triage
- Update your assessment if the new comments change the scope/severity
- Do NOT repeat your previous analysis verbatim

## Your previous triage (severity: high)
- Category: bug
- Suggested assignee: @sergiotejon
- Next steps: flatten elevenlabs proxy files...

## New discussion since last triage
@author: Actually we want to support subdirectories long-term
@reviewer: Agreed, the template needs to handle glob patterns
```

## Fix

Port the re-review context pattern from `pipeline/review_context.go` to the issue pipeline:

1. Query `LatestIssueReview` before building the prompt
2. Partition comments into before/after last triage
3. Format previous triage findings as structured context
4. Filter bot's own comments from "new discussion"
5. Inject RE-TRIAGE instruction into the prompt

## Files to modify

- `daemon/internal/issues/pipeline.go` — query previous review, build context
- `daemon/internal/issues/prompt.go` — add `ReviewContext` field to `IssueContext`, inject into template
- Reuse `partitionComments` from `daemon/internal/pipeline/review_context.go` (or extract to shared package)

## References

- PR re-review context: #90 (implemented in `pipeline/review_context.go`)
- Same pattern, applied to issues instead of PRs

sergiotejon:
Auto-generated by Heimdallm in response to #220.

Closes #220

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -163,8 +163,10 @@ func main() {
 
 	p := pipeline.New(s, ghClient, exec, &notifyWithSSE{notifier: notifier})
 
-	// Resolve bot login for re-review context filtering.
+	// Resolve bot login for re-review and re-triage context filtering.
+	var resolvedBotLogin string
 	if login, err := ghClient.AuthenticatedUser(); err == nil {
+		resolvedBotLogin = login
 		p.SetBotLogin(login)
 		slog.Info("bot login resolved", "login", login)
 	} else {
@@ -176,6 +178,9 @@ func main() {
 	// an issue that is classified as review_only, so this dep is harmless
 	// when auto_implement is not in use.
 	issuePipe := issuepipeline.New(s, ghClient, exec, issuepipeline.NewGitExec(), broker, &notifyWithSSE{notifier: notifier})
+	if resolvedBotLogin != "" {
+		issuePipe.SetBotLogin(resolvedBotLogin)
+	}
 	issueFetcher := issuepipeline.NewFetcher(ghClient, s, issuePipe)
 	srv := server.New(s, broker, p, apiToken)
 	srv.SetConfigPath(cfgPath)
--- a/daemon/internal/issues/pipeline.go
+++ b/daemon/internal/issues/pipeline.go
@@ -7,7 +7,9 @@ package issues
 
 import (
 	"context"
+	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
 	"strings"
@@ -171,13 +173,19 @@ type Pipeline struct {
 	git      GitOps
 	broker   Publisher
 	notify   Notifier
+	botLogin string
 }
+
+// SetBotLogin sets the GitHub login of the bot account. Used to filter
+// the bot's own comments from re-triage discussion context.
+func (p *Pipeline) SetBotLogin(login string) { p.botLogin = login }
 
 // issueStore is the subset of *store.Store the pipeline needs. Kept narrow
 // so tests can substitute a fake without bringing in SQLite.
 type issueStore interface {
 	UpsertIssue(i *store.Issue) (int64, error)
 	InsertIssueReview(r *store.IssueReview) (int64, error)
+	LatestIssueReview(issueID int64) (*store.IssueReview, error)
 	UpsertPR(pr *store.PR) (int64, error)
 }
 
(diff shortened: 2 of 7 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Well-structured fix that injects previous triage context into re-triages. Tests are thorough. Main concerns are code duplication between the two pipeline paths, silently-ignored JSON unmarshal errors, and a fallback gap when bot login resolution fails at startup.

### Issues

⚠️ **daemon/internal/issues/pipeline.go:293** — The 14-line re-triage context fetch/build block is duplicated verbatim between runReviewOnly (line ~293) and runAutoImplement (line ~441). Extract to a helper like p.buildReTriageContext(issueID, comments) to keep both paths in sync as the logic evolves.
⚠️ **daemon/internal/issues/triage_context.go:44** — json.Unmarshal errors are silently ignored in two places (prevTriageJSON and prevSuggestionsJSON). If stored JSON is malformed, triageFinding/suggestions silently become zero values and the LLM gets an incomplete context with no diagnostic trail. At minimum use `_ = json.Unmarshal(...)` to signal intent, and ideally slog.Warn on error since the store is expected to hold valid JSON.
⚠️ **daemon/cmd/heimdallm/main.go:165** — If ghClient.AuthenticatedUser() fails (transient GitHub error at startup), resolvedBotLogin stays empty for the entire daemon lifetime. Then in buildIssueTriageContext, strings.EqualFold(c.Author, "") is false for all real comments, so the bot's own previous triage comments will leak into the 'New discussion since last triage' section on every subsequent re-triage. Consider retrying on first use or logging more visibly that bot-comment filtering is disabled.
🟡 **daemon/internal/issues/triage_context.go:12** — partitionComments returns both before and after, but the single caller only uses `after` (discards `before` with `_`). Either inline the filter at the call site or keep the helper but document why `before` is exposed — as written it's dead output surface that may mislead future callers.
🟡 **daemon/internal/issues/triage_context.go:36** — The early-return guard `if prevTriageJSON == "" && lastTriageAt.IsZero()` is effectively unreachable from production callers — runReviewOnly/runAutoImplement only invoke this when prevReview != nil, which means CreatedAt is always non-zero. The guard is only exercised by tests. Consider either removing it or documenting that callers are expected to short-circuit on nil prevReview.
🟡 **daemon/internal/issues/triage_context.go:84** — Previous bot triage comments are filtered, but new bot comments (e.g. re-triage comments posted between two re-triages) will also be filtered — which is probably desired, but it means if bot misbehaves and a human responds to bot, only the human reply shows up without the bot context. Worth a brief comment explaining the filter's rationale so future maintainers don't relax it.

### Suggestions

- Extract the re-triage fetch+build logic into a Pipeline helper method to eliminate the duplication between runReviewOnly and runAutoImplement.
- Log json.Unmarshal failures in buildIssueTriageContext at slog.Warn level — malformed stored triage JSON is a real signal of upstream bugs and should not be swallowed.
- Consider passing resolvedBotLogin to both pipelines via constructor rather than post-hoc SetBotLogin, so the dependency is explicit and can't be forgotten for a future third pipeline.
- Add an integration-level assertion that the bot's own comment does not appear in the re-triage prompt (the unit tests cover it, but an end-to-end test would catch wiring regressions in main.go).

---

Muriano:
## ⚠️ MEDIUM Issue

json.Unmarshal errors are silently discarded (lines 51 and 72). If prevTriageJSON or prevSuggestionsJSON contain malformed JSON, the function silently proceeds with zero-value structs, producing misleading context (e.g. 'severity: unknown') with no log trace. Impact: hard to debug when stored JSON is corrupt. Fix: check the error and log at Warn level, similar to the sql.ErrNoRows pattern used in the caller.

**Location:** `daemon/internal/issues/triage_context.go` line 51

---

Muriano:
## ⚠️ MEDIUM Issue

The 15-line re-triage context building block (fetch LatestIssueReview, check error, call buildIssueTriageContext) is copy-pasted identically in runReviewOnly (line 293) and runAutoImplement (line 441). The comment even says 'same as review_only'. Impact: future changes must be applied in two places, risk of divergence. Fix: extract a helper method on Pipeline, e.g. `func (p *Pipeline) buildReTriageContext(issueID int64, comments []github.Comment) string`.

**Location:** `daemon/internal/issues/pipeline.go` line 293

---

Muriano (approved):

Well-structured feature that injects previous triage context into re-triage prompts, with good test coverage and proper bot-comment filtering. Two confirmed issues: duplicated logic block and silently discarded JSON parse errors.

### Suggestions

- partitionComments places comments with CreatedAt exactly equal to cutoff into 'before' (uses strict After). This is likely correct but worth a brief comment clarifying the boundary behavior, since a bot comment posted at the exact same instant as the triage timestamp would be excluded from new discussion.
- Consider adding a test case where prevTriageJSON is invalid JSON (e.g. '{broken') to document the current silent-fallback behavior and guard against regressions if error handling is added later.

---

Muriano (changes requested):
## 🤖 Heimdallm AI Review

Las dos issues principales reportadas en reviews anteriores siguen sin resolverse en el diff actual: el bug de la expresión CV que excluye servidores con 0 líderes, y el for: 1m demasiado agresivo para una alerta critical+oncall.

### Issues

🔴 **prometheus-rules/nats/leadership-skew.yaml:18** — Sin resolver desde review anterior. La expresión CV `count by (server_name)(nats_stream_subject_count{is_stream_leader="true"})` excluye servidores con 0 líderes de stddev()/avg(). Tras un scale-up (el escenario exacto que monitoriza), los nuevos nodos arrancan con 0 líderes y son invisibles, subestimando el CV (ej: 5→7 nodos → CV=0 en vez de ~0.63). Fix: `count by (server_name)(nats_stream_subject_count{is_stream_leader="true"}) or on(server_name) 0 * group by (server_name)(up{job=~"nats.*"})`.
⚠️ **prometheus-rules/nats/route-health.yaml:17** — Sin resolver desde review anterior. NatsRouteSlowConsumer usa `rate()[5m] > 0` con `for: 1m` y es critical+oncall. Un único incremento del contador en 5 minutos dispara un page tras solo 1 minuto. La alerta existente NatsServerSlowConsumers (slow-consumers.yaml) usa el mismo patrón pero con `for: 5m` y severity warning. Para una alerta que despierta gente, `for: 1m` es demasiado agresivo. Recomendación: subir a `for: 3m` o `for: 5m`, o usar umbral `> 0.01`.

---

sergiotejon:
Duplicate: issue #220 already resolved in PR #221 (merged).

PR #233, opened 2026-04-22
feat: implement #222 — bug: triage loop — RecomputeGrace too short for slow triages

### LINKED ISSUE #222: bug: triage loop — RecomputeGrace too short for slow triages

sergiotejon:
Auto-generated by Heimdallm in response to #222.

Closes #222

### DIFF
--- a/cli/internal/api/client.go
+++ b/cli/internal/api/client.go
@@ -111,6 +111,7 @@ type IssueReview struct {
 	ActionTaken string          `json:"action_taken"`
 	PRCreated   int             `json:"pr_created"`
 	CreatedAt   time.Time       `json:"created_at"`
+	CommentedAt time.Time       `json:"commented_at"`
 }
 
 type Issue struct {
--- a/daemon/internal/issues/pipeline.go
+++ b/daemon/internal/issues/pipeline.go
@@ -292,12 +292,16 @@ func (p *Pipeline) runReviewOnly(ctx context.Context, issue *github.Issue, issue
 	var triageCtx string
 	prevReview, _ := p.store.LatestIssueReview(issueID)
 	if prevReview != nil {
+		lastTriageRef := prevReview.CommentedAt
+		if lastTriageRef.IsZero() {
+			lastTriageRef = prevReview.CreatedAt
+		}
 		triageCtx = buildTriageContext(
 			prevReview.Triage,
 			prevReview.Suggestions,
 			prevReview.Summary,
 			extractSeverity(prevReview.Triage),
-			prevReview.CreatedAt,
+			lastTriageRef,
 			comments,
 			p.botLogin,
 		)
@@ -445,12 +449,16 @@ func (p *Pipeline) runAutoImplement(ctx context.Context, issue *github.Issue, is
 	var triageCtx string
 	prevReview, _ := p.store.LatestIssueReview(issueID)
 	if prevReview != nil {
+		lastTriageRef := prevReview.CommentedAt
+		if lastTriageRef.IsZero() {
+			lastTriageRef = prevReview.CreatedAt
+		}
 		triageCtx = buildTriageContext(
 			prevReview.Triage,
 			prevReview.Suggestions,
 			prevReview.Summary,
 			extractSeverity(prevReview.Triage),
-			prevReview.CreatedAt,
+			lastTriageRef,
 			comments,
 			p.botLogin,
 		)
--- a/daemon/internal/store/store.go
+++ b/daemon/internal/store/store.go
@@ -105,7 +105,8 @@ CREATE TABLE IF NOT EXISTS issue_reviews (
   suggestions  TEXT NOT NULL DEFAULT '[]',
   action_taken TEXT NOT NULL DEFAULT 'review_only',
   pr_created   INTEGER NOT NULL DEFAULT 0,
-  created_at   DATETIME NOT NULL
+  created_at   DATETIME NOT NULL,
+  commented_at DATETIME NOT NULL DEFAULT ''
 );
 
 CREATE TABLE IF NOT EXISTS activity_log (


ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

The change introduces a `CommentedAt` timestamp to better represent 'the last moment the bot spoke' as the reference point for detecting new comments after a triage. The core idea is sound, but the schema change lacks a migration path for existing databases and uses a questionable default value for a DATETIME column. Additionally, the diff does not show where `CommentedAt` is populated, making it impossible to verify the feature actually works end-to-end.

### Issues

🔴 **daemon/internal/store/store.go:109** — Schema change only applies to new databases via `CREATE TABLE IF NOT EXISTS`. Existing deployments will not get the `commented_at` column, so queries that SELECT or INSERT this column will fail with 'no such column' at runtime. A proper migration (ALTER TABLE ADD COLUMN, or a versioned migration step) is required.
🔴 **daemon/internal/store/store.go:109** — `commented_at DATETIME NOT NULL DEFAULT ''` uses an empty string as the default for a DATETIME column. SQLite has no native DATETIME type (affinity only), but scanning an empty string into Go's `time.Time` via database/sql typically fails or yields undefined behavior depending on the driver. A safer default would be `'0001-01-01 00:00:00'` (Go's zero time) or making the column NULLable with `sql.NullTime`, which would play better with the subsequent `IsZero()` check in pipeline.go.
🔴 **daemon/internal/store/store.go:109** — The diff adds the `commented_at` column but does not show any INSERT/UPDATE statements being updated to write it, nor the SELECT/scan path that materializes it into `IssueReview.CommentedAt`. If those call sites were not updated, `CommentedAt` will always be zero and the fallback to `CreatedAt` will always fire — silently nullifying the fix for #222. The PR should include the write path and, ideally, a test that verifies `CommentedAt` round-trips.
🟡 **cli/internal/api/client.go:114** — The `CommentedAt` field is added to the API struct without a documented meaning. Consider a short comment clarifying the semantic difference vs. `CreatedAt` (e.g., 'timestamp of the bot's last comment on the issue; used as the reference for detecting new comments').
🟡 **daemon/internal/issues/pipeline.go:295** — The same 4-line `lastTriageRef` fallback is duplicated in `runReviewOnly` and `runAutoImplement`. Extracting a small helper like `lastTriageReference(prevReview)` would reduce duplication and make the intent clearer at both call sites.

### Suggestions

- Add an explicit migration step (ALTER TABLE ... ADD COLUMN commented_at DATETIME) guarded by a schema-version check, so existing daemons pick up the new column on upgrade.
- Change the default to Go's zero time ('0001-01-01 00:00:00') or switch to a NULLable column with `sql.NullTime`, so the `IsZero()` fallback in pipeline.go behaves deterministically across drivers.
- Extend the PR with the write path: update the INSERT (and any UPDATE) that persists an `IssueReview` to set `commented_at` to the time the bot's triage comment was actually posted on GitHub, not `time.Now()` at row creation — otherwise the fix still loses the grace window when the triage takes longer than expected.
- Add a regression test reproducing the #222 scenario: a slow triage whose bot comment lands well after `CreatedAt`, then a user comment shortly after — the pipeline should NOT re-triage because the user comment is older than `CommentedAt`.
- Extract the `lastTriageRef` fallback into a helper to avoid drift between `runReviewOnly` and `runAutoImplement`.

---

Muriano:
## 🔴 HIGH Issue

The schema uses `CREATE TABLE IF NOT EXISTS`, so existing databases that already have the `issue_reviews` table will never get the new `commented_at` column. This will cause runtime SQL errors (column not found) on any existing installation. An `ALTER TABLE issue_reviews ADD COLUMN commented_at ...` migration is required. Impact: crashes on all existing deployments. Fix: add an explicit migration step (e.g., `ALTER TABLE IF NOT EXISTS` or a versioned migration) to add the column to pre-existing tables.

**Location:** `daemon/internal/store/store.go` line 108

---

Muriano:
## ⚠️ MEDIUM Issue

`commented_at DATETIME NOT NULL DEFAULT ''` uses an empty string as the default for a DATETIME column. While SQLite allows this due to type affinity, when the Go SQL driver scans an empty string into `time.Time` the behavior is driver-dependent — it may produce a parse error rather than a zero time. The Go code relies on `IsZero()` to detect missing values, but a failed parse won't yield a zero `time.Time`. Fix: use `DEFAULT '0001-01-01T00:00:00Z'` to produce a proper zero time that `IsZero()` will match, or make the column NULLable and handle `sql.NullTime` in Go.

**Location:** `daemon/internal/store/store.go` line 108

---

Muriano (changes requested):

Adds a `commented_at` timestamp to track when a review comment was posted, using it instead of `created_at` as the triage reference time. The logic is correct and the fallback to `created_at` when `commented_at` is zero is sound. However, there are two confirmed issues: a missing migration path for existing databases and a questionable SQLite default value for the new DATETIME column.

### Suggestions

- The `lastTriageRef` fallback logic is duplicated identically in `runReviewOnly` and `runAutoImplement`. Extract it into a small helper (e.g., `func triageReferenceTime(r *IssueReview) time.Time`) to keep the two paths in sync and reduce duplication.
- Consider adding a unit test that verifies the fallback behavior: when `CommentedAt` is zero, `CreatedAt` is used as the triage reference time.

---

sergiotejon:
Duplicate: issue #222 already resolved in PR #226 (merged).

PR #235, opened 2026-04-22
feat: implement #234 — bug: release pipeline state corrupted — tags without assets, release-please out of sync

### LINKED ISSUE #234: bug: release pipeline state corrupted — tags without assets, release-please out of sync
## Problem

The release pipeline has been in a corrupted state since ~v0.4.15. Multiple releases have been published with missing or incomplete assets (no CLI binaries, no DMG, no Linux packages). The root causes are interrelated:

### 1. Release-please state desync
Closing/reopening Release PRs manually caused release-please to lose track of its state. It creates tags pointing to wrong commits or believes a release was already published when it wasn't.

### 2. Tags pointing to wrong commits
Tags created during the chaos pointed to merge commits from feature branches instead of main HEAD. CI checks out the tag and finds missing files (`cli/go.mod not found`, `docker/ not found`).

### 3. Builds only run when release-please says so
All build jobs are gated on `release_created == true`. If release-please creates a tag but the builds fail, there's no way to re-trigger them. The release exists (with the tag) but has no assets. Manual tag recreation doesn't trigger builds.

### 4. GoReleaser Homebrew formula never worked
The `TAP_GITHUB_TOKEN` secret mapping and GoReleaser config had multiple issues (wrong secret name, `formats` vs `format` syntax, compilation errors). The formula has been updated manually twice.

## Current state

- v0.6.0 tag exists on correct commit (main HEAD) but has zero assets
- Homebrew formula manually set to v0.4.19 (last release with CLI binaries)
- `.release-please-manifest.json` says `"0.6.0"` — next release will be v0.6.1
- Code is healthy — all tests pass on main

### Short term
1. Add a manual dispatch trigger to the release workflow so builds can be re-run against any tag
2. Run it for v0.6.0 to populate assets

```yaml
on:
  push:
    branches: [main]
  workflow_dispatch:
    inputs:
      tag:
        description: 'Tag to build (e.g. v0.6.0)'
        required: true
```

### Long term
3. Consider separating release-please (tag creation) from build jobs (asset publishing) into two workflows. Build workflow triggers on tag push, not on release-please output.
4. Add a scheduled job that checks for releases with missing assets and re-triggers builds.
5. Automate Homebrew formula update as a separate simple step (curl checksums + update rb file) rather than relying on GoReleaser's brew integration.

## Affected releases

| Version | Status |
|---------|--------|
| v0.4.15 | Empty release (tag only) |
| v0.4.16 | Missing CLI binaries |
| v0.4.17 | Empty release |
| v0.4.18 | DMG + Linux OK, CLI failed |
| v0.4.19 | All assets present (last good) |
| v0.6.0 | Empty release (tag on correct commit) |

sergiotejon:
Auto-generated by Heimdallm in response to #234.

Closes #234

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,11 @@ name: Release
 on:
   push:
     branches: [main]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Existing tag to build assets for (e.g. v0.6.0)'
+        required: true
 
 env:
   REGISTRY: ghcr.io
@@ -12,6 +17,7 @@ env:
 
 jobs:
   release-please:
+    if: github.event_name == 'push'
     runs-on: ubuntu-latest
     permissions:
       contents: write
@@ -31,18 +37,23 @@ jobs:
   # ═══════════════════════════════════════════════════════════════════════════
   docker-build-push:
     needs: release-please
-    if: needs.release-please.outputs.release_created == 'true'
+    if: >-
+      always() &&
+      (needs.release-please.outputs.release_created == 'true' ||
+       github.event_name == 'workflow_dispatch')
     runs-on: ubuntu-latest
     permissions:
       contents: read
       packages: write
       attestations: write
       id-token: write
+    env:
+      TAG: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          ref: ${{ needs.release-please.outputs.tag_name }}
+          ref: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
@@ -61,9 +72,9 @@ jobs:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
             type=raw,value=latest
-            type=semver,pattern={{version}},value=${{ needs.release-please.outputs.tag_name }}
-            type=semver,pattern={{major}}.{{minor}},value=${{ needs.release-please.outputs.tag_name }}
-            type=semver,pattern={{major}},value=${{ needs.release-please.outputs.tag_name }}
+            type=semver,pattern={{version}},value=${{ env.TAG }}
+            type=semver,pattern={{major}}.{{minor}},value=${{ env.TAG }}
+            type=semver,pattern={{major}},value=${{ env.TAG }}
             type=sha,prefix=
 
       - name: Build and push Docker image
@@ -91,7 +102,10 @@ jobs:
   build-linux:
     name: Build Linux packages (.deb / .rpm / .AppImage)
     needs: release-please
-    if: needs.release-please.outputs.release_created == 'true'
+    if: >-
+      always() &&
+      (needs.release-please.outputs.release_created == 'true' ||
+       github.event_name == 'workflow_dispatch')
     runs-on: ubuntu-22.04
     permissions:
       contents: write
@@ -99,7 +113,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ needs.release-please.outputs.tag_name }}
+          ref: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
           fetch-depth: 0
 
       # ── Toolchains ──────────────────────────────────────────────────────────
@@ -149,7 +163,7 @@ jobs:
       - name: Build Flutter app (Linux release)
         working-directory: flutter_app
         run: |
-          VERSION="${{ needs.release-please.outputs.tag_name }}"
+          VERSION="${{ inputs.tag || needs.release-please.outputs.tag_name }}"
           VERSION="${VERSION#v}"
           flutter build linux --release \
             --build-name="$VERSION" \
@@ -158,7 +172,7 @@ jobs:
       # ── Prepare staging tree ──────────────────────────────────────────────────
       - name: Prepare staging directory
         run: |
-          VERSION="${{ needs.release-please.outputs.tag_name }}"
+          VERSION="${{ inputs.tag || needs.release-please.outputs.tag_name }}"
           VERSION="${VERSION#v}"
           BUNDLE="flutter_app/build/linux/x64/release/bundle"
 
@@ -305,7 +319,7 @@ jobs:
       - name: Upload to GitHub Release
         uses: softprops/action-gh-release@v2
         with:
-          tag_name: ${{ needs.release-please.outputs.tag_name }}
+          tag_name: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
           files: |
             ${{ env.DEB_PATH }}
             ${{ env.RPM_PATH }}
@@ -319,7 +333,10 @@ jobs:
   build-macos:
     name: Build Heimdallm.dmg
     needs: release-please
-    if: needs.release-please.outputs.release_created == 'true'
+    if: >-
+      always() &&
+      (needs.release-please.outputs.release_created == 'true' ||
+       github.event_name == 'workflow_dispatch')
     runs-on: macos-14
     permissions:
       contents: write
@@ -327,7 +344,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ needs.release-please.outputs.tag_name }}
+          ref: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
           fetch-depth: 0
 
       # ── Toolchains ──────────────────────────────────────────────────────────
@@ -369,7 +386,7 @@ jobs:
       - name: Build Flutter app (release)
         working-directory: flutter_app
         run: |
-          VERSION="${{ needs.release-please.outputs.tag_name }}"
+          VERSION="${{ inputs.tag || needs.release-please.outputs.tag_name }}"
           VERSION="${VERSION#v}"
           flutter build macos --release \
             --build-name="$VERSION" \
@@ -447,7 +464,7 @@ jobs:
       - name: Create DMG
         run: |
           APP="${APP}"
-          VERSION="${{ needs.release-please.outputs.tag_name }}"
+          VERSION="${{ inputs.tag || needs.release-please.outputs.tag_name }}"
           DMG="dist/Heimdallm-${VERSION}.dmg"
           mkdir -p dist
 
@@ -482,7 +499,7 @@ jobs:
       - name: Upload to GitHub Release
         uses: softprops/action-gh-release@v2
         with:
-          tag_name: ${{ needs.release-please.outputs.tag_name }}
+          tag_name: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
           files: ${{ env.DMG_PATH }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -493,14 +510,17 @@ jobs:
   goreleaser-cli:
     name: Build CLI binaries (GoReleaser)
     needs: release-please
-    if: needs.release-please.outputs.release_created == 'true'
+    if: >-
+      always() &&
+      (needs.release-please.outputs.release_created == 'true' ||
+       github.event_name == 'workflow_dispatch')
     runs-on: ubuntu-latest
     permissions:
       contents: write
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ needs.release-please.outputs.tag_name }}
+          ref: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
           fetch-depth: 0
       - uses: actions/setup-go@v5
         with:
@@ -514,4 +534,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HOMEBREW_TAP_TOKEN: ${{ secrets.HOMEBREW_TAP_TOKEN }}
-          GORELEASER_CURRENT_TAG: ${{ needs.release-please.outputs.tag_name }}
+          GORELEASER_CURRENT_TAG: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,7 +71,7 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
-            type=raw,value=latest
+            ${{ github.event_name != 'workflow_dispatch' && 'type=raw,value=latest' || '' }}
             type=semver,pattern={{version}},value=${{ env.TAG }}
             type=semver,pattern={{major}}.{{minor}},value=${{ env.TAG }}
             type=semver,pattern={{major}},value=${{ env.TAG }}
@@ -529,9 +529,9 @@ jobs:
         with:
           distribution: goreleaser
           version: '~> v2'
-          args: release --clean
+          args: release --clean ${{ github.event_name == 'workflow_dispatch' && '--skip=homebrew' || '' }}
           workdir: cli
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          HOMEBREW_TAP_TOKEN: ${{ secrets.HOMEBREW_TAP_TOKEN }}
+          HOMEBREW_TAP_TOKEN: ${{ secrets.TAP_GITHUB_TOKEN }}
           GORELEASER_CURRENT_TAG: ${{ inputs.tag || needs.release-please.outputs.tag_name }}


ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

Adds workflow_dispatch trigger to allow rebuilding assets for an existing tag, fixing #234. The conditional pattern is correct and the tag plumbing works, but two important issues: the `latest` Docker tag is unconditionally pushed on every dispatch (can regress latest to an older version when rebuilding old tags), and GoReleaser on dispatch may push incorrect formula updates to the Homebrew tap. Minor style duplication.

### Issues

🔴 **.github/workflows/release.yml:74** — Docker metadata always emits `type=raw,value=latest`. On a workflow_dispatch rebuild of an OLDER tag (e.g., rebuilding v0.5.0 when v0.6.0 is current), this will overwrite `:latest` with the older image, silently regressing users who pull `latest`. Gate the `latest` tag so it only applies on the push path (e.g., `type=raw,value=latest,enable=${{ github.event_name == 'push' }}`), or derive `enable` by comparing the dispatch tag to the repo's latest release.
🔴 **.github/workflows/release.yml:534** — goreleaser runs with `GORELEASER_CURRENT_TAG` set from inputs on dispatch, which can push updates to the Homebrew tap (`HOMEBREW_TAP_TOKEN`) for an older tag and overwrite a newer formula. Recommend either disabling the brew publisher on workflow_dispatch (pass `--skip=homebrew` or an equivalent flag for old tags) or confirming the goreleaser config has guards so rebuilding an old tag doesn't downgrade published artifacts/formulas.
🟡 **.github/workflows/release.yml:41** — The `always() && (release_created == 'true' || event_name == 'workflow_dispatch')` guard will run downstream jobs even if the `release-please` job explicitly fails on a push event (e.g., due to auth/quota errors). In that failure case `release_created` will not be `'true'` so the branch is fine, but consider using `!cancelled()` instead of `always()` to avoid running when the workflow has been cancelled.
🟡 **.github/workflows/release.yml:43** — Only `docker-build-push` gets a job-level `env.TAG`. The other three jobs (`build-linux`, `build-macos`, `goreleaser-cli`) repeat `${{ inputs.tag || needs.release-please.outputs.tag_name }}` at 7+ call sites, which is error-prone (easy to miss one when editing later). Apply the same `env.TAG` pattern to all four jobs and reference `${{ env.TAG }}` consistently.
🟡 **.github/workflows/release.yml:8** — No validation that the user-supplied `tag` input actually exists or matches the expected `vX.Y.Z` semver pattern. A typo will fail later in `actions/checkout` with a confusing error. Consider adding a short preflight step (e.g., `git ls-remote --tags` check or a regex validation) that fails fast with a clear message.

### Suggestions

- Gate the `latest` docker tag and the Homebrew publisher behind a `github.event_name == 'push'` condition so dispatch rebuilds never regress user-facing `latest` pointers.
- Hoist `TAG: ${{ inputs.tag || needs.release-please.outputs.tag_name }}` to a job-level `env` in build-linux / build-macos / goreleaser-cli to DRY up the workflow and reduce the risk of one step drifting.
- Add a preflight validation step that confirms the dispatch tag exists in git and matches the expected semver pattern before spending runner minutes on builds.

---

Muriano (approved):

Clean, well-structured PR that adds workflow_dispatch to recover from failed release builds. The always() + condition pattern is correctly implemented so skipped release-please doesn't block manual runs, and failed release-please still gates push-triggered builds. One confirmed issue with the Docker 'latest' tag being overwritten on manual re-runs of older releases.

### Suggestions

- Consider setting the `inputs.tag` into an environment variable at the job level and referencing it via `env.TAG` in `run:` blocks instead of using `${{ inputs.tag }}` directly in shell scripts. Direct `${{ }}` interpolation in `run:` is a known script-injection anti-pattern (even though here it's mitigated by requiring repo write access and the tag being validated by the checkout action).
- The `workflow_dispatch` input lacks a `type: string` declaration. While `string` is the default, being explicit improves readability for future maintainers.
- Consider adding a validation step at the start of dispatch-triggered jobs that confirms the tag actually exists in the repo (`git tag -l $TAG`) and matches the expected semver pattern, to fail fast with a clear message rather than a cryptic checkout error.

---

Muriano:
## ⚠️ MEDIUM Issue

Docker metadata always applies `type=raw,value=latest`. If workflow_dispatch is used to rebuild assets for an older tag (e.g. v0.5.0 after v0.6.0 is already released), the `latest` Docker image tag will be overwritten to point to the older version. Fix: conditionally include the `latest` tag only when the semver equals the highest release, or omit `latest` when event is workflow_dispatch (add a conditional line like `${{ github.event_name != 'workflow_dispatch' && 'type=raw,value=latest' || '' }}`).

**Location:** `.github/workflows/release.yml` line 74

---

Muriano:
## ⚠️ MEDIUM Issue

Secret reference renamed from `secrets.HOMEBREW_TAP_TOKEN` to `secrets.TAP_GITHUB_TOKEN`. This change affects ALL release runs (push and workflow_dispatch), not just the new manual path. If the repository secret has not been renamed to match, the normal push-triggered release will fail with goreleaser unable to push to the homebrew tap. Verify the secret name in repo settings matches `TAP_GITHUB_TOKEN`, or revert to `HOMEBREW_TAP_TOKEN` if that is the actual secret name.

**Location:** `.github/workflows/release.yml` line 537

---

Muriano (approved):

Well-structured fix that adds workflow_dispatch to rebuild release assets for existing tags. The approach is sound: skip release-please on manual dispatch, use always() to allow dependent jobs to run, and fall back to inputs.tag throughout. The existing reviewer comment about the latest Docker tag has been addressed in this revision. One real concern is the secret rename that affects all release paths.

### Suggestions

- Consider adding a description or validation note in the workflow_dispatch input to clarify that the tag must already exist (e.g. `description: 'Existing git tag to rebuild assets for (must already exist, e.g. v0.6.0)'`), since passing a non-existent tag will fail at checkout with a confusing error.
- The conditional `latest` tag line (`${{ github.event_name != 'workflow_dispatch' && 'type=raw,value=latest' || '' }}`) produces an empty line in the tags input when triggered manually. While docker/metadata-action handles this gracefully, a YAML comment explaining the intent would help future maintainers understand the conditional.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

The workflow correctly adds a `workflow_dispatch` path to rebuild assets for existing tags, and addresses the previously raised `latest` tag concern by gating `type=raw,value=latest` on event type. However, the same overwrite hazard still applies to the `{{major}}` tag when rebuilding an older release, and the secret rename for Homebrew is not explained. A few minor consistency issues as well.

### Issues

⚠️ **.github/workflows/release.yml:76** — The `{{major}}` docker tag has the same risk the existing review comment raised for `latest`: rebuilding v0.5.0 via workflow_dispatch when v0.6.0 is the current release will overwrite the `0` tag so it points to the older commit. `{{version}}` and `{{major}}.{{minor}}` are safe because they are unique to the rebuilt line, but `{{major}}` is not. Consider also gating `{{major}}` with `github.event_name != 'workflow_dispatch'`, or requiring the operator to verify the input tag is the highest semver before pushing shared moving tags.
⚠️ **.github/workflows/release.yml:536** — The Homebrew secret reference was renamed from `secrets.HOMEBREW_TAP_TOKEN` to `secrets.TAP_GITHUB_TOKEN` with no related change elsewhere and no mention in the PR description. If the secret was not actually renamed/created in repo settings, the next normal `push` release (non-dispatch path) will fail the Homebrew tap update silently or break Homebrew publication. Please confirm the secret exists under the new name, or revert the rename since it appears unrelated to the bug being fixed.
🟡 **.github/workflows/release.yml:43** — The job-level `env: TAG: ${{ inputs.tag || needs.release-please.outputs.tag_name }}` is defined for `docker-build-push`, but the `Checkout` step still uses the inline expression `ref: ${{ inputs.tag || needs.release-please.outputs.tag_name }}` instead of `${{ env.TAG }}`. Minor consistency nit — pick one and use it throughout so future edits don't drift.
🟡 **.github/workflows/release.yml:40** — There is no validation that `inputs.tag` refers to an existing tag on `workflow_dispatch`. A typo will only surface as a checkout failure deep into the job matrix after minutes of setup. Consider a tiny pre-step that runs `git rev-parse --verify "refs/tags/${{ inputs.tag }}"` (or `gh api repos/${{ github.repository }}/git/refs/tags/${{ inputs.tag }}`) early to fail fast with a clear message.
⚠️ **.github/workflows/release.yml:40** — The `always() && (release_created || workflow_dispatch)` guard runs the asset jobs even if `release-please` itself failed on a push event (not just when it was skipped for dispatch). That can produce a release with partial/missing assets on a bad release-please run. Safer form: `(github.event_name == 'workflow_dispatch') || (needs.release-please.result == 'success' && needs.release-please.outputs.release_created == 'true')`.

### Suggestions

- Document the new manual backfill flow (how to trigger, when to use, expectations about `latest`/`{{major}}` tags not moving) in a short note in the workflow file header or release docs.
- If the `HOMEBREW_TAP_TOKEN` → `TAP_GITHUB_TOKEN` rename is intentional, split it into its own commit with a clear message so it's auditable separately from the release-pipeline bugfix.
- Consider adding a boolean `workflow_dispatch` input like `update_latest` (default false) so an operator can explicitly opt in to moving `latest`/`{{major}}` when backfilling the current highest tag, rather than hardcoding the skip.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

Adds workflow_dispatch fallback to rebuild release assets for an existing tag. Logic is mostly sound (proper `always()` + OR guards, `--skip=homebrew` on manual runs, conditional `latest` Docker tag), but carries one high-risk change — the silent rename of the Homebrew tap secret — plus a few consistency/robustness gaps worth fixing before merge.

### Issues

🔴 **.github/workflows/release.yml:537** — Secret rename HOMEBREW_TAP_TOKEN → TAP_GITHUB_TOKEN is unrelated to this PR's stated purpose and impacts EVERY release (both push and workflow_dispatch). If the repo secret has not been renamed to match, every future automatic release will fail to publish to the Homebrew tap. This change should be split out, justified, or reverted — and the rename verified in repo settings before merge.
⚠️ **.github/workflows/release.yml:20** — No `concurrency:` group is declared. A manual workflow_dispatch run targeting a tag while a push-triggered release is in flight (or a retry targeting the same tag) can race on softprops/action-gh-release uploads and on GHCR pushes, producing partial/duplicate assets — the exact failure mode #234 is trying to prevent. Add `concurrency: { group: release-${{ inputs.tag || github.ref }}, cancel-in-progress: false }`.
🟡 **.github/workflows/release.yml:21** — `env.TAG` is defined at the docker-build-push job level but is only consumed in the metadata tags block; the checkout ref and other references still use the long `${{ inputs.tag || needs.release-please.outputs.tag_name }}` expression. Either use `env.TAG` consistently within this job or remove it, and consider hoisting the same pattern (as a workflow-level `env` or a computed job output) to the other three jobs where the expression is duplicated 4–5 times each.
🟡 **.github/workflows/release.yml:74** — The conditional-line trick `${{ ... && 'type=raw,value=latest' || '' }}` leaves a blank line in the `tags:` block on workflow_dispatch runs. docker/metadata-action tolerates this today but it's fragile and easy to misread. Safer: branch the whole tag list via an `if:` on two separate steps, or use `flavor: latest=false` + a computed `tags` input.
🟡 **.github/workflows/release.yml:8** — `inputs.tag` is consumed verbatim as a git ref and as the GoReleaser `GORELEASER_CURRENT_TAG` without validation. A typo ('v.0.6.0', '0.6.0' without the v, or a non-existent tag) fails late with confusing errors; worse, GoReleaser with `--clean` may still mutate the working tree before erroring. Add a first step that runs `git rev-parse --verify "refs/tags/${{ inputs.tag }}"` to fail fast, and/or validate the input matches `^v[0-9]+\.[0-9]+\.[0-9]+$`.
🟡 **.github/workflows/release.yml:532** — `--skip=homebrew` is correctly applied on workflow_dispatch, but GoReleaser also publishes other artifacts that may be undesirable on a manual rebuild (e.g. announce/scoop/nix if ever enabled, or re-triggering release notes). Worth reviewing `.goreleaser.yaml` to confirm no other publish-side-effect targets need skipping, and documenting in a comment which skips are intentional.

### Suggestions

- Split the HOMEBREW_TAP_TOKEN → TAP_GITHUB_TOKEN rename into its own PR with verification that the repo secret exists under the new name, or revert it here — it is outside the scope of issue #234.
- Add a `concurrency:` block keyed by the resolved tag to prevent manual reruns from racing against in-flight push releases.
- Factor `${{ inputs.tag || needs.release-please.outputs.tag_name }}` into a single workflow-level or job-level env var (e.g. `RELEASE_TAG`) to remove the 10+ duplicated expressions and reduce drift risk on future edits.
- Add a pre-flight validation step that verifies `inputs.tag` is a real, pushed tag before any build/upload step runs.
- Consider adding a short comment block at the top of the file explaining the dual-trigger model (push = release-please cuts tag + builds; workflow_dispatch = rebuild assets for existing tag, no new tag, no homebrew) so future maintainers don't need to infer it from the conditionals.

---

Muriano:
## ⚠️ MEDIUM Issue

`always()` in the `if` condition means the job runs even when `release-please` is **cancelled** or **fails** (not just skipped). On a push event where release-please fails unexpectedly, `release_created` will be undefined — the job won't run, so no immediate breakage, but `always()` is still broader than necessary and masks cancelled-workflow signals. Replace `always()` with `!cancelled()` in all four build jobs (`docker-build-push`, `build-linux`, `build-macos`, `goreleaser-cli`). `!cancelled()` still allows execution when the dependency is skipped (workflow_dispatch case) but correctly stops when it is cancelled or fails.

**Location:** `.github/workflows/release.yml` line 40

---

Muriano:
## ⚠️ MEDIUM Issue

Secret reference changed from `secrets.HOMEBREW_TAP_TOKEN` to `secrets.TAP_GITHUB_TOKEN`. This affects **all** release runs (push and workflow_dispatch). If the repository secret has not been renamed to `TAP_GITHUB_TOKEN`, the normal push-triggered release will fail when GoReleaser tries to push to the Homebrew tap. Verify the secret name matches in repo settings, or revert to `HOMEBREW_TAP_TOKEN`.

**Location:** `.github/workflows/release.yml` line 537

---

Muriano (approved):

The PR correctly adds workflow_dispatch support to re-run asset builds for existing tags. The conditional latest-tag logic is already handled. Two confirmed issues remain: the `always()` anti-pattern and the secret rename that affects the normal push path.

### Suggestions

- Consider adding a brief `validate-tag` job for workflow_dispatch that checks the tag exists and matches `v[0-9]+.[0-9]+.[0-9]+` before kicking off the expensive build matrix — saves runner minutes on typos.
- The `--skip=homebrew` flag for GoReleaser on workflow_dispatch is a good safety measure. Consider documenting in a PR comment or repo wiki when maintainers should use the manual dispatch (e.g., 'only when a push-triggered release created the tag but asset builds failed').

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

Sensible fix for the release pipeline corruption issue. Adds workflow_dispatch path to rebuild assets for an existing tag, guards the latest-tag clobber, and skips homebrew on manual re-runs. Two concerns worth flagging: an unrelated secret rename and the use of always() over !cancelled().

### Issues

🔴 **.github/workflows/release.yml:537** — Secret reference changed from HOMEBREW_TAP_TOKEN to TAP_GITHUB_TOKEN. This change affects ALL release runs (push and workflow_dispatch), not just the new manual path, and is unrelated to the PR's stated goal of handling pipeline corruption. If the repository secret in GitHub settings has not been renamed to TAP_GITHUB_TOKEN, the normal push-triggered release will fail when GoReleaser attempts to push to the Homebrew tap. Either verify/rename the secret in repo settings before merging, or revert this rename into a separate PR so the rollback is trivial if the tap publish breaks.
⚠️ **.github/workflows/release.yml:40** — Using always() in the if condition means the build jobs run even when release-please is cancelled or fails, not just when it is skipped (the workflow_dispatch case). Because release_created will be false/empty on failure the branches won't enter the build path, but always() still masks cancelled-workflow semantics and is broader than intended. Prefer !cancelled() which still permits execution when the dependency was skipped (workflow_dispatch) but correctly propagates cancellations. Apply to all four build jobs: docker-build-push, build-linux, build-macos, goreleaser-cli.
🟡 **.github/workflows/release.yml:74** — Minor: the conditional 'type=raw,value=latest' line expands to an empty string on workflow_dispatch, which docker/metadata-action tolerates, but a cleaner pattern is to put the raw tag on its own line guarded by a full ${{ ... && 'type=raw,value=latest' }} expression with no fallback, or to use the action's own flavor/latest=auto options. Current form works but relies on empty-line tolerance in the tags list.
🟡 **.github/workflows/release.yml:10** — workflow_dispatch input 'tag' has no validation. A typo (e.g. '0.6.0' without the leading 'v', or a nonexistent tag) will fail late at checkout or at softprops/action-gh-release upload, after a long Flutter/macOS build has already run. Consider adding a quick 'git rev-parse --verify refs/tags/${{ inputs.tag }}' guard as an early step, and/or documenting the expected 'vX.Y.Z' format in the input description beyond the example.
🟡 **.github/workflows/release.yml:47** — The env.TAG variable is defined at the docker-build-push job level and used only within that job's metadata step. The other three jobs (build-linux, build-macos, goreleaser-cli) repeat the 'inputs.tag || needs.release-please.outputs.tag_name' expression inline multiple times. Minor DRY improvement: define the same env.TAG at each job level (or promote to a workflow-level env) so the fallback expression lives in one place and future changes (e.g. adding input validation) don't require touching ~8 lines.

### Suggestions

- Before merge, explicitly confirm in the PR description whether the GitHub repo secret has been renamed to TAP_GITHUB_TOKEN, or split that rename into its own PR.
- Replace always() with !cancelled() in the four job-level if conditions — same semantics for the workflow_dispatch skip path, stricter for cancellations/failures.
- Add an early 'validate tag exists' step on workflow_dispatch to fail fast instead of after a 20-minute Flutter build.
- Verify that goreleaser v2 accepts '--skip=homebrew' (it does in v2; the v1 syntax was different) — worth a sanity check on the local cli/.goreleaser.yaml to ensure the homebrew section name matches the skip key.

---

PR #241, opened 2026-04-22
feat: implement #239 — bug(cli): dashboard doesn't refresh — r key and auto-tick don't update display

### LINKED ISSUE #239: bug(cli): dashboard doesn't refresh — r key and auto-tick don't update display
## Problem

`heimdallm-cli dashboard` loads initial data but:
1. **`r` key doesn't visibly refresh** — fetches data but the display doesn't update
2. **Auto-tick (10s) doesn't update display** — `tickMsg` triggers `fetchData` but no visual change
3. **`G` key only affects Logs tab** — expected to scroll-to-bottom in all tabs
4. **SSE events don't appear** — `connectSSE` runs but events don't trigger visual updates

The API returns data correctly (verified with `curl`). The issue is in the Bubbletea Update/View cycle.

## Likely causes

1. `fetchData` returns `dataMsg` which updates `d.prs` and `d.issues` but the View function may not re-render correctly
2. SSE errors are silently swallowed (`_ = d.client.StreamEvents(...)`)
3. Missing `tea.Cmd` chaining — after processing `dataMsg`, no follow-up command triggers a re-render
4. The `listenSSE` cmd waits on channel but if SSE connection fails, it blocks forever

## Expected behavior

- Dashboard updates in real-time as Heimdallm processes PRs/issues
- `r` forces immediate refresh with visual feedback
- Activity tab shows new events as they arrive via SSE
- Connection status indicator (connected/disconnected)

## Files

- `cli/internal/tui/dashboard.go` — Update/View cycle, SSE handling, key bindings

sergiotejon:
Auto-generated by Heimdallm in response to #239.

Closes #239

### DIFF
--- a/cli/internal/tui/dashboard.go
+++ b/cli/internal/tui/dashboard.go
@@ -43,11 +43,13 @@ type Dashboard struct {
 	logOffset int
 	logSeeded bool
 
-	err       error
-	connected bool
-	startTime time.Time
+	err        error
+	connected  bool
+	refreshing bool
+	startTime  time.Time
 
 	sseEvents chan api.SSEEvent
+	sseCtx    context.Context
 	sseCancel context.CancelFunc
 }
 
@@ -68,13 +70,18 @@ type dataMsg struct {
 	err      error
 }
 type sseMsg api.SSEEvent
+type sseDisconnectMsg struct{ err error }
+type sseReconnectMsg struct{}
 
 func NewDashboard(host, token string) *Dashboard {
+	ctx, cancel := context.WithCancel(context.Background())
 	return &Dashboard{
 		client:    api.New(host, token),
 		startTime: time.Now(),
 		sseEvents: make(chan api.SSEEvent, 32),
 		logFollow: true,
+		sseCtx:    ctx,
+		sseCancel: cancel,
 	}
 }
 
@@ -135,21 +142,24 @@ func (d *Dashboard) fetchData() tea.Msg {
 }
 
 func (d *Dashboard) connectSSE() tea.Msg {
-	ctx, cancel := context.WithCancel(context.Background())
-	d.sseCancel = cancel
-	go func() {
-		_ = d.client.StreamEvents(ctx, d.sseEvents)
-	}()
-	return nil
+	err := d.client.StreamEvents(d.sseCtx, d.sseEvents)
+	if d.sseCtx.Err() != nil {
+		return nil
+	}
+	return sseDisconnectMsg{err: err}
 }
 
 func (d *Dashboard) listenSSE() tea.Cmd {
 	return func() tea.Msg {
-		event, ok := <-d.sseEvents
-		if !ok {
+		select {
+		case event, ok := <-d.sseEvents:
+			if !ok {
+				return nil
+			}
+			return sseMsg(event)
+		case <-d.sseCtx.Done():
 			return nil
 		}
-		return sseMsg(event)
 	}
 }
 
@@ -193,6 +203,7 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return d, tea.Batch(d.fetchData, tickCmd())
 
 	case dataMsg:
+		d.refreshing = false
 		if msg.err != nil {
 			d.err = msg.err
 			d.connected = false
@@ -258,7 +269,16 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				}
 			}
 		}
+		d.connected = true
 		return d, d.listenSSE()
+
+	case sseDisconnectMsg:
+		return d, tea.Tick(5*time.Second, func(t time.Time) tea.Msg {
+			return sseReconnectMsg{}
+		})
+
+	case sseReconnectMsg:
+		return d, d.connectSSE
 	}
 
 	return d, nil
@@ -331,8 +351,14 @@ func (d *Dashboard) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 	case "G":
 		if d.activeTab == tabLogs {
 			d.logFollow = true
+		} else {
+			max := d.tabItemCount()
+			if max > 0 {
+				d.cursor = max - 1
+			}
 		}
 	case "r":
+		d.refreshing = true
 		return d, d.fetchData
 	case "1":
 		d.activeTab = tabActivity
@@ -396,6 +422,9 @@ func (d *Dashboard) View() string {
 }
 
 func (d *Dashboard) renderStatus() string {
+	if d.refreshing {
+		return lipgloss.NewStyle().Foreground(colorWarning).Bold(true).Render("● refreshing...")
+	}
 	if d.connected {
 		return statusOnline.Render("● online")
 	}


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

The fix adds a refreshing indicator, SSE auto-reconnect, and extends G-key behavior. Core mechanics look sound, but the connected-state tracking is inconsistent after SSE drops, the stored sseCancel is never invoked (goroutine leak on exit), and the reconnect has no backoff or state differentiation for the user.

### Issues

⚠️ **cli/internal/tui/dashboard.go:278** — On sseDisconnectMsg the code schedules a 5s reconnect Tick but never sets d.connected = false. The status bar continues to render '● online' during the reconnection window even though SSE is not flowing. Given the PR's intent is to make refresh state visible to the user, the disconnect/reconnect state should also surface (e.g. a '● reconnecting…' status, mirroring the new 'refreshing…' treatment).
⚠️ **cli/internal/tui/dashboard.go:272** — d.connected now has two independent sources of truth: dataMsg (HTTP fetch) toggles it to false on error, and sseMsg sets it to true whenever any SSE event arrives. These are two different channels with different failure modes — e.g. HTTP healthy + SSE dead leaves connected=true indefinitely (until the 5s reconnect cycle succeeds), and HTTP down + SSE healthy will bounce the flag on every event. Consider tracking httpConnected and sseConnected separately and composing the status, or at minimum flipping connected=false in sseDisconnectMsg.
🟡 **cli/internal/tui/dashboard.go:78** — sseCtx/sseCancel are created in NewDashboard but sseCancel is never called anywhere in this diff. Previously sseCancel was recreated per connectSSE call; now it's a single long-lived context for the lifetime of the process. If/when the TUI exits, the StreamEvents goroutine (now running synchronously inside the connectSSE Cmd) will not be torn down cleanly. Worth wiring sseCancel() into the quit path or whatever existing shutdown hook exists.
🟡 **cli/internal/tui/dashboard.go:280** — Reconnect is a fixed 5s tick with no backoff or jitter. If the server is down or the token is invalid, the client will hammer StreamEvents every 5s indefinitely. Not critical for a local TUI but an exponential backoff capped at ~30s would be friendlier on the daemon logs.
🟡 **cli/internal/tui/dashboard.go:206** — d.refreshing only clears on dataMsg. If fetchData ever hangs (no timeout) or returns an error path that doesn't reach this case, the status stays stuck on 'refreshing…'. Looking at the diff, dataMsg is the only exit so it's fine today — but a safety reset on dataMsg regardless of err path (which it does) combined with an HTTP timeout on the underlying client would close the gap.
🟡 **cli/internal/tui/dashboard.go:354** — The 'G' key now also jumps cursor to the last item on non-logs tabs. This is a new navigation feature bundled into a bugfix PR titled 'dashboard doesn't refresh'. Not wrong, but scope creep — worth calling out in the PR description or splitting out so the bisect trail stays clean.
🟡 **cli/internal/tui/dashboard.go:148** — connectSSE now blocks synchronously inside the Cmd goroutine on StreamEvents. The early-return guard uses d.sseCtx.Err() to distinguish intentional cancel from a real disconnect — good — but if StreamEvents ever returns err=nil on a legitimate remote-side close (non-cancel), sseDisconnectMsg will carry err=nil and the reconnect path will still fire, which is the desired behavior. Consider logging err so silent disconnects are debuggable from the operator side.

### Suggestions

- Introduce a dedicated sseConnected bool (or a small enum: connecting/online/reconnecting/offline) and have renderStatus drive entirely off that, so the user has clear visual feedback during the 5s reconnect window — it mirrors the pattern you just added for refreshing.
- Flip d.connected = false (or the new sseConnected) inside the sseDisconnectMsg case before scheduling the reconnect tick.
- Wire d.sseCancel() into the quit/teardown path so the streaming goroutine is released on exit.
- Add exponential backoff (e.g. 1s, 2s, 5s, 10s, 30s capped) for the reconnect Tick; keeps daemon logs clean when the server is genuinely unreachable.
- Split the 'G' key navigation enhancement into its own PR, or at least mention it explicitly in the PR description so the change set matches the title.

---

Muriano:
## ⚠️ MEDIUM Issue

sseDisconnectMsg handler does not set `d.connected = false`. When SSE drops (e.g., network blip while server is still reachable for REST), the UI continues showing '● online' for up to 5 seconds (reconnect delay) or longer if reconnect also fails. The `dataMsg` tick every 10s only corrects this if the REST endpoint also fails. For a dashboard whose status indicator explicitly tracks connectivity, this is misleading. Fix: add `d.connected = false` in the `sseDisconnectMsg` case before scheduling the reconnect tick.

**Location:** `cli/internal/tui/dashboard.go` line 272

---

Muriano (approved):

Solid fix for SSE disconnect/reconnect and dashboard refresh. The original connectSSE silently swallowed disconnects; the new code properly surfaces them and reconnects. One confirmed medium issue: SSE disconnect doesn't clear connected status, leaving a misleading 'online' indicator during the reconnect window.

### Suggestions

- Consider exponential backoff for SSE reconnection instead of a fixed 5-second delay — repeated failures (e.g., auth revoked, server down) will hammer the server every 5 seconds indefinitely.
- The `G` key 'go to end' behavior for non-log tabs (lines 354-358) is a feature addition unrelated to the bug described in #239. Consider splitting it into its own commit or PR for cleaner history.
- The `sseDisconnectMsg.err` field is captured but never logged or displayed. Consider surfacing it (e.g., in the log panel or via debug logging) to aid troubleshooting connection issues.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

The refactor correctly fixes the dashboard refresh bug by making connectSSE blocking and surfacing disconnect/reconnect as messages, and adds useful UX (refreshing indicator, G-to-end). However, the sseDisconnectMsg handler doesn't update d.connected, leaving the status indicator stale — matching Muriano's finding. Reconnect also lacks backoff and there's no visible error surfacing on repeated failure.

### Issues

⚠️ **cli/internal/tui/dashboard.go:278** — sseDisconnectMsg does not set d.connected = false, so '● online' is shown for at least 5s (reconnect delay) after SSE drops, longer if reconnect keeps failing and REST is still healthy. Set d.connected = false before scheduling the reconnect tick.
🟡 **cli/internal/tui/dashboard.go:280** — Reconnect uses a fixed 5s interval with no backoff or cap on retries. On a prolonged outage this hits the server every 5s indefinitely. Consider exponential backoff (e.g., 1s → 30s cap).
⚠️ **cli/internal/tui/dashboard.go:145** — connectSSE now blocks for the lifetime of the SSE connection inside a tea.Cmd goroutine. This is acceptable, but if StreamEvents returns err == nil and d.sseCtx.Err() == nil (clean server close), nil is returned and no reconnect is ever scheduled — the stream stays dead until a restart. Return sseDisconnectMsg unconditionally when ctx is not cancelled, even on nil error.
🟡 **cli/internal/tui/dashboard.go:75** — sseCtx is created once in NewDashboard and never reset. If anything cancels it (e.g., future Quit handling via d.sseCancel), the reconnect path calls connectSSE with an already-cancelled context and silently returns nil, so reconnect will never succeed. If cancel is intended only for shutdown this is fine, but the field pairing invites future bugs — add a comment or recreate the ctx on reconnect.
🟡 **cli/internal/tui/dashboard.go:354** — The 'G' else-branch for non-log tabs is unrelated to the refresh bug described by the PR title/#239. Not incorrect, but it should either be called out in the PR description or split into its own change.

### Suggestions

- In the sseDisconnectMsg case, set d.connected = false (and optionally d.err = msg.err for visibility) before returning the reconnect tick.
- Add exponential backoff for reconnects and bound retries, or at least surface persistent failure in the status line (e.g., '● reconnecting…').
- Return sseDisconnectMsg from connectSSE whenever ctx is not cancelled, regardless of whether err is nil, so clean server-side closes also trigger reconnect.
- Mention the 'G' navigation addition in the PR description, or move it to a separate PR since it's orthogonal to the refresh fix.

---

Muriano:
## ⚠️ MEDIUM Issue

sseDisconnectMsg handler does not set d.connected = false. When SSE drops but REST is still reachable, the status indicator shows '● online' for 5+ seconds (reconnect delay) until either reconnect succeeds or a REST fetch also fails. Fix: add d.connected = false before the tea.Tick call.

**Location:** `cli/internal/tui/dashboard.go` line 272

---

Muriano (approved):

Solid fix for the dashboard refresh bug. The SSE refactor from fire-and-forget goroutine to a blocking Cmd is the correct Bubble Tea pattern and fixes the root cause. One confirmed issue: the disconnect handler doesn't update the connection status indicator, which was already flagged in the existing PR discussion.

### Suggestions

- Consider exponential backoff on SSE reconnect instead of a fixed 5s delay — repeated fast failures (e.g., auth revoked) will spin reconnect attempts every 5s indefinitely.
- The G key in non-log tabs now jumps to the last item, which is a nice addition — consider documenting it in the help/keybinding view if one exists.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

The PR fixes the dashboard refresh/auto-tick issue by making connectSSE synchronous and returning proper disconnect/reconnect messages, adding a visible refreshing indicator, and adding 'G' jump-to-end for non-log tabs. The core fix is correct, but the SSE connectivity indicator is not updated on disconnect, reconnect attempts have no backoff or cap, and a single long-lived sseCtx introduces a subtle reconnect-after-quit concern.

### Issues

⚠️ **cli/internal/tui/dashboard.go:278** — sseDisconnectMsg handler doesn't set d.connected = false. When SSE drops but REST still works, the UI shows '● online' for up to 5s (or longer on repeated failures) until the next successful dataMsg or sseMsg. For a dashboard whose status explicitly tracks connectivity this is misleading. Add d.connected = false before scheduling the reconnect tick.
🟡 **cli/internal/tui/dashboard.go:280** — Reconnect uses a fixed 5s delay with no backoff and no max attempt cap. If the server is persistently down, this hammers it every 5s for the entire session. Consider exponential backoff with a ceiling (e.g., 5s → 30s → 60s).
🟡 **cli/internal/tui/dashboard.go:145** — sseCtx is created once in the constructor and reused across all reconnects. If sseCancel is ever invoked (e.g., on app quit) and an sseReconnectMsg is still in-flight in the tea runtime, connectSSE will immediately return sseDisconnectMsg because ctx is already done — but because the ctx.Err() branch returns nil, the reconnect chain terminates silently. This works today but is fragile: any future code path that cancels the ctx for reasons other than quit will permanently disable SSE with no visible signal. Consider creating a fresh context per connect attempt, or documenting the single-use lifecycle.
🟡 **cli/internal/tui/dashboard.go:206** — d.refreshing is cleared on any dataMsg, including the 10s auto-tick. If the user presses 'r' and the periodic tick's fetchData returns first, the refreshing indicator disappears before the user-triggered fetch completes. Minor UX race; could be avoided by tying the flag to a request id or clearing only after the user-initiated fetch returns.
🟡 **cli/internal/tui/dashboard.go:272** — d.connected is set to true inside the sseMsg handler, meaning the online indicator only appears after the first SSE event is received, not when the stream is successfully established. For low-event-rate environments the dashboard will show offline for a long time despite a healthy connection. Consider having connectSSE emit a connected message, or flipping d.connected on the first listenSSE invocation.

### Suggestions

- Set d.connected = false in the sseDisconnectMsg case to keep the status indicator truthful during the reconnect window.
- Introduce exponential backoff (capped) for reconnects rather than a hard-coded 5s.
- Consider emitting an explicit 'sse connected' message when StreamEvents is ready, so d.connected doesn't depend on receiving the first event.
- The 'G' jump-to-end addition for non-log tabs is a nice improvement but is unrelated to the bug in the PR title — worth calling out in the description or splitting into its own commit for traceability.

---

Muriano:
## ⚠️ MEDIUM Issue

sseDisconnectMsg handler still does not set d.connected = false. When SSE drops but REST remains reachable, the status indicator shows '● online' during the 5-second reconnect delay (and longer if reconnect also fails). The 10s data tick only corrects this if REST also fails. Fix: add `d.connected = false` before the tea.Tick call.

**Location:** `cli/internal/tui/dashboard.go` line 272

---

Muriano (approved):

The SSE reconnection and refresh indicator changes are solid. The previous finding about sseDisconnectMsg not setting d.connected = false remains unaddressed in the diff.

### Suggestions

- Consider logging or emitting a debug message on SSE disconnect/reconnect to aid troubleshooting connectivity issues in production.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

The PR fixes the dashboard refresh bug by restructuring SSE connection lifecycle with a persistent context, adding reconnect logic, and surfacing a 'refreshing...' state on manual refresh. Logic is sound and the Bubble Tea message flow is correct, but the disconnect handler fails to update connectivity state, which is a real UX regression for a status indicator.

### Issues

⚠️ **cli/internal/tui/dashboard.go:272** — sseDisconnectMsg handler does not set d.connected = false before scheduling the 5s reconnect tick. When SSE drops while REST remains reachable, the status indicator continues showing '● online' for at least 5 seconds (and longer if reconnect keeps failing, since dataMsg every 10s only flips connected=false when REST also fails). For a status indicator that explicitly tracks connectivity, this is misleading. Set d.connected = false in the sseDisconnectMsg case.
🟡 **cli/internal/tui/dashboard.go:145** — connectSSE now runs StreamEvents synchronously inside a tea.Cmd goroutine (good), but sseReconnectMsg returns d.connectSSE as the Cmd without also re-subscribing via listenSSE. On the initial connect path (Init/Update startup) listenSSE is kicked off elsewhere, but after a reconnect there is no explicit guarantee the listener loop is still pumping — verify that the existing listenSSE goroutine chain survives a disconnect/reconnect cycle (it uses the same sseEvents channel and sseCtx, so it should, but worth confirming with a quick manual disconnect test).
🟡 **cli/internal/tui/dashboard.go:354** — The added 'G' behavior for non-log tabs (jump to last item) is an unrelated UX change bundled into a bug-fix PR titled 'dashboard doesn't refresh'. Minor scope creep — consider splitting or calling it out in the PR description.
🟡 **cli/internal/tui/dashboard.go:206** — d.refreshing is cleared on dataMsg regardless of success/failure, which is correct, but if fetchData errors immediately the 'refreshing...' state may flash too briefly to notice. Not a bug, just worth confirming the UX on fast failures.

### Suggestions

- Add `d.connected = false` at the top of the sseDisconnectMsg case before the tea.Tick call to address the reviewer's repeated MEDIUM feedback.
- Consider also clearing d.connected on the reconnect attempt itself (sseReconnectMsg) so that repeated failed reconnects keep the UI honest.
- Split the 'G jumps to last item' change into its own commit/PR, or mention it in the PR description since the title only references the refresh bug.
- Add a brief comment near sseCtx/sseCancel explaining that the context now lives for the dashboard lifetime and cancel is only used on shutdown — the previous per-connection context pattern was replaced.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

The PR correctly addresses the core refresh bug by making SSE disconnects surface via messages (instead of silently dying in a goroutine) and by adding a visible refreshing indicator. However, the already-flagged issue about `d.connected` not being reset on SSE disconnect is still unaddressed, and there are a couple of smaller concerns around status precedence and identifier shadowing.

### Issues

⚠️ **cli/internal/tui/dashboard.go:279** — sseDisconnectMsg handler does not set `d.connected = false`. When SSE drops but REST is still reachable, the status indicator keeps showing '● online' during the 5s reconnect delay (and longer if reconnect keeps failing). Since this PR's whole point is making the dashboard correctly reflect live state, leaving stale 'online' on a known-broken SSE contradicts the intent. Set `d.connected = false` before scheduling the reconnect tick.
🟡 **cli/internal/tui/dashboard.go:284** — sseReconnectMsg dispatches `d.connectSSE` which will block inside `StreamEvents` until the stream ends again. If the underlying endpoint is down, `StreamEvents` likely returns quickly with an error, producing a tight reconnect loop of connect→disconnect→5s tick→connect. There is no backoff or cap — consider exponential backoff or at least a max retry counter to avoid hammering the server on sustained outages.
🟡 **cli/internal/tui/dashboard.go:425** — `renderStatus` checks `d.refreshing` before `d.connected`, so pressing 'r' while the dashboard is offline transiently shows '● refreshing...' instead of the true offline state. Given the fetch may fail and keep the UI in that state until dataMsg arrives, consider rendering refreshing only when also connected, or combining it with the connectivity indicator.
🟡 **cli/internal/tui/dashboard.go:355** — Local variable `max := d.tabItemCount()` shadows the Go 1.21+ builtin `max`. Not a bug today, but it is a readability footgun and can cause confusing errors if someone later tries to call the builtin in the same scope. Rename to `n`, `count`, or `last`.
🟡 **cli/internal/tui/dashboard.go:206** — `d.refreshing` is cleared on any incoming `dataMsg`, including the auto-tick. If the 10s tick fires right after the user presses 'r', the indicator can vanish almost immediately without ever visibly reflecting the user's manual refresh. Consider only clearing `refreshing` when the dataMsg is from a manually triggered fetch (e.g., tag the fetch with a flag), or ensure the indicator stays visible for a minimum duration.

### Suggestions

- Add `d.connected = false` in the sseDisconnectMsg case to keep the status indicator truthful during the reconnect window.
- Introduce exponential backoff (and optionally jitter) for SSE reconnect instead of a fixed 5s tick to avoid tight loops on sustained outages.
- Rename the local `max` variable in handleKey to avoid shadowing the Go builtin.
- Reorder or combine the refreshing/connected checks in renderStatus so 'refreshing...' does not mask an actual offline state.
- Consider tagging fetches triggered by the 'r' key (e.g. a dedicated message type or a counter) so the refreshing indicator is cleared only when the corresponding fetch completes, not by a concurrent auto-tick.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

PR correctly addresses the refresh bug by decoupling SSE streaming from context creation and adding explicit reconnect handling, plus a 'refreshing' status indicator and 'G' navigation for non-log tabs. The core fix is sound, but the sseDisconnectMsg handler fails to update the connectivity indicator, causing the status to lie for up to 5+ seconds after SSE drops.

### Issues

⚠️ **cli/internal/tui/dashboard.go:279** — sseDisconnectMsg handler does not set d.connected = false. When the SSE stream drops while REST is still reachable, '● online' remains shown during the 5s reconnect delay (and longer if reconnect keeps failing), since the 10s dataMsg tick only flips connected=false when REST itself fails. Set d.connected = false before scheduling the tea.Tick reconnect.
🟡 **cli/internal/tui/dashboard.go:272** — sseMsg handler unconditionally sets d.connected = true on every event. This is fine for recovery after a drop, but combined with the missing d.connected=false in sseDisconnectMsg, connectivity state is only ever flipped true, never false via the SSE path. Ensure disconnect path writes false so the sseMsg path's true write is meaningful.
🟡 **cli/internal/tui/dashboard.go:147** — sseCtx is created once in NewDashboard and reused across reconnects. If sseCancel is ever invoked (e.g., on shutdown), subsequent sseReconnectMsg triggers will call StreamEvents with an already-canceled context and return immediately, producing a tight reconnect loop via sseDisconnectMsg → 5s tick → sseReconnectMsg. Not currently triggered (cancel isn't called mid-session), but the invariant is fragile; consider recreating the ctx/cancel pair on each reconnect or guarding against ctx.Err() before dispatching reconnects.
🟡 **cli/internal/tui/dashboard.go:354** — The new 'G' branch for non-log tabs only handles jump-to-end; there is no corresponding 'g'/'gg' jump-to-top, which is the conventional vim pairing users will expect after seeing 'G' work. Minor UX inconsistency, not a correctness bug.

### Suggestions

- Add d.connected = false at the top of the sseDisconnectMsg case before the tea.Tick call — this is the fix the three prior review comments have been asking for.
- Consider recreating sseCtx/sseCancel inside the sseReconnectMsg handler (or at the top of connectSSE) so a future shutdown-and-restart flow can't wedge the reconnect loop.
- Reset d.refreshing in the err branch of dataMsg as well — it already runs before the err check so this is handled, but double-check that an early-return path (e.g., future edits) won't leave refreshing=true stuck on screen.
- Optionally add a 'g'/'gg' binding to mirror the new 'G' jump-to-end for non-log tabs.

---

Muriano:
## ⚠️ MEDIUM Issue

STILL UNADDRESSED (flagged 3x): sseDisconnectMsg handler does not set d.connected = false. When SSE drops but REST remains reachable, the status indicator shows '● online' during the 5-second reconnect delay and potentially longer if reconnect also fails repeatedly. The 10s data tick only corrects this if REST also fails. Impact: misleading connectivity status for up to 5+ seconds per reconnect attempt. Fix: add d.connected = false before the tea.Tick call in the sseDisconnectMsg case.

**Location:** `cli/internal/tui/dashboard.go` line 272

---

Muriano (approved):

Re-review of SSE reconnect and dashboard refresh fix. The previously flagged issue (d.connected = false missing in sseDisconnectMsg) remains unaddressed after 3 rounds of feedback. No new issues introduced.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Dashboard refresh fix introduces proper SSE reconnect flow and a 'refreshing' indicator, but the sseDisconnectMsg handler fails to flip d.connected=false (flagged 4x in prior review and still unaddressed), leaving the UI showing '● online' during the 5s reconnect window. A few minor concurrency/UX rough edges around connected-state tracking and racing fetches also warrant attention.

### Issues

⚠️ **cli/internal/tui/dashboard.go:278** — sseDisconnectMsg handler does not set d.connected = false before scheduling the 5s reconnect tick. When SSE drops but REST is still reachable, the status indicator continues rendering '● online' for at least 5 seconds (longer if reconnect keeps failing), which contradicts the whole purpose of the indicator. This was flagged four times in the existing PR discussion and remains unaddressed. Add `d.connected = false` at the top of the `case sseDisconnectMsg:` branch.
🟡 **cli/internal/tui/dashboard.go:272** — d.connected is only set to true inside `case sseMsg:` after an event is received. On a fresh/reconnected SSE where no events arrive immediately, the indicator may stay in the non-connected state even though the stream is healthy. Consider marking connected=true as soon as connectSSE starts (or via an explicit sseConnectedMsg) rather than on first event.
🟡 **cli/internal/tui/dashboard.go:206** — `d.refreshing = false` fires on any dataMsg, including dataMsgs produced by the periodic tickCmd. A tick-driven refresh that completes between the user pressing 'r' and the user-triggered fetchData returning will prematurely clear the 'refreshing...' indicator. Minor, but worth gating with a flag correlating the dataMsg to the manual fetch (e.g., include a source field in dataMsg).
🟡 **cli/internal/tui/dashboard.go:145** — connectSSE now blocks the Cmd goroutine until StreamEvents returns. If StreamEvents never returns on a half-open TCP connection (no keepalive), there is no way to trigger reconnect. Confirm the underlying client enforces a read deadline or ping timeout; otherwise the dashboard can wedge with no reconnect.
🟡 **cli/internal/tui/dashboard.go:80** — sseCancel is captured at construction but the diff shows no code path invoking it. If it is not already called from a quit/teardown handler elsewhere, the context leaks across the lifetime of the program. Verify a cleanup path exists (e.g., on tea.Quit / ctrl+c).

### Suggestions

- Address the repeatedly flagged connectivity-indicator bug in `case sseDisconnectMsg:` by setting `d.connected = false` before the tea.Tick reconnect.
- Emit an explicit sseConnectedMsg when connectSSE establishes the stream, instead of relying on the first sseMsg to mark connected=true.
- Consider exponential backoff (or at least a jittered delay) for reconnect rather than a fixed 5s, to avoid tight reconnect storms against a flapping daemon.
- Tag dataMsg with its origin (manual vs tick) so `refreshing` is only cleared by the fetch it actually corresponds to.
- Ensure sseCancel is invoked on shutdown to release the SSE goroutine.

---

PR #256, opened 2026-04-23
feat: implement #249 — feat(cli): add pr <id> and issue <id> detail commands

### LINKED ISSUE #249: feat(cli): add pr <id> and issue <id> detail commands
## Problem

No way to see the full content of a review or triage from the CLI.

## New commands

- `heimdallm-cli pr <id>` — full PR with review: summary, issues found, suggestions, severity, date
- `heimdallm-cli issue <id>` — full issue with triage: summary, classification, suggestions, action taken, PR created
- Formatted with Lipgloss (borders, colors, sections)
- `--json` flag for raw output

## Files

- `cli/internal/cli/pr_detail.go` (new)
- `cli/internal/cli/issue_detail.go` (new)
- `cli/internal/cli/root.go` (register commands)

sergiotejon:
Auto-generated by Heimdallm in response to #249.

Closes #249

### DIFF
--- a/cli/internal/api/client.go
+++ b/cli/internal/api/client.go
@@ -276,6 +276,44 @@ func (c *Client) UndismissIssue(id int64) error {
 	return err
 }
 
+// PRDetail is the response from GET /prs/{id}.
+type PRDetail struct {
+	PR      PR       `json:"pr"`
+	Reviews []Review `json:"reviews"`
+}
+
+// IssueDetail is the response from GET /issues/{id}.
+type IssueDetail struct {
+	Issue   Issue         `json:"issue"`
+	Reviews []IssueReview `json:"reviews"`
+}
+
+// GetPR fetches a single PR with all its reviews.
+func (c *Client) GetPR(id int64) (*PRDetail, error) {
+	data, err := c.do("GET", fmt.Sprintf("/prs/%d", id))
+	if err != nil {
+		return nil, err
+	}
+	var detail PRDetail
+	if err := json.Unmarshal(data, &detail); err != nil {
+		return nil, fmt.Errorf("parsing PR detail: %w", err)
+	}
+	return &detail, nil
+}
+
+// GetIssue fetches a single issue with all its reviews.
+func (c *Client) GetIssue(id int64) (*IssueDetail, error) {
+	data, err := c.do("GET", fmt.Sprintf("/issues/%d", id))
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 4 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

New CLI detail commands are clean and readable. The previous review covered unrelated daemon changes that are no longer in this diff, so findings there are moot. One real issue: printTriageMap iterates a Go map without sorting, producing non-deterministic output order. A few minor nits around error swallowing and ID validation.

### Issues

🟡 **cli/internal/cli/pr_detail.go:170** — printTriageMap iterates the decoded map with `for k, v := range m` after pulling out 'severity' — Go map iteration order is intentionally randomized, so the classification block will print fields in a different order on each invocation. For CLI output consumed by humans (and potentially diffed or scripted), sort the remaining keys before printing to get stable output.
🟡 **cli/internal/cli/pr_detail.go:139** — `b, _ := json.MarshalIndent(v, ...)` silently drops the error. The input was just successfully unmarshalled so a failure here is unlikely, but on failure the code will print an empty string after the index, yielding a misleading `  3. ` line. Either log/skip on error or document that the error is unreachable.
🟡 **cli/internal/cli/pr_detail.go:55** — strconv.ParseInt happily accepts negative values and 0, which will then hit the API and return a not-found error from the server. Validating id > 0 client-side would give a clearer 'invalid PR ID' message and save a round-trip. Same applies to issue_detail.go:37.
🟡 **cli/internal/cli/issue_detail.go:115** — formatJSONStringArray swallows unmarshal errors (`err != nil || len(items) == 0` both collapse to returning ''). If the labels/assignees column ever contains malformed JSON, the user sees nothing rather than a diagnostic. Acceptable for a display helper, but worth separating the two cases or logging on verbose.
🟡 **cli/internal/cli/pr_detail.go:153** — printJSONArray short-circuits on raw equal to '', '[]', or 'null' via exact string comparison. Any whitespace variant ('[ ]', '[\n]') will fall through and then the json.Unmarshal + len check will correctly handle it, so this is only a minor cosmetic asymmetry — the early-return branch is redundant with the unmarshal+empty check below it.

### Suggestions

- Sort map keys in printTriageMap before printing so classification output is stable across runs.
- Extract the shared header-printing pattern between pr_detail.go and issue_detail.go into a small helper to avoid future drift.
- Consider adding a simple test (even a snapshot-style one) for the JSON output path of each command, since the human-readable path is hard to assert on.

---

PR #257, opened 2026-04-23
feat: implement #250 — feat(cli): improve follow — filters by repo and event type, colors

### LINKED ISSUE #250: feat(cli): improve follow — filters by repo and event type, colors
## Problem

`follow` shows all events unfiltered with basic formatting.

## Improvements

- `--repo org/repo` — filter by repository
- `--type pr,issue` — filter by event type
- Color by type: PR=blue, Issue=yellow, Error=red
- Clear type badges: `[PR]`, `[ISSUE]`, `[SYSTEM]`
- Better SSE data parsing for richer context

## Files

- `cli/internal/cli/follow.go`

sergiotejon:
Auto-generated by Heimdallm in response to #250.

Closes #250

### DIFF
--- a/cli/internal/cli/follow.go
+++ b/cli/internal/cli/follow.go
@@ -4,14 +4,56 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
-	"github.com/theburrowhub/heimdallm/cli/internal/api"
+	"github.com/charmbracelet/lipgloss"
 	"github.com/spf13/cobra"
+	"github.com/theburrowhub/heimdallm/cli/internal/api"
 )
 
+var (
+	badgePR     = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("#3B82F6"))
+	badgeIssue  = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("#F59E0B"))
+	badgeError  = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("#EF4444"))
+	badgeSystem = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("#06B6D4"))
+	styleError  = lipgloss.NewStyle().Foreground(lipgloss.Color("#EF4444"))
+	styleMuted  = lipgloss.NewStyle().Foreground(lipgloss.Color("#6B7280"))
+)
+
+func eventCategory(eventType string) string {
+	switch {
+	case strings.HasPrefix(eventType, "pr_"),
+		strings.HasPrefix(eventType, "review_"),
+		strings.HasPrefix(eventType, "circuit_breaker_"):
+		return "pr"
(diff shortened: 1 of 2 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Clean, well-tested enhancement to the follow command: adds category-based filtering, repo filtering, and colored badges. Helpers (truncate, toInt) exist in util.go. One real visual bug: ANSI color codes inflate printf's byte-based padding so columns won't align as intended. A few minor edge cases around filter semantics and missing fields.

### Issues

⚠️ **cli/internal/cli/follow.go:111** — Column alignment will be broken by ANSI escape codes. `fmt.Printf("%s %-10s %-28s %s\n", ts, eventBadge(event.Type), event.Type, ...)` — printf's `%-10s` pads based on byte length, but `eventBadge` returns a lipgloss-rendered string containing ANSI escape sequences (~10+ invisible bytes). The visible badge width is 6–8 chars, but the padded field is much wider than 10 bytes once ANSI is included, so subsequent columns won't line up. Either render badges to a fixed visible width using lipgloss.NewStyle().Width(8) (so lipgloss handles ANSI-aware padding) or pad the plain label first and render color after.
🟡 **cli/internal/cli/follow.go:96** — Silent repo-filter bypass for events without a `repo` field. `if repo := extractRepo(event.Data); repo != "" && !strings.EqualFold(repo, repoFilter) { continue }` — when `extractRepo` returns "" (no field or JSON parse failure) the filter is skipped, so those events are always shown even when `--repo` is set. That's probably intentional for system-scoped events, but it also means an event with a malformed `data` payload will leak through the filter. Worth either documenting the behavior in the flag help or, stricter, skipping events whose repo can't be determined when `--repo` is active.
🟡 **cli/internal/cli/follow.go:170** — `from_label`/`to_label` pairing only emits when both are present. If only `to_label` exists (initial labeling) or only `from_label` (label removed), the transition is silently dropped. Minor; consider showing single-sided transitions as `→ ready` / `blocked →` if those events exist.
🟡 **cli/internal/cli/follow.go:149** — Type tightening on `severity` and `error` from `%v` to string-only is a subtle behavior change. Previously a numeric severity or non-string error would still render; now it's silently dropped. Almost certainly fine given the daemon's current event schema, but if any producer emits a non-string value it will now be invisible instead of ugly.
🟡 **cli/internal/cli/follow.go:129** — `--type` flag help lists only categories (pr, issue, system) but users may reasonably expect to filter by concrete event types like `pr_detected`. The category-only semantics are a reasonable design, but the help text could clarify it's not a substring match on event type names to avoid confusion.

### Suggestions

- For ANSI-safe column widths, use lipgloss.NewStyle().Width(N).Render(...) on the badge (lipgloss pads around the ANSI sequences correctly) instead of relying on fmt's %-Ns.
- Consider pre-parsing event.Data once per event (into a map) and passing it to both the filter and the formatter, instead of json.Unmarshal'ing twice when --repo is set.

---

PR #259, opened 2026-04-23
feat: implement #251 — feat(cli): human-readable config output instead of raw JSON

### LINKED ISSUE #251: feat(cli): human-readable config output instead of raw JSON
## Problem

`config` dumps raw JSON which is hard to read.

## Improvements

- Format by sections: Server, Repos, AI, Issue Tracking, Discovery
- Color keys vs values
- Hide empty fields
- `--json` flag to keep raw output when needed

## Files

- `cli/internal/cli/config.go`

sergiotejon:
Auto-generated by Heimdallm in response to #251.

Closes #251

### DIFF
--- a/cli/internal/cli/config.go
+++ b/cli/internal/cli/config.go
@@ -3,12 +3,21 @@ package cli
 import (
 	"encoding/json"
 	"fmt"
+	"sort"
+	"strings"
 
+	"github.com/charmbracelet/lipgloss"
 	"github.com/spf13/cobra"
 )
 
+var (
+	cfgSectionStyle = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("#7C3AED"))
+	cfgKeyStyle     = lipgloss.NewStyle().Foreground(lipgloss.Color("#06B6D4"))
+)
+
 func newConfigCmd() *cobra.Command {
-	return &cobra.Command{
+	var jsonFlag bool
+	cmd := &cobra.Command{
 		Use:   "config",
 		Short: "Show the daemon's running configuration",
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid refactor that adds a human-readable config renderer while keeping --json as an escape hatch. Good test coverage on helpers and section builders. Two concerns: potentially mismatched keys inside pr_metadata, and local_dir_base being rendered as a string list when the field is likely a single string on the daemon side.

### Issues

⚠️ **cli/internal/cli/config.go:113** — Inside pr_metadata the code reads keys 'reviewers' and 'labels' (no prefix) but then 'pr_assignee' and 'pr_draft' (with prefix). The daemon's pr_metadata block almost certainly uses consistent naming (either all-bare or all-prefixed). If the real keys are 'assignee'/'draft', those two fields will always be hidden silently because cfgEmpty(nil) is true. Verify against the daemon /config response schema before merging — this is the kind of bug that won't surface in unit tests because the tests feed the same (possibly wrong) keys the code looks for.
⚠️ **cli/internal/cli/config.go:74** — local_dir_base is passed to cfgStringList, which only renders when the value is []any. PR #184 exposed local_dir_base in /config, and based on its usage elsewhere (a single base directory path) it is likely a plain string, not a list. If so, this call is a silent no-op and the field will never appear in human output. The test sets it to []any{} which exercises the 'hidden when empty' branch but never the populated path, so the bug is not caught. Confirm the actual type in the daemon response and either switch to cfgKV or accept both shapes.
🟡 **cli/internal/cli/config.go:196** — cfgEmpty treats bool false as 'empty', so any explicitly-configured false-valued flag (e.g. activity_log_enabled: false, pr_draft: false, issue_tracking.enabled: false) is indistinguishable from 'not set' in the human view. This is the intended behavior per the tests (TestCfgIssueTrackingLinesDisabled expects 0 lines), but it means operators cannot verify a flag was explicitly disabled without falling back to --json. Worth documenting this behavior, or at minimum calling it out in the PR description so users know when to reach for --json.
🟡 **cli/internal/cli/config.go:220** — cfgFmtVal handles float64 integers and falls through to %v for everything else. When val is a []any or map[string]any (shouldn't happen via cfgKV because cfgEmpty filters empties, but could happen for a populated slice/map passed to cfgKV by mistake), the rendered output would be Go's default map/slice formatting like 'map[a:1]' — ugly but not dangerous. Low priority, just a nit that a defensive branch or a comment clarifying 'only scalars expected here' would prevent future surprises.
🟡 **cli/internal/cli/config_test.go:1** — Tests never exercise the top-level printHumanConfig orchestration (section ordering, blank-line separators between non-empty sections, suppression of entirely-empty sections). Given the section skipping logic has subtle 'first' state, a single test that captures stdout and asserts section ordering for a minimal and a full config would catch regressions cheaply.

### Suggestions

- Verify the exact key names the daemon emits under pr_metadata (assignee vs pr_assignee, draft vs pr_draft) against the /config response struct before merging.
- Confirm local_dir_base type in the daemon response; if it is a string, switch to cfgKV. If it is a list, update the test to include a populated case.
- Consider a snapshot-style test that feeds a realistic full config map to printHumanConfig and compares captured stdout — faster to maintain than many small builder tests.
- Minor: define the color constants (#7C3AED, #06B6D4) as named vars or alongside the existing CLI theme if one exists, to keep the palette centralized.

---

PR #263, opened 2026-04-23
feat: implement #262 — feat: unify all build artifacts under GoReleaser

### LINKED ISSUE #262: feat: unify all build artifacts under GoReleaser
## Problem

The release workflow has 4 separate build jobs with different tooling:

| Artifact | Current tooling | Job |
|----------|----------------|-----|
| CLI binaries | GoReleaser | `Build CLI binaries` |
| Docker image | docker/build-push-action | `docker-build-push` |
| Linux packages (.deb, .rpm, .AppImage) | fpm + manual scripting | `Build Linux packages` |
| macOS DMG | Flutter + create-dmg + codesign | `Build Heimdallm.dmg` |

This creates maintenance overhead: 4 different build pipelines, 4 different failure modes, inconsistent artifact naming, and the Homebrew formula update only works via GoReleaser.

## Proposal

Consolidate as much as possible under GoReleaser:

### Already GoReleaser
- CLI binaries (6 platforms) ✓
- Homebrew formula ✓ (when token works)

### Can move to GoReleaser
- **Docker image**: GoReleaser has native Docker support (`dockers:` config) — build + push to GHCR with semver tags
- **Linux packages**: GoReleaser has `nfpms:` for .deb and .rpm natively — replaces the fpm scripting
- **Checksums + SBOMs**: GoReleaser generates these automatically

### Stays separate
- **macOS DMG**: GoReleaser does not support Flutter builds or DMG creation. This stays as its own job.
- **AppImage**: GoReleaser does not support AppImage natively. Could use a custom publisher or keep separate.

## Benefits

- Single `.goreleaser.yml` config for CLI + Docker + Linux packages
- Consistent artifact naming and checksums
- Homebrew formula auto-updated on every release (not just CLI)
- Fewer CI jobs, faster releases
- One tool to debug instead of four

### Phase 1: Move Docker to GoReleaser
```yaml
dockers:
  - image_templates:
      - ghcr.io/theburrowhub/heimdallm:{{ .Version }}
      - ghcr.io/theburrowhub/heimdallm:latest
    dockerfile: docker/Dockerfile
    build_flag_templates:
      - --label=org.opencontainers.image.version={{ .Version }}
```

### Phase 2: Move Linux packages to GoReleaser
```yaml
nfpms:
  - id: heimdallm
    package_name: heimdallm
    formats: [deb, rpm]
    homepage: https://github.com/theburrowhub/heimdallm
    description: AI-powered GitHub PR review agent
    dependencies: [libgtk-3-0, libayatana-appindicator3-1]
```

### Phase 3: Simplify release.yml
Remove `docker-build-push` and `build-linux` jobs. Keep only:
1. `release-please` — tag creation
2. `goreleaser` — CLI + Docker + Linux packages + Homebrew
3. `build-macos` — DMG (Flutter, stays separate)

## Files

- `cli/.goreleaser.yml` → move to root `.goreleaser.yml` (covers all artifacts)
- `.github/workflows/release.yml` — simplify to 3 jobs
- `docker/Dockerfile` — no changes (GoReleaser uses it as-is)

sergiotejon:
Auto-generated by Heimdallm in response to #262.

Closes #262


Muriano:
## 🔴 HIGH Issue

RPM override dependency `libayatana-appindicator3-1` is a Debian package name, not an RPM package name. The correct RPM equivalent is `libayatana-appindicator-gtk3` (or similar, depending on distro). The old workflow used `gtk3`, `libnotify`, `libsecret` for RPM deps. This will cause RPM install failures when the dependency resolver cannot find the package. Fix: use the correct RPM package name (e.g., `libayatana-appindicator-gtk3`).

**Location:** `.goreleaser.yml` line 90

---

Muriano:
## ⚠️ MEDIUM Issue

Missing Linux package dependencies compared to old workflow. Old .deb had: `libgtk-3-0`, `libayatana-appindicator3-1`, `libnotify4`, `libsecret-1-0`. New nfpms only declares `libgtk-3-0` and `libayatana-appindicator3-1` — missing `libnotify4` and `libsecret-1-0`. Old RPM had `gtk3`, `libnotify`, `libsecret` — new RPM override is also missing `libnotify` and `libsecret`. This may cause runtime failures if the daemon depends on these libraries. Fix: add the missing dependencies to both default (deb) and rpm override sections.

**Location:** `.goreleaser.yml` line 83

---

Muriano:
## ⚠️ MEDIUM Issue

Docker `latest` tag is now always applied, including on `workflow_dispatch` runs. The old workflow conditionally excluded the `latest` tag on manual dispatch (`${{ github.event_name != 'workflow_dispatch' && 'type=raw,value=latest' || '' }}`). This behavioral change means manual re-releases of older tags could overwrite `latest` to point at an older version. Fix: consider using GoReleaser's `skip_push` or conditional template logic to preserve the old behavior, or confirm this is intentional.

**Location:** `.goreleaser.yml` line 64

---

Muriano (changes requested):

Solid consolidation of build artifacts under a single GoReleaser config. The workflow is much cleaner. However, there are confirmed issues with Linux package dependencies — an incorrect RPM package name and missing dependencies compared to the previous workflow.

### Suggestions

- The old pipeline ran `go test ./...` (daemon) and `flutter test` as a gate before building release artifacts. The new goreleaser job has no test step. Consider adding a test step before the GoReleaser action, or ensure tests run in a separate required job that gates the release.
- The old workflow generated artifact attestation via `actions/attest-build-provenance`. This supply-chain security feature is now removed. Consider adding GoReleaser's built-in signing/checksums verification or re-adding the attestation step after the GoReleaser action for the Docker image.
- In `Dockerfile.goreleaser`, AI CLI tool installs use `|| true` to silently swallow failures. Consider at minimum logging a warning when install fails, so operators can tell from build logs whether the tools are actually present in the image.
- The nfpms section only packages the daemon binary (`heimdallm-daemon`). The old `build-linux` job bundled the Flutter GUI app alongside the daemon into .deb/.rpm/.AppImage. If the intent is to ship a GUI-less daemon package, document this change in the PR description. If the Flutter app should still be included, this is a feature regression.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

The GoReleaser unification significantly simplifies the release workflow, but introduces several regressions in Linux packaging: the .deb/.rpm packages no longer include the Flutter GUI, desktop entry, or icons (they are now daemon-only), the AppImage target is dropped entirely, and the GTK dependencies declared in nfpms no longer match what the CGO_ENABLED=0 daemon actually needs. Supply-chain attestations for the Docker image are also lost. The previous Lua/ideogram findings are unrelated to this PR and do not apply.

### Issues

🔴 **.goreleaser.yml:78** — Linux nfpms packages (.deb/.rpm) now contain only the daemon binary. The previous workflow staged the Flutter GUI bundle under /opt/heimdallm, installed a /usr/bin/heimdallm symlink, a desktop entry, and icons at 48/128/256/512. All of this is gone. Also, the binary name shifts semantics: in the old .deb, `heimdallm` was the Flutter GUI and `heimdalld` was the daemon; in the new .deb, `heimdallm` is the daemon. This is a breaking change for anyone upgrading an existing installation and effectively removes desktop support on Linux. If this is intentional (daemon-only packaging), it should be documented in the PR description and release notes; otherwise this is a regression.
🔴 **.github/workflows/release.yml:0** — AppImage build target is removed with no replacement. The previous pipeline produced a universal Heimdallm-<version>-x86_64.AppImage that covered Arch, NixOS, and other distros without native package support. Losing this silently strands those users.
🔴 **.goreleaser.yml:86** — nfpms declares `libgtk-3-0` and `libayatana-appindicator3-1` (and `gtk3`/`libayatana-appindicator3-1` on RPM) as dependencies, but the daemon is built with CGO_ENABLED=0 and does not link against GTK. These dependencies are only meaningful if the Flutter GUI is shipped in the package — which it no longer is. As-is, apt/dnf will pull in GTK on servers that just want the daemon, which is both wrong and wasteful. Either drop the deps, or restore the GUI bundling.
⚠️ **.github/workflows/release.yml:47** — Build provenance attestation is lost. The old docker-build-push job ran `actions/attest-build-provenance@v2` with `push-to-registry: true` and had `attestations: write` permission. GoReleaser's docker integration does not produce SLSA provenance out of the box, and no replacement was added. Consumers verifying image provenance (cosign verify-attestation / gh attestation verify) will break on the next release. Consider re-adding a post-release attestation step using the image digest emitted by GoReleaser, or switch to `docker_signs`/cosign.
⚠️ **.goreleaser.yml:60** — Docker image_templates unconditionally include `ghcr.io/theburrowhub/heimdallm:latest`. The previous metadata-action explicitly suppressed `latest` on `workflow_dispatch` (`github.event_name != 'workflow_dispatch' && 'type=raw,value=latest'`) so that a manual re-run of an older tag would not clobber the rolling `latest` pointer. Under the new config, a workflow_dispatch on v1.2.3 after v1.3.0 is out will silently move `latest` backwards. Consider gating the `latest` template behind a condition or using GoReleaser's `skip_push` / `--skip` appropriately for manual runs.
🟡 **.goreleaser.yml:60** — The SHA-based Docker tag (`type=sha,prefix=`) from the old metadata config is not reproduced. If downstream deployments pinned images by short SHA, those tags will stop being produced. Low-impact if unused, but worth confirming.
🟡 **.github/workflows/release.yml:48** — `id-token: write` permission is retained but, with attestations removed, nothing in the job currently uses OIDC. Either re-introduce attestation/signing (preferred) or drop the unused permission to minimize token scope.
🟡 **docker/Dockerfile.goreleaser:1** — Two Dockerfiles now exist (`docker/Dockerfile` for local docker-compose builds and `docker/Dockerfile.goreleaser` for releases). They will drift over time — e.g., an upgrade to node:22-alpine or a new runtime dep added to one but not the other will produce behavioral divergence between local and published images. Consider consolidating via a multi-stage Dockerfile with a build arg that toggles between compile-in-container and copy-prebuilt-binary, or at minimum a CI check that asserts the runtime layers are identical.
⚠️ **docker/Dockerfile.goreleaser:9** — Each `npm install -g ... 2>/dev/null || true` swallows all failures including network errors, which means a broken release can ship an image missing claude-code / gemini-cli / codex / opencode without any CI signal. At minimum log the failure, and consider removing `|| true` so a transient npm failure fails the release rather than silently shipping a degraded image.

### Suggestions

- Clarify in the PR description whether dropping Flutter GUI from Linux .deb/.rpm and removing AppImage is intentional. If yes, add release notes; if no, restore nfpms `contents:` stanzas that pull the Flutter bundle, desktop file, and icons, and add a separate AppImage build step (GoReleaser has no native AppImage, but the existing shell block could be kept as a post-goreleaser step).
- Re-introduce SLSA provenance: either add a post-release `actions/attest-build-provenance` step keyed off the digest in GoReleaser's `artifacts.json`, or enable cosign signing via GoReleaser's `docker_signs`.
- Gate the `latest` docker tag on `workflow_dispatch == false` (e.g. via build_flag_templates or a conditional template) so manual re-releases of older tags don't rewind `latest`.
- Align nfpms dependencies with what's actually packaged: for a daemon-only build with CGO_ENABLED=0, drop GTK/appindicator deps; for a GUI-bundled build, restore them along with libnotify/libsecret as in the old workflow.
- Drop `|| true` on the npm global installs in Dockerfile.goreleaser, or at least pin versions and fail fast — silent partial installs are hard to debug in production.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

Consolidates CLI, Docker, and Linux-package builds under one GoReleaser config. The unification is clean, but it silently drops several artifacts and properties that the old workflow produced: the Flutter GUI bundle inside .deb/.rpm, the .AppImage entirely, SLSA build-provenance attestations, and the pre-build test gate. The nfpms RPM override also uses a Debian-style package name that will not resolve on Fedora/RHEL. These are real regressions relative to what the previous release workflow shipped and should be addressed before merge.

### Issues

🔴 **.goreleaser.yml:75** — Feature regression: .deb/.rpm packages no longer include the Flutter UI. The removed `build-linux` job built `flutter_app` (linux release) and staged the bundle at /opt/heimdallm alongside the daemon, so users installing heimdallm.deb/.rpm got the full GUI app plus the desktop entry and icons. The new nfpms block only packages the `heimdallm-daemon` binary — no UI bundle, no .desktop file, no icons, no /usr/bin symlink. If this is intentional (Linux desktop users expected to use AppImage/Flathub/etc.), document it in the PR description; otherwise these packages now ship a different product than before under the same name.
🔴 **.goreleaser.yml:85** — The old workflow produced a universal `.AppImage` (built via appimagetool) that targeted Arch, NixOS, and other non-deb/rpm distros. GoReleaser nfpms does not emit AppImage, and no replacement is configured. Users on those distros lose their only install path in this PR.
⚠️ **.goreleaser.yml:93** — RPM dependency name is wrong: `libayatana-appindicator3-1` is the Debian package name; Fedora/RHEL ship it as `libayatana-appindicator-gtk3`. The previous workflow already had this right (it declared `libayatana-appindicator3-1` only for deb and `gtk3`/`libnotify`/`libsecret` for rpm). As written, rpm installs will fail to resolve dependencies on stock Fedora/RHEL.
⚠️ **.goreleaser.yml:88** — Package declares GTK/appindicator runtime deps (gtk3, libayatana-appindicator3-1) but the package only contains the headless daemon binary now — there is no GUI in the package. Either the deps are vestigial from the old UI-bundling layout (and should be dropped), or the intent is still to bundle the UI (in which case see the high-severity finding above). Current state is internally inconsistent.
⚠️ **.github/workflows/release.yml:47** — Build provenance is silently dropped. The old `docker-build-push` job generated SLSA attestations via `actions/attest-build-provenance@v2` and had `attestations: write` + `id-token: write`. The new job keeps `id-token: write` but removes `attestations: write` and the attest step, and no GoReleaser equivalent (`sboms:`, `signs:`, or `--sign`) is configured. If attestations are a downstream requirement (supply-chain gates, cosign verification, etc.) this is a regression — worth confirming whether it was deliberate.
⚠️ **.goreleaser.yml:55** — `dockers` block only builds linux/amd64 (the daemon is declared amd64-only). This matches what the old Dockerfile produced, but `image_templates` includes `:latest` unconditionally, whereas the old workflow skipped `latest` on `workflow_dispatch` (see the removed docker/metadata-action `tags:` block). A manual dispatch re-run on an older tag will now overwrite `:latest` with that older image. Consider conditionalizing `:latest` or passing `--skip=docker` selectively on dispatch, similar to how homebrew is skipped.
🟡 **.github/workflows/release.yml:44** — The old `build-linux` job ran `go test ./... -timeout 60s` (daemon) and `flutter test` before producing artifacts. The new single GoReleaser job has no test step. If release is triggered via `workflow_dispatch` on a tag, nothing on the release path enforces that tests pass for that tag's tree — you're relying entirely on the PR-gating CI that ran when the commit landed on main. Worth a line of justification or an explicit test step.
🟡 **.github/workflows/release.yml:55** — `go-version-file: cli/go.mod` drives the toolchain version for the whole job, but `cache-dependency-path` lists both `cli/go.sum` and `daemon/go.sum`, and GoReleaser then builds the daemon too. If `daemon/go.mod` ever requires a newer Go version than `cli/go.mod`, the daemon build will fail with a confusing cache hit. Safer to either pin to `daemon/go.mod` (it's usually the stricter one) or add a sanity check that the two go directives agree.
🟡 **.goreleaser.yml:30** — Only `heimdallm-cli` is archived; `heimdallm-daemon` is not emitted as a tar.gz. Today that's fine because the daemon is distributed via Docker and .deb/.rpm, but if anyone wants to grab a raw binary from the release page (air-gapped install, custom systemd unit, etc.) there is no daemon artifact. Minor gap, flagging in case it was unintentional.

### Suggestions

- If the Flutter UI is genuinely no longer part of the .deb/.rpm, rename/repurpose the package (e.g. `heimdallm-daemon`) so users don't silently get a different product under the old name, and update the description/homepage accordingly. Alternatively, keep a second nfpms entry that packages the Flutter bundle + daemon and drop the current minimal one.
- Fix the RPM appindicator dep: move `libayatana-appindicator3-1` out of the rpm override and use `libayatana-appindicator-gtk3` (plus re-adding `libnotify` / `libsecret` if the bundled UI is restored).
- Either re-introduce SLSA attestations via `attestations:` in the goreleaser config (or a post-release `actions/attest-build-provenance` step over `dist/checksums.txt`), or note in the PR description that provenance was deliberately dropped.
- Condition the `:latest` Docker tag on non-`workflow_dispatch` runs (use a template guard on `image_templates`, or a separate workflow step) to preserve the previous behavior where manual dispatches didn't clobber latest.
- Consider restoring AppImage production (it was the only install path for Arch/NixOS users). GoReleaser itself doesn't build AppImage, but the old appimagetool step can live as a small follow-up job consuming `dist/` artifacts.

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Unifies CLI, daemon Docker image, Linux packages, and Homebrew under a single GoReleaser config, replacing fpm-based packaging and the previous per-job Docker build. The consolidation is clean and reduces CI complexity significantly. However, several regressions versus the old workflow were introduced: missing runtime dependencies in the Linux packages, an incorrect RPM dependency name, unconditional `latest` Docker tag on manual dispatch, loss of build-provenance attestation, loss of daemon/Flutter tests as a gate before packaging, and the .deb/.rpm no longer ship the Flutter GUI bundle that the previous fpm-based workflow staged under /opt/heimdallm.

### Issues

🔴 **.goreleaser.yml:90** — RPM override lists `libayatana-appindicator3-1`, which is a Debian package name. RPM distros use `libayatana-appindicator-gtk3` (Fedora) or equivalent. This will break RPM installation on dependency resolution.
🔴 **.goreleaser.yml:46** — The new nfpms packages only bundle the daemon binary (heimdallm). The old workflow built the Flutter Linux bundle and staged icons, a .desktop file, and a /usr/bin symlink into /opt/heimdallm. Users installing the .deb/.rpm will no longer get the GUI app, icons, or desktop entry — a major regression in what the Linux packages deliver. If the intent is to ship daemon-only packages, this should be called out explicitly in the PR description and release notes.
⚠️ **.goreleaser.yml:83** — Missing runtime deps vs. the old workflow: deb lost `libnotify4` and `libsecret-1-0`; rpm override is missing `libnotify` and `libsecret`. If the daemon (or the bundled GUI, if re-added) links to these, installs will succeed but runtime will fail.
⚠️ **.goreleaser.yml:57** — Docker `latest` tag is now applied unconditionally, including on `workflow_dispatch`. Old workflow explicitly excluded `latest` on manual dispatch so re-releasing an older tag would not overwrite latest. Use a `skip_push`/template guard (e.g., only push `latest` when the tag matches the current highest semver or when not dispatched manually) or document the behavior change.
⚠️ **.github/workflows/release.yml:47** — Lost `actions/attest-build-provenance@v2` for the Docker image. The old job generated a signed provenance attestation pushed to the registry; GoReleaser's Docker build does not replicate this. If supply-chain attestations are a requirement, wire provenance back in (e.g., a follow-up step using `cosign`/`attest-build-provenance` against the image digest GoReleaser emits).
⚠️ **.github/workflows/release.yml:47** — The old `build-linux` job ran `go test ./... -timeout 60s` and `flutter test` as a gate before building packages. The new unified job removes this gate entirely. If CI doesn't already guarantee tests have passed on the release SHA (release-please tags a merged commit, but workflow_dispatch can target arbitrary tags), broken releases could ship. Consider adding a test pre-job or confirm another workflow enforces this on tagged commits.
⚠️ **.goreleaser.yml:20** — Daemon build sets `CGO_ENABLED=0` and targets only linux/amd64. The previous daemon build used `make build-daemon`; confirm it did not require CGO (e.g., for sqlite) — switching to pure-Go silently can break features that depend on cgo-only drivers.
🟡 **.goreleaser.yml:53** — `dockers:` entry has no `goarch`/`goos` constraints and no `use: buildx` / platforms set. The old workflow used buildx with gha cache. Result: image is single-arch amd64 only and loses registry cache; if multi-arch was intended, add a matching `dockers_v2`/manifest or `use: buildx` with platforms.
🟡 **docker/Dockerfile.goreleaser:5** — Installs multiple global npm CLIs with `2>/dev/null || true`, so failures during image build are silently swallowed. If Anthropic/Gemini/OpenAI/opencode CLIs fail to install, the image will be shipped missing core tools with no indication. Either let the install fail hard or log the outcome.
🟡 **docker/Dockerfile.goreleaser:3** — Base image `node:20-alpine` differs from the previous `docker/Dockerfile`'s base (worth confirming parity). Also, installing many global npm packages inflates image size significantly; consider a multi-stage build or pinning versions for reproducibility.
🟡 **.goreleaser.yml:109** — `release.mode: keep-existing` means GoReleaser won't overwrite assets already uploaded by `release-please` or the macOS job — good — but it also means re-running the release after a partial failure won't replace bad artifacts. Document the re-run procedure (delete the asset, then re-dispatch).
🟡 **.github/workflows/release.yml:47** — `setup-go` uses `go-version-file: cli/go.mod` but caches both `cli/go.sum` and `daemon/go.sum`. If `cli/go.mod` and `daemon/go.mod` specify different Go toolchains, the daemon build may run under an unexpected version. Prefer aligning the toolchains or being explicit about which version is authoritative.

### Suggestions

- Clarify in the PR description whether Linux .deb/.rpm are intentionally daemon-only now (dropping the Flutter GUI bundle) — this is a user-visible packaging change.
- Fix RPM dependency naming: use `libayatana-appindicator-gtk3` (Fedora) and re-add `libnotify`/`libsecret` equivalents; mirror the full dep set on the deb side (`libnotify4`, `libsecret-1-0`).
- Restore build-provenance attestation for the Docker image, either via a post-release job keyed off GoReleaser's artifact metadata or by running `cosign attest`/`attest-build-provenance` against the pushed digest.
- Preserve the old `latest` tag behavior on `workflow_dispatch` by using a GoReleaser template guard on the `latest` image tag (e.g., skip it when `{{ .IsSnapshot }}` or via a conditional tag list driven by an env var set from the workflow).
- Re-introduce the test gate (daemon `go test` and `flutter test`) before packaging, either as a dependent job or inside the GoReleaser workflow, so a broken tag cannot produce packages.
- Verify the daemon actually builds correctly with `CGO_ENABLED=0`; if any dependency requires cgo, either enable it or switch to a pure-Go alternative explicitly.

---

PR #277, opened 2026-04-23
feat: implement #268 — step 1: fix Homebrew tap auto-update with org secret

### LINKED ISSUE #268: step 1: fix Homebrew tap auto-update with org secret
Part of epic #265

Verify org secret TAP_GITHUB_TOKEN works for GoReleaser brew publish. Repo-level secret was deleted. Next release should auto-update the formula.

## Acceptance
- [ ] Release triggers GoReleaser
- [ ] Formula auto-updated
- [ ] brew upgrade works

Order: first step, blocks all others.

sergiotejon:
Auto-generated by Heimdallm in response to #268.

Closes #268

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -521,6 +521,18 @@ jobs:
         with:
           ref: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
           fetch-depth: 0
+
+      - name: Verify Homebrew tap token
+        if: github.event_name != 'workflow_dispatch'
+        run: |
+          if [ -z "$HOMEBREW_TAP_TOKEN" ]; then
+            echo "::error::HOMEBREW_TAP_TOKEN is empty. Ensure the org secret TAP_GITHUB_TOKEN is granted to this repository."
+            exit 1
+          fi
+          echo "Homebrew tap token is available"
+        env:
+          HOMEBREW_TAP_TOKEN: ${{ secrets.TAP_GITHUB_TOKEN }}
+
       - uses: actions/setup-go@v5
         with:
           go-version-file: cli/go.mod


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Small, focused change that adds a pre-flight verification step to ensure the Homebrew tap token secret is available before proceeding with the release job. The logic is correct and the error message is actionable. Minor concerns around the workflow_dispatch bypass and token validation depth.

### Issues

🟡 **.github/workflows/release.yml:523** — The `if: github.event_name != 'workflow_dispatch'` condition skips the token verification entirely for manual dispatches. This means a manual release run could proceed with an empty token and fail later in a less clear way. Consider either running the check always (and only failing hard when the token is truly required) or documenting why workflow_dispatch is exempt.
🟡 **.github/workflows/release.yml:527** — The check only verifies the token is non-empty but not that it's valid. A malformed or expired token would still pass this check and fail later during the tap push with a less actionable error. Consider a lightweight validation like `gh auth status` or an API call to verify the token actually works against the tap repo.
🟡 **.github/workflows/release.yml:525** — The error message references 'org secret TAP_GITHUB_TOKEN' but the env var is named HOMEBREW_TAP_TOKEN. While the message is still understandable, it could confuse an operator debugging the failure since the two names don't obviously correspond. Consider clarifying the mapping explicitly.

### Suggestions

- Consider extracting the token verification into a reusable composite action if similar checks are needed in other workflow jobs.
- Add a comment above the step explaining why workflow_dispatch is exempt (presumably to allow testing without the secret), to prevent future confusion.
- If the downstream step that uses HOMEBREW_TAP_TOKEN already produces a clear error on empty token, this pre-flight check may be redundant — worth verifying.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Simple and well-placed fail-fast check that verifies the Homebrew tap token secret is available before the GoReleaser step runs. The implementation follows GitHub Actions best practices: token is accessed via env var (not interpolated into the shell script), the error message is actionable, and the check is correctly scoped to automated triggers only.

### Suggestions

- Consider also verifying the token on workflow_dispatch runs. Currently the `if: github.event_name != 'workflow_dispatch'` condition skips the check for manual triggers — if those also need the token for the Homebrew tap update, a missing token would fail later with a less clear error.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Small, focused CI change that adds a preflight check for the HOMEBREW_TAP_TOKEN (sourced from the org secret TAP_GITHUB_TOKEN) before the Homebrew tap release step. The guard fails fast with a clear error when the secret is not granted to the repo, which is a sensible improvement. Logic is correct and side-effect free.

### Issues

🟡 **.github/workflows/release.yml:523** — The condition `if: github.event_name != 'workflow_dispatch'` silently skips the verification for manual runs. If a manual dispatch is ever used to publish a real release, the token could still be missing and the later Homebrew step would fail with a less actionable error. Consider also verifying (or at least warning) on workflow_dispatch, or documenting why dispatch is exempt.
🟡 **.github/workflows/release.yml:526** — Using `[ -z "$HOMEBREW_TAP_TOKEN" ]` on a secret value is safe (GitHub masks it in logs), but the default shell lacks `set -euo pipefail`. Not a bug here since the only command after the check is `echo`, but adding `shell: bash` with pipefail is the project-wide hardening pattern worth considering.

### Suggestions

- Consider extending the check to workflow_dispatch runs, or add a comment explaining why dispatch is intentionally skipped (e.g., manual re-runs where the maintainer knowingly bypasses tap publishing).
- If other jobs in this workflow also depend on TAP_GITHUB_TOKEN, factor the verification into a reusable composite action or a single early job that gates downstream jobs via `needs:`, to avoid repeating the guard.
- Optionally log a non-sensitive confirmation like the token's length or first-4-chars-hash is unnecessary — the current `echo "Homebrew tap token is available"` is the right level of detail.

---

PR #278, opened 2026-04-23
feat: implement #269 — step 2: move Docker image build to GoReleaser

### LINKED ISSUE #269: step 2: move Docker image build to GoReleaser
Part of epic #265

Add dockers: section to .goreleaser.yml. Keep old docker-build-push job until verified.

## Important
- Keep SLSA attestation
- Match current tag pattern (latest, x.y.z, x.y, x)
- Do NOT remove old job until new one verified

## Acceptance
- [ ] GoReleaser builds+pushes Docker image
- [ ] Tags match current pattern
- [ ] Old job removed after verification

sergiotejon:
Auto-generated by Heimdallm in response to #269.

Closes #269

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -147,6 +147,69 @@ jobs:
           push-to-registry: true
 
   # ═══════════════════════════════════════════════════════════════════════════
+  # Docker image — legacy build-push-action (remove after GoReleaser verified)
+  # TODO(#269): Remove this job once goreleaser-daemon is verified in production
+  # ═══════════════════════════════════════════════════════════════════════════
+  docker-build-push:
+    name: Docker image (legacy — verify then remove)
+    needs: [release-please, tests]
+    if: always() && needs.tests.result == 'success'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+    env:
+      TAG: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.tag || needs.release-please.outputs.tag_name }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels)
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            ${{ github.event_name != 'workflow_dispatch' && 'type=raw,value=latest' || '' }}
+            type=semver,pattern={{version}},value=${{ env.TAG }}
+            type=semver,pattern={{major}}.{{minor}},value=${{ env.TAG }}
+            type=semver,pattern={{major}},value=${{ env.TAG }}
+            type=sha,prefix=
+
+      - name: Build and push Docker image
+        id: push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: docker/Dockerfile
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@v2
+        with:
+          subject-name: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          subject-digest: ${{ steps.push.outputs.digest }}
+          push-to-registry: true
+
+  # ═══════════════════════════════════════════════════════════════════════════
   # Linux packages — .deb, .rpm (GoReleaser nfpms) + .AppImage
   #
   # .deb/.rpm are built by GoReleaser's nfpms integration.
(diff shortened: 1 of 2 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR adds a legacy Docker build-push job alongside the new GoReleaser-based one as a safety net during the #269 transition. Changes are small, well-scoped, and clearly documented as temporary. Main concern is that both jobs push the same tags to the same registry in parallel, which is intentional per the PR narrative but introduces a race where the last-completing job wins and attestations may disagree.

### Issues

⚠️ **.github/workflows/release.yml:152** — Both `docker-build-push` (legacy) and `goreleaser-daemon` (new) run in parallel after the test gate and push to the same registry with overlapping tags (`latest`, semver, sha). Whichever job finishes last wins the tag pointer, so the digest published under e.g. `:latest` is non-deterministic during the verification window. This makes it hard to actually verify the GoReleaser output in production (consumers may be pulling either image) and defeats part of the purpose of keeping the legacy job as a comparison baseline. Consider pushing the legacy job to a distinct tag suffix (e.g. `-legacy`) or a different image name during the verification period, or serialize them so the new job always wins.
🟡 **.github/workflows/release.yml:197** — Two separate `attest-build-provenance` steps (one here, one in the existing goreleaser job) will publish attestations for the same registry subject but for different digests produced by different builders. Registry attestation consumers and signature-verification tooling can get confused when multiple provenance statements exist for overlapping tags. Worth confirming the attestation push-to-registry behavior is additive and not conflicting, and consider dropping attestation from the legacy job since it's marked for removal.
🟡 **.github/workflows/release.yml:151** — `if: always() && needs.tests.result == 'success'` is correct for handling the skipped `release-please` dependency on workflow_dispatch, but it's inconsistent with how other sibling jobs likely gate themselves — worth a quick check that this matches the pattern used by `goreleaser-daemon` so both jobs have identical activation conditions; otherwise one could run while the other is skipped, silently regressing coverage.
🟡 **.github/workflows/release.yml:184** — The `type=raw,value=latest` conditional uses `github.event_name != 'workflow_dispatch'` but the existing `goreleaser-daemon` job may have different latest-tag semantics. If the two jobs disagree on whether `:latest` is pushed for a given trigger, the tag state becomes dependent on which job runs, amplifying the race condition above. Verify both jobs apply `:latest` under identical conditions.

### Suggestions

- Consider pushing the legacy image under a distinct tag namespace (e.g. `:latest-legacy`, `:v1.2.3-legacy`) during the verification window so both images are independently observable and the GoReleaser output owns the canonical tags from day one.
- Add a tracking checklist or acceptance criteria in issue #269 describing exactly what 'verified in production' means (matching digests? passing smoke tests? N releases observed?) so the removal of this job isn't indefinitely deferred.
- Since the job is explicitly temporary, consider dropping the attestation step from the legacy job to avoid double-provenance and simplify eventual removal.
- The TODO comment references #269 — good. Consider also adding a dated reminder (e.g. 'remove by 2026-05-15') to prevent the 'temporary' job from lingering.

---

Muriano:
## ⚠️ MEDIUM Issue

Tag collision between goreleaser-daemon and docker-build-push: both jobs push to the same REGISTRY/IMAGE_NAME with identical semver tags (e.g., `v1.2.3`, `1.2`, `1`, `latest`). Since they run in parallel, the last one to finish overwrites the other's tags. The attestation from the first job then references a digest that no longer matches the tag. Fix: either (a) use a distinct tag suffix for the legacy job (e.g., `{{version}}-legacy`) so images can be compared without collision, or (b) add an explicit dependency so one always runs after the other.

**Location:** `.github/workflows/release.yml` line 195

---

Muriano (approved):

Clean transitional PR that adds a legacy Docker build job as safety net while migrating to GoReleaser. Well-documented with TODO for removal. One confirmed medium issue: both Docker jobs push identical tags to the same registry, creating a race condition where the last writer wins and attestations may reference the wrong digest.

### Suggestions

- Consider adding a step that compares the GoReleaser image manifest against the legacy image manifest (e.g., `docker manifest inspect` + `cosign verify`) to automate the verification rather than relying on manual checks before removing the legacy job.
- The TODO comment references issue #269 — consider also adding a timeline or criteria for removal (e.g., 'remove after 2 successful releases') to prevent the legacy job from lingering indefinitely.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

PR adds a legacy docker-build-push job alongside the new GoReleaser-based Docker build as a temporary safety net during migration per issue #269. Documentation is updated to explain the transition. The approach is reasonable, but the parallel jobs push to identical tags causing race conditions and attestation/tag mismatch.

### Issues

⚠️ **.github/workflows/release.yml:195** — Tag collision: both goreleaser-daemon and docker-build-push push identical semver tags (v1.2.3, 1.2, 1, latest) to the same REGISTRY/IMAGE_NAME in parallel. The last finisher overwrites the other's tags, and the earlier job's attestation then references a digest that no longer matches the published tag. Use a distinct tag suffix for the legacy job (e.g., {{version}}-legacy) or serialize via needs: to avoid races and preserve attestation integrity.
⚠️ **.github/workflows/release.yml:151** — `if: always() && needs.tests.result == 'success'` together with `needs: [release-please, tests]` will run this job even when release-please fails or is skipped, as long as tests passed. On push to main without a release, TAG resolves to an empty string and the build will fail or mis-tag. Consider also gating on release-please success/outputs like the sibling jobs do.
🟡 **.github/workflows/release.yml:184** — The `type=raw,value=latest` tag is conditionally emitted via a templated empty-string fallback. An empty line in the tags list is typically ignored by docker/metadata-action, but this is fragile; prefer a proper `enable=${{ github.event_name != 'workflow_dispatch' }}` condition on the raw tag entry.

### Suggestions

- Suffix legacy tags (e.g., `{{version}}-legacy`, `latest-legacy`) so both images coexist without overwriting each other and attestations remain valid.
- Alternatively, add `needs: [goreleaser-daemon]` on docker-build-push (or vice versa) to serialize the two, accepting the longer pipeline during the verification window.
- Align the `if:` guard with the sibling GoReleaser jobs so the legacy job doesn't run when release-please hasn't produced a tag.
- Add a clear removal checklist (issue/TODO referencing PR + date) so the temporary job isn't forgotten in production.

---

PR #258, opened 2026-04-23
feat(heimdallm): persist in-flight reviews + reload delay + multi-instance regression test + breaker banner

ivanmunozruiz:
Refs #243 (consolidated Task 5 + Task 6 per user instruction to minimise team review cycles).

## Task 5 — Restart / reload hardening

- Persistent `reviews_in_flight` table in SQLite keyed on (pr_id, head_sha); the previous in-memory map wiped on daemon restart / config reload.
- `ClaimInFlightReview` / `ReleaseInFlightReview` / `ClearStaleInFlight(30m)` (startup cleanup).
- `RunTier2` + `Pipeline.Start` gain `coldStart bool` — initial startup fires the first tick immediately for UX; reload waits one `PollInterval` so UI config PATCHes cannot re-trigger the whole fleet.

## Task 6 — Multi-instance regression test + UI banner

- `pipeline_reloop_test.go` adds `TestRun_TwoInstancesSharingStoreDoNotDoubleReview` — two adapters on the same SQLite; Instance B must dedup against Instance A's `PublishedAt`.
- Flutter: subscribe to the `circuit_breaker_tripped` SSE event (event constant landed with Task 2 in PR #246) and render a dismissible MaterialBanner. Cost events no longer slip by silently.
- `CircuitBreakerBanner` widget + widget test cover the show-and-dismiss contract.

## Test plan

- [x] Go: claim/release/clear-stale unit tests, multi-instance regression test, existing Tier 2 tests updated for new signature — all green via `make test-docker`
- [x] Flutter: widget test for banner + existing tests — all green
- [ ] Manual: kill daemon mid-review, restart, verify the review does not re-run (same SHA) and stale claims are cleared after 30 minutes
- [ ] Manual: PATCH /config, verify Tier 2 does not immediately re-poll every repo
- [ ] Manual: trip the circuit breaker (seed 3 reviews on a test PR), verify banner appears in Flutter and dismisses cleanly

## Why consolidated

Combining Task 5 and Task 6 into one PR rather than stacking two separate PRs — one review cycle from the team instead of two. The commits are logically grouped: Task 5's \`38ab91e\` is the daemon-side restart hardening, followed by Task 6's regression test + UI banner commits.

PR #247, opened 2026-04-23
fix(daemon): anchor PR dedup on PublishedAt with 2-minute grace

ivanmunozruiz:
Refs #243 (Fix 3).

Closes the secondary dedup defense by anchoring on the actual post-to-GitHub timestamp. Depends on the PublishedAt column migration.

## Summary

Before: \`PRAlreadyReviewed\` compared \`updated_at\` against \`rev.CreatedAt + 30s\`. But \`CreatedAt\` is stamped BEFORE the Claude call — for any review taking longer than 30s the grace was already exhausted when the review actually posted, causing the daemon to re-review the same commit on the next tick (the 2026-04-22 cost-runaway).

After:
- New \`PublishedAt\` column on \`reviews\`, stamped immediately after \`SubmitReview\` returns success.
- Dedup prefers \`PublishedAt\`; falls back to \`CreatedAt\` for legacy rows.
- Grace widened to 2 minutes — wide enough to absorb GitHub replication + peer-bot submission timing, narrow enough not to suppress legitimate human activity.
- Shared helper \`pipeline.ReviewFreshEnough\` + constant \`pipeline.GraceDefault\` so PR and issue dedup paths cannot drift apart.

## Test plan

- [x] Slow-review scenario (review takes 3 min, dedup still holds)
- [x] Legacy row fallback (PublishedAt zero -> CreatedAt anchor)
- [x] Activity past grace is treated as new change
- [x] \`make test-docker\` green (all 13 packages)
- [ ] Manual: deploy to dev, review a PR, assert exactly one review regardless of subsequent \`updated_at\` bumps

Issue #243, opened 2026-04-23
[P0 INCIDENT] PR review feedback loop costing money — team-wide multi-instance amplification


ivanmunozruiz:
## Summary

On 2026-04-22 the Heimdallm fleet entered a cost-runaway feedback loop on [freepik-company/ai-bumblebee-proxy#1051](https://github.com/freepik-company/ai-bumblebee-proxy/pull/1051). A single team member's daemon posted **15 reviews on the same PR within 57 minutes**. With ~7 team instances running, the total is ~100+ reviews. Estimated cost: **€1,300** in Claude API charges before the team disabled Heimdallm on every machine to stop the bleed.

This is the same class of bug that #222 reported for issues and #226 fixed — the PR code path was never patched. `daemon/internal/issues/fetcher.go:20` already documents the debt explicitly: *"The PR pipeline in main.go currently duplicates this 30 s window inline."*

> **Note to implementer:** the original audit had gaps. This body has been re-audited by multiple reviewers and the fix list was materially revised. Read the whole thing — the fix that looks scariest (pre-review cross-instance check) was **removed** after audit; the fix that looks boring (fail-closed on HEAD SHA) is the one that probably closes the actual leak.

## Evidence

From `~/.local/share/heimdallm/heimdallm.log` on one machine (others expected similar):

```
19:15:15  pipeline: starting review  pr=1051
19:17:44  pipeline: review published to GitHub  severity=low     (review 1, 2m 29s)
19:22:55  pipeline: starting review  pr=1051                     (loop begins)
19:23:32  pipeline: review published to GitHub  severity=medium  (review 2)
...
20:12:40  pipeline: starting review  pr=1051                     (review 15)
```

- Each review: 2–5 minutes on Claude Opus
- Severity oscillates (`low`/`medium`/`high`) as Claude re-examines
- Every 3–5 min a fresh cycle fires

Multi-instance totals: ~7 team daemons × 15 reviews ≈ 100+ reviews. Estimated cost ≈ €1,300.

## Root cause — four layered defenses, all failed

The daemon has **four PR-dedup defenses** that should have stopped this loop. All four failed for different reasons. Fixes must address every layer.

### Defense 1 (PRIMARY) — HEAD SHA guard, `pipeline.go:215`

```go
if prevReview.HeadSHA == pr.Head.SHA {
    return prevReview, nil // skip, no change
}
```

[...]

PR #283, opened 2026-04-23
fix(daemon): plumb HEAD SHA through scheduler so in-flight claim fires (closes #264)

### LINKED ISSUE #264: bug(daemon): in-flight review claim silently bypassed — tier2/tier3 never set Head.SHA when calling runReview
## Summary

The persistent in-flight review lock added in #258 (as part of the #243 fix plan) **never fires on the normal polling path**. Every review that flows through the tier2 poll → `ProcessPR` → `runReview` chain skips the claim silently because the `gh.PullRequest` built by the caller has no `Head.SHA`, and `runReview` guards the claim on `pr.Head.SHA != ""`. Same gap on the tier3 `HandleChange` path.

Net effect: if two consecutive poll ticks (default 1 min apart) both see the same PR in the "to review" list before the first review finishes, **both reviews run concurrently and both publish to GitHub**. Exactly the pattern that #243 was meant to stop.

## Reproduction (observed 2026-04-23 on main)

PR theburrowhub/heimdallm#263 received two back-to-back reviews from a single daemon instance:

```
12:14:14 pipeline: starting review pr=263              ← first review starts
12:14:15 pipeline: using CLI cli=claude
                                                        ← first review still running
12:15:14 pipeline: starting review pr=263              ← second review starts 60 s later
12:15:15 pipeline: using CLI cli=claude
12:15:37 pipeline: review stored locally review_id=247
12:15:38 pipeline: review published to GitHub pr=263 github_review_id=4161561917 state=CHANGES_REQUESTED
12:16:39 pipeline: review stored locally review_id=248
12:16:40 pipeline: review published to GitHub pr=263 github_review_id=4161567552 state=CHANGES_REQUESTED
```

Both reviews landed on GitHub as `CHANGES_REQUESTED` with the same severity. No `runReview: already in flight (persistent), skipping` log line was emitted — the claim was never attempted.

## Root cause

**`daemon/cmd/heimdallm/main.go:301`** — claim is gated on a non-empty Head SHA:

```go
if stored != nil && pr.Head.SHA != \"\" {
    ok, err := s.ClaimInFlightReview(stored.ID, pr.Head.SHA)
    ...
}
```

**`daemon/cmd/heimdallm/main.go:1424-1435`** (tier2 `ProcessPR`) — constructs `gh.PullRequest` **without** `Head.SHA` and calls `runReview`:

```go
ghPR := &gh.PullRequest{
    ID:        pr.ID,
    Number:    pr.Number,
    Repo:      pr.Repo,
    Title:     pr.Title,
    HTMLURL:   pr.HTMLURL,
    User:      gh.User{Login: pr.Author},
    State:     pr.State,
    Draft:     pr.Draft,
    UpdatedAt: pr.UpdatedAt,
}
a.runReview(ghPR, aiCfg)
```

**`daemon/cmd/heimdallm/main.go:1705-1718`** (tier3 `HandleChange`) — same omission.

**`daemon/internal/scheduler/tier2.go:19-29`** — `Tier2PR` has no `HeadSHA` field, so the scheduler contract cannot propagate it today even if `FetchPRsToReview` had it:

```go
type Tier2PR struct {
    ID        int64
    Number    int
    Repo      string
    Title     string
    HTMLURL   string
    Author    string
    State     string
    Draft     bool
    UpdatedAt time.Time
}
```

Result: `pr.Head.SHA == \"\"` on every normal invocation → claim is skipped without a log line → no in-flight protection.

## Why the other defenses didn't save us

The layered fixes from #243 each have a gap against this specific race:

- **PR #245 (fail-closed HEAD SHA in `pipeline.Run`)** — each goroutine does its own SHA lookup independently; neither sees the other's in-flight state, so both pass.
- **PR #246 (SQLite circuit breaker, 3/PR/24h)** — caps worst-case cost but still allows 2–3 duplicate reviews before tripping.
- **PR #247 (PublishedAt + 2-min grace dedup)** — only helps after a review has been *published*, not while one is mid-flight.

The persistent in-flight claim from #258 was the defense meant to stop concurrent runs on the same PR. It's wired into `runReview` correctly; the bug is that callers don't pass the SHA.

## Proposed fixes (in order of preference)

1. **Add `HeadSHA` to `Tier2PR` and plumb it through.** Requires `FetchPRsToReview` to surface the SHA (GitHub already returns it on the PR list), `Tier2PR.HeadSHA` field, and set `ghPR.Head.SHA` in both `ProcessPR` (line 1434) and `HandleChange` (line 1712). Smallest behavioral change, matches the original design intent of `(prID, headSHA)` as the claim key.
2. **Fetch the SHA inside `runReview`** before the claim, if the caller didn't provide one. Costs one GitHub API call per review (acceptable — we already make many); keeps the scheduler contract unchanged.
3. *(Not recommended)* Claim by `prID` alone. Would block legit re-reviews on a new commit pushed during an active review, which is exactly what `inflight.go`'s comment warns against.

## Related

- Closed incident: #243
- Closed bug: #242
- In-flight lock implementation: #258
- Complementary fixes: #245, #246, #247, #255

## Suggested acceptance

- [ ] Regression test that simulates two rapid `ProcessPR` calls with the same `(prID, headSHA)` and asserts only one `runReview` reaches `pipeline.Run`.
- [ ] Manual verification: trigger a review via the normal tier2 poll path and confirm `runReview: already in flight (persistent), skipping` appears when a second tick hits the same PR mid-flight.

ivanmunozruiz:
## Summary

Closes #264. Repairs the wiring gap in #258 that silently bypassed the persistent in-flight review claim, reopening the #243 double-review pattern for every PR flowing through the tier2 poll loop and tier3 watch loop.

Both call sites that reach `runReview` used to construct `gh.PullRequest` without setting `Head.SHA`, which the claim guard gates on (`if stored != nil && pr.Head.SHA != ""`). On the Search-API path (tier2) `HeadSHA` was not even available on `scheduler.Tier2PR`; on the Pulls-API path (tier3) `scheduler.ItemSnapshot` was dropping it even though `GetPRSnapshot` had already fetched it.

Reproduced live on 2026-04-23 against PR #263: two concurrent reviews from a single daemon, 60 s apart, both published `CHANGES_REQUESTED` to GitHub. No `runReview: already in flight (persistent), skipping` log line appeared — the claim was never attempted.

## Changes

- **`scheduler.Tier2PR.HeadSHA`** (new field). Populated in `tier2Adapter.FetchPRsToReview` after the review-guard filter, via `GetPRHeadSHA(repo, number)`. One extra `/pulls/N` call per PR that cleared the gate — bounded by the small number of review-requested PRs per cycle. Fail-open on resolver error (logs a warning, leaves SHA empty).
- **`scheduler.ItemSnapshot.HeadSHA`** (new field). `CheckItem` forwards it from the `github.PRSnapshot` already fetched for `State`/`Draft` — free copy, no extra GitHub API cost.
- **`ProcessPR` / `HandleChange`**: both now set `ghPR.Head = gh.Branch{SHA: ...}` before calling `runReview`.
- **`runReview`**: adds a defensive INFO log when the claim is skipped (either because the PR is not yet upserted or the caller supplied empty SHA), so any future wiring regression — or a transient GH API outage that leaves SHA empty — surfaces in logs instead of silently re-opening the #243 window. Downstream defenses (fail-closed SHA in `pipeline.Run` from #245, circuit breaker from #246, PublishedAt grace from #247) still cap the cost in that fallback path.
- **Regression tests** in `daemon/cmd/heimdallm/main_inflight_plumb_test.go`:
  - `TestTier2Adapter_FetchPRsToReview_PopulatesHeadSHA` — drives `/pulls/N` once per passing PR and asserts `Tier2PR.HeadSHA` is populated.
  - `TestTier2Adapter_FetchPRsToReview_EmptyHeadSHAWhenResolverFails` — locks in the fail-open posture on HTTP 503.
  - `TestTier2Adapter_ProcessPR_PlumbsHeadSHAIntoGhPR` — asserts the Tier2PR → ghPR wiring.
  - `TestTier2Adapter_HandleChange_PlumbsHeadSHAIntoGhPR` — same for the tier3 side.
  - `TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview` — end-to-end property: two goroutines entering `ProcessPR` with the same `(pr_id, head_sha)` result in exactly **one** review body execution. This is the regression the original #258 tests missed.

## Scope notes

- The **manual-trigger review endpoint** (HTTP POST) is intentionally untouched. It already has a nil-SHA branch and an explicit `review already in progress` error return; expanding its SHA plumbing is a separate concern.
- Comment on `inflight.go`'s `(pr_id, head_sha)` keying is preserved — this PR is purely a wiring fix, not a design change.

## Test plan

- [x] `make test-docker` (full suite): 13 packages, all green.
- [x] `go vet ./...` inside the docker image: clean.
- [ ] Manual verification post-merge: trigger a review via normal tier2 poll, confirm `runReview: already in flight (persistent), skipping` fires on the next tick while the first review is mid-flight.
- [ ] Confirm no `runReview: in-flight claim skipped` INFO lines appear under normal operation (they should only appear on first-upsert or GH API outage).

## Related

- Closed incident: #243
- Wiring bug this PR closes: #264
- In-flight claim implementation this PR completes: #258
- Complementary layered defenses: #245, #246, #247, #255


Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-documented fix that correctly plumbs HEAD SHA through both Tier 2 and Tier 3 scheduler paths to enable the persistent in-flight claim. The fail-open design on SHA resolver error is sound given the documented layered defenses. Test coverage is thorough — happy path, error path, both tiers, and concurrent behavior are all exercised.

### Suggestions

- TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview reimplements the claim/release logic inline rather than exercising the production runReview code path. If the claim logic in main.go ever changes (e.g. adding a new precondition), this test won't catch the drift. Consider extracting the claim guard into a shared helper that both production and tests call, or at minimum add a comment noting the duplication risk.
- The time.Sleep(50ms) in the concurrent test is reasonable for local runs but could be fragile on heavily-loaded CI runners. A channel-based synchronization (e.g. one goroutine signals it has acquired the claim before the second goroutine proceeds) would make the test deterministic without relying on timing.
- The defensive log block (main.go ~line 325) could emit at Warn level instead of Info — an empty Head.SHA from the caller is a wiring regression signal, not normal operation, and Warn would make it more visible in production dashboards.

---

sergiotejon:
## 🟡 LOW Issue

GetPRHeadSHA makes one extra /pulls/N API call per PR per fetch cycle. The comment says it's bounded by review-requested PRs, but under rate-limit pressure this could compound. Consider caching the SHA briefly or batching.

**Location:** `daemon/cmd/heimdallm/main.go` line 1408

---

sergiotejon:
## ⚠️ MEDIUM Issue

TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview uses time.Sleep(50ms) for synchronization, making it potentially flaky under load. The two goroutines also start sequentially so one may finish before the other begins — this doesn't reliably test true concurrency.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 340

---

sergiotejon:
## ⚠️ MEDIUM Issue

The concurrent test lacks a sync barrier (e.g. sync.WaitGroup or channel) to ensure both goroutines actually contend simultaneously. One goroutine could complete ProcessPR entirely before the second starts, making the test pass trivially without exercising the claim contention.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 289

---

sergiotejon (approved):

Correct plumbing fix for a real bug. The code changes are minimal and well-placed. Comments are excessively verbose but the logic is sound. One minor concern about the extra API call in the hot path.

### Suggestions

- Add a sync barrier (e.g. a channel or sync.WaitGroup countdown) before the two concurrent ProcessPR goroutines to guarantee they actually race, rather than relying on goroutine scheduling.
- Trim the inline comments to essential information — the issue references are helpful but the multi-paragraph explanations belong in commit messages or PR descriptions, not inline.

---

sergiotejon:
## 🟡 LOW Issue

The defensive log comment block is 10 lines for 5 lines of code. The comments throughout this PR are disproportionately long relative to the changes — consider trimming to essential context only.

**Location:** `daemon/cmd/heimdallm/main.go` line 312

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Solid, narrowly-scoped fix that plumbs HEAD SHA end-to-end so the persistent in-flight claim keyed on (pr_id, head_sha) actually fires from both Tier 2 and Tier 3 call sites. Struct fields, wiring, and tests all match. Main concerns are test-quality (the concurrent regression test doesn't reliably force contention) and comment verbosity.

### Issues

⚠️ **daemon/cmd/heimdallm/main_inflight_plumb_test.go:289** — TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview spawns two goroutines back-to-back without a start barrier. Because the first call acquires the DB claim immediately and only then sleeps 50ms, scheduler latency could let it finish before the second goroutine even enters runReview — the test would pass without ever exercising the claim-contention path it claims to cover. Add a sync.WaitGroup or a chan struct{} that both goroutines block on before calling ProcessPR, or use a fake runReview that blocks on a channel released only after both have entered.
⚠️ **daemon/cmd/heimdallm/main_inflight_plumb_test.go:340** — time.Sleep(50ms) inside the fake runReview is a timing-based synchronization primitive. Under CI load this can be flaky (goroutine starvation) and the comment's 'generous vs 60s tick' justification conflates production cadence with test-goroutine scheduling. Prefer a channel: goroutine 1 signals 'claim acquired', waits on a 'release' channel that the test closes only after confirming goroutine 2 tried and failed to claim.
🔴 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:1** — The test file references a runReview field on tier2Adapter (a.runReview = runReview; `a := &tier2Adapter{... runReview: runReview}`). The diff does not show this field being added to the tier2Adapter struct. If the field doesn't already exist the file won't compile; if it was added in a separate change it should be visible here or called out. Verify the struct change lands together.
🟡 **daemon/cmd/heimdallm/main.go:1408** — GetPRHeadSHA adds one /pulls/N call per PR that clears the review gate. On a repo with many review-requested PRs and secondary-rate-limit pressure, this roughly doubles GitHub traffic per tick. A short (one-tick) in-memory cache keyed on (repo, number, updated_at) would eliminate redundant lookups when the same PR lingers in the queue across ticks, without complicating the fail-open story.
🟡 **daemon/cmd/heimdallm/main.go:312** — Comment blocks are disproportionate to the code. The 10-line justification in the else branch, the ~12-line block in FetchPRsToReview, and the multi-line block in ProcessPR/HandleChange repeat the same #264/#243/#258 lineage. One reference per call site plus a single canonical explanation in scheduler/tier2.go's Tier2PR doc (which already has it) would be enough.
🟡 **daemon/cmd/heimdallm/main.go:315** — The defensive-log reason logic says 'stored PR not found' when stored == nil and 'empty Head.SHA from caller' otherwise. That is correct for the current outer guard (stored != nil && pr.Head.SHA != ""), but it is load-bearing on that exact condition. If the guard ever gains a third clause, the reason string will silently mislead. Consider computing reason from the actual predicates rather than a single else-branch nil check.
🟡 **daemon/cmd/heimdallm/main.go:1415** — Fail-open on SHA resolver error means that during a sustained GitHub /pulls outage every PR skips the claim guard for the duration. The comment correctly notes downstream defenses still apply, but it would be worth emitting a metric/counter (not just a slog.Warn) so an ops dashboard can detect elevated 'SHA lookup failed' rates before they translate into duplicate-review incidents.

### Suggestions

- Replace time.Sleep + bare goroutine spawn in the concurrent test with an explicit start barrier (chan struct{} closed after both goroutines are registered) and a channel-based hold inside the fake runReview; assert that the second goroutine observed ClaimInFlightReview returning false rather than asserting only on the review-body counter.
- Verify and surface the tier2Adapter.runReview field addition in this PR; if it lives in another commit, include it here so the test file compiles from a single diff.
- Trim the per-call-site #264/#243/#258 commentary to one-liners pointing at the canonical doc on scheduler.Tier2PR; the current comment-to-code ratio will rot as context shifts.
- Add a lightweight per-tick cache (or reuse an existing cache layer) for GetPRHeadSHA keyed on (repo, number, updated_at) to soften rate-limit pressure on large queues without weakening the fail-open behavior.
- Emit a counter metric alongside the 'HEAD SHA lookup failed' warn log so sustained resolver failure is observable before it manifests as a duplicate-review incident.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Correct and well-tested fix that plumbs HEAD SHA through Tier2PR and ItemSnapshot so the persistent in-flight claim can actually key on (pr_id, head_sha). The code changes are minimal and the fail-open posture is sensible. Main concerns are excessive inline comments and one extra GitHub API call per eligible PR per poll cycle.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:1399** — Excessive inline comment block (13 lines) for a straightforward API call + error fallback. The issue history and layered-defense rationale belong in the commit message or PR description, not inline — these references to #258, #264, #243 will rot as the codebase evolves. A 1-2 line comment explaining the fail-open choice would suffice.
🟡 **daemon/cmd/heimdallm/main.go:311** — 11-line comment for a 6-line defensive log block. The comment-to-code ratio is extreme. The slog.Info message itself is already descriptive enough; the comment mostly restates the PR description.
🟡 **daemon/cmd/heimdallm/main.go:1468** — 5-line comment on a single struct field assignment (Head: gh.Branch{SHA: pr.HeadSHA}). This pattern repeats at lines 1651 and 1757 — each site has a multi-line comment explaining the same issue-chain. Consider a single doc comment on the HeadSHA field in the struct definition (which already exists) and trusting readers to follow it.
🟡 **daemon/cmd/heimdallm/main.go:1404** — GetPRHeadSHA makes one /pulls/N call per PR that clears the review gate, every poll cycle. If the SHA hasn't changed since last cycle, this call is redundant. Consider caching the last-known SHA per PR ID (e.g., in a map alongside lastSkippedUpdatedAt) and skipping the lookup when UpdatedAt is unchanged. This would reduce GitHub API quota consumption for stable PRs.
🟡 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:312** — TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview uses time.Sleep(50ms) inside the mock runReview to hold the claim during contention. Under heavy CI load this could be flaky — consider using a sync barrier (e.g., two channels) to guarantee both goroutines are inside ProcessPR before either proceeds, rather than relying on timing.
🟡 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:1** — 456-line test file with extensive multi-line comment blocks restating the same issue history (#243, #258, #264). The test names and assertion messages are already descriptive. The top-of-file block comment, per-function doc comments, and inline narrative could be trimmed by ~60% without losing test clarity.

### Suggestions

- Trim inline comments to 1-2 lines each — the issue history and layered-defense rationale are already well-documented in the PR description and commit message. Inline comments should explain 'why this specific code choice' not 'the full bug timeline'.
- Consider caching the resolved HEAD SHA per PR ID between poll cycles (keyed on UpdatedAt) to avoid redundant /pulls/N calls for PRs whose UpdatedAt hasn't changed since last tick.
- Replace time.Sleep-based concurrency contention in TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview with a channel-based barrier to eliminate potential CI flakiness.
- The three identical comment blocks at ProcessPR, CheckItem, and HandleChange all explain the same Head.SHA plumbing rationale — deduplicate by referencing the Tier2PR.HeadSHA / ItemSnapshot.HeadSHA doc comment instead.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Correct and well-tested fix that plumbs HEAD SHA through Tier2PR and ItemSnapshot so the persistent in-flight claim can actually key on (pr_id, head_sha). The code changes are minimal and the fail-open posture is sensible. Main concerns are excessive inline comments and one extra GitHub API call per eligible PR per poll cycle.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:1399** — Excessive inline comment block (13 lines) for a straightforward API call + error fallback. The issue history and layered-defense rationale belong in the commit message or PR description, not inline — these references to #258, #264, #243 will rot as the codebase evolves. A 1-2 line comment explaining the fail-open choice would suffice.
🟡 **daemon/cmd/heimdallm/main.go:311** — 11-line comment for a 6-line defensive log block. The comment-to-code ratio is extreme. The slog.Info message itself is already descriptive enough; the comment mostly restates the PR description.
🟡 **daemon/cmd/heimdallm/main.go:1468** — 5-line comment on a single struct field assignment (Head: gh.Branch{SHA: pr.HeadSHA}). This pattern repeats at lines 1651 and 1757 — each site has a multi-line comment explaining the same issue-chain. Consider a single doc comment on the HeadSHA field in the struct definition (which already exists) and trusting readers to follow it.
🟡 **daemon/cmd/heimdallm/main.go:1404** — GetPRHeadSHA makes one /pulls/N call per PR that clears the review gate, every poll cycle. If the SHA hasn't changed since last cycle, this call is redundant. Consider caching the last-known SHA per PR ID (e.g., in a map alongside lastSkippedUpdatedAt) and skipping the lookup when UpdatedAt is unchanged. This would reduce GitHub API quota consumption for stable PRs.
🟡 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:312** — TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview uses time.Sleep(50ms) inside the mock runReview to hold the claim during contention. Under heavy CI load this could be flaky — consider using a sync barrier (e.g., two channels) to guarantee both goroutines are inside ProcessPR before either proceeds, rather than relying on timing.
🟡 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:1** — 456-line test file with extensive multi-line comment blocks restating the same issue history (#243, #258, #264). The test names and assertion messages are already descriptive. The top-of-file block comment, per-function doc comments, and inline narrative could be trimmed by ~60% without losing test clarity.

---

sergiotejon:
## ⚠️ MEDIUM Issue

Concurrent test has no sync barrier (e.g. channel or sync.WaitGroup gate) to ensure both goroutines actually contend on the claim simultaneously. One goroutine can complete before the other starts, making the test pass without exercising the race it claims to cover.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 289

---

sergiotejon:
## ⚠️ MEDIUM Issue

time.Sleep(50ms) for synchronization is fragile. Under CI load this can flake. Use a channel or condition variable to coordinate instead of wall-clock sleeps.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 340

---

sergiotejon:
## 🟡 LOW Issue

GetPRHeadSHA adds an extra API call per PR per cycle. No caching or conditional fetch (e.g. If-None-Match). Under rate-limit pressure this compounds. Consider caching the SHA for a short TTL.

**Location:** `daemon/cmd/heimdallm/main.go` line 1408

---

sergiotejon:
## 🟡 LOW Issue

10-line comment block for 5 lines of code. The comments throughout this PR are disproportionately verbose — trim to essential context.

**Location:** `daemon/cmd/heimdallm/main.go` line 312

---

sergiotejon (approved):

Correct fix for a real bug (empty HEAD SHA bypassing in-flight claim). The plumbing is straightforward and the fail-open posture is reasonable. Main concerns are the concurrent test's lack of a proper barrier and excessive comment volume.

### Suggestions

- Add a sync barrier (e.g. a closed channel after both goroutines are ready) in the concurrent test so both goroutines actually contend on ClaimInFlightReview simultaneously.
- Replace time.Sleep(50ms) with deterministic synchronization primitives to avoid flaky tests under load.
- Consider caching the HEAD SHA briefly (e.g. per fetch cycle) to avoid redundant /pulls/N calls if the same PR appears in consecutive cycles.

---

Muriano:
## ⚠️ MEDIUM Issue

STILL UNADDRESSED: TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview lacks a sync barrier to guarantee both goroutines contend on ClaimInFlightReview simultaneously. One goroutine can finish ProcessPR before the second starts, making the test pass trivially without exercising the claimed race. Fix: add a channel or sync gate that both goroutines block on before entering the claim path, e.g. a sync.WaitGroup with Add(2) released after both goroutines are ready.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 289

---

Muriano:
## ⚠️ MEDIUM Issue

STILL UNADDRESSED: time.Sleep(50ms) inside the fake runReview is the only mechanism forcing overlap between the two goroutines. Under CI load or slow machines this is flaky — if the first goroutine completes before the second even calls ProcessPR, the claim is already released and both reviews run. Replace with a deterministic coordination primitive (e.g. a channel the first goroutine blocks on until the second has attempted the claim).

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 340

---

Muriano (approved):

Re-review: the diff is unchanged since the last review. The core plumbing fix is correct and well-tested. The two MEDIUM issues about the concurrent test's lack of a sync barrier and reliance on time.Sleep remain unaddressed. The LOW issues (verbose comments, uncached SHA lookup) also remain but do not block merge.

### Suggestions

- Consider trimming the 10-line comment blocks (e.g. lines 312-322 in main.go) to essential context — the issue references already carry the full story.
- GetPRHeadSHA (line 1408) could benefit from a short TTL cache or If-None-Match conditional fetch to reduce API pressure under rate-limit contention, though the current bounded-by-review-gate argument is reasonable for now.

---

ivanmunozruiz:
Follow-up to the AI review comments — responding one by one:

**MEDIUM × 2 (sleep-based concurrent test):** Agreed. Pushed `0b32e6f` to replace the `time.Sleep(50 ms)` with a deterministic channel hand-off: goroutine A signals `holdingClaim` via `sync.Once` immediately after `ClaimInFlightReview` returns ok, and blocks on `release`; the test then launches goroutine B, which is now guaranteed to hit `ClaimInFlightReview` while the row is still held and fail atomically via SQLite `INSERT OR IGNORE`. Removes the scheduler-timing dependency entirely. Verified stable across `-count=10` runs.

**LOW (extra `/pulls/N` API call):** Net zero in aggregate. `pipeline.Run` already did this lookup inside the #245 fail-closed guard (`pipeline.go:241`, `if pr.Head.SHA == ""`). After this PR, `FetchPRsToReview` shifts the same call earlier so the in-flight claim can see the SHA; `pipeline.Run` then skips its own fetch because the field is populated. Same API call count, just earlier in the flow. No caching/batching needed.

**LOW (comment verbosity):** Leaving as-is. #264 reopened the pattern of a €1,300 production incident (#243). Each block documents which invariant is load-bearing, which defense takes over if it fails, and how the regression arose — context a future contributor would need to avoid re-introducing the same wiring bug. Trimming would save ~20 lines at the cost of that context; not the trade I'd make on this fix.

ivanmunozruiz:
@vbuenog's review — point by point against the pushed changes:

**⚠️ MEDIUM line 289 & 340 (no start barrier, `time.Sleep` in fake runReview):** Fixed in `0b32e6f`. The test now uses a deterministic two-phase channel hand-off — goroutine A signals `holdingClaim` via `sync.Once` as soon as `ClaimInFlightReview` returns ok, then blocks on a `release` channel; goroutine B is launched only after that signal, so it's guaranteed to hit `ClaimInFlightReview` while the row is still held and fail atomically (SQLite `INSERT OR IGNORE`). `time.Sleep` is gone from the fake `runReview`. Verified stable across `-count=10`.

**🔴 HIGH line 1 (`tier2Adapter.runReview` field missing from diff):** False positive — the field is pre-existing at `main.go:1176` (declared long before this PR, used in the earlier `main_tier3_guards_test.go`). The diff doesn't include it because this PR doesn't touch the struct definition. No code change needed; test compiles cleanly (see the CI on this push).

**🟡 LOW line 315 (reason logic coupled to outer guard):** Valid point, fixed in `1251d7e`. Swapped the `if stored != nil` branch for an explicit `switch` on both predicates so a future edit to the outer guard can't silently produce misleading log output.

**🟡 LOW line 1408 (extra `/pulls/N` call, consider cache):** Net-zero in aggregate — `pipeline.Run` was already doing this lookup in the #245 fail-closed guard (`pipeline.go:241`, `if pr.Head.SHA == ""`). This PR shifts it earlier so the in-flight claim can see the SHA; `pipeline.Run` then skips its own fetch. Same API call count. A per-tick cache would help if we ever move back to resolving in both places, but today it'd just add state for no savings.

**🟡 LOW line 312 (comment verbosity):** Staying as-is. The context documents a recurring-regression fix (#264 reopened the pattern of the €1,300 #243 incident). Each block explains which invariant is load-bearing and what fails-over if the primary defense breaks. The maintenance cost of ~20 extra comment lines is much lower than the cost of someone refactoring this path without that context and re-introducing the same wiring gap.

**🟡 LOW line 1415 (metric counter on SHA lookup failure):** Good operational suggestion but requires a metrics pipeline we don't have wired yet. Filed as a follow-up: a dedicated issue for emitting counters on all claim-adjacent fail-open paths (not just this one) would land as its own PR.

CI should go back to green on this push; re-review welcome.

sergiotejon:
## 🟡 LOW Issue

528-line test file for a ~30-line production fix is heavy. Consider extracting shared test helpers (adapter construction, store seeding) to reduce duplication across the 5 test functions.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 1

---

sergiotejon:
## 🟡 LOW Issue

GetPRHeadSHA error is logged at Warn but the empty SHA silently disables the claim guard for that tick. Consider logging at Error or emitting a metric so operators can distinguish 'claim skipped because SHA lookup failed' from normal operation.

**Location:** `daemon/cmd/heimdallm/main.go` line 1420

---

sergiotejon (approved):

Solid bug fix that plumbs HEAD SHA through the scheduler to enable the in-flight claim guard. The concurrent test synchronization was improved per review feedback. Comments are verbose but justified given the production incident history.

### Suggestions

- Extract repeated tier2Adapter test construction into a helper to reduce boilerplate across the 5 test functions.
- Add a brief inline comment on the Tier2PR.HeadSHA field noting it may be empty on resolver failure, rather than relying solely on the doc block above the struct.

---

sergiotejon:
## 🟡 LOW Issue

The else branch's defensive log block (14 lines of comments + 12 lines of code) is disproportionate for what is essentially a debug log. The switch computing 'reason' is fine but the essay-length comments could be trimmed to 2-3 lines referencing the issue number.

**Location:** `daemon/cmd/heimdallm/main.go` line 314

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review. The two MEDIUM concurrency-test findings are genuinely fixed: the sleep-based coordination is replaced with a deterministic channel hand-off (holdingClaim via sync.Once + release) that guarantees goroutine B races an already-held claim row. Author's rebuttals on comment verbosity and /pulls/N caching are reasonable (pipeline.Run already made the call under #245, so it's net-zero). A couple of earlier LOWs remain unaddressed but are non-blocking; one small new nit in the revised test.

### Issues

🟡 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:439** — In the 'setup broken' timeout branch, close(release) is called before wg.Wait(). If goroutine A never reached the claim path (the very scenario this branch catches), close(release) is a no-op for A — but goroutine B is not yet started at this point, so wg.Wait() will deadlock forever because wg.Add(2) expects two Done()s. Either start B before the holdingClaim wait, or wg.Add(1) before each go-statement, or drop wg.Wait() from this fatal branch and rely on t.Fatalf to abort.
🟡 **daemon/cmd/heimdallm/main.go:315** — Unchanged from prior review: the reason-string derivation ('stored PR not found' vs 'empty Head.SHA from caller') is load-bearing on the exact outer guard `stored != nil && pr.Head.SHA != ""`. If a third clause is ever added to the guard, this else branch will silently mislabel. Compute reason from the actual predicates rather than a lone stored==nil check.
🟡 **daemon/cmd/heimdallm/main.go:1415** — Unchanged from prior review: a slog.Warn on SHA resolver failure is invisible to ops dashboards. During a sustained /pulls outage every PR fails open and skips the claim guard; a counter/metric would let elevated 'SHA lookup failed' rates be detected before they translate into duplicate-review incidents.

### Suggestions

- Tiny test hardening: in the 2s timeout fatal branch (line ~439) call t.Fatalf before wg.Wait(), or start goroutine B before waiting on holdingClaim, to avoid a potential wg.Wait() deadlock if A never enters runReview.
- Consider a metric (counter) on the 'tier2: HEAD SHA lookup failed' warn path so a sustained /pulls outage is observable without scraping logs.
- Derive the defensive-log reason in main.go:315 from the individual predicates rather than a bare else-with-nil-check, so future guard changes cannot silently mislabel.

---

sergiotejon:
## 🟡 LOW Issue

14 lines of comments for a 12-line defensive log block is excessive. 2-3 lines referencing #264 plus the switch would suffice.

**Location:** `daemon/cmd/heimdallm/main.go` line 314

---

sergiotejon:
## 🟡 LOW Issue

528-line test file with significant boilerplate duplication across 5 tests (adapter construction, broker setup, login/cfg vars). Extract a helper like newTestTier2Adapter(t, opts) to cut ~150 lines.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 1

---

sergiotejon:
## 🟡 LOW Issue

SHA lookup failure logged at Warn but silently disables the claim for that tick. Error level or a future metric would better surface this operationally.

**Location:** `daemon/cmd/heimdallm/main.go` line 1420

---

sergiotejon:
## ⚠️ MEDIUM Issue

The concurrent test's runReview reimplements production claim logic (GetPRByGithubID, ClaimInFlightReview, ReleaseInFlightReview). If production claim logic changes, this test silently drifts. Consider calling the real runReview or extracting the claim logic into a shared function.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 289

---

sergiotejon (approved):

Solid bug fix that plumbs HeadSHA through the scheduler to enable the in-flight claim guard. The core production changes are correct and well-defended. Test quality improved significantly after replacing time.Sleep with channel-based synchronization. Main concerns are comment verbosity and test helper duplication.

### Suggestions

- Extract a newTestTier2Adapter(t, ...Option) helper to deduplicate broker/store/login/cfg setup across all 5 tests.
- Consider extracting the claim/release logic from runReview into a named method so the concurrent test can call it directly rather than reimplementing it, reducing drift risk.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Correct and well-tested fix for a real bug (#264): HEAD SHA was missing from both tier2 and tier3 code paths, silently disabling the in-flight claim guard and allowing double reviews. The plumbing is sound and the concurrent test now uses deterministic channel-based synchronization (addressing prior reviewer feedback about time.Sleep). Main concerns are excessive comment verbosity and the uncached extra GitHub API call.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:314** — Excessive comment-to-code ratio: 10-line comment block for 5 lines of defensive logging. The historical context (issue numbers, prior bugs, downstream defenses) belongs in the commit message or PR description, not inline — it will rot as the code evolves. Trim to 1-2 lines explaining the non-obvious 'why'.
🟡 **daemon/cmd/heimdallm/main.go:1402** — Comment blocks at every plumbing site (lines 1402-1413, 1468-1472, 1651-1654, 1757-1762) repeat the same narrative about #243/#258/#264 and downstream defenses. One canonical comment at the struct definition (already present in tier2.go) is sufficient; the call sites should be self-explanatory from the field name HeadSHA.
🟡 **daemon/cmd/heimdallm/main.go:1408** — GetPRHeadSHA issues one uncached /pulls/N call per PR per fetch cycle. While bounded by review-gated PRs, under rate-limit pressure (or a burst of review-requested PRs) this compounds. Consider caching the SHA briefly (keyed on pr_id, TTL matching the poll interval) or using a conditional request (If-None-Match) to reduce API consumption.
⚠️ **daemon/cmd/heimdallm/main_inflight_plumb_test.go:289** — The concurrent test's runReview reimplements production claim logic (ClaimInFlightReview/ReleaseInFlightReview) rather than exercising the real runReview code path. If the production claim logic moves, changes signature, or adds conditions, this test will still pass — giving false confidence. Consider testing through the actual code path or at minimum adding a comment noting this coupling.
🟡 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:1** — 528-line test file with extensive narrative comments (the test file has more comment lines than code lines). The prose duplicates the PR description and will drift from reality. Test names and assertion messages are already descriptive — trim the block comments to essential contract notes only.

### Suggestions

- Trim inline comments to 1-2 lines of non-obvious context per site; move the historical narrative (#243/#258/#264 chain, downstream defense catalog) into the PR description or a single canonical doc comment on the HeadSHA field.
- Consider caching GetPRHeadSHA results for one poll cycle (keyed on pr_id) to avoid redundant /pulls/N calls when the same PR appears in consecutive FetchPRsToReview cycles — especially relevant under GitHub rate-limit pressure.
- For the concurrent regression test, consider injecting a claim-checking hook or wrapping the real runReview closure rather than reimplementing the claim logic from scratch, so the test stays coupled to the actual production behavior.
- The fail-open posture on SHA resolution failure is a sound design choice given the layered defenses — consider adding a metric/counter (not just a log line) for SHA resolution failures so they can be monitored in aggregate without log scraping.

---

sergiotejon:
## ⚠️ MEDIUM Issue

The concurrent test's runReview reimplements production claim logic (GetPRByGithubID → ClaimInFlightReview → ReleaseInFlightReview). If the real runReview's claim sequence changes, this test silently drifts and may pass while the actual behavior is broken. Extract the claim guard into a shared helper or call the real runReview with a stubbed pipeline.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 289

---

sergiotejon:
## 🟡 LOW Issue

~150 lines of duplicated adapter construction boilerplate (broker, loginMu, cfgMu, etc.) across 5 tests. A newTestTier2Adapter(t, ...opts) helper would improve maintainability.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 1

---

sergiotejon:
## 🟡 LOW Issue

SHA lookup failure is logged at Warn but silently disables the in-flight claim for that tick. Consider logging at Error level to make this more visible operationally, since it degrades a key safety mechanism.

**Location:** `daemon/cmd/heimdallm/main.go` line 1420

---

sergiotejon:
## 🟡 LOW Issue

14 lines of comments for a 12-line defensive log block. The context is valuable but could be condensed to 3-4 lines referencing #264 without losing the essential information.

**Location:** `daemon/cmd/heimdallm/main.go` line 314

---

sergiotejon (approved):

Correct fix for a real plumbing bug (#264). The production changes are small and well-reasoned. Test file is thorough but heavy with duplicated boilerplate. The concurrent test reimplements production claim logic, creating drift risk. Comment verbosity is a style preference the author has defended.

### Suggestions

- Extract a newTestTier2Adapter(t, ...Option) helper to deduplicate broker/login/cfg setup across the 5 test functions.
- Extract the claim guard (stored lookup → ClaimInFlightReview → defer Release) into a named function in production code so the concurrent test can call it directly instead of reimplementing it.
- Promote the SHA-lookup-failure log from Warn to Error, or add a TODO for a metric counter, so operators can alert on claim degradation.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of the PR fixing #264 plumbing of HEAD SHA through scheduler. The concurrent test's sleep-based synchronization has been replaced with a deterministic channel hand-off (addresses prior MEDIUM). The reason-string switch in the defensive log now checks both predicates explicitly (addresses prior LOW on main.go:315). Remaining issues are minor.

### Issues

🟡 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:439** — In the 'setup broken' timeout branch, close(release) + wg.Wait() can deadlock: if goroutine A timed out without ever reaching the claim path, close(release) is a no-op for A (A isn't blocked on release), but wg.Add(2) requires two Done()s and goroutine B hasn't been launched yet, so wg.Wait() blocks forever. Safer to drop wg.Wait() from this fatal branch or restructure to wg.Add(1) per goroutine at launch time.
🟡 **daemon/cmd/heimdallm/main.go:1420** — SHA resolver failure logged at Warn but silently disables the claim guard for that tick. During a sustained /pulls outage every PR fails open; without a counter/metric, elevated 'SHA lookup failed' rates are invisible on ops dashboards until a duplicate-review incident surfaces. Consider Error level or filing a follow-up to emit a metric.
⚠️ **daemon/cmd/heimdallm/main_inflight_plumb_test.go:289** — The concurrent test's fake runReview reimplements production claim logic (GetPRByGithubID → ClaimInFlightReview → defer Release). If the real runReview's claim sequence evolves (e.g. added grace-period check, different key shape), this test will silently drift and continue passing while real behavior regresses. Consider extracting the claim guard into a shared helper the real runReview and the test both call.

### Suggestions

- Extract the claim sequence (GetPRByGithubID + ClaimInFlightReview + defer Release) into a reusable helper so the concurrent test exercises the same code path as production.
- File a follow-up issue to emit a counter/metric on SHA-lookup failure so fail-open behavior is observable.
- Consider a small newTestTier2Adapter(t, opts) helper to reduce the ~150 lines of duplicated adapter/broker/loginMu/cfgMu boilerplate across the 5 tests.

---

sergiotejon:
## ⚠️ MEDIUM Issue

The concurrent test's runReview reimplements the production claim sequence (GetPRByGithubID → ClaimInFlightReview → ReleaseInFlightReview). If the real runReview's claim logic changes, this test silently drifts. Extract the claim guard into a shared helper or test against the real runReview with a stubbed pipeline.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 289

---

sergiotejon:
## 🟡 LOW Issue

~150 lines of duplicated adapter construction boilerplate (broker, loginMu, cfgMu, store, etc.) across 5 tests. Extract a newTestTier2Adapter(t, ...opts) helper.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 1

---

sergiotejon:
## 🟡 LOW Issue

14 lines of comments for a 12-line defensive log block is disproportionate. Could be condensed to 3-4 lines referencing #264.

**Location:** `daemon/cmd/heimdallm/main.go` line 314

---

sergiotejon (approved):

Solid bug fix that plumbs HEAD SHA through the scheduler to enable the in-flight claim guard. The core production changes are correct and well-defended. The concurrent test was improved with channel-based synchronization. Main concerns are test boilerplate duplication and the reimplemented claim logic in tests drifting from production.

### Suggestions

- Extract a newTestTier2Adapter(t, opts) helper to eliminate ~150 lines of duplicated boilerplate across the 5 test functions
- Extract the claim guard logic (GetPRByGithubID → ClaimInFlightReview → defer Release) into a shared function used by both production runReview and the concurrent test, so they cannot drift
- Promote the SHA-lookup-failure log from Warn to Error, since it silently degrades the in-flight claim safety mechanism for that tick

---

sergiotejon:
## 🟡 LOW Issue

SHA lookup failure logged at Warn but silently disables the claim guard for that tick. Consider Error level since it degrades a key safety mechanism.

**Location:** `daemon/cmd/heimdallm/main.go` line 1420

---

sergiotejon:
## ⚠️ MEDIUM Issue

The concurrent test's runReview reimplements the production claim sequence (GetPRByGithubID → ClaimInFlightReview → ReleaseInFlightReview). If production claim logic changes, this test silently drifts. Extract the claim guard into a shared helper or call the real runReview with a stubbed pipeline.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 289

---

sergiotejon:
## 🟡 LOW Issue

~150 lines of duplicated adapter construction boilerplate across 5 tests. Extract a newTestTier2Adapter helper to reduce duplication and improve maintainability.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 1

---

sergiotejon:
## 🟡 LOW Issue

14-line comment block for a 12-line defensive log is disproportionate. The issue references are valuable but the prose could be condensed to 3-4 lines.

**Location:** `daemon/cmd/heimdallm/main.go` line 314

---

sergiotejon:
## 🟡 LOW Issue

SHA lookup failure logged at Warn but silently disables the claim guard for that tick. Since this degrades a key safety mechanism, Error level would better surface the condition operationally.

**Location:** `daemon/cmd/heimdallm/main.go` line 1420

---

sergiotejon (approved):

Solid bug fix plumbing HeadSHA through scheduler to enable the in-flight claim guard. The core production changes are correct and well-defended. The concurrent test was improved with channel-based synchronization. Main concerns are test maintainability and comment verbosity.

### Suggestions

- Extract a newTestTier2Adapter(t, ...opts) helper to eliminate ~150 lines of duplicated boilerplate across the 5 test functions
- Extract the claim guard logic (GetPRByGithubID → ClaimInFlightReview → defer Release) into a shared function used by both production runReview and the concurrent test, so the test cannot silently drift from production behavior
- Promote the SHA lookup failure log from Warn to Error, or add a metric counter, since it silently degrades the in-flight claim for that tick

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid, well-scoped bug fix that correctly plumbs HeadSHA through the scheduler structs so the persistent in-flight claim can actually fire. The production changes are minimal and correct. The fail-open posture on transient SHA-lookup errors is well-reasoned with appropriate defense-in-depth. Test coverage is thorough, though the concurrent test carries a drift risk by reimplementing claim logic. Comment verbosity is high but justified by the author given the incident history.

### Issues

⚠️ **daemon/cmd/heimdallm/main_inflight_plumb_test.go:289** — The concurrent test's runReview reimplements the production claim sequence (GetPRByGithubID → ClaimInFlightReview → defer ReleaseInFlightReview). If the real runReview in main.go ever adds, reorders, or changes claim logic (e.g. switching from INSERT OR IGNORE to a SELECT-then-INSERT with a different isolation level), this test will keep passing while the actual behavior diverges. Consider either (a) extracting the claim guard into a named helper that both production and test code call, or (b) adding a brief comment in main.go's runReview pointing to this test so future editors know to update both.
🟡 **daemon/cmd/heimdallm/main.go:1420** — SHA lookup failure is logged at slog.Warn but silently disables the in-flight claim for that tick — the primary deduplication mechanism. Since this degrades a key safety mechanism (the one this PR exists to fix), slog.Error would better match the operational severity and make it easier to alert on.
🟡 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:1** — ~150 lines of duplicated adapter construction boilerplate (broker start/stop, loginMu, cfgMu, login, cfg vars) across 5 tests. A newTestTier2Adapter(t, ...Option) helper would reduce duplication and make it harder to forget a required field when adding future tests.
🟡 **daemon/cmd/heimdallm/main.go:314** — The else-branch defensive log block has 14 lines of comments for 12 lines of code. The switch-on-predicates approach is good engineering (the reason string stays truthful if the outer guard changes), but the multi-paragraph comment explaining why could be condensed to 2-3 lines referencing #264 without losing the essential context.

### Suggestions

- Extract the claim guard (GetPRByGithubID → ClaimInFlightReview → defer Release) into a small named helper (e.g. tryClaimReview) so the concurrent test can call the real logic instead of reimplementing it — eliminates the drift risk entirely.
- Promote the SHA-lookup-failure log from slog.Warn to slog.Error to match its operational impact: it disables the primary deduplication mechanism for that tick.
- Extract a newTestTier2Adapter(t, ...opts) helper in a test_helpers_test.go file to cut ~150 lines of duplicated broker/login/cfg setup across the 5 tests.
- Consider adding a short comment in main.go's runReview claim guard pointing to main_inflight_plumb_test.go's concurrent test, so future editors of the claim logic know to update both sites.

---

sergiotejon:
## ⚠️ MEDIUM Issue

The concurrent test's runReview reimplements the production claim sequence (GetPRByGithubID → ClaimInFlightReview → ReleaseInFlightReview). If production runReview's claim logic changes, this test silently drifts and passes while actual behavior is broken. Extract the claim guard into a shared helper or test against the real runReview with a stubbed pipeline.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 289

---

sergiotejon:
## 🟡 LOW Issue

~150 lines of duplicated adapter construction boilerplate (broker, loginMu, cfgMu, store setup) across 5 tests. A newTestTier2Adapter(t, ...opts) helper would cut significant duplication.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 1

---

sergiotejon:
## 🟡 LOW Issue

14 lines of comments for 12 lines of defensive logging code is disproportionate. 3-4 lines referencing #264 plus the switch would preserve the essential context.

**Location:** `daemon/cmd/heimdallm/main.go` line 314

---

sergiotejon:
## 🟡 LOW Issue

SHA lookup failure logged at Warn but silently disables the in-flight claim for that tick. Since this degrades a key safety mechanism, Error level would better surface it operationally.

**Location:** `daemon/cmd/heimdallm/main.go` line 1420

---

sergiotejon (approved):

Correct fix for a real production bug (empty HEAD SHA bypassing in-flight claim). The plumbing through Tier2PR and ItemSnapshot is sound. Main concerns are test maintainability and comment verbosity, plus one medium issue around test-production drift.

### Suggestions

- Extract a newTestTier2Adapter(t, ...opts) helper to deduplicate broker/login/cfg/store setup across all 5 tests
- Extract the claim guard logic (GetPRByGithubID → ClaimInFlightReview → defer Release) into a named function shared between production runReview and the concurrent test, so they can't drift apart
- Condense the defensive-log comment block to 3-4 lines referencing #264 and the layered defenses

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Correct and well-reasoned fix for a real production bug (#264): HEAD SHA was missing from the scheduler plumbing, silently disabling the in-flight claim mechanism. The core wiring through Tier2PR.HeadSHA and ItemSnapshot.HeadSHA is sound, the fail-open posture on SHA-lookup failure is defensible, and the deterministic channel hand-off in the concurrent test is a solid improvement over the earlier sleep-based approach. Main concerns are the test's reimplementation of production claim logic (drift risk) and general verbosity.

### Issues

⚠️ **daemon/cmd/heimdallm/main_inflight_plumb_test.go:289** — The concurrent test's runReview reimplements the full production claim sequence (GetPRByGithubID → ClaimInFlightReview → ReleaseInFlightReview). If the real runReview's claim logic changes — e.g. adding a retry, changing the key, or reordering guards — this test silently drifts and can pass while the actual behavior is broken. Extract the claim guard into a shared helper callable by both production and test code, or test against the real runReview with a stubbed pipeline.
🟡 **daemon/cmd/heimdallm/main.go:1420** — SHA lookup failure is logged at Warn but silently disables the in-flight claim for the entire tick (headSHA = ""). Since this degrades a key safety mechanism that prevents the €1,300 double-review pattern, consider logging at Error level so operators can distinguish 'claim skipped due to SHA failure' from normal skip paths in alerting rules.
🟡 **daemon/cmd/heimdallm/main_inflight_plumb_test.go:1** — ~150 lines of duplicated adapter construction boilerplate (broker start/stop, loginMu, cfgMu, login, cfg declarations) across 5 test functions. A newTestTier2Adapter(t, ...opts) helper would reduce duplication and make future test additions cheaper.
🟡 **daemon/cmd/heimdallm/main.go:314** — The defensive else-branch has a 14-line comment block for 12 lines of code. The issue references and invariant documentation are valuable, but the prose restates the same points (wiring regression, layered defenses, truthful log) multiple times. 3–4 lines referencing #264 plus the reason switch would preserve the essential context at half the length.

### Suggestions

- Extract the claim-guard sequence (GetPRByGithubID → ClaimInFlightReview → defer ReleaseInFlightReview) into a shared helper so the concurrent test calls the same code as production, eliminating drift risk.
- Extract a newTestTier2Adapter(t, ...Option) helper to DRY up the ~150 lines of repeated broker/login/cfg/store setup across the 5 test functions.
- Promote the SHA-lookup-failure log from Warn to Error (or add a metric counter) so operators can alert on sustained failures that silently disable the claim guard — distinguishing it from the normal 'claim skipped' path.
- Trim the defensive-log comment block to 3–4 lines: reference #264 and the fail-over chain, drop the repeated rationale about truthful logs and future edits.

---

sergiotejon:
## ⚠️ MEDIUM Issue

The concurrent test's runReview reimplements the production claim sequence (GetPRByGithubID → ClaimInFlightReview → ReleaseInFlightReview). If production claim logic changes, this test silently drifts. Extract the claim guard into a shared helper or test against the real runReview with a stubbed pipeline.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 289

---

sergiotejon:
## 🟡 LOW Issue

~150 lines of duplicated adapter construction boilerplate (broker, loginMu, cfgMu, store) across 5 tests. Extract a newTestTier2Adapter(t, ...opts) helper.

**Location:** `daemon/cmd/heimdallm/main_inflight_plumb_test.go` line 1

---

sergiotejon:
## 🟡 LOW Issue

14-line comment block for a 12-line defensive log is disproportionate. 3-4 lines referencing #264 plus the switch would suffice.

**Location:** `daemon/cmd/heimdallm/main.go` line 314

---

sergiotejon:
## 🟡 LOW Issue

SHA lookup failure logged at Warn but silently disables the claim guard for that tick. Consider Error level since it degrades a key safety mechanism.

**Location:** `daemon/cmd/heimdallm/main.go` line 1420

---

sergiotejon (approved):

Solid bug fix plumbing HeadSHA through scheduler to enable the in-flight claim guard. The core fix is correct and well-tested. Main concerns are test maintainability and comment verbosity.

### Suggestions

- Extract duplicated tier2Adapter test setup into a newTestTier2Adapter(t, opts) helper to cut boilerplate and improve maintainability
- Extract the claim guard logic (GetPRByGithubID → ClaimInFlightReview → defer Release) into a shared function so the concurrent test calls the same code as production rather than reimplementing it
- Promote the SHA-lookup-failure log from Warn to Error, or add a metric counter, so operators can distinguish degraded claim behavior from normal operation

---

PR #290, opened 2026-04-23
feat: implement #288 — bug: GoReleaser nfpms fails — Flutter GUI artifacts missing at build time

### LINKED ISSUE #288: bug: GoReleaser nfpms fails — Flutter GUI artifacts missing at build time
## Problem

The `Build Linux packages` job fails at `GoReleaser — build .deb + .rpm` because nfpms expects Flutter GUI artifacts (binary, icons, desktop entry) that do not exist when GoReleaser runs.

GoReleaser builds the Go daemon but does NOT build the Flutter app. The nfpms `contents:` entries reference files from `flutter_app/build/` which is empty.

## Fix

Add a Flutter build step BEFORE GoReleaser in the Linux packages job:

```yaml
- name: Build Flutter app
  working-directory: flutter_app
  run: flutter build linux --release

- name: Build daemon
  run: make build-daemon

```

Alternatively, the nfpms config should reference pre-built artifacts from a previous step, not assume GoReleaser builds them.

## Files

- `.github/workflows/release.yml` — add Flutter build step before GoReleaser in Linux job
- `.goreleaser.yml` or `cli/.goreleaser.yml` — verify nfpms contents paths match actual build output

Ready for auto_implement.

sergiotejon:
Auto-generated by Heimdallm in response to #288.

Closes #288

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -162,7 +162,7 @@ jobs:
         with:
           distribution: goreleaser
           version: '~> v2'
-          args: release --clean --skip=nfpm,archive
+          args: release --clean --config .goreleaser-docker.yml --skip=archive
           workdir: daemon
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -372,7 +372,7 @@ jobs:
         with:
           distribution: goreleaser
           version: '~> v2'
-          args: release --clean --skip=docker,archive
+          args: release --clean --skip=archive
           workdir: daemon
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/daemon/.goreleaser-docker.yml
+++ b/daemon/.goreleaser-docker.yml
@@ -1,0 +1,49 @@
+version: 2
+
+project_name: heimdallm
+
+builds:
+  - id: daemon
+    main: ./cmd/heimdallm
+    binary: heimdalld
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+    goarch:
+      - amd64
+    ldflags:
+      - -s -w
+
+# ── Docker image — pushed to GHCR ────────────────────────────────────────────
+# Used by: goreleaser-daemon job (release.yml)
+dockers:
+  - ids:
+      - daemon
+    image_templates:
+      - "ghcr.io/theburrowhub/heimdallm:{{ .Version }}"
+      - "ghcr.io/theburrowhub/heimdallm:{{ .Major }}.{{ .Minor }}"
+      - "ghcr.io/theburrowhub/heimdallm:{{ .Major }}"
+      - "ghcr.io/theburrowhub/heimdallm:latest"
+      - "ghcr.io/theburrowhub/heimdallm:{{ .ShortCommit }}"
+    dockerfile: ../docker/Dockerfile.goreleaser
+    use: buildx
+    build_flag_templates:
+      - "--pull"
+      - "--cache-from=type=gha"
+      - "--cache-to=type=gha,mode=max"
+      - "--label=org.opencontainers.image.created={{.Date}}"
+      - "--label=org.opencontainers.image.title={{.ProjectName}}"
+      - "--label=org.opencontainers.image.revision={{.FullCommit}}"
+      - "--label=org.opencontainers.image.version={{.Version}}"
+      - "--label=org.opencontainers.image.source=https://github.com/theburrowhub/heimdallm"
+
+changelog:
+  disable: true
+
+release:
+  github:
+    owner: theburrowhub
+    name: heimdallm
+  mode: keep-existing
+  extra_files: []
--- a/daemon/.goreleaser.yml
+++ b/daemon/.goreleaser.yml
@@ -15,31 +15,8 @@ builds:
     ldflags:
       - -s -w
 
-# ── Docker image — pushed to GHCR ────────────────────────────────────────────
-# Used by: goreleaser-daemon job (release.yml --skip=nfpm,archive)
-dockers:
-  - ids:
-      - daemon
-    image_templates:
-      - "ghcr.io/theburrowhub/heimdallm:{{ .Version }}"
-      - "ghcr.io/theburrowhub/heimdallm:{{ .Major }}.{{ .Minor }}"
-      - "ghcr.io/theburrowhub/heimdallm:{{ .Major }}"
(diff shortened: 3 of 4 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Clean architectural fix that splits the GoReleaser config into two files — one for Docker, one for nfpms — so each CI job only sees the sections it needs. This eliminates the root cause (nfpms requiring Flutter staging artifacts during the Docker-only job). The release.mode: keep-existing in the new docker config correctly prevents overwriting the GitHub release, and changelog.disable: true avoids duplicate entries. No issues found.

### Suggestions

- The build section in .goreleaser-docker.yml duplicates the one in .goreleaser.yml (binary name, ldflags, CGO_ENABLED). If these drift apart in the future it could cause subtle inconsistencies. Consider adding a comment in both files pointing to the other, or documenting in release-pipeline.md that build settings must be kept in sync.
- The docker config hardcodes linux/amd64 only, which is correct for the current GHCR image but worth a comment noting this is intentional (no multi-arch Docker image) so a future contributor doesn't accidentally add arm64 without also updating the Dockerfile.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Reasonable fix that splits docker and nfpm goreleaser runs into separate config files so the nfpm job can require the Flutter staging/ directory without the docker job needing it. Logic is sound, but watch for release-mode coordination between the two goreleaser invocations and the duplicate daemon build.

### Issues

⚠️ **daemon/.goreleaser-docker.yml:45** — The docker-only config uses `mode: keep-existing`, which is correct for appending to a release created by the other job. However, correctness depends on `daemon/.goreleaser.yml` (nfpm run) also being compatible — i.e. either also using `keep-existing` or being the job that first creates the release. The diff doesn't show the release stanza of the nfpm config, so there's a risk of the second job overwriting or failing to find the release. Verify both configs use a consistent release mode (ideally both `keep-existing`) and that the docker job (which runs first per release.yml ordering) is allowed to create the release if absent.
🟡 **daemon/.goreleaser-docker.yml:5** — The `builds:` block duplicates the daemon build definition from `daemon/.goreleaser.yml`. This compiles the daemon binary twice across the two jobs (once for docker, once for nfpm), wasting CI time. Not a bug, but worth noting — the daemon binary is identical between the two configs (same ldflags, same target). Acceptable tradeoff for config isolation, but flag drift risk: future changes to build flags must be mirrored in both files.
🟡 **daemon/.goreleaser-docker.yml:1** — No comment/header explains why this config exists separately from `.goreleaser.yml`. Given the original bug (nfpms failing because Flutter artifacts were missing at the docker job), a 2-line comment at the top referencing issue #288 and stating 'docker-only config so nfpm's staging/ dependency doesn't block docker builds' would save the next maintainer a git blame dive.
🟡 **.github/workflows/release.yml:372** — The nfpm job now runs without `--skip=docker`. Since `daemon/.goreleaser.yml` had its `dockers:` section removed, there's nothing to skip, so the flag removal is correct. But confirm no other `dockers:` entries remain in that file (diff only shows removal of one block) — a lingering docker stanza there would cause the nfpm job to try pushing images a second time.

### Suggestions

- Add a comment at the top of `daemon/.goreleaser-docker.yml` linking to issue #288 and explaining the split, so the rationale survives.
- Verify the nfpm `daemon/.goreleaser.yml` release section uses `mode: keep-existing` (or equivalent) to safely coexist with the docker job's release creation; consider adding a test or dry-run step.
- Consider extracting the shared `builds:` block via a YAML anchor or a shared includable snippet if GoReleaser supports it, to avoid drift between the two configs.
- Document in `docs/release-pipeline.md` the execution order (docker job first, nfpm second) and the release-mode contract between them, since the split now makes the ordering load-bearing.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Splits the daemon GoReleaser config into two files to fix the issue where nfpms packaging requires Flutter GUI artifacts that aren't available during the Docker-image release job. The Docker job now uses a dedicated `.goreleaser-docker.yml` (with only the binary build and docker image), while the Linux packages job uses the original `.goreleaser.yml` (now without the docker block). Workflow `--skip` flags and docs are updated accordingly. Change is focused and logically consistent.

### Issues

🟡 **daemon/.goreleaser-docker.yml:6** — The `daemon` build block is duplicated between `.goreleaser.yml` and `.goreleaser-docker.yml`. Any future change (ldflags, version injection, build tags) must be kept in sync manually, which is a maintenance hazard and easy to miss. Consider documenting this coupling with a short comment in both files, or using a shared snippet via `includes:` if feasible.
🟡 **daemon/.goreleaser-docker.yml:44** — `mode: keep-existing` on the release block is appropriate to avoid clobbering the release created by the CLI job, but this relies on the CLI release job running first. If job ordering changes (e.g. the daemon job runs before the CLI release is created), `keep-existing` will still attempt to attach to/create a release and the behavior may differ from expectations. Worth confirming via a `needs:` dependency in the workflow or a brief comment.
⚠️ **.github/workflows/release.yml:165** — The Docker job now runs goreleaser with the full release lifecycle (only `--skip=archive`), which means it may publish/tag a GitHub release in addition to the Docker image. Combined with the nfpms job also running `--skip=archive` on the default config, both jobs could race on release asset uploads. Verify that `mode: keep-existing` + job ordering (or explicit `needs:`) prevents duplicate release creation or asset-upload conflicts.
🟡 **daemon/.goreleaser-docker.yml:1** — No `dist:` directory override is set, so both goreleaser invocations (docker config and default config) will write to `daemon/dist/`. If these two jobs ever run on the same runner/workspace (or share caches), artifacts could collide. In the current workflow they run in separate jobs/runners so this is fine, but a distinct `dist: dist-docker` would be safer and self-documenting.

### Suggestions

- Add a short header comment to each of the two goreleaser config files stating which workflow job consumes it and which `--skip` flags are expected, so the split is obvious to future maintainers.
- Consider adding `needs: [cli-release-job]` (or whichever job creates the GitHub release) to both the daemon docker and linux-packages jobs to make the `keep-existing` contract explicit.
- Add a CI check or comment ensuring the `builds:` block stays in sync between `.goreleaser.yml` and `.goreleaser-docker.yml`, since divergence would silently produce different binaries in the docker image vs. the .deb/.rpm packages.
- Update the referenced comment in `.goreleaser.yml` (`# Used by: build-linux job (release.yml)`) to also mention the exact `--skip` flags used, mirroring the detail previously present (`--skip=docker,archive`), for discoverability.

---

PR #294, opened 2026-04-23
feat: implement #293 — bug: Linux packages build fails — GoReleaser rejects dirty tree from pubspec.lock

### LINKED ISSUE #293: bug: Linux packages build fails — GoReleaser rejects dirty tree from pubspec.lock
## Problem

The `Build Linux packages` job fails because GoReleaser detects a dirty working tree:

```
M flutter_app/pubspec.lock
Learn more at https://goreleaser.com/errors/dirty
```

`flutter pub get` modifies `pubspec.lock` before GoReleaser runs. GoReleaser refuses to build when there are uncommitted changes.

## Fix

Add a step to reset the lockfile before GoReleaser:

```yaml
- name: Reset Flutter lockfile changes
  run: git checkout -- flutter_app/pubspec.lock

```

Alternatively, pass `--skip=validate` to GoReleaser args, but resetting the file is cleaner.

## Files

- `.github/workflows/release.yml` — add git checkout step before GoReleaser in the Linux packages job

Ready for auto_implement.

sergiotejon:
Auto-generated by Heimdallm in response to #293.

Closes #293

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -367,6 +367,9 @@ jobs:
           echo "VERSION=$VERSION"  >> "$GITHUB_ENV"
 
       # ── .deb + .rpm (GoReleaser nfpms) ──────────────────────────────────────
+      - name: Reset Flutter lockfile changes
+        run: git checkout -- flutter_app/pubspec.lock
+
       - name: GoReleaser — build .deb + .rpm
         uses: goreleaser/goreleaser-action@v6
         with:


Muriano (approved):
## 🤖 Heimdallm AI Review

Simple and targeted fix for a CI build failure caused by Flutter's pubspec.lock dirtying the git tree before GoReleaser runs. The approach of resetting the single file with `git checkout --` is correct and minimal. However, this is a symptom-level fix rather than a root-cause solution — the lockfile gets dirty because a prior Flutter build step modifies it, and a .gitignore entry or GoReleaser `--dirty` flag could be more robust long-term.

### Suggestions

- Consider adding `flutter_app/pubspec.lock` to `.goreleaser.yaml`'s `before.hooks` or using `--dirty` flag if GoReleaser supports it, so the pipeline is resilient to any future lockfile changes without needing an explicit reset step.
- Investigate why the Flutter build step modifies pubspec.lock — if it's expected, consider whether pubspec.lock should be .gitignored (if not committed) or whether `flutter pub get --enforce-lockfile` should be used in the build step to prevent modifications.
- Add a comment above the reset step explaining WHY the lockfile gets dirty (e.g., 'Flutter build updates pubspec.lock; GoReleaser rejects dirty trees') so future maintainers understand the purpose of this step.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Small, targeted fix that resets pubspec.lock before GoReleaser runs to avoid a dirty-tree rejection. The change is minimal and correct for the immediate symptom, but it's a workaround rather than a root-cause fix and only covers pubspec.lock — any other Flutter-generated file changes would re-trigger the same failure.

### Issues

⚠️ **.github/workflows/release.yml:370** — This treats the symptom, not the cause. pubspec.lock is being mutated by an earlier Flutter step in the job; the workflow should identify which step dirties the tree and either avoid the mutation (e.g., ensure lockfile is already up-to-date upstream) or commit the regenerated lockfile intentionally. Leaving a 'git checkout --' band-aid means future Flutter upgrades will silently diverge between CI and the committed lockfile.
🟡 **.github/workflows/release.yml:370** — The reset is narrow: `git checkout -- flutter_app/pubspec.lock` only restores that one file. If `flutter pub get` or build steps also touch `.dart_tool/`, `pubspec_overrides.yaml`, generated files, or a future second lockfile, GoReleaser will fail again with a less obvious error. Consider either `git stash --include-untracked` / `git checkout -- flutter_app/` (scoped to the Flutter dir) or, better, using GoReleaser's `--skip=validate` only where justified, or configuring the build so the tree stays clean.
🟡 **.github/workflows/release.yml:370** — No guard or verification: if the file doesn't exist or git is not in the expected state, the step will fail opaquely. A short `git status --porcelain flutter_app/` log line before/after would make CI debugging much easier next time this breaks.
🟡 **.github/workflows/release.yml:370** — Step name 'Reset Flutter lockfile changes' doesn't explain WHY. Since this is explicitly a workaround for GoReleaser's dirty-tree check (issue #293), the name or a one-line comment above it should reference the issue so a future maintainer doesn't delete it as cleanup.

### Suggestions

- Investigate and document why pubspec.lock drifts during the release job — is it a Flutter SDK version mismatch between dev machines and the CI runner? Pinning the Flutter version or committing the CI-generated lockfile would remove the need for this reset entirely.
- Broaden the reset to `git checkout -- flutter_app/` (or use `git stash -u` scoped appropriately) to be resilient against other auto-generated Flutter artifacts dirtying the tree.
- Add a comment above the step referencing issue #293 so the workaround's intent is preserved: `# Workaround for #293: Flutter build mutates pubspec.lock, GoReleaser rejects dirty tree.`
- Consider emitting `git status --porcelain` before and after the reset for CI log visibility when this step does anything.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

The change adds a step to reset pubspec.lock before GoReleaser runs, addressing the dirty tree issue. The fix is minimal and targeted, but it masks the root cause rather than preventing the lockfile modification, and only resets one specific file which may not cover all potential dirty-tree sources.

### Issues

🟡 **.github/workflows/release.yml:370** — Using `git checkout --` discards changes without logging what was reset. Consider logging the diff first (e.g., `git diff flutter_app/pubspec.lock || true`) for debugging CI issues when lockfile changes are unexpected.
⚠️ **.github/workflows/release.yml:370** — This only resets pubspec.lock but other Flutter build artifacts (e.g., .dart_tool, build/, .flutter-plugins) could also dirty the tree. Consider using `git stash` or a broader reset, or alternatively pass `--skip=validate` / `GORELEASER_CURRENT_TAG` to GoReleaser to bypass the dirty-check intentionally.
⚠️ **.github/workflows/release.yml:370** — The fix treats the symptom rather than the cause. The Flutter build step upstream is modifying pubspec.lock — ideally that step should run with a flag that avoids lockfile mutation (e.g., `flutter pub get --enforce-lockfile`), or pubspec.lock should be regenerated deterministically before being committed.
🟡 **.github/workflows/release.yml:370** — No comment explaining WHY the reset is needed. A future maintainer seeing `git checkout -- flutter_app/pubspec.lock` won't understand it's a workaround for GoReleaser's dirty-tree check. Add a brief comment referencing the issue.

### Suggestions

- Add a comment above the step explaining it's a workaround for GoReleaser's dirty-tree validation after Flutter build mutates pubspec.lock
- Consider `flutter pub get --enforce-lockfile` in the upstream Flutter build step to prevent the lockfile mutation at the source
- Broaden the reset to cover other potentially-dirty Flutter artifacts, or use `git stash -u` before GoReleaser and `git stash pop` after if the changes are needed later
- Log the diff before resetting (`git diff flutter_app/pubspec.lock`) to aid debugging if the lockfile drift becomes a recurring issue

---

Issue #238, opened 2026-04-22
feat: comment-based control markers + pre-push state check to prevent duplicate processing


sergiotejon:
## Problem

Heimdallm reprocesses issues that have already been resolved, creating duplicate PRs. Two root causes:

1. **No state check before push**: Heimdallm starts auto_implement, takes 19 minutes, meanwhile the issue gets closed (PR merged). Heimdallm finishes and creates a duplicate PR because it never rechecked the issue state.

2. **No control mechanism**: There's no way to tell Heimdallm "this issue is done, don't touch it again" without closing the issue or using the dismiss API.

### Fix 1: Pre-push state check

Before `git push` + `CreatePR`, verify the issue is still open:

```go
// Just before push in auto_implement
issue, err := gh.GetIssue(repo, number)
if issue.State == "closed" {
    slog.Info("issues pipeline: issue closed during implementation, aborting",
        "repo", repo, "number", number)
    git.DeleteLocalBranch(branch)
    return nil, nil // no error, just skip
}
```

### Fix 2: Comment-based control markers

Heimdallm reads and writes HTML comment markers in issues:

**Markers:**

| Marker | Meaning | Who writes it |
|--------|---------|---------------|
| `<!-- heimdallm:done -->` | Issue fully processed, don't reprocess | Heimdallm (after successful auto_implement) or human |
| `<!-- heimdallm:skip -->` | Skip this issue permanently | Human |
| `<!-- heimdallm:retry -->` | Force reprocessing (overrides done) | Human |

**How it works:**

1. Before processing, Heimdallm scans the last N comments for markers
2. If `done` or `skip` found (and no `retry` after them) → skip
3. After successful auto_implement → Heimdallm posts a comment with `<!-- heimdallm:done -->`
4. If a human wants to force reprocessing → comment `<!-- heimdallm:retry -->`

**Visual output:**

The marker is an HTML comment (invisible in GitHub UI), but Heimdallm also writes visible text:

```markdown
<!-- heimdallm:done -->
✅ Implementation complete — PR #157 created on branch `heimdallm/issue-137`.
This issue will not be reprocessed unless `<!-- heimdallm:retry -->` is added.
```

**Priority (highest to lowest):**
1. Latest `<!-- heimdallm:retry -->` → force reprocess
2. Latest `<!-- heimdallm:skip -->` → permanent skip
3. Latest `<!-- heimdallm:done -->` → skip (normal completion)
4. Dismiss via API → skip (existing mechanism)
[...]

PR #296, opened 2026-04-23
fix(daemon): stop issue-triage cost-runaway (closes #292)

### LINKED ISSUE #292: bug(daemon): issue-triage cost-runaway — marker scan 404s on issue numbers, 47 re-triages in 46 min observed on #264
## TL;DR

The issue-triage pipeline has the same cost-runaway risk that #243 had on the PR side. The idempotency mechanism from #238 (control-marker scan on the issue's own comments) is **dead code on issue numbers**: `github.Client.FetchComments` unconditionally hits the PR-only `/pulls/:n/comments` endpoint, which always 404s for issues, and the whole call fails. The fetcher then falls through to a time-based dedup (#222's `CommentedAt + 30s` grace, `daemon/internal/issues/fetcher.go:214-221`) that is not robust against `updated_at` bumps from linked-PR activity. Every bump causes a re-triage; every re-triage posts a new comment; every new comment bumps `updated_at` again → infinite loop bounded only by the issue being closed.

**Live reproduction on 2026-04-23**, seen on this very repo's issue #264: from `2026-04-23T11:32:39Z` until `2026-04-23T12:18:31Z` (when PR #283 merged and closed the issue), the daemon posted **47 triage comments at severity=high**, one per minute. Trigger was opening PR #283 referencing `#264`, which started bumping the issue's `updated_at` via cross-reference timeline events and subsequent CI/bot activity on the linked PR. Total Claude burn: ~47 high-context runs on a single issue in 46 minutes. None of the PR-side defenses added in #246/#247/#258/#264 protect this path.

### GitHub side
```bash
$ gh api "repos/theburrowhub/heimdallm/issues/264/comments?per_page=100" --jq 'length'
48

$ gh api "repos/theburrowhub/heimdallm/issues/264/comments?per_page=100" \
  --jq '{first: .[0].created_at, last: .[-1].created_at}'
{
  "first": "2026-04-23T10:27:40Z",   # original triage when issue was opened
  "last":  "2026-04-23T12:18:31Z"    # 21 s before the merge that closed the issue
}
```

First comment at 10:27:40Z was legitimate (triage on open). Then nothing between 10:27 and 11:32 because the fix PR was being drafted. At 11:32:39Z the first re-triage landed (coinciding with PR #283 being opened against `#264`), and then **every ~60 s** until close. All 47 loop comments are from the heimdallm bot account (`ivanmunozruiz`), all severity=high.

### Daemon log (truncated)

```
time=2026-04-23T14:00:39.613+02:00 level=INFO msg="issues pipeline: triage complete" repo=theburrowhub/heimdallm number=264 severity=high posted=true
time=2026-04-23T14:01:15.360+02:00 level=WARN msg="issues fetcher: marker scan failed, falling through to dedup checks" repo=theburrowhub/heimdallm number=264 err="github: fetch review comments: status 404: {\"message\":\"Not Found\", \"documentation_url\": \"https://docs.github.com/rest/pulls/comments#list-review-comments-on-a-pull-request\", ... }"
time=2026-04-23T14:01:15.363+02:00 level=INFO msg="notify" title="Issue Triage Started" message="theburrowhub/heimdallm #264"
time=2026-04-23T14:01:15.737+02:00 level=WARN msg="issues pipeline: failed to fetch comments, proceeding without" err="github: fetch review comments: status 404: ..."
time=2026-04-23T14:01:36.002+02:00 level=INFO msg="issues pipeline: triage complete" repo=theburrowhub/heimdallm number=264 severity=high posted=true
```

Every tick shows the same pattern: `marker scan failed` (404) → fallthrough → re-triage → `posted=true`.

### Primary bug — `FetchComments` calls a PR-only endpoint on issues

`daemon/internal/github/client.go:470-501`:

```go
func (c *Client) FetchComments(repo string, number int) ([]Comment, error) {
    reviewCh := make(chan result, 1)
    issueCh := make(chan result, 1)
    go func() { reviewCh <- result{c.fetchReviewComments(repo, number)} }()   // /pulls/:n/comments
    go func() { issueCh  <- result{c.fetchIssueComments(repo, number)} }()    // /issues/:n/comments
    r1 := <-reviewCh
    r2 := <-issueCh
    if r1.err != nil { return nil, r1.err }   // <-- fails entire call when review-comments 404s
    if r2.err != nil { return nil, r2.err }
    ...
}

func (c *Client) fetchReviewComments(repo string, number int) ([]Comment, error) {
    path := fmt.Sprintf("/repos/%s/pulls/%d/comments", repo, number)   // PR-only endpoint
    ...
    if resp.StatusCode != http.StatusOK {
        return nil, fmt.Errorf("github: fetch review comments: status %d: %s", resp.StatusCode, errBody)
    }
    ...
}
```

`/repos/owner/repo/pulls/:n/comments` exists **only for PRs**. On an issue number it returns 404. `FetchComments` treats that as a hard error, aborting the whole call.

### Secondary bug — fetcher's marker scan depends on `FetchComments`

`daemon/internal/issues/fetcher.go:163-182`:

```go
if f.comments != nil {
    comments, cmErr := f.comments.FetchComments(issue.Repo, issue.Number)
    if cmErr != nil {
        slog.Warn("issues fetcher: marker scan failed, falling through to dedup checks",
            "repo", issue.Repo, "number", issue.Number, "err", cmErr)
    } else {
        switch ScanMarkers(comments) {
        case MarkerResultRetry: return false, "", nil
        case MarkerResultSkip:  return true, "skip marker", nil
        case MarkerResultDone:  return true, "done marker", nil
        }
    }
}
```

Every marker scan on an issue errors out (primary bug above) → marker check is skipped → control falls through to the time-based dedup at line 214-221:

```go
ref := latest.CommentedAt
if ref.IsZero() { ref = latest.CreatedAt }
cutoff := ref.Add(RecomputeGrace) // 30 s
}
```

### Why the fallback dedup isn't enough

`RecomputeGrace = 30 * time.Second` (`fetcher.go:24`). This was the fix for #222 — it addressed the race between `CreatedAt` (stored before the comment posted) and `UpdatedAt` (bumped when GitHub accepts the comment). It relies on the assumption that **the next 30 seconds after the bot's comment land is the only time `updated_at` moves**. But a linked PR on the issue generates cross-reference timeline events that bump `updated_at` arbitrarily — commits, CI status comments, AI reviews posted on the PR mentioning the issue, merge-base syncs, etc. Every minute something bumps #264 past the 30 s cutoff, every minute the fetcher re-dispatches, every re-triage posts a new comment that re-anchors the race, ad infinitum.

### Why the PR-side defenses don't cover this

- **#246 (SQLite circuit breaker, 3/PR/24h)** — keyed on `pr_id`, no issue-equivalent table.
- **#258 (persistent in-flight claim)** — table `reviews_in_flight` keyed on `(pr_id, head_sha)`, no issue-equivalent row type; the pipeline.Run this PR just completed (#264, via PR #283) is PR-only.
- **#247 (PublishedAt + 2-min grace)** — pipeline.Run-level dedup on PR reviews; issue triage has its own (weaker) `CommentedAt + 30s` grace in `fetcher.go`.
- **#245 (fail-closed HEAD SHA)** — pure PR concept.

### Fix 1 — root cause (mandatory)

Split comment fetching so the issue-triage path stops calling the PR-only endpoint.

**Files:**
- Modify `daemon/internal/github/client.go:470-501`
- Modify `daemon/internal/issues/fetcher.go:61-63, 168`

**Sketch (option A, preferred):** add a dedicated `FetchIssueCommentsOnly` method that returns only issue comments, and wire the issue fetcher to use it instead of `FetchComments`. Leaves `FetchComments` untouched for PR callers who need both lists.

```go
// github/client.go — new method
func (c *Client) FetchIssueCommentsOnly(repo string, number int) ([]Comment, error) {
    return c.fetchIssueComments(repo, number) // /repos/:repo/issues/:n/comments
}

// issues/fetcher.go — widen the interface
type issueMarkerFetcher interface {
    FetchIssueCommentsOnly(repo string, number int) ([]github.Comment, error)
}

// in alreadyProcessed:
if f.comments != nil {
    comments, cmErr := f.comments.FetchIssueCommentsOnly(issue.Repo, issue.Number)
    ...
}
```

**Sketch (option B, if B is simpler in context):** make `FetchComments` tolerate a 404 from the review-comments leg and return just the issue comments in that case. Less isolated but shorter diff.

Pick A. Review comments never exist on an issue, so calling the PR endpoint on it is genuinely wrong — a dedicated "issue-only" method makes the API explicit.

**Test:** unit test against a fake HTTP server that returns 404 on `/pulls/:n/comments` and a valid issue-comments payload on `/issues/:n/comments`. Assert the new method returns the issue comments with no error. Add a test that drives `alreadyProcessed` with the new fetcher wired in and a triage-marker-bearing comment — assert `MarkerResultDone` short-circuits the dedup regardless of `updated_at` value.

### Fix 2 — defense-in-depth: persistent in-flight claim for issues (recommended)

Extend the pattern from #258 to issue triage. Either add a sibling table `issue_triage_in_flight (issue_id INTEGER, updated_at TEXT, started_at DATETIME, PRIMARY KEY(issue_id, updated_at))` or generalize `reviews_in_flight` to a polymorphic `(entity_type, entity_id, identity_key)` table.

**Files:**
- New: `daemon/internal/store/issue_inflight.go` (or extend `inflight.go`)
- Schema migration: `daemon/internal/store/store.go` (new CREATE TABLE)
- Wire-in: `daemon/internal/issues/pipeline.go` (gate at the top of `runReviewOnly` and `runAutoImplement`)
- Clear stale claims on startup, mirroring `ClearStaleInFlight`

**Identity key for issues:** use `issue.UpdatedAt` truncated to ISO-seconds, so a genuinely new activity (new updated_at) produces a new claim, but two ticks on the same snapshot collapse.

**Test:** mirror `TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview` from `daemon/cmd/heimdallm/main_inflight_plumb_test.go`. Two concurrent `fetcher.ProcessRepo` calls on the same issue → only one `pipeline.Run` reaches Claude.

### Fix 3 — defense-in-depth: extend circuit breaker to issues (recommended)

Mirror #246's SQLite-backed circuit breaker for issue triage:

- Per-issue: cap at 3 triages / 24 h (same as the PR equivalent).
- Per-repo-issue-triage: cap at 10 triages / hour (tighter than PR's 20, because triages cost a full-context Claude run each).

**Files:**
- Extend `daemon/internal/store/circuitbreaker.go` with per-entity-type counters, or add a parallel function set.
- Wire in `daemon/internal/issues/pipeline.go` at the same gate as Fix 2.

**Test:** seed 3 failed/successful triages within 24 h, assert the 4th is rejected with a `CircuitBreakerError` instead of dispatching.

### Fix 4 — observability: counter on marker-scan fallthrough (nice-to-have)

Expose a Prometheus-style counter (`heimdallm_issue_marker_scan_fallthrough_total{repo=...}`) that increments whenever `alreadyProcessed` hits the `falling through to dedup checks` WARN. An operator alarm on sustained growth catches the next incident before it accumulates cost.

**Files:**
- `daemon/internal/issues/fetcher.go:170` — increment counter alongside the existing WARN.
- Expose the counter on the existing `/metrics` endpoint (or whatever mechanism is already in place).

## Acceptance criteria

- [ ] `make test-docker` passes (no raw `go test` on host per team policy — EDR will complain).
- [ ] Fix 1: Unit test for `FetchIssueCommentsOnly` using `httptest.NewServer`.
- [ ] Fix 1: Integration test driving `alreadyProcessed` against an issue whose comments carry a done-marker; asserts skip regardless of `updated_at`.
- [ ] Fix 1: No `issues fetcher: marker scan failed` WARN appears in daemon logs against issue numbers during normal operation (acceptable on transient 5xx only).
- [ ] Fix 2: Regression test mirroring `TestTier2Adapter_ProcessPR_ConcurrentCallsCollapseToOneReview` — two concurrent triage dispatches on the same `(issue_id, updated_at)` collapse to one Claude call.
- [ ] Fix 3: Test that 4 triages on the same issue in 24 h trip the circuit breaker on the 4th attempt; `CircuitBreakerError` is returned and no Claude call is made.
- [ ] Post-merge verification: on a live daemon, tail logs during PR-linked activity on an open issue and confirm the re-triage loop does not recur (0 `posted=true` lines on an issue whose last triage's marker is still present).

## References

- **This issue's immediate ancestor:** #264 (PR-side wiring fix closed via PR #283 on 2026-04-23). That fix solved the PR half of the cost-runaway; this one extends the same defenses to the issue half.
- **Original P0 incident:** #243 (€1,300 review-loop runaway).
- **Defense lineage to mirror:**
  - #246: SQLite-backed circuit breaker (PR review).
  - #258: persistent in-flight review claim (PR review).
  - #247: PublishedAt + 2-min grace (PR dedup).
  - #245: fail-closed HEAD SHA lookup (PR).
- **Earlier iterations of the triage-loop bug:**
  - #222: `RecomputeGrace` grace-window race (fixed by PR #226 — addresses the same file at `fetcher.go:214-221` but a different root cause; this new bug sits on top of that fix).
  - #220: re-triage loses context from previous triage (unrelated behavior but same pipeline).
  - #176: issues re-processed after implement (separate concern, auto_implement side).
- **Idempotency mechanism disabled by this bug:** #238 (comment-based control markers + pre-push state check).

## Scope note

Fix 1 is the minimum viable patch to close the bleeding. Fixes 2 and 3 are defense-in-depth and map cleanly to the PR-side work we already shipped — recommended to do them as separate PRs right after Fix 1 lands (same approach as the #243 plan: root-cause fix first, layered defenses after). Fix 4 can wait until there is a metrics pipeline to land it into.

## Severity

**HIGH** — actively cost money on 2026-04-23 (one incident already observed). Structural weakness mirrors #243. No upstream defense prevents recurrence; next linked PR against any tracked issue can restart the loop.

ivanmunozruiz:
## Summary

Closes #292. Root-cause fix + two defense-in-depth layers, all in one PR so the issue-side of the cost-runaway (mirror of #243 / #264 but for issue triage) gets fully closed in one merge.

### Fix 1 — root cause (commit `27218f2`)
`github.Client.FetchComments` calls both `/pulls/:n/comments` and `/issues/:n/comments` in parallel and fails the whole call when either leg errors. On an issue number `/pulls/:n/comments` always 404s, so the issue-triage marker scan (`daemon/internal/issues/fetcher.go:163-182`) silently aborted every tick and fell through to the weaker `CommentedAt + 30s` dedup — which a linked-PR's cross-reference activity can bump past repeatedly (47 re-triages observed on #264 in 46 minutes on 2026-04-23).

- New `Client.FetchIssueCommentsOnly` that only hits `/issues/:n/comments`.
- `issueMarkerFetcher` interface (fetcher) and `IssueCommentFetcher` interface (pipeline) now require the issue-only method; `FetchComments` stays unchanged for PR callers.
- The pipeline's triage-context fetch in `runReviewOnly` / `runAutoImplement` also moved to the issue-only method, eliminating a silent quality regression (triage was always running without prior comment context because every call 404'd).

### Fix 2 — in-flight claim for issues (commit `6aa7cba`)
Mirror of the PR-side persistent claim (#258). Two concurrent fetcher ticks observing the same `(github_issue_id, updated_at)` snapshot now collapse to one `pipeline.Run`.

- New table `issue_triage_in_flight (issue_id, updated_at, started_at)` with PK `(issue_id, updated_at)`.
- `ClaimIssueTriageInFlight` / `ReleaseIssueTriageInFlight` / `ClearStaleIssueTriageInFlight` store methods, symmetric with the PR trio.
- `issues.Pipeline.Run` claims at the top using `issue.ID` and `issue.UpdatedAt.UTC().Format(time.RFC3339)` so ticks on the same snapshot collapse but a genuine update produces a new claim. Fail-open on claim errors.
- Startup sweeps stale claims older than 30 min, same policy as the PR sweep.
- Regression test `TestIntegration_ConcurrentPipelineRunsCollapseToOneDispatch` uses a deterministic hand-off (blocking executor + channels + `sync.Once`) to assert exactly one executor call even when two goroutines dispatch concurrently. Stress-tested at `-count=10` — no flakes.

### Fix 3 — circuit breaker for issues (commit `4a22fc0`)
Mirror of the PR-side circuit breaker (#246). Hard cap on triages per issue and per repo.

- Store: `CountIssueReviewsForIssue`, `CountIssueTriagesForRepo`, `IssueCircuitBreakerLimits`, `CheckIssueCircuitBreaker`.
- Pipeline: `ErrCircuitBreakerTripped` + `*CircuitBreakerError`, `SetCircuitBreakerLimits`. Check runs AFTER the claim and the upsert so the counter keys on the internal store ID.
- Config: new `per_issue_24h` / `per_issue_repo_hr` knobs under `[circuit_breaker]`. Defaults **3** and **10** — tighter per-repo cap than the PR cap (20) because every triage is a full-context Claude run.
- `main.go` wires config → `store.IssueCircuitBreakerLimits` → `issuePipe.SetCircuitBreakerLimits`.
- "0 = unlimited" contract locked in with a regression test, same pattern as the PR-side breaker.
- Integration test seeds 3 prior triages and asserts the 4th returns `*issues.CircuitBreakerError` without reaching the executor.

Fix 4 (Prometheus metric on marker-scan fallthrough) from the issue is **not** included — no metrics pipeline exists in the daemon today; filing it as a separate observability follow-up makes more sense than landing a stub counter here.

## Test plan

- [x] `make test-docker` green (EDR policy, no raw `go test` on host)
- [x] New store tests: `TestIssueInflight_*`, `TestCountIssue*`, `TestCheckIssueCircuitBreaker_*`
- [x] New github client test: `TestFetchIssueCommentsOnly_IgnoresPREndpoint`, `TestFetchIssueCommentsOnly_PropagatesRealErrors`
- [x] New integration tests: `TestIntegration_ConcurrentPipelineRunsCollapseToOneDispatch`, `TestIntegration_IssueCircuitBreakerTripsAfterCap`
- [x] Concurrent test stress-tested at `-count=10` (zero flakes)
- [ ] Post-merge live verification: tail daemon logs during PR-linked activity on an open issue; confirm no `issues fetcher: marker scan failed` WARN on issue numbers and no re-triage loop.

## References

- Closes #292
- Mirrors #258 (PR in-flight claim), #246 (PR circuit breaker)
- Related lineage: #243 (P0 cost incident), #264 / #283 (PR-side wiring fix)
- Idempotency mechanism restored: #238 (control-marker scan)


sergiotejon:
## 🟡 LOW Issue

claim uses issue.ID (GitHub ID) but circuit breaker uses issueID (store ID). The claim key space and breaker key space are mismatched — if two issues ever share a GitHub ID across installations the claim would collide, though unlikely in practice.

**Location:** `daemon/internal/issues/pipeline.go` line 295

---

sergiotejon:
## 🟡 LOW Issue

On claim error the code proceeds (fail-open) but also sets claimed=false, so the defer won't release. If the claim actually succeeded but the error was transient/misleading, the row will remain in issue_triage_in_flight until the stale sweep. Acceptable given the stale sweep exists, but worth a comment.

**Location:** `daemon/internal/issues/pipeline.go` line 307

---

sergiotejon:
## ⚠️ MEDIUM Issue

blockingExec.called is a plain bool written from goroutines without synchronization. Should be atomic or protected by a mutex to avoid a data race under -race.

**Location:** `daemon/internal/issues/integration_test.go` line 155

---

sergiotejon:
## 🟡 LOW Issue

FetchIssueCommentsOnly is called on p.gh which is typed as issueGitHub interface containing FetchIssueCommentsOnly — but the interface also still declares FetchComments in the full issueGitHub interface (not shown in diff). Confirm FetchComments was removed from issueGitHub or this is a partial interface; otherwise dead method on the interface.

**Location:** `daemon/internal/issues/pipeline.go` line 400

---

sergiotejon:
## 🟡 LOW Issue

CheckIssueCircuitBreaker checks per-issue THEN per-repo sequentially. If per-issue passes but per-repo query fails, the error propagates and the caller falls open. This is fine but means a repo-level DB error won't trip the breaker even when the per-issue cap is not yet hit.

**Location:** `daemon/internal/store/issue_circuitbreaker.go` line 55

---

sergiotejon (approved):

Solid defensive fix mirroring PR-side circuit breaker and in-flight claim patterns to the issue-triage path. Well-tested. A few minor concerns.

### Suggestions

- Make blockingExec.called an atomic.Bool to avoid data races in TestIntegration_IssueCircuitBreakerTripsAfterCap and concurrent test scenarios.
- Consider adding an index on (issue_id, created_at) for issue_reviews and (repo, created_at) via the join for the circuit breaker COUNT queries if not already present — these run on every triage attempt.
- Add a brief integration test verifying ClearStaleIssueTriageInFlight actually unblocks a subsequent claim on the same key.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid, well-tested fix for the issue-triage cost runaway — mirrors the proven PR-side defenses (inflight claim, circuit breaker, stale sweep) and correctly decomposes FetchComments to avoid the 404 cascade that caused #292. A few real concerns: the inflight claim key uses issue.ID (GitHub ID) while the breaker uses the internal store ID, creating two distinct key spaces for the same logical resource; a data race on blockingExec.called in the new concurrent test; and breaker trips will re-notify on every tick because the claim is released on trip.

### Issues

⚠️ **daemon/internal/issues/pipeline.go:295** — Claim key uses issue.ID (GitHub ID), but the breaker, store seeding, and all existing _test_helpers use the internal store ID returned by UpsertIssue. If issue_triage_in_flight.issue_id is a FK to issues.id (as the stale-sweep helper suggests), this is a type-level mismatch: claims insert with one ID space, the sweep/telemetry read another. At minimum document the contract; ideally claim AFTER UpsertIssue so both paths share the store ID.
⚠️ **daemon/internal/issues/integration_test.go:155** — blockingExec.called is a plain bool written by the executor goroutine and read by the test goroutine without synchronization. Will trip go test -race. Use sync/atomic (atomic.Bool) or a mutex, consistent with execCount which already uses atomic.
⚠️ **daemon/internal/issues/pipeline.go:348** — On circuit-breaker trip the defer releases the inflight claim, so the next fetcher tick on the same (issue, updated_at) will re-acquire the claim, re-hit the breaker, and re-send a notification. Consider keeping the claim held on trip (don't release), or suppress p.notify.Notify when the most recent trip on this (issue, updated_at) was already notified. Otherwise a single stuck runaway scenario spams the notifier once per poll.
🟡 **daemon/internal/issues/pipeline.go:307** — On claim error the code fails-open and sets claimed=false, so the defer skips release. If the underlying claim actually landed but the error was transient (e.g. Scan error on a successful INSERT), the row stays until the 30-minute stale sweep. Acceptable given the sweep, but worth an inline comment so future readers don't 'fix' it.
🟡 **daemon/internal/store/issue_circuitbreaker.go:55** — CheckIssueCircuitBreaker evaluates per-issue then per-repo sequentially and returns on the first error. A transient DB error on the per-repo query bubbles up as an error (caller fails-open) even when the per-issue cap is clearly tripped. Consider: if per-issue returns tripped=true, return immediately; only query per-repo if per-issue passed cleanly.
🟡 **daemon/internal/issues/pipeline.go:88** — IssueCommentFetcher now only declares FetchIssueCommentsOnly. Confirm no remaining call site on p.gh uses FetchComments against an issue number — a grep for p.gh.FetchComments should return zero hits in the issue package after this change. If issueGitHub still embeds FetchComments, the method becomes dead weight on the issue-side interface.
🟡 **daemon/internal/issues/integration_test.go:310** — The comment 'But wait — the pipeline checks the claim on gh.ID...' reads like an authoring aside left in production test code. Keep the assertion, but tighten the comment to a one-liner stating the contract (pipeline resolves store ID before breaker check) without the conversational framing.
🟡 **daemon/internal/config/circuit_breaker.go:22** — PerIssueRepoHr default (10) is tighter than PerRepoHr (20) with justification 'each triage is a full-context Claude run'. Reasonable, but users running both PR review and issue triage on the same repo now have two independent 1-hour axes — a busy repo could consume 20 PR reviews AND 10 triages per hour. Worth noting in the TOML comment that these axes do not share a budget.

### Suggestions

- Move the ClaimIssueTriageInFlight call to AFTER UpsertIssue so the claim and breaker share the internal store ID — eliminates the key-space mismatch and simplifies the stale-sweep FK story.
- Convert blockingExec.called to atomic.Bool (matching the atomic.Int32 already used for execCount) so the new integration test is clean under -race.
- Either hold the claim through a breaker trip (don't release in defer when the trip fired) or rate-limit the breaker-trip notification per (issue_id, updated_at) so a single runaway doesn't spam the notifier each poll.
- Short-circuit CheckIssueCircuitBreaker: return (true, reason, nil) as soon as the per-issue axis trips, without also hitting the per-repo query — cheaper and avoids masking a trip behind a transient repo-query error.
- Add one regression test that drives Run twice on the SAME (issue, updated_at) without concurrency, asserts the second call returns (nil, nil), and then drives a third call after bumping updated_at to confirm the new snapshot is allowed through — locks in the claim-key contract explicitly.

---

Muriano:
## 🟡 LOW Issue

slog.Error used for circuit breaker trip, but this is expected operational behavior under load — not an error. slog.Warn is more appropriate and avoids polluting error-level alerting. The PR-side circuit breaker should be checked for consistency, but using Error here risks triggering on-call alerts for a normal cost-control mechanism.

**Location:** `daemon/internal/issues/pipeline.go` line 360

---

Muriano (approved):

Well-structured fix for issue-triage cost-runaway (#292). Mirrors the existing PR-side defenses (in-flight claims, circuit breaker, stale cleanup) onto the issue pipeline. Root cause (FetchComments 404ing on issue numbers) is correctly identified and fixed with FetchIssueCommentsOnly. Tests are thorough — concurrent dispatch collapse, circuit breaker trip, and the new GitHub client method all have dedicated coverage. No high-severity issues found.

### Suggestions

- Verify the migration (truncated from diff) includes an index on issue_reviews(issue_id, created_at) — CountIssueReviewsForIssue and CountIssueTriagesForRepo do COUNT(*) with time-range filters that will degrade on large tables without a covering index. The PR-side reviews table has idx_reviews_pr_created for the same reason.
- In blockingExec (integration_test.go), the `called` field is a plain bool while `execCount` correctly uses atomic.Int32. If a future test refactor introduces concurrent reads of `called`, the race detector will flag it. Consider using atomic.Bool for consistency, though it's not a bug today since assertions run after goroutine completion with proper happens-before via channels.
- The doc comment on CircuitBreakerConfig fields says '0 = unlimited' but applyDefaults() replaces 0 with the default value, making it impossible for users to configure unlimited via TOML. This is consistent with the existing PR-side behavior so not a regression, but the doc comment is misleading — consider updating it to '0 = use default' or documenting a sentinel like -1 for unlimited.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Solid, well-structured fix that addresses the #292 cost-runaway through three complementary layers: root-cause fix (FetchIssueCommentsOnly avoids the /pulls 404 cascade), persistent in-flight claims (dedup concurrent fetcher ticks), and a circuit breaker (hard cap on triage count). All three mirror established PR-side patterns. Comprehensive test coverage including concurrency, circuit breaker trip/allow, stale-claim cleanup, and error propagation. No blocking issues found.

### Suggestions

- applyDefaults() in config.go treats 0 as 'unset' and overwrites it with the default (3/10), but the doc comments on CircuitBreakerConfig say '0 = unlimited'. This means a user who explicitly sets `per_issue_24h = 0` in TOML to get unlimited triages will silently get the default cap of 3 instead. This is the same pre-existing pattern as the PR-side caps (PerPR24h/PerRepoHr), so it's consistent — but worth a follow-up to allow explicit opt-out, e.g. via a negative sentinel or a separate boolean flag.
- blockingExec.called in integration_test.go is a plain bool written inside a goroutine (ExecuteRaw) without synchronization. It works today because the concurrent test only checks the atomic execCount and the circuit-breaker test is single-goroutine, but if a future test reads `called` while a goroutine writes it, the race detector will flag it. Consider making it an atomic.Bool for hygiene.
- The issue_triage_in_flight table schema lacks an index on started_at, which ClearStaleIssueTriageInFlight scans with a range predicate. At current volumes this is negligible, but if the table grows (e.g. many repos with high-frequency polling), an index on started_at would keep the startup sweep fast. Same observation applies to the PR-side reviews_in_flight table.

---

sergiotejon:
## ⚠️ MEDIUM Issue

blockingExec.called is a plain bool written/read across goroutines without synchronization. Will fail under -race.

**Location:** `daemon/internal/issues/integration_test.go` line 155

---

sergiotejon:
## 🟡 LOW Issue

slog.Error for circuit breaker trip is misleading — this is expected operational behavior, not an error. Use slog.Warn to avoid false alerts.

**Location:** `daemon/internal/issues/pipeline.go` line 360

---

sergiotejon:
## 🟡 LOW Issue

claimIssueID uses issue.ID (GitHub ID) but CheckIssueCircuitBreaker uses issueID (store ID). Different key spaces for claim vs breaker is confusing and could mask bugs if the mapping ever diverges.

**Location:** `daemon/internal/issues/pipeline.go` line 295

---

sergiotejon:
## 🟡 LOW Issue

On claim error, claimed stays false so defer won't release. If the row was actually inserted, it leaks until the stale sweep. A brief comment acknowledging this would help.

**Location:** `daemon/internal/issues/pipeline.go` line 307

---

sergiotejon:
## 🟡 LOW Issue

CheckIssueCircuitBreaker runs two sequential queries without a transaction. A concurrent insert between the two checks could let a triage slip through just as the repo cap is hit.

**Location:** `daemon/internal/store/issue_circuitbreaker.go` line 55

---

sergiotejon (approved):

Solid defensive fix for a real cost-runaway incident. The three-layer approach (in-flight claim, marker scan fix, circuit breaker) is sound. A few concurrency and logging issues remain.

### Suggestions

- Change blockingExec.called to atomic.Bool or protect with a mutex to fix the data race
- Downgrade slog.Error to slog.Warn for circuit breaker trips — reserve Error for unexpected failures
- Consider wrapping the two circuit breaker count queries in a read transaction for consistency

---

Muriano:
## ⚠️ MEDIUM Issue

blockingExec.called is a plain bool while execCount uses atomic.Int32. Although current tests establish happens-before via channels, if the claim dedup ever regresses and both goroutines reach ExecuteRaw, the concurrent writes to `called` would be flagged by -race. Use atomic.Bool or int32 for consistency with execCount and to be safe under -race in failure scenarios. Impact: flaky -race failures if claim logic regresses. Fix: change `called bool` to `called atomic.Bool` and use `.Store(true)` / `.Load()` at call sites.

**Location:** `daemon/internal/issues/integration_test.go` line 155

---

Muriano:
## ⚠️ MEDIUM Issue

Doc comments state '0 = unlimited' for PerIssue24h and PerIssueRepoHr, but applyDefaults() in config.go overwrites 0 with the default value (3 and 10 respectively), making it impossible to configure unlimited via TOML. This is the same pre-existing issue as PerPR24h/PerRepoHr, but the new fields perpetuate a misleading doc contract. Impact: users who set per_issue_24h=0 expecting unlimited get the default cap instead. Fix: either use a pointer/sentinel (-1) for unlimited, or correct the doc to say '(omit or set to 0 to use the default)'.

**Location:** `daemon/internal/config/circuit_breaker.go` line 18

---

Muriano (approved):

Well-structured PR that correctly mirrors PR-side cost-runaway defenses (in-flight claim, circuit breaker, stale sweep) to the issue-triage pipeline, and fixes the root-cause 404 cascade from FetchComments hitting /pulls/:n/comments on issue numbers. Architecture is consistent with existing patterns. No high-severity issues found; the existing reviewer comments are accurate but low-impact.

### Suggestions

- In TestIntegration_IssueCircuitBreakerTripsAfterCap, the variable `gh` (line ~261) shadows the conventional meaning of `gh` as the GitHub client used elsewhere in the file (fakeGH). Consider naming it `issueFixture` or `ghIssue` for clarity.
- The claim key space (GitHub issue ID) diverges from the PR-side pattern which uses store IDs. While functional, adding a one-line comment in pipeline.go explaining why GitHub IDs are used for claims (snapshot dedup) vs store IDs for the circuit breaker (historical count) would help future maintainers.
- Consider reconciling the '0 = unlimited' contract across both PR-side and issue-side circuit breaker configs in a follow-up — using -1 as the unlimited sentinel would make the TOML contract honest and avoid surprising users.

---

sergiotejon:
## ⚠️ MEDIUM Issue

blockingExec.called is a plain bool written/read across goroutines. Should be atomic.Bool to avoid -race failures.

**Location:** `daemon/internal/issues/integration_test.go` line 155

---

sergiotejon:
## ⚠️ MEDIUM Issue

Doc says '0 = unlimited' but applyDefaults() replaces 0 with the default value, making unlimited impossible to configure. Misleading contract.

**Location:** `daemon/internal/config/circuit_breaker.go` line 18

---

sergiotejon:
## 🟡 LOW Issue

slog.Error for circuit breaker trip is expected operational behavior, not an error. Use slog.Warn to avoid false alerts.

**Location:** `daemon/internal/issues/pipeline.go` line 360

---

sergiotejon:
## 🟡 LOW Issue

claim uses issue.ID (GitHub ID) while circuit breaker uses issueID (store ID). Different key spaces is confusing and could mask bugs.

**Location:** `daemon/internal/issues/pipeline.go` line 295

---

sergiotejon:
## 🟡 LOW Issue

Two sequential queries without a transaction; a concurrent insert between checks could let a triage slip through at the repo cap boundary.

**Location:** `daemon/internal/store/issue_circuitbreaker.go` line 55

---

sergiotejon (approved):

Solid fix for issue-triage cost runaway. Adds in-flight claims, circuit breaker, and fixes the FetchComments 404 cascade. The PR discussion already surfaced the key issues; the most actionable are the data race on blockingExec.called and the misleading 0=unlimited doc.

### Suggestions

- Change blockingExec.called from bool to atomic.Bool for -race safety
- Fix the 0=unlimited doc contract: either use -1 as the sentinel for unlimited or document that 0 means 'use default'
- Downgrade slog.Error to slog.Warn for circuit breaker trips to avoid polluting error-level alerting

---

ivanmunozruiz:
Thanks for the reviews — addressed in `e97c417`. Summary of what changed and what didn't:

**Applied (8 items):**
- 🛑 **Notification spam on breaker trip** (vbuenog MEDIUM): claim is now held across a trip, so the next tick on the same `(issue, updated_at)` short-circuits at the claim check instead of re-tripping and re-notifying. Stale-claim sweep (30 min) eventually reclaims; a real activity bump produces a new claim key. New regression test asserts the second `Run` returns `(nil, nil)` and notifier count stays at 0.
- ⚠️ **`blockingExec.called` data race** (4 reviewers MEDIUM): now `atomic.Bool` with `Store(true)` / `Load()`. Matches the `atomic.Int32` pattern already used for `execCount`.
- ⚠️ **"0 = unlimited" doc lying** (3 reviewers MEDIUM): doc rewritten to "Default N; set to 0 to apply the default". Honest about the actual `applyDefaults` behaviour.
- 🟡 **Missing index trio** (Muriano LOW): added `idx_issue_reviews_issue_created`, `idx_issue_reviews_created`, `idx_issues_repo` — mirrors of the PR-side indexes from #243.
- 🟡 **Key-space mismatch (claim uses GitHub ID, breaker uses store ID)** (sergiotejon ×4, vbuenog, Muriano LOW): kept the split (it's intentional — claim must run before upsert to gate Run early; breaker must use store ID because `issue_reviews.issue_id` references `issues.id`) but added a clear inline comment in `pipeline.go` explaining the contract so future readers don't try to "unify" them.
- 🟡 **Claim-error fail-open leaks row** (sergiotejon ×2 LOW): added inline comment documenting that the row may leak until the 30-min sweep, and why the alternative (assume insert landed and release) would be worse.
- 🟡 **`gh` variable shadowing in test** (Muriano LOW): renamed to `ghIssue`.
- 🟡 **"But wait…" authoring aside** (vbuenog LOW): rewritten as a tight contract statement about the upsert-before-breaker ordering.

**Not in this PR (filed as follow-up considerations):**
- `slog.Error` → `slog.Warn` for breaker trip: the PR-side breaker uses `slog.Error` too — needs to change in both for consistency. Out of scope here.
- Negative sentinel (`-1`) for "unlimited" caps: behaviour change across PR + issue paths. Out of scope here.
- Transactional `CheckIssueCircuitBreaker`: the store opens with `db.SetMaxOpenConns(1)` so the two queries are already serialized — concurrent insert can't slip between them.
- Short-circuit when per-issue trips: `CheckIssueCircuitBreaker` already returns immediately on trip; the suggestion is already implemented.
- Standalone integration test for `ClearStaleIssueTriageInFlight`: covered by the unit test in `store/issue_inflight_test.go:TestIssueInflight_StaleEntriesAreCleared`.

`make test-docker` green across all daemon packages, plus stress run of `TestIntegration_*` at `-count=10` with no flakes.

PR #297, opened 2026-04-23
fix: reset entire worktree before GoReleaser

### LINKED ISSUE #293: bug: Linux packages build fails — GoReleaser rejects dirty tree from pubspec.lock

sergiotejon:
Fixes #293. git checkout -- . instead of just pubspec.lock.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -367,8 +367,8 @@ jobs:
           echo "VERSION=$VERSION"  >> "$GITHUB_ENV"
 
       # ── .deb + .rpm (GoReleaser nfpms) ──────────────────────────────────────
-      - name: Reset Flutter lockfile changes
-        run: git checkout -- flutter_app/pubspec.lock
+      - name: Reset worktree for GoReleaser (Flutter build dirties tree)
+        run: git checkout -- .
 
       - name: GoReleaser — build .deb + .rpm
         uses: goreleaser/goreleaser-action@v6


Muriano (approved):
## 🤖 Heimdallm AI Review

The PR broadens a targeted git checkout of a single file (flutter_app/pubspec.lock) to reset the entire worktree (`git checkout -- .`). This fixes GoReleaser's dirty-tree check more robustly but is a wider blast radius — any intentionally modified or generated file in earlier steps will also be discarded. In this CI context the risk is low because the workflow is linear and earlier artifacts (Flutter build outputs, VERSION env var) are either consumed already or stored outside the worktree, but the approach trades precision for convenience.

### Suggestions

- Consider scoping the reset more narrowly (e.g., `git checkout -- flutter_app/` or listing the known dirty paths) to avoid silently discarding files that a future workflow step might intentionally modify. A broad `git checkout -- .` is safe today but fragile as the workflow evolves.
- Add a brief inline comment or echo statement listing *why* the tree is dirty (Flutter build artifacts, lockfile changes) so future maintainers understand the intent without reading git blame.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Small CI fix that broadens the pre-GoReleaser cleanup from a single file to the whole tracked tree. Pragmatic and likely correct for the immediate problem, but the wider blast radius and the fact that untracked files are not cleaned deserve a callout.

### Issues

⚠️ **.github/workflows/release.yml:371** — `git checkout -- .` discards ALL tracked-file modifications in the worktree, not just Flutter's pubspec.lock. If any earlier step in this job intentionally edits a tracked file that GoReleaser needs (e.g., a templated version file, generated metadata, or a manifest), those edits will be silently reverted. Worth verifying nothing upstream of this step relies on a dirty tracked file.
🟡 **.github/workflows/release.yml:370** — The step name says "Reset worktree" but `git checkout -- .` only resets tracked files; untracked files produced by the Flutter build (build/, .dart_tool/, generated assets) are left in place. If GoReleaser's dirty-tree complaint is ever triggered by an untracked file, this fix won't catch it. Consider `git clean -fdx` (scoped, e.g. to flutter_app/) alongside the checkout, or at minimum align the step name with what it actually does.
🟡 **.github/workflows/release.yml:371** — Root-cause is masked rather than fixed: the Flutter build step is dirtying tracked files. A more durable fix is to make the Flutter step not mutate tracked state (e.g., commit the lockfile drift, or run the build with flags that don't rewrite pubspec.lock), so the workflow doesn't accumulate "reset everything" band-aids over time.

### Suggestions

- Add a brief comment above the step explaining WHY a full checkout is needed (which prior step dirties the tree) so future maintainers don't narrow it back down and reintroduce the GoReleaser failure.
- If the goal is truly a clean tree for GoReleaser, pair `git checkout -- .` with `git clean -fd -- flutter_app` (scoped to avoid nuking unrelated artifacts) and/or run `git status --porcelain` immediately after for diagnostic visibility in CI logs.
- Investigate fixing the Flutter step so it doesn't mutate tracked files in the first place — that removes the need for any reset step.

---

PR #318, opened 2026-04-23
feat(nats): PR publish worker with ack/nak retry (#305)

Muriano:
## Summary

- Create `PublishWorker` NATS consumer with ack/nak error handling:
  - `nil` return → Ack (success or permanent failure)
  - Non-nil error → NakWithDelay(30s) for NATS retry on transient GitHub errors
  - Panic recovery via safeHandle (panic → nak, not crash)
- Add `PRPublishPublisher` with dedup by review ID
- Review worker now publishes `PRPublishMsg` after successful review (happy path)
- Simplified `PublishPending()` to enqueue via NATS instead of direct GitHub submit (catch-up scanner)
- Export `BuildGitHubBody` and `SeverityToEvent` from pipeline package
- Change `runReview` to return `*store.Review` for publish message creation
- Add `store.GetReview(id)` method

**Part of:** #298 (epic: embed NATS in backend)  
**Closes:** #305

## Test plan

- [ ] `go test ./internal/worker/ -v` — 6 tests pass (3 review + 3 publish: ack/nak/panic)
- [ ] `go test ./internal/bus/ -run TestPRPublishPublisher -v` — publish + dedup tests pass
- [ ] `go test ./... -count=1` — full suite passes, no regressions
- [ ] Smoke test: review completes → publish worker enqueues → GitHub submit via NATS

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #306, opened 2026-04-23
feat(nats): task 7: issue triage worker (NATS consumer)


Muriano:
Subscribe to `heimdallm.issue.triage`. Calls existing issues.Pipeline.Run() with ReviewOnly mode. Publish events.issue_review_completed.

Blocked by: task 4
Part of #298

Issue #298, opened 2026-04-23
epic: embed NATS in backend — event-driven pipeline replacing tier-based polling


Muriano:
## Overview

Replace the current tier-based synchronous polling architecture with an embedded NATS server. Each polling cycle publishes events to NATS subjects, and async consumers handle the actual work (reviews, triage, implementation, state updates). Concurrency is globally configurable.

## Motivation

The current architecture has several structural problems:
- **Unbounded goroutine creation**: Tier 2 spawns a goroutine per PR with no semaphore (#242 root cause)
- **Tight coupling**: Tiers communicate via Go channels and shared state (mutex-protected pointers), making the flow hard to trace and debug
- **No durability**: If the daemon restarts mid-review, in-progress work is lost
- **SSE bottleneck**: The in-process SSE broker has a 10-connection limit and zombie connections kill notifications (#286)
- **Sequential issue processing**: Issues are processed repo-by-repo in a for-loop; a slow repo blocks all others
- **No backpressure**: Rate limiter gates API calls but doesn't limit work queue depth

### Current (Tier-based)
\`\`\`
Tier 1 (discovery) → channel → Tier 2 (poll + process) → WatchQueue → Tier 3 (watch)
                                    ↓ goroutines (unbounded)
                                Pipeline.Run() / Issues.Run()
\`\`\`

### Proposed (NATS event-driven)
\`\`\`
Pollers (discovery, PR poll, issue poll, state check)
    ↓ publish events
NATS embedded server (JetStream)
    ↓ subscribe + consume
Workers (review, triage, implement, state update)
    ↓ configurable concurrency (global worker pool)
\`\`\`

### NATS Subjects

| Subject | Producer | Consumer | Payload |
|---------|----------|----------|---------|
| \`heimdallm.discovery.repos\` | Discovery poller | Repo list updater | \`["org/repo1", ...]\` |
| \`heimdallm.pr.review\` | PR poller | Review worker | \`{repo, number, github_id, head_sha}\` |
| \`heimdallm.pr.publish\` | Review worker | Publish worker | \`{review_id}\` |
| \`heimdallm.issue.triage\` | Issue poller | Triage worker | \`{repo, number, github_id}\` |
| \`heimdallm.issue.implement\` | Issue poller / promote | Implement worker | \`{repo, number, github_id}\` |
| \`heimdallm.state.check\` | State poller | State worker | \`{type, repo, number, github_id}\` |
[...]

PR #319, opened 2026-04-23
feat(nats): issue triage worker + Fetcher NATS dispatch (#306)

Muriano:
## Summary

- Refactor `Fetcher.ProcessRepo` to publish classified issues to NATS instead of calling `pipeline.Run` directly
- Add `IssuePublisher` interface with `PublishIssueTriage` / `PublishIssueImplement`
- Add `NATSIssuePublisher` implementation with dedup by github_id
- Create `TriageWorker` consumer for `heimdallm.issue.triage` — fetches issue from GitHub, resolves per-repo config, runs in ReviewOnly mode
- Backward compatible: when publisher is nil, Fetcher falls back to direct pipeline.Run
- Wire publisher on Fetcher + triageHandler + TriageWorker in main.go

**Closes:** #306

**Stacks on:** #318 (PR publish worker)

## Test plan

- [ ] `go test ./internal/bus/ -run TestIssuePublisher -v` — triage + implement publish tests pass
- [ ] `go test ./internal/issues/ -count=1` — existing tests pass (publisher not set, fallback path)
- [ ] `go test ./internal/worker/ -run TestTriageWorker -v` — consumer test passes
- [ ] Smoke test: issues with triage labels trigger `triage-worker: processing` in logs

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -433,6 +433,8 @@ func main() {
 	// ── Multi-tier Pipeline ──────────────────────────────────────────────
 	js := eventBus.JetStream()
 	publishPub := bus.NewPRPublishPublisher(js)
+	issuePublisher := bus.NewIssuePublisher(js)
+	issueFetcher.SetPublisher(issuePublisher)
 
 	// tier2Adapter bridges main.go's concrete types to the Pipeline's
 	// Tier 2 / Tier 3 interfaces.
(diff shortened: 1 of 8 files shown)


sergiotejon:
## ⚠️ MEDIUM Issue

Dedup msgID uses only githubID (`issue-triage:<id>`). If the same issue is re-opened or updated and should be re-triaged, the dedup window will silently drop it. Consider including a timestamp or event-specific identifier.

**Location:** `daemon/internal/bus/publisher.go` line 111

---

sergiotejon:
## ⚠️ MEDIUM Issue

Messages are always acked even when the handler fails (pipeline error). If the NATS consumer has a reasonable MaxDeliver, nacking on handler failure would enable automatic retries for transient errors (e.g. GitHub API rate limits).

**Location:** `daemon/internal/worker/triage.go` line 67

---

sergiotejon:
## 🟡 LOW Issue

Config is read under cfgMu.Lock() but aiCfg (a value copy) is mutated after unlock (LocalDir assignment on line 669). This is safe today since it's a local copy, but the pattern is fragile — a future refactor returning a pointer would introduce a race.

**Location:** `daemon/cmd/heimdallm/main.go` line 659

---

sergiotejon:
## 🟡 LOW Issue

triageHandler duplicates ~50 lines of config resolution and RunOptions construction that likely exists elsewhere in main.go for the direct-pipeline path. Consider extracting a shared helper to reduce drift risk.

**Location:** `daemon/cmd/heimdallm/main.go` line 641

---

sergiotejon:
## 🟡 LOW Issue

Test relies on time.Sleep(200ms) for worker startup and time.Sleep(500ms) for message delivery. This is flaky under CI load — consider polling with a short deadline instead.

**Location:** `daemon/internal/worker/triage_test.go` line 33

---

sergiotejon (approved):

The PR is a clean NATS-based issue triage worker implementation. Previous findings were about Flutter files not in this diff. The Go code is well-structured with minor concerns around dedup key design and always-ack semantics.

### Suggestions

- Consider nacking failed messages instead of always acking, so transient failures (GitHub API errors) get retried automatically up to MaxDeliver.
- Extract the config-resolution + RunOptions construction in main.go into a shared helper to avoid duplication between the triage handler and the direct-pipeline path.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

PR adds a NATS-based issue triage worker and refactors the Fetcher to publish classified issues to JetStream instead of invoking the pipeline directly. The implementation is well-structured with decent test coverage, but has a few real concerns around dedup semantics, error handling (always-ack), duplicated config-resolution logic in main.go, and flaky time.Sleep-based tests. Existing reviewer comments correctly identify the main issues.

### Issues

⚠️ **daemon/internal/bus/publisher.go:111** — Dedup msgID uses only githubID (issue-triage:<id>). If an issue is re-opened or updated and should be re-triaged, the JetStream dedup window will silently drop the re-publish. Include issue.UpdatedAt (or a content-version hash / event identifier) in the msgID so legitimate re-classifications are not deduped away.
⚠️ **daemon/internal/worker/triage.go:67** — Messages are unconditionally Ack()'d even when the handler fails (e.g. GetIssue returns an error or the pipeline fails). Transient failures (GitHub rate limits, network blips) are then lost with no retry. Prefer msg.Nak() / NakWithDelay() on transient errors and rely on the consumer's MaxDeliver for bounded retries; Ack() should only happen on success or on permanent/poison messages (decode errors, panics).
🟡 **daemon/cmd/heimdallm/main.go:641** — triageHandler duplicates ~50 lines of config resolution and RunOptions construction that already exists for the direct-pipeline path. Extract a shared helper (e.g. buildIssueRunOptions(repo)) to prevent drift — otherwise the two code paths will silently diverge when config fields are added.
🟡 **daemon/cmd/heimdallm/main.go:659** — aiCfg.LocalDir is mutated after cfgMu.Unlock(). Safe today because AIForRepo returns a value copy, but fragile: a future refactor returning a pointer would introduce a data race. Either do the ResolveLocalDir inside the lock or add a comment documenting the value-copy invariant.
🟡 **daemon/internal/worker/triage_test.go:33** — Test uses time.Sleep(200ms) for worker startup and time.Sleep(500ms) for message delivery. Flaky under CI load — replace with polling (e.g. require.Eventually / a loop with short deadline) on len(received) and a sync signal for worker readiness.
🟡 **daemon/cmd/heimdallm/main.go:716** — triageWCtx uses context.Background() with a deferred cancel. On shutdown the worker's Start will return (iter.Stop), but the in-flight triageHandler run (which can be long — a full pipeline execution) is not cancelled because the ctx passed to handler is the same triageWCtx that gets cancelled only via defer. Verify ordering: ensure the main shutdown flow cancels triageWCtx before waiting on in-flight work, and consider passing a graceful-shutdown ctx distinct from the consume-loop ctx.
🟡 **daemon/internal/issues/fetcher.go:92** — NewFetcher still takes PipelineRunner as a required param, and SetPublisher toggles behavior after construction. Mixing a required pipeline with an optional publisher (that fully replaces it) makes the Fetcher's contract unclear. Consider either making pipeline optional (both nil-checked) or constructing two distinct Fetchers, so misconfiguration is a compile-time error rather than a silent no-op branch.
🟡 **daemon/internal/worker/triage.go:37** — PullMaxMessages(1) caps concurrency to a single in-flight triage at a time per worker. Given triage can invoke long-running pipelines, this may be intentional, but it should be a documented/configurable choice rather than a hard-coded 1.

### Suggestions

- Change msgID to include issue.UpdatedAt or a content hash so legitimate re-triages are not deduped (e.g. fmt.Sprintf("issue-triage:%d:%d", githubID, updatedAtUnix)).
- Nak transient handler failures and only Ack on success or permanent errors; rely on the consumer's MaxDeliver for bounded retries and a DLQ for poison messages.
- Extract the config-resolution + RunOptions construction in triageHandler into a shared helper used by both the direct pipeline path and the NATS handler.
- Replace time.Sleep-based synchronization in triage_test.go with polling (require.Eventually or a channel the handler signals on).
- Make TriageWorker concurrency (PullMaxMessages) configurable and document the expected concurrency model.
- Clarify Fetcher's pipeline-vs-publisher contract — either make both optional or split into two types — so the runtime branch can't silently skip work.

---

PR #320, opened 2026-04-23
feat(nats): issue implement worker (#307)

Muriano:
## Summary

- Create `ImplementWorker` NATS consumer for `heimdallm.issue.implement`
- Handler fetches issue from GitHub, sets Mode=Develop, resolves per-repo config, runs `issuePipe.Run`
- Same pattern as TriageWorker (Task 7) — panic recovery, always-ack, cancellable context
- Wire implementHandler + ImplementWorker startup in main.go

**Part of:** #298 (epic: embed NATS in backend)  
**Closes:** #307

**Stacks on:** #319 (issue triage worker)

## Test plan

- [ ] `go test ./internal/worker/ -run TestImplementWorker -v` — consumer test passes
- [ ] `go test ./... -count=1` — full suite passes
- [ ] Smoke test: develop-labeled issues trigger `implement-worker: processing`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #308, opened 2026-04-23
feat(nats): task 9: state check poller → NATS


Muriano:
Refactor Tier 3 watch cycle to publish `heimdallm.state.check` per item. Consumer checks GitHub state and publishes state_changed events. Replace WatchQueue with NATS KV for backoff state (survives restart).

Blocked by: task 2
Part of #298

PR #321, opened 2026-04-23
feat(nats): state check with NATS KV — replaces WatchQueue (#308)

Muriano:
## Summary

- Replace in-memory WatchQueue with durable NATS KV bucket `HEIMDALLM_WATCH`
- `WatchKV` wrapper: Enroll, Get, ResetBackoff, IncreaseBackoff, ScanReady, EvictStale
- State poller goroutine: scans KV every 30s, publishes `StateCheckMsg` for ready items
- `StateWorker` consumer: checks GitHub state via existing CheckItem/HandleChange, updates KV backoff (reset on change, double on no change)
- Replace `pipe.Queue().Push()` in reviewHandler with `WatchKV.Enroll()`
- Watch state now survives daemon restarts (KV persisted to disk)

**Closes:** #308

**Stacks on:** #320 (issue implement worker)

## Test plan

- [ ] `go test ./internal/bus/ -run TestWatchKV -v` — 6 KV CRUD/scan/evict tests pass
- [ ] `go test ./internal/bus/ -run TestStateCheckPublisher -v` — publisher test passes
- [ ] `go test ./internal/worker/ -run TestStateWorker -v` — consumer test with backoff verification
- [ ] Smoke test: PR review → KV enroll → state poller publishes → state worker checks

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -845,6 +845,9 @@ func main() {
 		entry, err := eventBus.WatchKV().Get(ctx, key)
 		if err == nil {
 			item.LastSeen = entry.LastSeen
+		} else {
+			slog.Warn("state-handler: KV get failed, using zero LastSeen",
+				"key", key, "err", err)
 		}
 
 		changed, snap, err := adapter.CheckItem(ctx, item)
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -538,13 +538,11 @@ func main() {
 			}
 		}
 
-		// Maintain Tier 3 watching (Task 9 replaces WatchQueue entirely).
-		cfgMu.Lock()
-		q := pipe.Queue()
-		cfgMu.Unlock()
-		q.Push(&scheduler.WatchItem{
-			Type: "pr", Repo: pr.Repo, Number: pr.Number, GithubID: pr.ID,
-		})
+		// Enroll for state watching via NATS KV.
+		if err := eventBus.WatchKV().Enroll(ctx, "pr", pr.Repo, pr.Number, pr.ID); err != nil {
+			slog.Warn("review-worker: failed to enroll watch",
+				"repo", pr.Repo, "pr", pr.Number, "err", err)
+		}
 	}
 
 	reviewWorker := worker.NewReviewWorker(eventBus.JetStream(), reviewHandler)
@@ -791,6 +789,83 @@ func main() {
 	go func() {
 		if err := implementW.Start(implementWCtx); err != nil {
 			slog.Error("implement worker stopped", "err", err)
+		}
+	}()
+
+	// ── State check poller ──────────────────────────────────────────────
+	// Scans the NATS KV watch bucket every 30s and publishes StateCheckMsg
+	// for items due for a state check. Replaces the in-memory WatchQueue.
+	stateCheckPub := bus.NewStateCheckPublisher(js)
+	statePollerCtx, statePollerCancel := context.WithCancel(context.Background())
+	defer statePollerCancel()
+	go func() {
+		ticker := time.NewTicker(30 * time.Second)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-statePollerCtx.Done():
+				return
+			case <-ticker.C:
+				watchKV := eventBus.WatchKV()
+				if evicted, err := watchKV.EvictStale(statePollerCtx); err != nil {
+					slog.Warn("state-poller: evict failed", "err", err)
+				} else if evicted > 0 {
+					slog.Debug("state-poller: evicted stale items", "count", evicted)
+				}
+
+				ready, err := watchKV.ScanReady(statePollerCtx)
+				if err != nil {
+					slog.Warn("state-poller: scan failed", "err", err)
+					continue
+				}
+				for _, entry := range ready {
+					if err := stateCheckPub.PublishStateCheck(statePollerCtx, entry.Type, entry.Repo, entry.Number, entry.GithubID); err != nil {
+						slog.Warn("state-poller: publish failed",
+							"type", entry.Type, "repo", entry.Repo, "number", entry.Number, "err", err)
+					}
+				}
+			}
+		}
+	}()
+
+	// ── NATS state check worker ─────────────────────────────────────────
+	// Consumes state check requests, calls GitHub API, updates KV backoff.
+	// Reuses the existing CheckItem/HandleChange logic from tier2Adapter.
+	stateHandler := func(ctx context.Context, msg bus.StateCheckMsg) (bool, error) {
+		item := &scheduler.WatchItem{
+			Type:     msg.Type,
+			Repo:     msg.Repo,
+			Number:   msg.Number,
+			GithubID: msg.GithubID,
+		}
+
+		// Read LastSeen from KV for the dedup check inside CheckItem.
+		// Key separator is "." (NATS KV doesn't allow ":").
+		key := fmt.Sprintf("%s.%d", msg.Type, msg.GithubID)
+		entry, err := eventBus.WatchKV().Get(ctx, key)
+		if err == nil {
+			item.LastSeen = entry.LastSeen
+		}
+
+		changed, snap, err := adapter.CheckItem(ctx, item)
+		if err != nil {
+			return false, err
+		}
+		if !changed {
+			return false, nil
+		}
+		if err := adapter.HandleChange(ctx, item, snap); err != nil {
+			return true, err
+		}
+		return true, nil
+	}
+
+	stateW := worker.NewStateWorker(js, eventBus.WatchKV(), stateHandler)
+	stateWCtx, stateWCancel := context.WithCancel(context.Background())
+	defer stateWCancel()
+	go func() {
+		if err := stateW.Start(stateWCtx); err != nil {
+			slog.Error("state worker stopped", "err", err)
 		}
 	}()
 
--- a/daemon/internal/bus/bus.go
+++ b/daemon/internal/bus/bus.go
@@ -21,10 +21,11 @@ type Config struct {
 
 // Bus wraps an embedded NATS server with a JetStream-enabled client.
 type Bus struct {
-	server *natsserver.Server
-	conn   *nats.Conn
-	js     jetstream.JetStream
-	cfg    Config
+	server  *natsserver.Server
+	conn    *nats.Conn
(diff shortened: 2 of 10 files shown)


sergiotejon:
## ⚠️ MEDIUM Issue

ResetBackoff/IncreaseBackoff have read-modify-write races: two concurrent calls on the same key can clobber each other. Consider using KV revision-based CAS (Update with last revision) instead of Get+Put.

**Location:** `daemon/internal/bus/watch.go` line 88

---

sergiotejon:
## 🟡 LOW Issue

json.Marshal error is silently ignored (`data, _ := json.Marshal(entry)`). Same in IncreaseBackoff and ForceUpdate.

**Location:** `daemon/internal/bus/watch.go` line 100

---

sergiotejon:
## 🟡 LOW Issue

IncreaseBackoff/ResetBackoff errors are silently discarded after handler completes. At minimum log them.

**Location:** `daemon/internal/worker/state.go` line 82

---

sergiotejon:
## ⚠️ MEDIUM Issue

stateHandler reads KV entry by key but silently ignores all errors including non-NotFound errors. A transient KV failure would result in a zero LastSeen, potentially causing a false 'changed' detection.

**Location:** `daemon/cmd/heimdallm/main.go` line 841

---

sergiotejon:
## 🟡 LOW Issue

ScanReady iterates all keys sequentially with individual Gets. This is O(n) round-trips and will degrade as the bucket grows. Consider using a KV watcher or ListKeys with batch fetch.

**Location:** `daemon/internal/bus/watch.go` line 130

---

sergiotejon (approved):

NATS KV watch implementation is solid. Key separator inconsistency from prior review is fixed. A few minor issues remain around race conditions in read-modify-write KV operations and missing error handling.

### Suggestions

- Use KV Update (CAS) with revision in ResetBackoff/IncreaseBackoff to prevent concurrent write races
- Distinguish jetstream.ErrKeyNotFound from other errors in stateHandler's KV Get to avoid masking transient failures

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid replacement of the in-memory WatchQueue with a durable NATS KV-backed watch store, poller, and worker. Tests cover the core KV operations and worker flow. Main concerns are concurrency safety on KV updates (read-modify-write races), silently-swallowed errors in several places, and a couple of correctness/robustness issues in the state handler and poller lifecycle.

### Issues

⚠️ **daemon/internal/bus/watch.go:88** — ResetBackoff and IncreaseBackoff perform Get+Put without CAS. Concurrent invocations (e.g., poller re-publishing while worker processes, or multiple workers) can clobber each other and lose backoff progression or LastSeen updates. Use kv.Update(key, data, revision) with a retry loop on jetstream.ErrKeyExists/wrong-revision to make the update atomic.
⚠️ **daemon/cmd/heimdallm/main.go:841** — stateHandler calls eventBus.WatchKV().Get(ctx, key) and treats any error as 'entry absent', leaving item.LastSeen as zero. For a transient KV error this makes CheckItem see LastSeen=0 and likely report a false 'changed', triggering spurious HandleChange side effects. Only tolerate jetstream.ErrKeyNotFound; on other errors, log and return err so the message is redelivered.
🟡 **daemon/cmd/heimdallm/main.go:795** — The state-check poller uses context.Background() with its own cancel rather than the main daemon context. On shutdown it will only stop via the deferred statePollerCancel(), but if any earlier defer blocks or main returns via os.Exit, the poller can continue publishing against a shutting-down JetStream. Prefer deriving from the daemon's root ctx and ordering shutdown so the poller stops before the bus. Also, if stateCheckPub.PublishStateCheck fails mid-batch, the loop keeps publishing the remaining ready items — acceptable, but also consider that a slow GitHub/JetStream response under 30s cadence can cause duplicate in-flight checks for the same item (no dedup on the consumer side).
🟡 **daemon/internal/worker/state.go:82** — Return values of watchKV.IncreaseBackoff / ResetBackoff are discarded. On KV error the backoff silently fails to advance, which can cause the same item to be republished every 30s indefinitely. At minimum log the error; ideally surface it so ops can detect KV health issues.
🟡 **daemon/internal/worker/state.go:40** — StateWorker.Start uses PullMaxMessages(1), effectively serializing all state checks through a single goroutine. Given GitHub API calls can take seconds, this limits throughput. Consider making concurrency configurable or at least documenting the choice.
🟡 **daemon/internal/worker/state.go:68** — After a decode error the message is Ack'd and skipped, but the KV entry is never updated, so the poller will keep re-publishing the same malformed message (until the publisher stops producing it). Since messages are produced by the poller from KV entries this is unlikely in practice, but worth a comment or a TermMessage instead of Ack for poison-pill semantics.
🟡 **daemon/internal/bus/watch.go:100** — json.Marshal errors are discarded in ResetBackoff, IncreaseBackoff, and ForceUpdate (`data, _ := json.Marshal(entry)`). Marshalling a WatchEntry cannot realistically fail, but swallowing the error violates the project's handling pattern used in Enroll and masks future struct changes that might introduce a non-marshalable field.
🟡 **daemon/internal/bus/watch.go:130** — ScanReady does Keys() + per-key Get(), which is O(n) round-trips every 30s. Fine for small watch sets but will degrade as the bucket grows. Consider using jetstream.KeyValue.ListKeys or a KV watcher with a local cache refreshed on deltas.
🟡 **daemon/internal/bus/watch.go:41** — Enroll unconditionally overwrites an existing entry, resetting backoff and LastSeen. If a PR is re-enrolled by a retry path after already being watched with a long backoff, this will reset progress. Consider Create-if-absent (kv.Create) or a check-and-skip for existing keys, unless reset-on-reenroll is intentional — if so, a comment would help.
🟡 **daemon/internal/worker/state_test.go:29** — Test relies on time.Sleep for synchronization, which is flaky under load. Prefer a channel signal from the handler to deterministically wait for the call before cancelling.

### Suggestions

- Switch ResetBackoff/IncreaseBackoff to CAS using kv.Update(key, data, revision) with a small retry loop; the revision is available from kv.Get's KeyValueEntry.Revision().
- In stateHandler, distinguish jetstream.ErrKeyNotFound from other errors; return non-NotFound errors so the message is NAK'd/redelivered rather than running CheckItem with LastSeen=0.
- Log (and optionally surface via metrics) errors from IncreaseBackoff/ResetBackoff in the state worker so silent KV failures don't cause republish loops.
- Derive the state poller and state worker contexts from the daemon's root context so shutdown ordering is deterministic; stop the poller before the bus.
- Propagate json.Marshal errors in ResetBackoff/IncreaseBackoff/ForceUpdate for consistency with Enroll.
- Consider a dedup/in-flight guard (e.g., short-lived KV lock or skipping items whose NextCheck was just advanced) so a slow state check doesn't queue duplicates on the next 30s tick.
- Replace time.Sleep-based synchronization in state_test.go with a done channel from the handler for reliability.

---

Muriano:
## Respuesta a los comentarios de review

Fixes en `2699d53`.

### ✅ Arreglados

**json.Marshal error ignorado** (@sergiotejon 🟡) → Ahora todos los Marshal retornan error envuelto.

**KV errors no logueados en StateWorker** (@sergiotejon 🟡) → IncreaseBackoff/ResetBackoff errors ahora se logean con warning.

**stateHandler ignora KV Get errors** (@sergiotejon ⚠️) → Ahora logea warning cuando Get falla, explicando que usa zero LastSeen como fallback.

### 📝 Anotados

**Read-modify-write race** (@sergiotejon ⚠️) — El state-worker es single-consumer (MaxAckPending=Workers*2) y el poller solo lee. El único race posible es enrollment simultáneo con un check, que es improbable y el efecto es benigno (un backoff reset perdido). CAS via KV revision es la solución correcta pero añade complejidad para un race teórico. Anotado para futuro.

**ScanReady O(n)** (@sergiotejon 🟡) — En práctica <100 items watched simultáneamente. Un watcher o batch fetch es optimización prematura.

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of #321. Author's commit 2699d53 addresses the visibility gaps (json.Marshal errors wrapped, StateWorker KV errors logged, stateHandler KV Get error logged). The two annotated items (RMW race without CAS; O(n) ScanReady) remain intentionally deferred with reasonable justification given single-consumer state worker and small watch-set size. The stateHandler change logs but still falls back to zero LastSeen on any KV error rather than distinguishing ErrKeyNotFound from transient errors — this is a partial fix: now observable, but a transient KV blip can still drive a false 'changed' on the next check. Overall: acceptable to merge; remaining items are LOW.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:841** — stateHandler now logs on KV Get error (good), but still unconditionally uses zero LastSeen as fallback. A transient KV error will still cause CheckItem to see LastSeen=0 and likely emit a false 'changed', triggering spurious HandleChange side effects. Prefer: tolerate only jetstream.ErrKeyNotFound with zero-LastSeen fallback; on other errors return err so the message is redelivered. Partial fix from previous review.
🟡 **daemon/internal/bus/watch.go:88** — ResetBackoff/IncreaseBackoff still use Get+Put without CAS. Author's rationale (single state-consumer, benign lost-reset on enrollment race) is defensible but not enforced by code — if MaxAckPending or concurrency ever increases, silent backoff clobbering returns. Consider at minimum a TODO/comment at the call sites or an assertion on the consumer's expected single-flight semantics so the invariant is discoverable.
🟡 **daemon/internal/bus/watch.go:41** — Enroll still unconditionally overwrites an existing entry (not mentioned by author). A re-enroll from a retry path resets backoff and LastSeen on an item already being watched with a long backoff, undoing progress. If reset-on-reenroll is intentional, a one-line comment clarifying would help; otherwise prefer kv.Create + skip-if-exists.
🟡 **daemon/internal/worker/state.go:40** — PullMaxMessages(1) still serializes all state checks through one goroutine. With GitHub calls taking seconds this caps throughput; worth documenting the choice or making it configurable before the watch set grows.

### Suggestions

- Split the stateHandler KV Get: if errors.Is(err, jetstream.ErrKeyNotFound) log+proceed with zero LastSeen; otherwise return the error so NATS redelivers.
- Add a short comment at ResetBackoff/IncreaseBackoff noting the single-consumer assumption so the lack of CAS is intentional and discoverable in code review.
- Consider kv.Create in Enroll (or explicitly document overwrite semantics) to avoid resetting backoff progress on retry-driven re-enrollment.
- When revisiting ScanReady under load, prefer a KV watcher maintaining an in-memory min-heap keyed on NextCheck over Keys()+per-key Get().

---

PR #324, opened 2026-04-24
fix(daemon): silence PR SHA-skip side effects + respect re-request review (closes #322 bugs 3-5)

### LINKED ISSUE #322: bug: PR re-review side effects — notify/activity-log spam, re-request ignored, locked-PR retry loop
## Overview

Originally tracked two unrelated retry-loop bugs. Bug 2 was fixed in #296 (mergeado el 24/04 07:37 CEST). Scope now expanded to cover three additional PR-side defects observed in production logs on 2026-04-24 against `freepik-company/ai-api-specs#557`.

Status snapshot:

| Bug | Status |
|---|---|
| 1. PublishPending retries 422 "lock prevents review" forever | OPEN |
| 2. Issue re-triage every poll cycle (marker scan 404) | **FIXED** by #296 |
| 3. Notify "PR Review Started" spam every poll on stable PRs | OPEN — added 2026-04-24 |
| 4. Activity log shows N false "high review" entries per poll on stable PRs | OPEN — added 2026-04-24 |
| 5. Re-request review explícito ignorado mientras HEAD SHA no cambie | OPEN — added 2026-04-24 |

---

## Bug 1: PublishPending retries locked PR forever (review_id=681)

**Symptom:** Every ~1 minute:
```
WARN pipeline: retry publish failed review_id=681 err="github: submit review: status 422: lock prevents review"
```

**Root cause:** `PublishPending()` treats all GitHub errors as transient. HTTP 422 "lock prevents review" is a **permanent** error — the PR's conversation is locked (merged/closed with lock). Retrying will never succeed.

**Impact:** Burns 1 GitHub API call per poll cycle indefinitely. The review will never be published but is never marked as permanently failed.

**Fix:** In `PublishPending()` (and the new NATS publish worker), treat 422 as a permanent failure — mark the review as orphaned (`MarkReviewPublished(rev.ID, -1, "", ...)`) and stop retrying. Specifically, 422 with "lock prevents review" means the PR was locked after the review was stored but before it could be published.

---

## Bug 2: Issue #144 re-triaged every poll cycle (~2 min) — **FIXED by #296**

**Symptom (historical):** Every poll cycle:
```
WARN issues fetcher: marker scan failed, falling through to dedup checks
  repo=freepik-company/fc-vertexai-api number=144
  err="github: fetch review comments: status 404"
...
INFO issues pipeline: triage complete repo=freepik-company/fc-vertexai-api number=144 severity=medium posted=true
```

**Root cause (historical, two layers):**
1. **Wrong API endpoint for marker scan:** The marker scanner called `FetchComments` which used the Pull Request review comments API (`/repos/{repo}/pulls/{number}/comments`). Issue #144 is an **issue**, not a PR — this endpoint returned 404. The scanner fell through to dedup checks.
2. **Dedup check not catching re-triage:** After the marker scan failed, `alreadyProcessed()` should still detect that the issue was already triaged via the `LatestIssueReview` + grace window check. However, the triage pipeline itself bumped `updated_at` when it posted its comment, which pushed the issue past the `RecomputeGrace` window (30s) on the next poll cycle.

**Resolution:** Fixed by #296 (mergeado 2026-04-24 07:37 CEST):
- New `FetchIssueCommentsOnly` method that bypasses the PR-only endpoint for issue numbers.
- Persistent in-flight claim for issue triage keyed on `(github_issue_id, updated_at)` mirroring #258.
- SQLite circuit breaker with caps per issue/24h (3) and per repo/hour (10), mirroring #246.

Closing this sub-bug. Bug 1 above remains.

---

## Bug 3: Notify "PR Review Started" desktop spam every poll on stable PRs

**Symptom:** Daemon emits a desktop notification every `PollInterval` (60 s) for any PR whose `updated_at` is being bumped (CI bots, cross-references) but whose HEAD SHA has not changed. The actual review is correctly skipped by the `#245` fail-closed SHA check, but the notify fires *before* that check.

Live log on 2026-04-24 against PR `freepik-company/ai-api-specs#557`:
```
07:44:41 starting review pr=557
07:44:41 notify "PR Review Started" #557           ← notify dispara
07:44:42 skipping re-review, HEAD SHA unchanged    ← review se ataja
07:44:42 review done pr=557 severity=high github_review_id=4164377925
... y se repite a 07:45, 07:46, 07:47, 07:48, 07:49, 07:50, 07:51, 07:52, 07:53 (10 notificaciones para una sola review estable)
```

**Root cause:** `daemon/internal/pipeline/pipeline.go:222` — `p.notify.Notify("PR Review Started", ...)` se dispara incondicionalmente al entrar a `Run`, *antes* del SHA fail-closed check (líneas 276–280).

**Impact:** UX-only (no money burned). Operator desktop is spammed once per minute per stable PR. With multiple stable PRs in flight the daemon can produce dozens of phantom notifications per hour.

**Fix:** Move the `notify("PR Review Started", ...)` call from line 222 to *after* the SHA-skip check (line 280), so it only fires when a real review is about to be dispatched.

---

## Bug 4: Activity log records false "review" entries every poll on stable PRs

**Symptom:** Activity log UI shows N rows of "high review" for the same PR — one per poll cycle — when in reality only the original review was performed and all subsequent entries are skip-paths. Operator-visible spam.

Live evidence on 2026-04-24:
- PR `freepik-company/ai-api-specs#557` was reviewed once on 2026-04-23 17:11 (`CHANGES_REQUESTED`, `github_review_id=4164377925`).
- Daemon restarted 2026-04-24 07:44.
- Activity log between 07:44:42 and 07:53:39 shows **10 separate "high review" entries** for #557, all referencing the same `github_review_id=4164377925`.
- Zero new Claude calls were made — the SHA fail-closed check skipped every one.

**Root cause:** Two-layer wiring bug.
1. `daemon/internal/pipeline/pipeline.go:265–279` — both the legacy-backfill path (line 274) and the SHA-skip path (line 279) return `(prevReview, nil)`, i.e. the *previous* review object with `nil` error.
2. `daemon/cmd/heimdallm/main.go:399–426` (`runReview` wrapper) — the defensive `if rev == nil { return }` at line 417 is the only thing that suppresses the SSE `EventReviewCompleted` emission. Because pipeline.Run returned `prevReview` (not `nil`), the wrapper proceeds to emit `EventReviewCompleted` with the previous review's severity and ID.
3. `daemon/internal/activity/recorder.go:68` — the activity recorder subscribes to `EventReviewCompleted` and inserts a row into `activity_log` for every event received, regardless of whether the review is genuinely new.

Net effect: every poll cycle on a stable PR produces one fake `activity_log` row, one fake desktop notification (Bug 3), and one fake SSE event consumed by any other UI subscribers.

**Fix:** In both branches (lines 274 and 279), return `(nil, nil)` instead of `(prevReview, nil)`. The caller's existing defensive check (`if rev == nil { return }`) already handles this exit cleanly without emitting completed/error events. The skip itself is still logged inside `pipeline.Run` (`slog.Info("pipeline: skipping re-review, HEAD SHA unchanged")`) so visibility is preserved.

Note: callers do not consume `prevReview` for any other purpose on these paths — the only downstream uses are the `slog.Info("review done")` line and the `EventReviewCompleted` emission, both of which we explicitly want to suppress on a skip.

---

## Bug 5: Explicit re-request review is silently ignored while HEAD SHA is unchanged

**Symptom:** Operator presses **"Re-request review"** in the GitHub UI for a PR whose HEAD SHA hasn't moved. GitHub re-adds the bot to `requested_reviewers` and emits a `review_requested` timeline event. Tier 2 sees the PR re-listed by Search Issues. `runReview` is called. `pipeline.Run` runs the SHA fail-closed check (`#245`), notes `prevReview.HeadSHA == pr.Head.SHA`, and skips. **The user's explicit request is silently dropped.**

**Severity:** This is the **functional** bug in the trio. The previous two are cosmetic UX/log issues; this one means the daemon does not respond to a direct user instruction.

**Why this happens:** The `#245` fail-closed check exists to defend against `#243`-style cost runaways (linked-PR cross-refs, CI bots, dependabot bumps). It conflates *all* sources of `updated_at` bumps into a single decision: same SHA → skip. There is no distinction between "spurious bump" and "explicit human asking for a re-review".

**Legitimate use cases for re-request on the same SHA:**
- Operator dismissed the bot's previous review and wants a fresh one.
- Prompt config changed (different agent, different model) and operator wants a re-test against the same code.
- Operator wants a second opinion under a different angle.

**Fix plan:**
1. Add `GetPRTimelineEvents(repo, number)` to `github.Client` returning at minimum `{event, actor, created_at}` from `GET /repos/{repo}/issues/{number}/timeline` filtered to `review_requested` and `review_dismissed` for the bot login.
2. In `pipeline.Run`, before the SHA-skip return at line 279, query the timeline. If the most recent event for the bot login is a `review_requested` newer than `prevReview.CreatedAt` (and not superseded by a later `review_dismissed` followed by no fresh `review_requested`), bypass the SHA skip and proceed with the review.
3. Same `#245`-style fail-closed posture: if the timeline query fails, treat as "no re-request detected" and keep the skip — never widen the cost surface on a transient API error.

The PR-side circuit breaker (`#246`, 3/PR/24h, 20/repo/h) still caps the worst case if a misbehaving operator presses re-request 50 times in a row.

---

## Acceptance criteria

- [ ] Bug 1: `PublishPending` recognises 422 "lock prevents review" as permanent and marks the review as orphaned.
- [x] Bug 2: closed by #296.
- [ ] Bug 3: no desktop "PR Review Started" notification fires on a SHA-unchanged skip.
- [ ] Bug 4: no `activity_log` row, `EventReviewCompleted` SSE, or "review done" log line is emitted on a SHA-unchanged skip.
- [ ] Bug 5: pressing "Re-request review" in GitHub on a stable PR triggers a fresh review on the next poll cycle, even with the same HEAD SHA. The cost cap from `#246` still applies.
- [ ] All wired with `make test-docker` (no raw `go test` on host per team policy).
- [ ] Live verification on a real production PR after merge.

## Plan for this PR

Bugs 3, 4, 5 share the same code site (`pipeline.go:Run` SHA-skip path) so they will land together in a single PR. Bug 1 lives in `PublishPending` and is unrelated mechanically — left for a separate PR if a different teammate wants to take it, or appended to the same PR if no churn arises.

Self-assigning to @ivanmunozruiz.

ivanmunozruiz:
## Summary

Closes 3 of the 4 bugs tracked in #322 — the PR-side ones, all sharing the SHA-skip code site in \`pipeline.Run\`. Bug 1 (PublishPending 422 lock loop) lives in a different code path and stays in #322 for a separate PR.

| Bug | Symptom | Status in this PR |
|---|---|---|
| 3. Notify \"PR Review Started\" desktop spam every poll on stable PRs | Operator received a phantom desktop notification once per minute per stable PR | ✅ Fixed |
| 4. Activity log records false \"high review\" entries every poll on stable PRs | UI showed 10 fake reviews on PR #557 in 10 min — none were actually run | ✅ Fixed |
| 5. Re-request review explícito ignorado mientras HEAD SHA no cambie | The functional bug — daemon silently dropped the operator's explicit \"Re-request review\" action | ✅ Fixed |
| 1. PublishPending retries 422 \"lock prevents review\" forever | Independent code site (publish loop, not review loop) | Out of scope — left in #322 |

Self-assigned the issue to me before starting work.

## Commit 1 — \`11852b5\` Bugs 3 + 4

Two changes in \`pipeline.go\`:

1. **Move the unconditional notify out of the function entry.** \`p.notify.Notify(\"PR Review Started\", ...)\` was firing at the top of \`Run\`, before the SHA-skip dedup check. Moved to immediately AFTER the SHA-skip block so it only runs when a real review is about to be dispatched.
2. **Make SHA-skip and legacy-backfill paths return \`(nil, nil)\`** instead of \`(prevReview, nil)\`. The caller in \`cmd/heimdallm/main.go\` has a defensive \`if rev == nil { return }\` filter (line 417) that suppresses the \`EventReviewCompleted\` SSE / \`activity_log\` row / \`\"review done\"\` log line. Returning \`prevReview\` bypassed that filter — which is why the activity log on PR #557 showed 10 false \"high review\" entries in 10 minutes despite zero Claude credits being spent. The new contract matches the gate-skip path on line 219 (also \`(nil, nil)\`). The backfill side-effect (\`UpdateReviewHeadSHA\`) still runs.

Updated 3 existing tests for the new contract:
- \`TestPipeline_Run_SkipsReviewOnSameHeadSHA\`: now asserts \`rev2 == nil\` AND that \`fakeNotify\` events for \"PR Review Started\" total exactly 1 (one real review, one silent skip) — locks in Bug 3 and Bug 4.
- \`TestPipeline_Run_Tier3PathSkipsOnSameHeadSHA\`: same updated assertions for the Tier 3 watch path.
- \`TestRun_LegacyRowWithEmptyHeadSHAIsBackfilledAndSkipped\`: legacy backfill is now also a silent skip; verifies the side-effect via the store directly.

## Commit 2 — \`4cc9590\` Bug 5

The functional one. Adds an explicit-action bypass on top of the SHA fail-closed dedup.

**\`github/client.go\`**:
- New \`TimelineEvent\` type and \`GetPRTimelineEventsForReviewer(repo, number, login)\` method.
- Pages through \`GET /repos/:repo/issues/:n/timeline\`, filters to \`review_requested\` (where \`requested_reviewer.login == login\`) and \`review_dismissed\` (where \`dismissed_review.user.login == login\`), returns sorted ascending by created_at.
- 50-page hard cap to defend against malformed pagination responses.

**\`pipeline/pipeline.go\`**:
- Optional \`TimelineFetcher\` dependency on \`Pipeline\` + \`SetTimelineFetcher\` setter, mirroring \`SetCircuitBreakerLimits\`. Nil-safe — disables the bypass when unset, preserving pre-#322 behaviour for tests that don't care.
- New helper \`shouldBypassSHASkipForReReview\` encapsulates the decision: bypass iff the most recent timeline event for the bot login (newer than \`prevReview.CreatedAt\`) is a \`review_requested\`. A subsequent \`review_dismissed\` cancels the bypass (operator no longer wants our review).
- **Same fail-closed posture as #245**: a transient timeline API error keeps the SHA skip in place. The PR-side circuit breaker (\`#246\`, 3/PR/24h, 20/repo/h) still caps the worst case if a misbehaving operator presses re-request 50 times in a row.

**\`cmd/heimdallm/main.go\`**:
- Wires the production \`*github.Client\` as the \`TimelineFetcher\` at startup, right after \`SetCircuitBreakerLimits\`.

**Tests added**:
- 2 GitHub-client tests: filter-by-login behaviour with mixed payload + empty-login guard.
- 4 pipeline tests: happy-path bypass, stale-request rejection, dismiss-after-request rejection, fail-closed-on-timeline-error.

## Test plan

- [x] \`make test-docker\` green across all daemon packages
- [x] All 6 new tests pass; all 3 updated tests pass
- [ ] Live verification post-merge: re-request review on a stable PR → confirm fresh review is dispatched on next poll cycle. Bumps from CI bots / cross-refs alone (no re-request) → confirm no notify, no activity_log row, no Claude call.

## References

- Closes #322 bugs 3, 4, 5 (bug 1 left open for a separate PR)
- Builds on \`#245\` (HEAD SHA fail-closed dedup) — preserves the cost-runaway protection while honouring deliberate user actions
- Builds on \`#246\` (PR-side circuit breaker) — still the hard cap if re-request is abused
- Live evidence on PR \`freepik-company/ai-api-specs#557\` on 2026-04-24

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -191,6 +191,13 @@ func main() {
 	}
 	p.SetCircuitBreakerLimits(&cbLimits)
 
+	// Wire the GitHub client as the timeline fetcher so the SHA-skip
+	// path can detect explicit re-request review actions and bypass the
+	// dedup. See theburrowhub/heimdallm#322 Bug 5. Requires the bot
+	// login resolved below; the pipeline no-ops the bypass if either
+	// p.timeline or p.botLogin is unset.
+	p.SetTimelineFetcher(ghClient)
+
 	// Issue-side circuit-breaker caps (theburrowhub/heimdallm#292) — mirrors
 	// the PR-side defenses against runaway triage loops.
 	issueCBLimits := store.IssueCircuitBreakerLimits{
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -198,6 +198,14 @@ func main() {
 	// p.timeline or p.botLogin is unset.
 	p.SetTimelineFetcher(ghClient)
 
+	// Wire the SSE broker as the lifecycle publisher so Run emits
+	// pr_detected / review_started / review_completed / review_skipped
+	// at the correct semantic point (after the SHA-skip + gate
+	// decisions). The caller used to publish these blindly at function
+	// entry, leaving Flutter spinners colgados on every SHA-skip and
+	// firing phantom desktop notifications. See #322 Bugs 3+4.
+	p.SetPublisher(broker)
+
 	// Issue-side circuit-breaker caps (theburrowhub/heimdallm#292) — mirrors
 	// the PR-side defenses against runaway triage loops.
 	issueCBLimits := store.IssueCircuitBreakerLimits{
@@ -401,8 +409,13 @@ func main() {
 		slog.Info("pipeline: reviewing PR",
 			"repo", pr.Repo, "number", pr.Number, "github_id", pr.ID, "title", pr.Title)
 
-		broker.Publish(sse.Event{Type: sse.EventPRDetected, Data: sseData(map[string]any{"pr_number": pr.Number, "repo": pr.Repo})})
-		broker.Publish(sse.Event{Type: sse.EventReviewStarted, Data: sseData(map[string]any{"pr_number": pr.Number, "repo": pr.Repo})})
+		// Lifecycle SSEs (pr_detected, review_started, review_completed,
+		// review_skipped) are published from within p.Run via its
+		// Publisher dependency — this caller only handles the error
+		// paths because they need contextual error data the pipeline
+		// doesn't pre-shape (the err.Error() string and the
+		// CircuitBreakerError discriminant). See theburrowhub/heimdallm#322
+		// Bugs 3+4 for the regression that made emitting from here unsafe.
 		rev, err := p.Run(pr, buildRunOpts(pr, aiCfg))
 		if err != nil {
 			slog.Error("pipeline run failed", "repo", pr.Repo, "pr", pr.Number, "err", err)
@@ -422,20 +435,15 @@ func main() {
 			return
 		}
 		if rev == nil {
-			// Defense-in-depth gate in pipeline.Run rejected this PR. Callers are
-			// expected to filter upstream, so this is the safety net — exit quietly
-			// without emitting a completed/error event.
+			// Pipeline took a skip path and already emitted
+			// EventReviewSkipped with the correct reason
+			// (sha_unchanged / legacy_backfill / not_open / draft /
+			// self_authored). Nothing else to do.
 			return
 		}
 		slog.Info("pipeline: review done",
 			"repo", pr.Repo, "number", pr.Number, "severity", rev.Severity,
 			"github_review_id", rev.GitHubReviewID)
-		broker.Publish(sse.Event{Type: sse.EventReviewCompleted, Data: sseData(map[string]any{
-			"pr_number": pr.Number,
-			"repo":      pr.Repo,
-			"pr_id":     rev.PRID,
-			"severity":  rev.Severity,
-		})})
 	}
 
 	// ── Multi-tier Pipeline ──────────────────────────────────────────────
@@ -799,6 +807,13 @@ func main() {
 			}
 		}()
 
+		// Lifecycle SSEs (pr_detected, review_started, review_completed,
+		// review_skipped with the actual reason) are published by p.Run
+		// via its Publisher dependency. Trigger only owns error paths so
+		// it can attach the err.Error() string the caller surfaces.
+		// Pre-#322 the trigger fabricated review_skipped(not_open) on
+		// every nil return — that lied for SHA-skip / legacy-backfill
+		// paths added in #322 Bug 4.
 		rev, err := p.Run(ghPR, buildRunOpts(ghPR, aiCfg))
 		if err != nil {
 			var cbErr *pipeline.CircuitBreakerError
@@ -816,28 +831,10 @@ func main() {
 			broker.Publish(sse.Event{Type: sse.EventReviewError, Data: sseData(map[string]any{"pr_id": prID, "error": err.Error()})})
 			return err
 		}
-		if rev == nil {
-			// pipeline.Run's defense-in-depth gate rejected this PR (e.g. an
-			// operator triggered a re-review on a closed/merged PR). Emit a
-			// review_skipped event so the UI can surface the reason instead of
-			// leaving the request hanging silently.
-			broker.Publish(sse.Event{
-				Type: sse.EventReviewSkipped,
-				Data: sseData(map[string]any{
-					"repo":      pr.Repo,
-					"pr_number": pr.Number,
-					"pr_title":  pr.Title,
-					"reason":    string(pipeline.SkipReasonNotOpen),
-				}),
-			})
-			return nil
-		}
-		broker.Publish(sse.Event{Type: sse.EventReviewCompleted, Data: sseData(map[string]any{
-			"pr_number": pr.Number,
-			"repo":      pr.Repo,
-			"pr_id":     prID,
-			"severity":  rev.Severity,
-		})})
+		// rev == nil → pipeline already emitted EventReviewSkipped with
+		// the actual reason. rev != nil → pipeline already emitted
+		// EventReviewCompleted. Either way, nothing else to do here.
+		_ = rev
 		return nil
 	})
 
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -833,7 +833,10 @@ func main() {
 		}
 		// rev == nil → pipeline already emitted EventReviewSkipped with
 		// the actual reason. rev != nil → pipeline already emitted
-		// EventReviewCompleted. Either way, nothing else to do here.
+		// EventReviewCompleted. Either way the trigger callback only
+		// has to report success/failure (its signature is
+		// `func(prID int64) error`, see SetTriggerReviewFn) so the
+		// review payload itself is not needed here.
 		_ = rev
 		return nil
 	})
--- a/daemon/internal/activity/recorder.go
+++ b/daemon/internal/activity/recorder.go
@@ -214,6 +214,20 @@ func (r *Recorder) recordIssueReviewError(ev sse.Event) error {
 	return err
 }
 
+// dedupSkipReasons are review_skipped reasons that fire on routine
+// poll cycles rather than user-visible policy decisions, and therefore
+// should NOT produce activity_log rows. Recording them would spam the
+// activity feed with one row per poll for every stable PR — exactly the
+// regression theburrowhub/heimdallm#322 Bug 4 was meant to close (the
+// pre-fix path emitted EventReviewCompleted on those skips, which was
+// then routed here). Keep policy skips (not_open / draft /
+// self_authored) recorded — those reflect the bot deciding not to
+// review and are useful in the audit trail.
+var dedupSkipReasons = map[string]bool{
+	"sha_unchanged":   true,
+	"legacy_backfill": true,
+}
+
 func (r *Recorder) recordReviewSkipped(ev sse.Event) error {
 	var p struct {
 		Repo     string `json:"repo"`
@@ -223,6 +237,11 @@ func (r *Recorder) recordReviewSkipped(ev sse.Event) error {
 	}
 	if err := decode(ev.Data, &p); err != nil {
 		return err
+	}
+	if dedupSkipReasons[p.Reason] {
+		// Routine dedup skip — UI still gets the SSE so the spinner can
+		// clear, but the activity log stays free of poll-cycle noise.
+		return nil
 	}
 	_, err := r.store.InsertActivity(time.Now(), orgOf(p.Repo), p.Repo, "pr",
 		p.PRNumber, p.PRTitle, "review_skipped", p.Reason, map[string]any{
--- a/daemon/internal/activity/recorder_test.go
+++ b/daemon/internal/activity/recorder_test.go
@@ -304,6 +304,29 @@ func TestRecorder_ReviewSkipped(t *testing.T) {
 	}
 }
 
+// TestRecorder_ReviewSkippedDedupReasonsAreNotRecorded locks in the
+// fix from theburrowhub/heimdallm#322 review feedback: dedup-flavoured
+// skips (sha_unchanged, legacy_backfill) MUST NOT generate activity_log
+// rows even though they fire as review_skipped SSE events. They run on
+// every poll cycle on stable PRs — recording them would produce one
+// spam row per minute per stable PR, which is the activity-log spam
+// regression Bug 4 was supposed to close.
+func TestRecorder_ReviewSkippedDedupReasonsAreNotRecorded(t *testing.T) {
+	_, fs, events := newTestRecorder(t)
+
+	for _, reason := range []string{"sha_unchanged", "legacy_backfill"} {
+		events <- sse.Event{
+			Type: sse.EventReviewSkipped,
+			Data: `{"repo":"org/name","pr_number":42,"pr_title":"Fix X","reason":"` + reason + `"}`,
+		}
+	}
+	// Give the recorder a beat to drain any rows it might have inserted.
+	time.Sleep(50 * time.Millisecond)
+	if got := fs.count(); got != 0 {
+		t.Errorf("dedup skips should NOT produce activity_log rows; got %d rows", got)
+	}
+}
+
 func TestRecorder_StoreFailureIsLoggedAndDropped(t *testing.T) {
 	_, fs, events := newTestRecorder(t)
 	fs.setFailNext(true)
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -510,6 +510,103 @@ func (c *Client) FetchIssueCommentsOnly(repo string, number int) ([]Comment, err
 	return c.fetchIssueComments(repo, number)
 }
 
+// TimelineEvent is a slim view of a GitHub PR timeline entry. Only the
+// two events the pipeline needs are surfaced: review_requested (someone
+// asked the reviewer for a review) and review_dismissed (someone
+// dismissed the reviewer's prior review). Other event types in the
+// timeline (commits, labels, comments, assignments…) are filtered out
+// at fetch time so callers don't have to reason about them.
+type TimelineEvent struct {
+	Event     string    // "review_requested" or "review_dismissed"
+	Actor     string    // login of the user who triggered the event
+	CreatedAt time.Time
+}
+
+// GetPRTimelineEventsForReviewer returns the review_requested and
+// review_dismissed events on a PR that target the given reviewer login,
+// sorted ascending by created_at. Used by the pipeline to detect
+// explicit re-request review actions on PRs whose HEAD SHA hasn't
+// changed (theburrowhub/heimdallm#322 Bug 5): a re-request bumps
+// updated_at and re-adds the bot to requested_reviewers, but the SHA
+// fail-closed dedup (#245) would otherwise silently skip the review
+// regardless of the user's explicit intent.
+//
+// Returns an empty slice (not nil) when no relevant events exist —
+// callers can range over the result without a nil guard.
+func (c *Client) GetPRTimelineEventsForReviewer(repo string, number int, login string) ([]TimelineEvent, error) {
+	if login == "" {
+		return nil, fmt.Errorf("github: GetPRTimelineEventsForReviewer: empty login")
+	}
+	out := []TimelineEvent{}
+	page := 1
+	for {
+		path := fmt.Sprintf("/repos/%s/issues/%d/timeline?per_page=100&page=%d", repo, number, page)
+		resp, err := c.do("GET", path, "application/vnd.github+json")
+		if err != nil {
+			return nil, fmt.Errorf("github: fetch timeline: %w", err)
+		}
+		body, _ := io.ReadAll(io.LimitReader(resp.Body, maxBodyBytes))
+		resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			errBody := safeTruncate(string(body), maxErrBodyLen)
+			return nil, fmt.Errorf("github: fetch timeline: status %d: %s", resp.StatusCode, errBody)
+		}
+		var raw []struct {
+			Event     string    `json:"event"`
+			CreatedAt time.Time `json:"created_at"`
+			Actor     struct {
+				Login string `json:"login"`
+			} `json:"actor"`
+			RequestedReviewer *struct {
+				Login string `json:"login"`
+			} `json:"requested_reviewer,omitempty"`
+			DismissedReview *struct {
+				User struct {
+					Login string `json:"login"`
+				} `json:"user"`
+			} `json:"dismissed_review,omitempty"`
+		}
+		if err := json.Unmarshal(body, &raw); err != nil {
+			return nil, fmt.Errorf("github: decode timeline: %w", err)
+		}
+		for _, ev := range raw {
+			switch ev.Event {
+			case "review_requested":
+				if ev.RequestedReviewer != nil && strings.EqualFold(ev.RequestedReviewer.Login, login) {
+					out = append(out, TimelineEvent{
+						Event:     ev.Event,
+						Actor:     ev.Actor.Login,
+						CreatedAt: ev.CreatedAt,
+					})
+				}
+			case "review_dismissed":
+				if ev.DismissedReview != nil && strings.EqualFold(ev.DismissedReview.User.Login, login) {
+					out = append(out, TimelineEvent{
+						Event:     ev.Event,
+						Actor:     ev.Actor.Login,
+						CreatedAt: ev.CreatedAt,
+					})
+				}
+			}
+		}
+		// Stop paging when we get a short page (under per_page=100).
+		if len(raw) < 100 {
+			break
+		}
+		page++
+		// Hard cap so a misbehaving server can't make us iterate forever.
+		if page > 50 {
+			slog.Warn("github: timeline pagination cap hit, returning partial results",
+				"repo", repo, "pr", number, "pages", page)
+			break
+		}
+	}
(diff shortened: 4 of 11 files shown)


ivanmunozruiz:
Reviewer tenía razón en ambos findings. Arreglado en \`27972b6\` con un cambio estructural: **el pipeline ahora es dueño de todos los lifecycle SSEs** (pr_detected, review_started, review_completed, review_skipped) — el caller solo emite los paths de error (necesitan err.Error() y CircuitBreakerError discriminant que el pipeline no pre-shapea).

**Cambios**:
- \`pipeline/pipeline.go\`: nuevo \`Publisher\` interface optional (mismo patrón que \`TimelineFetcher\`). \`*sse.Broker\` lo satisface implícitamente — cero adapter en main.go. Helpers \`publish\` y \`publishSkipped\` para centralizar payload schemas.
- \`pipeline/guards.go\`: dos nuevos \`SkipReason\` — \`SkipReasonSHAUnchanged\` y \`SkipReasonLegacyBackfill\` — para que la UI distinga skips de dedup de los gates pre-existentes.
- \`pipeline/Run\`: emite \`pr_detected\` + \`review_started\` justo en el punto del notify (post-SHA-skip), \`review_completed\` al final del happy path, y \`review_skipped(reason)\` en cada path de skip silente (gate / legacy-backfill / sha_unchanged).
- \`main.go runReview\`: quita las emisiones pre-Run y post-Run; solo error/circuit-breaker queda.
- \`main.go trigger handler\`: quita el \`SkipReasonNotOpen\` fabricado para todo \`rev == nil\` — pipeline ahora emite el reason real.
- \`main.go startup\`: \`p.SetPublisher(broker)\`.

**Tests añadidos** (todos con \`fakePublisher\` que captura los eventos):
- \`TestPipeline_Run_SHASkipEmitsReviewSkipped\`: asserta el orden exacto en happy path \`[pr_detected, review_started, review_completed]\` y que el segundo run (SHA-skip) emite **solo** \`[review_skipped]\` con \`reason=\"sha_unchanged\"\` — sin phantom \`review_started\` que dejaría el spinner colgado.
- \`TestPipeline_Run_LegacyBackfillEmitsReviewSkipped\`: \`review_skipped\` con \`reason=\"legacy_backfill\"\`.
- \`TestPipeline_Run_GateSkipEmitsReviewSkipped\`: \`review_skipped\` con la reason real del Evaluate (e.g. \`not_open\`).
- \`TestPipeline_Run_NilPublisherIsNoop\`: legacy contract preservado, sin panic cuando no hay publisher wired.

\`make test-docker\` verde en todo el daemon. Esperando CI + tu visto bueno.

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Well-structured PR that correctly centralizes lifecycle SSE emission into Pipeline.Run and adds re-request review bypass for SHA-skip dedup. The design is sound — optional Publisher/TimelineFetcher interfaces preserve backward compatibility, fail-closed posture is maintained, and tests cover the key paths. A few issues worth addressing: a silent event-drop on JSON marshal failure, an unbounded timeline pagination loop risk, and some comment verbosity.

### Issues

🟡 **daemon/internal/pipeline/pipeline.go:0** — publish() silently swallows json.Marshal errors. While these events are 'observability not correctness', a marshal failure on a map[string]any with basic types (string, int) would indicate a serious bug in the caller. At minimum, log the error at Warn level so it doesn't become a silent black hole during debugging.
🟡 **daemon/internal/pipeline/pipeline.go:0** — shouldBypassSHASkipForReReview walks the entire sorted events slice to find the latest relevant event but uses a linear scan starting from the beginning. This is correct but the comment says 'Walk it to find the most recent event whose timestamp is strictly newer than prevReview.CreatedAt' — iterating backward from the end would be clearer and marginally faster. Minor readability nit.
🟡 **daemon/internal/github/client.go:0** — GetPRTimelineEventsForReviewer paginates up to 50 pages (5000 events) with no early-exit once it has enough data. For very active PRs, this could be expensive. Consider breaking early once you've seen an event newer than prevReview.CreatedAt (since events are sorted by created_at in the API response), or passing the cutoff timestamp so the method can stop paging once it's past the relevant window.
🟡 **daemon/internal/github/client.go:0** — The sort.Slice at the end of GetPRTimelineEventsForReviewer re-sorts events that the GitHub API already returns in chronological order. The sort is a defensive measure (good), but worth a comment explaining it's a safety net against GitHub's undocumented ordering rather than a functional necessity.
⚠️ **daemon/cmd/heimdallm/main.go:0** — The trigger handler now has `_ = rev` to suppress the unused-variable warning. This is fine mechanically, but the rev return value carries the GitHubReviewID which callers previously used. If the trigger handler's return channel or any future logging needs the review ID (e.g., for the trigger-review API response), this information is now discarded. Verify no downstream consumer of the trigger callback expects a review result.
⚠️ **daemon/internal/pipeline/pipeline.go:0** — The pipeline now imports and directly references sse.Event types, creating a coupling between the pipeline package and the SSE package. Previously the pipeline was transport-agnostic. The Publisher interface partially abstracts this, but the pipeline still builds sse.Event values and references sse.EventReviewSkipped etc. constants directly. Consider defining the event type constants in the pipeline package (or a shared constants package) and having the Publisher interface accept a pipeline-defined event type, keeping the pipeline decoupled from the SSE transport layer.
🔴 **daemon/internal/pipeline/pipeline.go:0** — The notification (p.notify.Notify) and the SSE events (pr_detected, review_started) are emitted AFTER the SHA-skip decision but BEFORE the circuit breaker check. If the circuit breaker trips, the user sees a desktop notification 'PR Review Started' and SSE events for pr_detected + review_started, but the review never actually starts — the pipeline returns a CircuitBreakerError. This creates a similar phantom-spinner problem that Bug 3 was trying to fix, just on a different path.
🟡 **daemon/internal/pipeline/pipeline_test.go:0** — fakePublisher uses a sync.Mutex for thread safety, but fakeTimeline does not protect its `calls` counter. If tests ever run with parallel goroutines hitting the pipeline, fakeTimeline.calls would have a data race. Currently tests are sequential so this is latent, not active.

### Suggestions

- Move the p.notify.Notify and p.publish(EventPRDetected/EventReviewStarted) calls to AFTER the circuit breaker check (after line 341 in the current base). This ensures no phantom 'PR Review Started' notification or SSE event fires when the circuit breaker immediately rejects the review, which is the same class of bug that Bug 3 addresses for SHA-skip.
- Consider defining a minimal event-type enum or string constants within the pipeline package itself, and have the Publisher interface accept those. This keeps the pipeline package decoupled from the SSE transport layer and makes it easier to add non-SSE publishers (e.g., metrics, webhooks) in the future.
- Add a slog.Warn in the publish() helper when json.Marshal fails, rather than silently returning. The payloads are simple map[string]any so this should never happen in practice, but a silent swallow makes debugging very painful if it ever does.
- Add a test for the circuit-breaker-after-SSE interaction: wire a fakePublisher, set circuit breaker limits to 0, and verify that no review_started event is emitted when the breaker trips. This would catch the notify/SSE ordering issue described above.

---

ivanmunozruiz:
Aplicados los findings de los dos reviews en \`e397fcb\`. Tres HIGH atajados + las 5 menores.

### HIGH cerradas

**1. notify + SSE antes del circuit breaker (sergiotejon HIGH)** → movido el bloque \`notify\` + \`pr_detected\` + \`review_started\` a DESPUÉS del breaker check. Si el breaker trippa, el run sale vía \`*CircuitBreakerError\` sin lifecycle event previo. \`TestRun_CircuitBreakerTripStopsExecute\` ahora wirea \`fakePublisher\` y asserta que 3 reviews reales + 1 breaker-trip producen exactamente 3 \`PR Review Started\` notificaciones y 3 \`review_started\` SSE — lock-in del invariant.

**2. Flutter no consume \`review_skipped\` (colega HIGH)** → añadido \`case 'review_skipped'\` en \`dashboard_providers.dart\` que limpia el spinner del \`reviewingPRsProvider\` (mismo patrón que \`review_completed\` pero sin notification de \"complete\" ni list refresh). Trigger manual sobre PR con SHA-unchanged ya no deja spinner colgado.

**3. activity_log graba \`review_skipped\` por poll (colega HIGH)** → añadida lista \`dedupSkipReasons = {sha_unchanged, legacy_backfill}\` en \`recorder.go\`. Esos reasons NO insertan fila — son skips de dedup (no acción de usuario), exactamente el spam que Bug 4 quería cerrar. Los reasons de policy (\`not_open\`/\`draft\`/\`self_authored\`) sí siguen grabándose porque reflejan decisiones deliberadas del bot. Test nuevo \`TestRecorder_ReviewSkippedDedupReasonsAreNotRecorded\` asserta cero filas para ambos.

### Menores cerradas

- \`pipeline.publish()\` ahora hace \`slog.Warn\` en marshal failure (no más silent swallow)
- \`shouldBypassSHASkipForReReview\` walks events backward y short-circuits en primer cutoff miss — más claro y marginalmente más rápido en PRs activos
- \`GetPRTimelineEventsForReviewer\`'s \`sort.Slice\` lleva ahora un comentario explicando que es safety net (no necesidad funcional)
- \`fakeTimeline.calls\` es mutex-guarded (mismo posture que \`fakePublisher\`)
- \`_ = rev\` del trigger handler tiene un comentario apuntando a la firma \`func(prID int64) error\` de \`SetTriggerReviewFn\` para que se entienda que el discard es deliberado

### NO atajado en este PR (sergiotejon ⚠️)

- **Acoplamiento \`pipeline → sse.Event\`**: dejado por consistencia con \`issues/pipeline.go\` que ya hace lo mismo. Refactorizar a un event-type enum local + adapter sería sensato, pero rompería la simetría con el otro pipeline y el scope de este PR ya creció lo suyo. Lo dejo como follow-up.
- **Timeline pagination 50-page early-exit con cutoff**: la API de GitHub devuelve eventos en orden ASCENDENTE — para encontrar el último event > cutoff necesitas llegar al final de la pagination en cualquier caso. La sugerencia de break-early sólo aplicaría si paginásemos descendente (que GitHub no soporta para timeline). Documento la limitación con el cap de 50 páginas.

\`make test-docker\` verde. Esperando CI + segundo round de review.

sergiotejon:
## 🟡 LOW Issue

In the SHA-unchanged branch, when shouldBypassSHASkipForReReview returns true the code falls through to the rest of Run, but the condition `pr.Head.SHA != "" && prevReview.HeadSHA == pr.Head.SHA` means prevReview is non-nil with matching SHA. Execute will run and InsertReview/UpdateReview semantics should be verified — ensure that re-review on same SHA doesn't create a duplicate row or violate any (pr_id, head_sha) uniqueness constraint. Worth a test that asserts the happy-path on re-request with unchanged SHA produces a fresh review row as expected.

**Location:** `daemon/internal/pipeline/pipeline.go` line 419

---

sergiotejon:
## 🟡 LOW Issue

The 50-page cap (5000 events) silently returns partial results with only a Warn log. For the re-request bypass use case, partial results from the head of the timeline could miss the most recent review_requested event (since results are sorted ASC and the cutoff is at the tail), causing fail-closed skip on a legitimate re-request. Consider returning an error or a truncation flag when the cap is hit so shouldBypassSHASkipForReReview can fail-closed explicitly rather than on stale data.

**Location:** `daemon/internal/github/client.go` line 565

---

sergiotejon:
## 🟡 LOW Issue

dedupSkipReasons is a package-level mutable map — although never written after init, using a map literal leaves it mutable. Prefer a small switch statement or a function for clarity and to prevent accidental mutation by future code.

**Location:** `daemon/internal/activity/recorder.go` line 217

---

sergiotejon (approved):

Re-review: the author's structural fix is solid. Lifecycle SSEs are now owned by the pipeline, emitted post-breaker/post-SHA-skip, with distinct SkipReasons (sha_unchanged, legacy_backfill) routed correctly to UI and filtered from activity_log. Previous HIGH findings (notify/SSE before breaker, Flutter not consuming review_skipped, activity_log spam) are all addressed with lock-in tests. The remaining concerns are minor and the deferred items (sse coupling, ASC-only pagination) have reasonable rationale.

### Suggestions

- Add an integration-style test covering the full re-request flow: first run produces review, second run with same SHA but a review_requested timeline event proceeds to Execute and emits pr_detected/review_started/review_completed (not review_skipped).
- Consider returning an explicit truncation sentinel from GetPRTimelineEventsForReviewer when the page cap is hit, so the bypass logic can fail-closed on incomplete data instead of potentially missing a tail event.
- Follow-up ticket for the pipeline→sse.Event coupling is reasonable; track it explicitly so the symmetry argument with issues/pipeline.go is revisited together.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Well-structured fix that correctly centralizes lifecycle SSE emission inside Pipeline.Run, eliminating phantom spinners and notification spam. The core logic (re-request bypass, dedup skip reasons, activity log filtering) is sound. Several issues deserve attention before merge: the Run docstring is now stale, the timeline API is called on every poll for stable PRs (rate-limit concern), and the backward-iteration loop is deceptively not a loop.

### Issues

🔴 **daemon/internal/pipeline/pipeline.go:176** — Run's return-contract docstring is stale after this PR. It still says (nil, nil) only means gate rejection and that 'Skip-event publication is the caller's responsibility; the pipeline only logs on this path'. Both statements are now wrong: (nil, nil) also covers SHA-skip and legacy-backfill, and the pipeline now publishes skip events itself via publishSkipped. This is the primary contract reference for callers — leaving it stale will mislead future integrators.
⚠️ **daemon/internal/pipeline/pipeline.go:215** — shouldBypassSHASkipForReReview calls the timeline API on every poll cycle for every open PR whose SHA hasn't changed. For a bot watching N repos with M stable open PRs, that's N*M extra GitHub API calls per poll interval. At 1-minute polls with e.g. 200 stable PRs, that's 12,000 calls/hour — a meaningful fraction of GitHub's rate limit. Consider caching the last-checked timestamp per PR or only querying the timeline when pr.UpdatedAt changed since the last review's PublishedAt (a signal that something actually happened on the PR).
⚠️ **daemon/internal/pipeline/pipeline.go:227** — The backward for-loop in shouldBypassSHASkipForReReview always returns on the very first iteration — it never actually iterates. The loop structure and the lengthy comment about 'walking backward' and 'O(1) for the common case' obscure what is simply a check on the last element. Replace with: `if len(events) == 0 { return false }; ev := events[len(events)-1]; return ev.CreatedAt.After(prevReview.CreatedAt) && ev.Event == "review_requested"`
⚠️ **daemon/internal/pipeline/pipeline_test.go:362** — The existing TestPipeline_Run_SkipsReviewOnSameHeadSHA expects `rev2` to be the previous review (`rev2.ID == rev1.ID`), but this PR changes the SHA-skip path from `return prevReview, nil` to `return nil, nil`. This test must be updated to expect `rev2 == nil`. The diff is truncated so this may already be addressed — please verify.
🟡 **daemon/internal/activity/recorder.go:226** — dedupSkipReasons uses string literals ("sha_unchanged", "legacy_backfill") that must stay in sync with pipeline.SkipReasonSHAUnchanged and pipeline.SkipReasonLegacyBackfill. The activity package can't import pipeline (would create coupling), so the strings are unavoidable — but add a comment referencing the canonical constants so a rename gets caught in review.
🟡 **daemon/internal/activity/recorder_test.go:325** — time.Sleep(50 * time.Millisecond) for test synchronization is a flaky-test pattern. Under CI load the recorder goroutine might not drain within 50ms, or might not even be scheduled. Consider draining the channel explicitly or using a synchronization primitive.
🟡 **daemon/internal/github/client.go:545** — GetPRTimelineEventsForReviewer does not accept a context.Context, so pagination of up to 50 pages cannot be cancelled or timed out. This matches the existing client methods (FetchDiff etc.), but since the timeline fetch runs on every poll for stable PRs, a stuck HTTP connection would block the entire pipeline goroutine indefinitely. Worth noting as follow-up tech debt.

### Suggestions

- Update the Run docstring (lines 176-184) to reflect the new return contract: (nil, nil) now covers gate-skip, SHA-skip, and legacy-backfill, and the pipeline itself publishes skip events when a Publisher is wired.
- Add a lightweight cache or guard to avoid calling the timeline API when the PR's UpdatedAt hasn't changed since the last stored review — this would reduce unnecessary API calls from O(stable_PRs * polls/hour) to near-zero for truly idle PRs.
- Replace the misleading backward for-loop in shouldBypassSHASkipForReReview with a direct check on the last element — the current loop body always returns on the first iteration.
- Consider adding a comment on dedupSkipReasons noting the canonical constants are pipeline.SkipReasonSHAUnchanged and pipeline.SkipReasonLegacyBackfill, to flag the coupling during future renames.

---

PR #326, opened 2026-04-24
fix(daemon): mark review orphan on permanent submit failure (closes #325)

### LINKED ISSUE #325: bug(daemon): PublishPending retries 422 'lock prevents review' forever — burns API calls per poll cycle
## TL;DR

\`pipeline.PublishPending()\` treats every GitHub error as transient and retries on the next poll cycle. HTTP 422 with body \"lock prevents review\" is a **permanent** error — the PR's conversation is locked (merged/closed-and-locked) and no future retry will succeed. The unpublished review row stays in SQLite forever, the daemon burns one GitHub API call per poll cycle attempting to re-submit it indefinitely.

Originally tracked as Bug 1 in #322 (now closed); the other three bugs in #322 landed in #324, this one was scoped out for a separate PR.

## Symptom

Live observation in production logs (PR review_id=681):

\`\`\`
WARN pipeline: retry publish failed review_id=681 err=\"github: submit review: status 422: lock prevents review\"
WARN pipeline: retry publish failed review_id=681 err=\"github: submit review: status 422: lock prevents review\"
WARN pipeline: retry publish failed review_id=681 err=\"github: submit review: status 422: lock prevents review\"
... (repeats every poll cycle, indefinitely)
\`\`\`

## Root cause

\`daemon/internal/pipeline/pipeline.go:629-684\` — \`PublishPending\` iterates over every row from \`ListUnpublishedReviews\` and calls \`SubmitReview\` again. The error handling at line 663-665 is uniform:

\`\`\`go
ghID, ghState, err := p.gh.SubmitReview(pr.Repo, pr.Number, ..., ...)
if err != nil {
    slog.Warn(\"pipeline: retry publish failed\", \"review_id\", rev.ID, \"err\", err)
    continue  // ← never marks the row, will retry on next tick forever
}
\`\`\`

\`SubmitReview\` (\`daemon/internal/github/client.go:202-238\`) collapses every non-200 response into a single error string:

\`\`\`go
if resp.StatusCode != http.StatusOK {
    errBody := safeTruncate(string(respBody), maxErrBodyLen)
    return 0, \"\", fmt.Errorf(\"github: submit review: status %d: %s\", resp.StatusCode, errBody)
}
\`\`\`

There is no way for the caller to distinguish \"permanent lock\" from a transient 5xx without string-matching the error message.

The same path on the \`Run\` happy-path (\`pipeline.go:559-590\`) silently logs the publish failure and leaves the row unpublished, which then feeds back into the same \`PublishPending\` loop above.

## Why \"lock prevents review\" is permanent

GitHub's PR conversation lock (set via the \"Lock conversation\" UI action, or implicitly when certain merge/close flows fire) blocks all review submissions on that PR. Unlocking requires manual operator action. From the daemon's perspective: never going to succeed. There is no exponential-backoff scenario that helps; the row should be marked done so the audit trail records \"review existed locally but PR was locked before publish\" and the retry loop stops.

## Why the existing orphan-row escape doesn't catch this

\`PublishPending\` already handles one orphan case (\`pipeline.go:643-647\`):

\`\`\`go
if pr.Repo == \"\" {
    _ = p.store.MarkReviewPublished(rev.ID, -1, \"\", time.Now().UTC())
    slog.Info(\"pipeline: skipping pending review for PR with no repo\", \"review_id\", rev.ID)
    continue
}
\`\`\`

It uses sentinel \`(-1, \"\")\` to mark a row as \"never going to publish, do not retry\". The 422-lock fix should reuse that exact pattern.

### Approach: typed error in the GitHub client

Add a \`*github.PermanentSubmitError\` type that \`SubmitReview\` returns when the response is HTTP 422 AND the body mentions a lock condition. Callers (\`Run\` + \`PublishPending\`) use \`errors.As\` to detect it and mark the row orphan via the existing \`MarkReviewPublished(rev.ID, -1, \"\", ...)\` sentinel.

Why typed (not string-match in the caller): the lock message is the most common permanent-422 today but GitHub may add others (e.g. archived repo, deleted PR). Centralising \"is this 422 permanent?\" inside the client lets us extend the predicate without touching the pipeline.

**Files**:
- \`daemon/internal/github/client.go\`: new \`PermanentSubmitError struct { StatusCode int; Reason string }\` with \`Error()\` and \`Unwrap()\`. \`SubmitReview\` checks \`StatusCode == 422 && strings.Contains(body, \"lock\")\` (or a small predicate \`isPermanent422(body)\`) and returns \`&PermanentSubmitError{...}\` instead of the generic \`fmt.Errorf\`.
- \`daemon/internal/pipeline/pipeline.go\`:
  - In \`PublishPending\`, after the \`SubmitReview\` call: \`var permErr *github.PermanentSubmitError; if errors.As(err, &permErr) { p.store.MarkReviewPublished(rev.ID, -1, \"\", time.Now().UTC()); slog.Info(\"pipeline: marking review as orphaned (PR locked or otherwise unrecoverable)\", \"review_id\", rev.ID, \"reason\", permErr.Reason); continue }\`
  - In \`Run\` (around line 564 \`if publishErr != nil\`): same \`errors.As\` check — mark orphan immediately so the row never enters the retry loop in the first place.

### Acceptance criteria

- [ ] Unit test against \`SubmitReview\` with httptest returning 422 + body \`{\"message\":\"lock prevents review\"}\` asserts \`*PermanentSubmitError\` is returned (with the right \`StatusCode\` and a populated \`Reason\`).
- [ ] Unit test against \`SubmitReview\` with httptest returning 5xx asserts a generic error (NOT \`PermanentSubmitError\`) so transient failures continue to retry.
- [ ] Unit test for \`PublishPending\`: feed it a row whose \`SubmitReview\` mock returns \`PermanentSubmitError\` — assert one \`MarkReviewPublished(reviewID, -1, \"\", _)\` call landed and the retry loop does not pick the row up again.
- [ ] Same coverage for \`Run\` — a 422 lock on the initial publish marks orphan immediately, no extra \`PublishPending\` tick required.
- [ ] \`make test-docker\` green (no raw \`go test\` per team policy — EDR fires alerts).
- [ ] Live verification post-merge: tail daemon logs against a real locked PR; confirm the \"retry publish failed\" warning appears at most once and the row disappears from \`SELECT * FROM reviews WHERE published_at = '' OR github_review_id = 0\`.

### Scope notes

- \`Run\` and \`PublishPending\` get the same fix because both can hit 422-lock — the lock can be set after the daemon stores the review locally but before the first publish attempt completes.
- Fix is local to \`pipeline\` + \`github\` packages — no new tables, no schema migration. The orphan sentinel \`(reviewID, -1, \"\")\` already exists and is exercised by \`pipeline.go:643-647\`, so this PR just teaches \`PublishPending\` (and \`Run\`'s retry path) to use it for permanent 422s.
- If GitHub introduces other permanent 422 reasons later, extend the predicate in \`SubmitReview\` rather than the pipeline.

### References

- Originally tracked as Bug 1 in #322 (closed by #324; the other three bugs were the SHA-skip side effects and re-request bypass).
- Sentinel orphan pattern: \`pipeline.go:643-647\` (\"PR with no repo\").
- Same defensive posture as the fail-closed work in #245 / #246 / #258 / #283: cap the cost surface, never let a transient failure mode burn API calls forever.

### Severity

**MEDIUM** — does not burn Claude credits (the row is already stored locally), but does burn one GitHub API call per poll cycle per locked PR forever, and the retry warning hides legitimate publish failures in operator log noise.

ivanmunozruiz:
## Summary

Closes #325. \`PublishPending\` and the initial publish path in \`Run\` treated every GitHub error as transient. HTTP 422 with \"lock prevents review\" body is permanent — the PR's conversation has been locked and no future retry will succeed without operator intervention. Pre-fix the unpublished review row stayed in SQLite forever and the daemon burned one GitHub API call per poll cycle per locked PR (observed in production logs as a continuous \`WARN pipeline: retry publish failed\` for \`review_id=681\`).

This was bug 1 of the original #322 quartet; the other three landed in #324.

## Approach

Typed error in the GitHub client + matching detection in both pipeline paths. Reuses the existing \`MarkReviewPublished(reviewID, -1, \"\", ...)\` sentinel that \`PublishPending\` already uses for PRs with no repo (see \`pipeline.go:643-647\`).

### Files

- **\`daemon/internal/github/client.go\`** — new \`*PermanentSubmitError\` struct (\`StatusCode\`, \`Reason\`, \`Body\`) plus a \`classifyPermanentSubmit422\` predicate that case-insensitively matches \`\"lock prevents review\"\` and \`\"pull request is locked\"\`. \`SubmitReview\` returns the typed error in the 422-lock case and keeps the generic \`fmt.Errorf\` for everything else (transient 5xx, unknown 422s) so the retry loop continues working there. Same fail-closed posture as #245.

- **\`daemon/internal/pipeline/pipeline.go\` (PublishPending)** — \`errors.As(&permErr)\` on the \`SubmitReview\` error; on hit, marks orphan via the existing sentinel and continues. \`slog.Info\` records \`reason\` and \`status\` so the operator log says \"orphaned because pr_locked\" rather than the previous \"retry publish failed\" loop.

- **\`daemon/internal/pipeline/pipeline.go\` (Run)** — same \`errors.As\` check on the initial publish error. A 422 lock during the first \`SubmitReview\` now marks orphan in-place, so the row never enters the PublishPending retry queue. Transient errors still take the existing retry path.

### Tests

3 new client tests:
- \`TestSubmitReview_LockedPRReturnsPermanentSubmitError\`: 422 + lock body → typed error with \`StatusCode=422, Reason=\"pr_locked\"\`, body populated.
- \`TestSubmitReview_TransientErrorIsNotPermanent\`: 503 stays as generic error so retries continue.
- \`TestSubmitReview_422WithoutLockIsNotPermanent\`: 422 with malformed-body error message is NOT classified (guards against over-rotation).

3 new pipeline tests (\`pipeline_orphan_test.go\`):
- \`TestRun_LockedPRMarksReviewOrphanImmediately\`: \`Run\` with a locked PR → row marked orphan in-place, \`ListUnpublishedReviews\` returns empty, subsequent \`PublishPending\` is a no-op.
- \`TestPublishPending_LockedPRStopsRetrying\`: pre-seeded unpublished row → first tick marks orphan, next 3 ticks must NOT re-attempt \`SubmitReview\`.
- \`TestPublishPending_TransientErrorStillRetries\`: 5xx error keeps the row in the queue and \`SubmitReview\` is attempted on every tick (regression guard against over-rotation).

## Test plan

- [x] \`make test-docker\` green across all daemon packages
- [x] All 6 new tests pass
- [ ] Live verification post-merge: tail daemon logs against a real locked PR; expect one \"review marked orphan\" \`INFO\` line and zero subsequent \"retry publish failed\" warnings for that \`review_id\`.

## References

- Closes #325
- Original bug-1 spec from #322 (closed by #324, where the other 3 bugs landed)
- Same \`(-1, \"\")\` orphan sentinel: \`pipeline.go:643-647\` (PRs with no repo)
- Same fail-closed-on-unknown-error posture: #245 / #246 / #258 / #283

### DIFF
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -192,6 +192,49 @@ func (c *Client) fetchByQualifier(username, qualifier string, repos []string) ([
 	return result.Items, nil
 }
 
+// PermanentSubmitError signals that SubmitReview hit a state from
+// which no retry will recover — e.g. the PR's conversation is locked,
+// the PR has been deleted, or the repo was archived. Callers should
+// mark the local review row as orphaned (via the same sentinel as
+// the no-repo path in pipeline.PublishPending) instead of retrying
+// every poll cycle. See theburrowhub/heimdallm#325.
+type PermanentSubmitError struct {
+	StatusCode int    // HTTP status code returned by GitHub (always 422 today)
+	Reason     string // short human-readable label, e.g. "pr_locked"
+	Body       string // safe-truncated response body for diagnostics
+}
+
+func (e *PermanentSubmitError) Error() string {
+	return fmt.Sprintf("github: submit review permanent failure (status %d, reason %s): %s",
+		e.StatusCode, e.Reason, e.Body)
+}
+
+// classifyPermanentSubmit422 inspects an HTTP 422 response body and
+// returns (reason, true) when the failure is known to be permanent for
+// the daemon — i.e. no future retry can succeed without operator
+// intervention. Returns ("", false) for transient or unknown 422s so
+// the caller falls back to the existing retry path.
+//
+// Currently classified as permanent:
+//   - "lock prevents review" — PR conversation is locked. Unlocking
+//     requires the operator to use the GitHub UI; the daemon cannot
+//     recover on its own.
+//
+// Extend with care: a false positive here marks a recoverable review
+// as orphan and loses it permanently. When in doubt, leave the case
+// out and let the retry loop run — it is rate-limited by the poll
+// interval and only burns 1 GitHub API call per cycle.
+func classifyPermanentSubmit422(status int, body []byte) (string, bool) {
+	if status != http.StatusUnprocessableEntity {
+		return "", false
+	}
+	lower := strings.ToLower(string(body))
+	if strings.Contains(lower, "lock prevents review") || strings.Contains(lower, "pull request is locked") {
+		return "pr_locked", true
+	}
+	return "", false
+}
+
 // SubmitReview posts an AI-generated review to GitHub as a PR review.
 // event should be "REQUEST_CHANGES", "COMMENT", or "APPROVE".
 // Returns the GitHub review ID and the review state reported by the API —
@@ -199,6 +242,12 @@ func (c *Client) fetchByQualifier(username, qualifier string, repos []string) ([
 // event and on GitHub's server-side rules. We pass the state through to the
 // store so the web UI can show a review-decision badge sourced from GitHub
 // rather than derived locally from severity.
+//
+// On HTTP 422 with a body that indicates the PR is in a state from which
+// no retry can recover (e.g. the conversation has been locked), returns
+// a *PermanentSubmitError so the caller can mark the local review row
+// as orphaned instead of retrying every poll cycle. See
+// theburrowhub/heimdallm#325.
 func (c *Client) SubmitReview(repo string, number int, body, event string) (int64, string, error) {
 	path := fmt.Sprintf("/repos/%s/pulls/%d/reviews", repo, number)
 
@@ -225,6 +274,13 @@ func (c *Client) SubmitReview(repo string, number int, body, event string) (int6
 	respBody, _ := io.ReadAll(io.LimitReader(resp.Body, maxBodyBytes))
 	if resp.StatusCode != http.StatusOK && resp.StatusCode != 200 {
 		errBody := safeTruncate(string(respBody), maxErrBodyLen)
+		if reason, ok := classifyPermanentSubmit422(resp.StatusCode, respBody); ok {
+			return 0, "", &PermanentSubmitError{
+				StatusCode: resp.StatusCode,
+				Reason:     reason,
+				Body:       errBody,
+			}
+		}
 		return 0, "", fmt.Errorf("github: submit review: status %d: %s", resp.StatusCode, errBody)
 	}
 
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -275,10 +275,14 @@ func (c *Client) SubmitReview(repo string, number int, body, event string) (int6
 	if resp.StatusCode != http.StatusOK && resp.StatusCode != 200 {
 		errBody := safeTruncate(string(respBody), maxErrBodyLen)
 		if reason, ok := classifyPermanentSubmit422(resp.StatusCode, respBody); ok {
+			// Body carries the FULL response (not safe-truncated) so an
+			// operator inspecting *PermanentSubmitError can see the full
+			// GitHub error payload — the reason substring may live past
+			// the maxErrBodyLen cutoff that the generic-error path uses.
 			return 0, "", &PermanentSubmitError{
 				StatusCode: resp.StatusCode,
 				Reason:     reason,
-				Body:       errBody,
+				Body:       string(respBody),
 			}
 		}
 		return 0, "", fmt.Errorf("github: submit review: status %d: %s", resp.StatusCode, errBody)
--- a/daemon/internal/github/poller_test.go
+++ b/daemon/internal/github/poller_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
@@ -358,5 +359,33 @@ func TestSubmitReview_422WithoutLockIsNotPermanent(t *testing.T) {
 	var permErr *gh.PermanentSubmitError
 	if errors.As(err, &permErr) {
 		t.Errorf("422 without lock body must NOT classify as PermanentSubmitError, got %+v", permErr)
+	}
+}
+
+// TestSubmitReview_LockedPRBodyIsNotTruncated covers the
+// post-review-feedback fix to #325: when SubmitReview returns a
+// *PermanentSubmitError, the Body field must carry the FULL response
+// body (not safe-truncated) so an operator inspecting the error can
+// see the complete GitHub payload — the lock substring may live past
+// the maxErrBodyLen cutoff used by the generic-error path.
+func TestSubmitReview_LockedPRBodyIsNotTruncated(t *testing.T) {
+	// Build a body where the lock substring sits well past 200 bytes
+	// (maxErrBodyLen) so a truncation regression would lose it.
+	padding := strings.Repeat("x", 500)
+	bigBody := `{"message":"Validation Failed","padding":"` + padding + `","errors":[{"message":"lock prevents review"}]}`
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusUnprocessableEntity)
+		w.Write([]byte(bigBody))
+	}))
+	defer srv.Close()
+
+	client := gh.NewClient("fake-token", gh.WithBaseURL(srv.URL))
+	_, _, err := client.SubmitReview("org/repo", 1, "body", "COMMENT")
+	var permErr *gh.PermanentSubmitError
+	if !errors.As(err, &permErr) {
+		t.Fatalf("expected PermanentSubmitError on big locked body, got %v", err)
+	}
+	if !strings.Contains(permErr.Body, "lock prevents review") {
+		t.Errorf("permErr.Body lost the lock substring (truncated at boundary?); body=%q", permErr.Body)
 	}
 }
--- a/daemon/internal/github/poller_test.go
+++ b/daemon/internal/github/poller_test.go
@@ -2,6 +2,7 @@ package github_test
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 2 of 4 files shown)


sergiotejon:
## 🟡 LOW Issue

The condition `resp.StatusCode != http.StatusOK && resp.StatusCode != 200` is redundant since http.StatusOK == 200. Pre-existing but worth noting.

**Location:** `daemon/internal/github/client.go` line 275

---

sergiotejon:
## ⚠️ MEDIUM Issue

Substring matching on response body text ('lock prevents review', 'pull request is locked') is fragile — GitHub could change the wording at any time, silently reverting this fix to the broken retry-forever behavior. Consider also matching on the structured errors[].code/message field if stable, or adding a log/metric when a 422 is NOT classified so regressions in GitHub's message wording are detectable.

**Location:** `daemon/internal/github/client.go` line 235

---

sergiotejon:
## 🟡 LOW Issue

On initial-publish permanent failure, if MarkReviewPublished itself fails, the code falls through and only logs a warning — but the review row remains in the unpublished queue where PublishPending will re-encounter the same PermanentSubmitError and eventually mark it orphan. This is actually correct self-healing behavior, but the log message 'will fall through to PublishPending' could be clearer that this is intentional recovery.

**Location:** `daemon/internal/pipeline/pipeline.go` line 589

---

sergiotejon:
## 🟡 LOW Issue

Tests don't cover the 422-without-lock case at the pipeline level — only at the client level. A pipeline-level test asserting a generic 422 still retries would strengthen the fail-closed guarantee.

**Location:** `daemon/internal/pipeline/pipeline_orphan_test.go`

---

sergiotejon (approved):

Well-crafted fix for issue #325 that introduces a typed PermanentSubmitError to distinguish unrecoverable GitHub submit failures (PR locked) from transient ones, preventing indefinite retry loops. The implementation is conservative, well-documented, and thoroughly tested on both the initial publish path and the PublishPending retry path.

### Suggestions

- Consider parsing the 422 response body as JSON and matching on the structured errors[].message or errors[].code field rather than lowercased substring matching, which would be more robust to GitHub wording changes.
- Add a metric or structured log when a 422 is seen but NOT classified as permanent, so operators can detect if GitHub changes its error wording and the fix silently regresses.
- Consider extracting the `MarkReviewPublished(rev.ID, -1, "", time.Now().UTC())` orphan-sentinel into a small helper like `markReviewOrphan(rev.ID)` since it's now used in three places (no-repo path, initial-publish permanent failure, PublishPending permanent failure) — reduces the risk of future sentinel drift.
- The PermanentSubmitError comment says 'always 422 today' for StatusCode — consider whether 404 (PR deleted) or 410 (repo archived/gone) should also be classified as permanent, as the doc comment hints at these cases but only 422-lock is implemented.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured fix that correctly introduces a typed PermanentSubmitError for unrecoverable GitHub 422s (locked PRs) and marks affected reviews as orphans in both the initial-publish and retry-loop paths. Conservative fail-open classification design is appropriate. Test coverage is thorough with both positive and negative cases. A few minor concerns around duplication and fragility.

### Issues

🟡 **daemon/internal/pipeline/pipeline.go:597** — The orphan-marking block (errors.As + MarkReviewPublished with (-1, "") sentinel + structured logging) is duplicated nearly verbatim between Run (~line 597) and PublishPending (~line 690). Consider extracting a small helper like `markOrphanOnPermanentFailure(reviewID, permErr)` to reduce duplication and ensure both sites stay in sync if the sentinel convention or logging changes.
🟡 **daemon/internal/github/client.go:231** — classifyPermanentSubmit422 relies on substring matching against GitHub's error body ("lock prevents review", "pull request is locked"). If GitHub changes this wording the classification silently degrades to the transient path, which is safe but means the fix stops working. Consider additionally matching on structured fields (e.g. the JSON `errors[].code` or `errors[].resource`) which are more stable across API versions, falling back to the substring check.
🟡 **daemon/internal/github/client.go:276** — classifyPermanentSubmit422 receives the raw respBody (pre-truncation) for classification, which is correct. However, inside SubmitReview the errBody passed to PermanentSubmitError.Body is the truncated version. If the lock message appears past the truncation boundary, the diagnostic Body field would not contain the reason string, which could confuse operators inspecting the error. This is an unlikely edge case given typical GitHub error sizes but worth noting.
🟡 **daemon/internal/pipeline/pipeline_orphan_test.go:1** — The test file duplicates several fake types (fakeExecOrphan, fakeGHTransientSubmit, fakeNotify) that are structurally identical to fakes in sibling test files. If the GithubDeps or Executor interface gains a method, every fake must be updated independently. Consider a shared testutil file with configurable fakes (e.g., function fields for SubmitReview) to reduce maintenance burden.

### Suggestions

- Extract the orphan-marking logic (PermanentSubmitError check + MarkReviewPublished + logging) into a shared helper method on Pipeline to keep the Run and PublishPending paths in sync.
- Add a direct unit test for classifyPermanentSubmit422 covering edge cases: empty body, body exceeding typical size, mixed-case variations, and partial substring matches — the current tests only exercise it indirectly through the full HTTP round-trip.
- Consider matching on GitHub's structured JSON error fields (resource, code) in addition to substring matching for more resilient classification against API wording changes.

---

ivanmunozruiz:
Aplicados 3 de 4 LOW findings en \`377cda7\`:

✅ **Helper compartido** \`markOrphanIfPermanent(reviewID, err, source)\`: \`Run\` y \`PublishPending\` ahora rutan por el mismo sitio — sin drift posible entre la sentinela \`(-1, \"\")\`, el log shape o la lógica de detección. El parámetro \`source\` deja en el log si el orphan vino de \"initial publish\" o \"retry publish\".

✅ **Body sin truncar en \`*PermanentSubmitError\`**: pasaba la versión safe-truncated (200 bytes), ahora pasa la response completa. La razón puede vivir más allá del cutoff y truncarla silenciaría el diagnóstico que el operador inspecciona cuando recibe el error tipado. La path genérica (\`fmt.Errorf\`) mantiene el safe-truncate para higiene de logs.

✅ **Test directo del classifier**: \`TestSubmitReview_LockedPRBodyIsNotTruncated\` construye un cuerpo de 500 bytes con la substring \"lock prevents review\" plantada más allá de la frontera de truncado. Asserta que \`permErr.Body\` aún contiene la substring → lock-in del contrato.

❌ **No incluido en este PR (filed mentally as follow-up):**

- **JSON-structured field matching** para \`classifyPermanentSubmit422\` (\`errors[].code\` / \`errors[].resource\`). Es una mejora real — el substring matching es frágil ante cambios de wording de GitHub — pero el cambio toca el predicate, requiere parsear el body, y necesita su propio set de tests. Más invasivo que el scope original del PR. Lo abriré como ticket separado si quieres.
- **Shared testutil** para los fakes duplicados (\`fakeExecOrphan\`, \`fakeGHTransientSubmit\`, etc). Refactor amplio que tocaría todos los \`*_test.go\` del paquete pipeline. Mismo razonamiento — out-of-scope para este PR.

Como **comportamiento defensivo**: el substring matching actual ya es fail-closed (un wording change → cae al path transient → la review se intenta de nuevo, sin pérdida). El JSON parsing sería una mejora, no un fix.

\`make test-docker\` verde. CI re-disparado.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

The author addressed 3 of 4 previous findings well: extracted markOrphanIfPermanent helper to eliminate duplication, passes the full (non-truncated) response body in PermanentSubmitError, and added a truncation-boundary regression test. The two deferred items (JSON-structured field matching, shared test fakes) are reasonable follow-ups given fail-closed behavior. The revised code is clean and well-tested. One minor observation remains.

### Issues

🟡 **daemon/internal/github/client.go:278** — The comment says 'Body carries the FULL response (not safe-truncated)' but respBody is still bounded by io.LimitReader(resp.Body, maxBodyBytes). For typical GitHub error payloads this is fine, but the comment is slightly misleading — it's 'full up to maxBodyBytes', not unbounded. A one-word tweak ('full LimitReader-bounded response') would prevent a future reader from assuming there's no upper bound on the Body field's size.

### Suggestions

- Consider opening a tracked ticket for the JSON-structured field matching follow-up (errors[].code) as discussed, so it doesn't get lost — the substring approach is fail-closed but a wording change silently degrades the fix to retry-forever behavior without any signal.
- The redundant condition `resp.StatusCode != http.StatusOK && resp.StatusCode != 200` (line 275) is pre-existing but could be simplified to just `resp.StatusCode != http.StatusOK` in a follow-up cleanup.

---

sergiotejon:
## 🟡 LOW Issue

Pre-existing redundant condition `resp.StatusCode != http.StatusOK && resp.StatusCode != 200` (StatusOK == 200) remains unchanged. Not introduced by this PR but still worth a one-line cleanup.

**Location:** `daemon/internal/github/client.go` line 275

---

sergiotejon:
## ⚠️ MEDIUM Issue

Substring matching on GitHub's human-readable error body remains fragile. Author acknowledges and defers to a follow-up; the fail-closed posture (unknown 422 → transient retry path) is correct, so this is acceptable for this PR but should be tracked. Consider adding a debug-level log when a 422 falls through unclassified to make future wording drift observable.

**Location:** `daemon/internal/github/client.go` line 235

---

sergiotejon:
## 🟡 LOW Issue

Pipeline-level coverage for the 422-without-lock case is still absent. The client-level test exists, but a pipeline test asserting a generic 422 error keeps the row in the unpublished queue would strengthen the fail-closed contract end-to-end. The existing TestPublishPending_TransientErrorStillRetries uses a 503-shaped error, not a 422 without lock substring.

**Location:** `daemon/internal/pipeline/pipeline_orphan_test.go`

---

sergiotejon (approved):

Re-review of #326: the author addressed 3 of 4 prior LOW findings via a shared markOrphanIfPermanent helper, untruncated Body in PermanentSubmitError, and a dedicated truncation-boundary test. The remaining MEDIUM (substring fragility) is acknowledged as follow-up with sound fail-closed reasoning. New code is well-structured, well-documented, and well-tested. Minor residual nits only.

### Suggestions

- Add a debug/info log inside classifyPermanentSubmit422 when a 422 is received but not classified as permanent — cheap observability for detecting GitHub wording drift before it becomes a silent regression.
- Consider extending TestPublishPending_TransientErrorStillRetries (or adding a sibling) to drive a 422 body without the lock substring through the pipeline, locking in that only the classified reason triggers orphan-marking.
- File the deferred follow-ups (JSON-structured errors[].code matching, shared testutil for pipeline fakes) as tracked issues so they are not lost.

---

PR #333, opened 2026-04-24
feat: implement #328 — feat(cli): PRs tab — show GitHub number instead of DB ID, add author and date columns

### LINKED ISSUE #328: feat(cli): PRs tab — show GitHub number instead of DB ID, add author and date columns
## Current state

PRs tab shows internal DB ID (e.g. 634) instead of GitHub PR number (#237). Missing author and review date.

## Improvements

1. **Show PR number** — replace DB ID with `#237` format (the Number field exists in the API)
2. **Add author column** — `Author` field exists in API, not rendered
3. **Add date column** — show when the review was done (`LatestReview.CreatedAt`) or when PR was last updated
4. **Sort by date** — most recent reviews first (currently unclear order)
5. **Add URL hint** — show how to open PR in browser (e.g. press `o` to open)
6. **Filter by repo** — keybinding to filter to a specific repo
7. **Show dismissed badge** — if PR is dismissed, show it

## Files

- `cli/internal/tui/dashboard.go` — PRs rendering

sergiotejon:
Auto-generated by Heimdallm in response to #328.

Closes #328

### DIFF
--- a/cli/internal/tui/dashboard.go
+++ b/cli/internal/tui/dashboard.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os/exec"
+	"runtime"
 	"sort"
 	"strings"
 	"time"
@@ -59,6 +61,8 @@ type Dashboard struct {
 	showDetail   bool
 	detailScroll int
 	detailLines  []string
+
+	repoFilter string
 }
 
 type activityLine struct {
@@ -230,6 +234,9 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 					d.prs = append(d.prs, pr)
 				}
 			}
+			sort.Slice(d.prs, func(i, j int) bool {
+				return d.prs[i].LatestReview.CreatedAt.After(d.prs[j].LatestReview.CreatedAt)
+			})
 			d.issues = nil
 			for _, iss := range msg.issues {
 				if iss.LatestReview != nil {
@@ -375,10 +382,29 @@ func (d *Dashboard) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 			}
 		}
 	case "enter":
-		if d.activeTab == tabPRs && d.cursor < len(d.prs) {
+		if d.activeTab == tabPRs && d.cursor < len(d.visiblePRs()) {
 			d.openDetail()
 		} else if d.activeTab == tabIssues && d.cursor < len(d.issues) {
 			d.openDetail()
+		}
+	case "o":
+		if d.activeTab == tabPRs {
+			prs := d.visiblePRs()
+			if d.cursor < len(prs) && prs[d.cursor].URL != "" {
+				return d, openURLCmd(prs[d.cursor].URL)
+			}
+		}
+	case "f":
+		if d.activeTab == tabPRs {
+			if d.repoFilter != "" {
+				d.repoFilter = ""
+			} else {
+				prs := d.visiblePRs()
+				if d.cursor < len(prs) {
+					d.repoFilter = prs[d.cursor].Repo
+				}
+			}
+			d.cursor = 0
 		}
 	case "r":
 		d.refreshing = true
@@ -442,7 +468,7 @@ func (d *Dashboard) openDetail() {
 	d.detailScroll = 0
 	switch d.activeTab {
 	case tabPRs:
-		d.detailLines = buildPRDetailLines(d.prs[d.cursor], d.width)
+		d.detailLines = buildPRDetailLines(d.visiblePRs()[d.cursor], d.width)
 	case tabIssues:
 		d.detailLines = buildIssueDetailLines(d.issues[d.cursor], d.width)
 	}
@@ -500,6 +526,35 @@ func (d *Dashboard) clampCursor() {
 	}
 	if d.cursor < 0 {
 		d.cursor = 0
+	}
+}
+
+func (d *Dashboard) visiblePRs() []api.PR {
+	if d.repoFilter == "" {
+		return d.prs
+	}
+	var filtered []api.PR
+	for _, pr := range d.prs {
+		if pr.Repo == d.repoFilter {
+			filtered = append(filtered, pr)
+		}
+	}
+	return filtered
+}
+
+func openURLCmd(url string) tea.Cmd {
+	return func() tea.Msg {
+		var cmd *exec.Cmd
+		switch runtime.GOOS {
+		case "darwin":
+			cmd = exec.Command("open", url)
+		case "windows":
+			cmd = exec.Command("rundll32", "url.dll,FileProtocolHandler", url)
+		default:
+			cmd = exec.Command("xdg-open", url)
+		}
+		_ = cmd.Start()
+		return nil
 	}
 }
 
@@ -652,43 +707,63 @@ func (d *Dashboard) renderActivity(height int) string {
 }
 
 func (d *Dashboard) renderPRs(height int) string {
-	if len(d.prs) == 0 {
+	prs := d.visiblePRs()
+	if len(prs) == 0 {
+		if d.repoFilter != "" {
+			return lipgloss.NewStyle().Foreground(colorMuted).Render(
+				fmt.Sprintf("  No PRs matching filter: %s  [f to clear]", d.repoFilter))
+		}
 		return lipgloss.NewStyle().Foreground(colorMuted).Render("  No PRs found.")
 	}
 
 	var b strings.Builder
-	header := fmt.Sprintf("  %-6s %-25s %-35s %-8s %-8s", "ID", "REPO", "TITLE", "SEVERITY", "STATE")
+	header := fmt.Sprintf("  %-7s %-18s %-24s %-12s %-8s %-10s %s", "PR", "REPO", "TITLE", "AUTHOR", "SEVERITY", "REVIEWED", "STATE")
+	if d.repoFilter != "" {
+		header += "  " + lipgloss.NewStyle().Foreground(colorWarning).Render(fmt.Sprintf("[filter: %s]", d.repoFilter))
+	}
 	b.WriteString(headerStyle.Render(header))
 	b.WriteString("\n")
-	b.WriteString("  " + strings.Repeat("─", 86))
+	b.WriteString("  " + strings.Repeat("─", 90))
 	b.WriteString("\n")
 
 	maxVisible := height - 2
 	if maxVisible < 1 {
 		maxVisible = 1
 	}
-	start, end := visibleRange(d.cursor, len(d.prs), maxVisible)
+	start, end := visibleRange(d.cursor, len(prs), maxVisible)
 
 	for i := start; i < end; i++ {
-		pr := d.prs[i]
+		pr := prs[i]
 		sev := "---"
+		reviewed := "---"
 		if pr.LatestReview != nil {
 			sev = pr.LatestReview.Severity
+			reviewed = pr.LatestReview.CreatedAt.Format("Jan 02")
 		}
-		title := truncateRunes(pr.Title, 33)
-		repo := truncateRunes(pr.Repo, 23)
+		title := truncateRunes(pr.Title, 22)
+		repo := truncateRunes(pr.Repo, 16)
+		author := truncateRunes(pr.Author, 10)
+		prNum := fmt.Sprintf("#%d", pr.Number)
 		sevRendered := severityStyle(sev).Render(fmt.Sprintf("%-8s", sev))
-		line := fmt.Sprintf("  %-6d %-25s %-35s %s %-8s", pr.ID, repo, title, sevRendered, pr.State)
+
+		state := pr.State
+		if pr.Dismissed {
+			state += " ⊘"
+		}
+
+		line := fmt.Sprintf("  %-7s %-18s %-24s %-12s %s %-10s %s", prNum, repo, title, author, sevRendered, reviewed, state)
 
 		if i == d.cursor {
 			b.WriteString(selectedRowStyle.Render(line))
+		} else if pr.Dismissed {
+			b.WriteString(lipgloss.NewStyle().Foreground(colorMuted).Render(line))
 		} else {
 			b.WriteString(line)
 		}
 		b.WriteString("\n")
 	}
 
-	if ind := scrollIndicator(start, end, len(d.prs)); ind != "" {
+	if ind := scrollIndicator(start, end, len(prs)); ind != "" {
 		b.WriteString(lipgloss.NewStyle().Foreground(colorMuted).Render(ind))
 	}
 	return b.String()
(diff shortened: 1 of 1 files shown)


ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

The PR cleanly adds repo filtering, browser-open, PR-number display, and author/date columns. Logic is mostly sound, but there is a potential nil-pointer dereference in the new sort and a few smaller ergonomic/robustness issues.

### Issues

🔴 **cli/internal/tui/dashboard.go:237** — sort.Slice dereferences d.prs[i].LatestReview.CreatedAt unconditionally. The surrounding code (e.g. renderPRs at line ~725 and the symmetric issues loop that explicitly guards with `if iss.LatestReview != nil`) suggests a PR with nil LatestReview is possible. If any such PR reaches d.prs, the sort will panic. Either guard the PR append with `if pr.LatestReview != nil` (mirroring issues) or make the less-func nil-safe (treat nil as zero time).
🟡 **cli/internal/tui/dashboard.go:553** — openURLCmd silently swallows exec.Start errors with `_ = cmd.Start()`. If the URL fails to open (missing xdg-open on Linux, sandboxed env, etc.) the user gets no feedback. Consider surfacing the error via a tea.Msg/status line instead of dropping it.
🟡 **cli/internal/tui/dashboard.go:958** — The new PRs-tab help string drops `[pgup/pgdn]page` that the previous shared line (and the Issues tab) advertises, even though Update() still handles pgup/pgdn. Either keep the binding documented or remove the handler — don't let help drift from behavior.
🟡 **cli/internal/tui/dashboard.go:720** — Column widths vs. the separator are slightly inconsistent: the new format uses 2+7+1+18+1+24+1+12+1+8+1+10+1+state ≈ 93+ chars, while the separator is 90 `─`. Minor cosmetic drift; consider computing or aligning both from a shared constant.
🟡 **cli/internal/tui/dashboard.go:399** — `f` toggles the filter, but when clearing it the cursor is reset to 0 rather than tracking the previously selected PR. For large lists this is a mild UX regression — consider preserving the highlighted PR across toggle when possible.
🟡 **cli/internal/tui/dashboard.go:391** — Pressing `o` on a PR with empty URL silently no-ops. Worth a brief status message to make the inertness discoverable, especially now that `o` is advertised in the help line.

### Suggestions

- Make the new sort.Slice nil-safe or explicitly filter nil-LatestReview PRs at append time to match the issues handling.
- Plumb openURLCmd errors back as a tea.Msg and render a transient status line (and consider validating the URL scheme is http/https before launching an external handler).
- Extract the PR row format string + separator width into named constants so header/rows/separator stay in sync.
- Add a small unit test for visiblePRs() covering empty filter, matching filter, and non-matching filter to lock in behavior.

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Adds PR number, author, reviewed-date columns, repo filtering, open-in-browser action, and sort by latest review. Functionality looks reasonable, but there are a few correctness and robustness concerns around nil dereferences, process handling, and help text.

### Issues

🔴 **cli/internal/tui/dashboard.go:237** — sort.Slice dereferences d.prs[i].LatestReview.CreatedAt without a nil guard. The surrounding append loop in the diff does not clearly filter out PRs with LatestReview == nil (unlike the issues loop below it which does), so this will panic if any PR lacks a latest review. Either filter nil reviews before sorting or guard inside the less-func.
⚠️ **cli/internal/tui/dashboard.go:552** — openURLCmd calls cmd.Start() and returns immediately without cmd.Wait() or reaping. On Linux/macOS this leaks a zombie child process for every 'o' press. Either Wait in a goroutine, set Process.Release, or use cmd.Run inside the tea.Cmd goroutine.
🟡 **cli/internal/tui/dashboard.go:958** — Help text for the PRs tab no longer advertises [pgup/pgdn]page, though the handler still supports paging. Inconsistent with the Issues/Activity help lines and regresses discoverability.
🟡 **cli/internal/tui/dashboard.go:532** — visiblePRs() is invoked multiple times per render (renderPRs, tabItemCount, handlers) and rebuilds the filtered slice each call. Consider caching the filtered slice on state change, or at minimum share one snapshot per Update/View cycle.
🟡 **cli/internal/tui/dashboard.go:741** — State column may contain the multi-byte rune '⊘' (' ⊘' suffix). Because fmt '%-Ns' pads by rune count but terminal display width varies, the last column alignment can drift for dismissed rows. Consider using lipgloss width-aware padding or a fixed ASCII marker.
🟡 **cli/internal/tui/dashboard.go:748** — Author is truncated to 10 runes but formatted with %-12s, leaving a 2-space gap that is inconsistent with other columns where truncate width equals format width. Harmless but worth aligning.

### Suggestions

- Add a nil check before sorting, e.g. filter PRs with LatestReview == nil out (or sort them to the end) to avoid a latent panic.
- In openURLCmd, spawn a goroutine that calls cmd.Wait() (or use cmd.Run) so child processes are reaped; optionally redirect stdout/stderr to /dev/null to avoid terminal corruption under xdg-open.
- Restore [pgup/pgdn]page in the PRs help string to match the Issues tab.
- Compute visiblePRs() once per Update/View pass and pass the slice down, rather than recomputing in every helper.
- Consider preserving cursor position when toggling the repo filter off (return to the previously focused PR) for better UX.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean PR that delivers what it claims: replaces DB ID with GitHub PR number, adds author/date columns, implements repo filtering and browser-open features. Code follows existing Bubble Tea patterns. No blocking issues found.

### Suggestions

- openURLCmd (line ~545): `_ = cmd.Start()` silently discards the error. If the browser command fails (e.g., xdg-open not installed on Linux), the user gets zero feedback. Consider returning a tea.Msg with the error so the dashboard can show a brief status message.
- visiblePRs() allocates a new filtered slice on every call when repoFilter is set. It's called multiple times per render/key cycle. Consider caching the result (invalidate on data refresh or filter change) to avoid repeated allocations, especially as the PR list grows.
- The separator line width is hardcoded to 90 (`strings.Repeat("─", 90)`). If terminal width is narrower than the header, the line will overflow. Consider deriving separator width from `d.width` or at least from the header format string length.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: the HIGH nil-pointer risk in the PR sort is addressed via nil guards in the less-func. However, the MEDIUM zombie-process issue in openURLCmd and several LOW findings remain unchanged in this diff.

### Issues

⚠️ **cli/internal/tui/dashboard.go:559** — openURLCmd still calls cmd.Start() and returns without cmd.Wait() or Process.Release(). On Linux/macOS this leaks a zombie child for every 'o' press. Use cmd.Run() inside the tea.Cmd goroutine, or spawn a goroutine that calls Wait, or call Process.Release() after Start.
🟡 **cli/internal/tui/dashboard.go:966** — PRs tab help line dropped [pgup/pgdn]page even though the handler still supports paging; Issues/Activity help lines still advertise it. Regresses discoverability and diverges from the other tabs.
🟡 **cli/internal/tui/dashboard.go:539** — visiblePRs() is still rebuilt on every call (renderPRs, tabItemCount, enter/o/f handlers). For any PRs change or a repoFilter toggle, cache the filtered slice (e.g. recompute once per Update and store on the struct) to avoid repeated O(n) scans per frame.
🟡 **cli/internal/tui/dashboard.go:751** — The State column still appends the multi-byte rune ' ⊘' for dismissed PRs. fmt width specifiers pad by rune count, not terminal cell width, so dismissed rows may misalign in some terminals. Use lipgloss width-aware padding or an ASCII marker.
🟡 **cli/internal/tui/dashboard.go:746** — Author is truncated to 10 runes but formatted with %-12s, leaving a 2-space gap inconsistent with other columns where truncate width matches format width. Align truncate/format widths (e.g. 12/12 or 10/10).

### Suggestions

- For openURLCmd, prefer `if err := cmd.Start(); err == nil { go cmd.Wait() }` or `cmd.Process.Release()` to avoid child reaping issues on *nix.
- Consider computing `d.filteredPRs` once at the end of Update (when d.prs or d.repoFilter changes) and reading from it in render/handlers to keep rendering O(n) per frame instead of O(n·k).
- Restore [pgup/pgdn]page in the PRs help string so all list tabs document the same navigation shortcuts.
- When clearing repoFilter via 'f', consider preserving the cursor by mapping the currently selected PR back to its index in the full list, instead of always resetting to 0.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean implementation of PR tab improvements: GitHub number display, author/date columns, repo filtering, and browser open. No blocking issues found. Minor suggestions for robustness.

### Suggestions

- openURLCmd (line ~551): The error from cmd.Start() is silently discarded. Consider returning a tea.Msg with the error so the dashboard can show a brief status message if the browser fails to open.
- visiblePRs() is called multiple times per render/key-handling cycle, each time allocating a new filtered slice. For large PR lists, consider caching the result and invalidating on data change or filter toggle.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

HIGH nil-deref in sort.Slice was addressed in the follow-up commit (ri/rj nil guards now present). The remaining LOW findings from the prior review are still unaddressed in the current diff: silent error swallowing in openURLCmd, pgup/pgdn dropped from PRs help text despite handler still firing, column/separator width drift, cursor reset to 0 on filter toggle, and silent no-op when pressing 'o' on a PR with empty URL. No new regressions introduced.

### Issues

🟡 **cli/internal/tui/dashboard.go:562** — openURLCmd still discards cmd.Start() error with `_ = cmd.Start()`. On sandboxed/headless Linux without xdg-open, or in SSH sessions without a DISPLAY, the user presses `o` and nothing visibly happens. Return a status tea.Msg on error so the UI can surface the failure.
🟡 **cli/internal/tui/dashboard.go:966** — PRs-tab help string drops `[pgup/pgdn]page` that the Issues tab still advertises, yet handleKey (earlier in Update) still processes pgup/pgdn for PRs. Either re-add to the help line or remove the handler to keep help and behaviour in sync.
🟡 **cli/internal/tui/dashboard.go:727** — Column widths (`%-7s %-18s %-24s %-12s %-8s %-10s %s` ≈ 87+state chars incl. separators) and the 90-`─` separator drift by a few columns. Derive both from shared constants to keep them locked together as future edits happen.
🟡 **cli/internal/tui/dashboard.go:411** — On filter toggle (`case "f"`) the cursor is always reset to 0, including when clearing the filter. Preserving the currently highlighted PR across toggle — e.g. by remembering its Number/ID and re-seeking after recomputing the slice — would avoid the UX regression in large repos.
🟡 **cli/internal/tui/dashboard.go:396** — Pressing `o` on a PR with empty URL silently no-ops. Since `o` is now advertised in the help line, the inertness is surprising — emit a brief status/tea.Msg so the user knows the action was received but the URL is missing.

### Suggestions

- Consider factoring the PR-table column layout into a small struct (name,width) so the header, row formatter, and separator width all derive from one source of truth.
- Route openURLCmd errors through the same status/footer mechanism used elsewhere, so both 'no URL' and 'open failed' share a single visible feedback path.
- When `f` clears the filter, snapshot the current PR identifier (Number+Repo) before recomputing `visiblePRs()` and re-seek to it in the unfiltered list; fall back to 0 if not found.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean UI enhancement that adds PR number, author, date columns, repo filtering, and browser-open functionality. No blocking issues found. The code follows existing Bubble Tea patterns and the sorting/filtering logic is correct.

### Suggestions

- openURLCmd silently discards the error from cmd.Start(). Consider returning a tea.Msg with the error so the status bar can show feedback if the browser fails to open.
- visiblePRs() is called multiple times per render cycle (renderPRs, tabItemCount, clampCursor, handleKey). For large PR lists with an active filter, caching the filtered result once per Update cycle would avoid redundant allocations — though this is unlikely to matter at typical PR counts.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

PR adds useful TUI features (PR # display, author/date columns, repo filter, browser open, sort by review date) but introduces several bugs around cursor/index handling and a destructive sort that mutates the underlying slice used for filtering.

### Issues

⚠️ **cli/internal/tui/dashboard.go:237** — sort.Slice comparator is incorrect: when ri == nil it returns false unconditionally, but it should return false only when rj is also non-nil (so nil entries are pushed to the end). As written, two nil entries return false for both (i,j) and (j,i) which violates strict weak ordering and may produce unstable/incorrect ordering; more importantly, a nil ri vs non-nil rj returns false, placing nils before non-nils — the opposite of the intended 'most recent first, nils last'.
🟡 **cli/internal/tui/dashboard.go:392** — Enter handler uses len(d.visiblePRs()) for bounds check but openDetail() indexes d.visiblePRs()[d.cursor] again — calling visiblePRs() twice allocates a new filtered slice each time. Minor perf/consistency issue; cache the slice once and reuse.
⚠️ **cli/internal/tui/dashboard.go:408** — Toggling repoFilter resets cursor to 0 but does not reset detailScroll or close an open detail view. If the user has the detail view open and presses 'f', the detail will reference the wrong PR or panic if the filtered list shrinks below cursor index.
⚠️ **cli/internal/tui/dashboard.go:414** — openURLCmd uses exec.Command with a URL passed directly to 'open'/'xdg-open'/'rundll32'. While Start() doesn't go through a shell, a URL beginning with '-' could be interpreted as a flag by these handlers. Consider validating the URL scheme (http/https) before launching, or passing '--' separator where supported.
🟡 **cli/internal/tui/dashboard.go:422** — openURLCmd silently ignores errors from cmd.Start(). User pressing 'o' on a system without xdg-open gets no feedback. Consider surfacing a status message via tea.Msg.
🟡 **cli/internal/tui/dashboard.go:244** — After sorting d.prs, the cursor is not clamped/reset, so if the user was navigating PRs and a refresh re-sorts them, the cursor now points at a different PR than before with no visual indication.
🟡 **cli/internal/tui/dashboard.go:730** — Header column widths (PR=7, REPO=18, TITLE=24, AUTHOR=12, SEVERITY=8, REVIEWED=10) sum + separators exceed the 90-char separator line, and truncation widths (repo=16, title=22, author=10) leave inconsistent spacing. Verify the layout aligns at typical terminal widths and update the separator length accordingly.
🟡 **cli/internal/tui/dashboard.go:754** — pr.Dismissed field is referenced but not visible in the diff context — ensure the api.PR struct actually has a Dismissed field, otherwise this won't compile. Same for pr.Author and pr.Number being referenced as new columns.
🟡 **cli/internal/tui/dashboard.go:965** — PR help text drops [pgup/pgdn]page and [G]follow that are present in other tabs' help, but those keybindings may still be functional. Inconsistent help text confuses users about which keys work where.

### Suggestions

- Fix sort comparator to: if ri == nil && rj == nil { return false }; if ri == nil { return false }; if rj == nil { return true }; return ri.CreatedAt.After(rj.CreatedAt) — current logic places nils incorrectly.
- Cache d.visiblePRs() result in a local variable in handleKey to avoid repeated O(n) filtering on each keystroke and ensure index consistency.
- Validate URL scheme (https?://) in openURLCmd before invoking the OS handler to prevent argument injection via a maliciously crafted URL.
- When toggling repoFilter, also close any open detail view (d.showDetail = false) to avoid stale references.
- Consider extracting the column layout constants (widths, separator length) to named constants to keep header, separator, and row formatting in sync.
- Add a unit test for visiblePRs() and the new sort comparator to lock in the intended ordering and filter semantics.

---

sergiotejon:

El issue origen #328 queda abierto para que Heimdallm regenere la implementación contra el estado actual de `main`.

PR #336, opened 2026-04-24
feat: implement #331 — feat(cli): Config tab — show all fields, sections, expand repos and maps

### LINKED ISSUE #331: feat(cli): Config tab — show all fields, sections, expand repos and maps
## Current state

Config tab shows only 7 hardcoded keys, truncates arrays/maps to 60 chars making them unreadable. Missing most of the config surface.

### 1. Show ALL config fields
Add missing: `discovery_topic`, `discovery_orgs`, `discovery_interval`, `local_dir_base`, `execution_timeout`, `agent_configs`, `repo_overrides`, `effective_repositories`, `local_dirs_detected`, `non_monitored`, `pr_reviewers`, `pr_labels`, `pr_assignee`.

### 2. Organize by sections
```
── Server ──────────────────────
  Port:           7842
  Bind:           0.0.0.0
  Poll interval:  1m

── Repositories (18) ───────────
  • theburrowhub/heimdallm
  • freepik-company/ai-bumblebee-proxy
  • ...

── Discovery ───────────────────
  Topic:    heimdallm-review
  Orgs:     freepik-company, theburrowhub
  Interval: 15m

── AI ──────────────────────────
  Primary:   claude
  Fallback:  —
  Mode:      single
  Timeout:   30m

── Issue Tracking ──────────────
  Enabled:      true
  Filter mode:  exclusive
  Develop:      enhancement, feature, bug
  Review only:  question, discussion, analysis
  Skip:         wontfix, duplicate, invalid

── PR Metadata (defaults) ──────
  Reviewers: Muriano, ivanmunozruiz, vbuenog
  Labels:    auto-generated
  Assignee:  sergiotejon
```

### 3. Expand repos as bullet list
Do not truncate — show all repos, one per line.

### 4. Color keys vs values
Keys in muted color, values in bright. Sections with colored headers.

### 5. Scrollable
Already scrollable — just needs more content to scroll through.

## Files

- `cli/internal/tui/dashboard.go` — `buildConfigLines()` and `renderConfig()`

sergiotejon:
Auto-generated by Heimdallm in response to #331.

Closes #331

### DIFF
--- a/cli/internal/tui/dashboard.go
+++ b/cli/internal/tui/dashboard.go
@@ -744,18 +744,13 @@ func (d *Dashboard) renderConfig(height int) string {
 		return lipgloss.NewStyle().Foreground(colorMuted).Render("  No configuration loaded.")
 	}
 
-	var b strings.Builder
-	b.WriteString(headerStyle.Render("  Configuration"))
-	b.WriteString("\n")
-	b.WriteString("  " + strings.Repeat("─", 60))
-	b.WriteString("\n")
-
 	lines := d.buildConfigLines()
 	if len(lines) == 0 {
-		return b.String()
+		return lipgloss.NewStyle().Foreground(colorMuted).Render("  No configuration loaded.")
 	}
 
-	maxVisible := height - 2
+	var b strings.Builder
+	maxVisible := height
 	if maxVisible < 1 {
 		maxVisible = 1
 	}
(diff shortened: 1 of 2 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Mostly a presentational refactor that expands the Config tab with sections and helpers. Behavior is largely additive and safe, but a few issues stand out: aggressive value filtering can hide meaningful config, some dead code is left behind, and long values are no longer truncated at all.

### Issues

⚠️ **cli/internal/tui/dashboard.go:1145** — Filtering with `vs == "" || vs == "0" || vs == "false"` in the Agent Configs and Repo Overrides sections hides values that are often meaningful — e.g. an explicit `enabled: false`, `pr_draft: false`, `timeout: 0` (‘disabled’), or `retries: 0`. A user auditing overrides would silently not see disabled features. Prefer only skipping missing/empty-string values, or make the skip rules per-key rather than string-based.
🟡 **cli/internal/tui/dashboard.go:1215** — `formatConfigValue` is no longer called anywhere in the new `buildConfigLines` (which uses its own `fmtVal` closure). After this change it's effectively dead code; remove it or wire `fmtVal` to call through it to avoid two near-identical formatters drifting apart.
🟡 **cli/internal/tui/dashboard.go:1000** — Truncation was removed from `formatConfigValue` (60-rune cap) and the new `fmtVal`/kv helpers don't cap anywhere either. Long values — `issue_prompt`/`implement_prompt` paths, a large `discovery_orgs` list, or a JSON-marshaled map — will wrap/overflow on narrow terminals. Consider truncating at render time based on viewport width, or marking the line as collapsible.
🟡 **cli/internal/tui/dashboard.go:752** — After the refactor, `buildConfigLines` always appends at least the Server section when `d.config != nil`, so the `if len(lines) == 0` early return in `renderConfig` is unreachable. Either drop the dead branch or guard `buildConfigLines` to actually return nil on a completely empty payload.
🟡 **cli/internal/tui/dashboard.go:1115** — In the Local Directories section, `fmt.Sprintf("      %s → %v", name, localDetected[name])` uses raw `%v` instead of the `fmtVal` helper. If a detected entry is a nested map/slice it will render in Go's native `map[key:val]` syntax rather than the JSON format used everywhere else, producing inconsistent UI.
🟡 **cli/internal/tui/dashboard.go:747** — The old `  Configuration` header and separator line were removed from `renderConfig`. If the surrounding tab chrome doesn't already render a title for this tab, the pane becomes headerless and inconsistent with other tabs. Worth confirming the parent layout still labels the section.
🟡 **cli/internal/tui/dashboard.go:960** — `section()` computes padding as `40 - len([]rune(label))`. That's fine for the current ASCII labels, but the count uses rune length, not terminal cell width; if a future section title contains any wide/emoji glyph the separator length will be off. Minor but easy to lock in now by using `runewidth.StringWidth` (lipgloss already depends on it transitively).

### Suggestions

- Tighten the skip rule in Agent Configs / Repo Overrides to only omit empty strings (and perhaps `nil`), so explicit booleans and zero values remain visible — or make the filter per-key based on known defaults.
- Delete `formatConfigValue` now that `fmtVal` supersedes it, or have `fmtVal` delegate to it so they can't diverge.
- Reintroduce a width-aware truncation (based on `height`/viewport width) for long string values to avoid layout breakage on narrow terminals.
- Use `fmtVal` (not `%v`) for nested values in the Local Directories detected section to keep JSON-style rendering consistent.
- Confirm whether removing the `  Configuration` header was intentional; if the tab chrome doesn't provide a title, add it back or move it to the parent renderer.
- Consider extracting the repeated ‘sort map keys, render bold name, iterate entries, apply filter’ block shared by Agent Configs and Repo Overrides into a helper — they are nearly identical.

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Config tab expansion is a clear UX improvement, but the agent_configs and repo_overrides sections silently drop any field whose value stringifies to '0' or 'false', which hides meaningful configuration. A few minor redundancy/robustness issues also exist.

### Issues

🔴 **cli/internal/tui/dashboard.go:1155** — `if vs == "" || vs == "0" || vs == "false" { continue }` hides valid config values. A field explicitly set to false/0 (e.g. an `enabled: false`, `max_concurrency: 0`, `pr_draft: false`) is meaningful and should be visible — otherwise users can't tell from the Config tab whether a flag is unset vs. deliberately disabled. Only the empty-string case should be skipped (or render '—').
🔴 **cli/internal/tui/dashboard.go:1175** — Same filtering bug duplicated in the `repo_overrides` block; same fix needed.
⚠️ **cli/internal/tui/dashboard.go:1213** — `formatConfigValue` no longer truncates slices/maps to 60 runes. With the new Config tab layout this is intended, but `formatConfigValue` still has other callers (grep the file) and now may emit very long/multi-line JSON into contexts that previously relied on truncation to keep the UI from wrapping. Confirm all remaining callers tolerate untruncated output, or keep truncation here and let `buildConfigLines` use its own `fmtVal`.
🟡 **cli/internal/tui/dashboard.go:751** — The `len(lines) == 0` branch now returns 'No configuration loaded.', but `buildConfigLines` always appends at least the Server section headers when `d.config != nil`, so this branch is effectively unreachable and misleading if ever hit. Either drop the branch or return a distinct message like 'Empty configuration.'
🟡 **cli/internal/tui/dashboard.go:935** — `strSlice` only matches `[]any`; if any producer ever puts a `[]string` directly into `d.config` (e.g. a future in-process path that bypasses JSON round-trip) the field will silently render as empty. Consider handling `[]string` too, or assert the config source is always JSON-decoded.
🟡 **cli/internal/tui/dashboard.go:966** — `section()` computes padding from the raw label string before `headerStyle.Render`, which is fine for width, but if `headerStyle` ever includes background color the trailing spaces after the label name will be styled differently from the `─` fill. Minor cosmetic risk — consider padding then rendering the full line once.
🟡 **cli/internal/tui/dashboard.go:913** — `keyStyle` and `boldStyle` are re-instantiated on every call to `buildConfigLines`; cheap, but they belong at package scope alongside the other style vars for consistency.
🟡 **cli/internal/tui/dashboard.go:1083** — `Log retention` appends ' days' only when the string is non-empty, but `retention_days` above is rendered raw without a unit — inconsistent presentation for two closely-related fields.

### Suggestions

- In the agent_configs/repo_overrides loops, skip only empty strings (or render '—') so `false`/`0` values remain visible; if the intent is to hide defaults, compare against the documented default rather than string literals.
- Add a tiny unit test for `buildConfigLines` with a representative `map[string]any` fixture (including an `enabled: false` and `pr_draft: false`) to lock in the rendered output and prevent future filtering regressions.
- Consider moving `keyStyle`, `boldStyle`, and the `section`/`kv`/`kvWide` helpers to package level — they're pure formatting and will be reused as the config tab grows.
- Unify units handling: either always suffix durations/days with their unit, or never (and label the field accordingly).
- Optional: when `execution_timeout` or `poll_interval` arrive as numeric seconds, format them as `time.Duration` strings for readability.

---

Muriano:
## ⚠️ MEDIUM Issue

Falsy-value filter hides legitimate config: the condition `vs == "" || vs == "0" || vs == "false"` suppresses display of agent config and repo override fields that are intentionally set to 0 or false (e.g., a timeout of 0, or an explicit disable flag). Users will not see these values in the Config tab, which defeats the purpose of 'show all fields'. Fix: only skip empty strings, or display all keys unconditionally and let the user judge.

**Location:** `cli/internal/tui/dashboard.go` line 1155

---

Muriano (approved):

UI-focused PR that restructures the Config tab to show all configuration fields in organized sections. The daemon correctly exposes additional config fields. One confirmed logic bug in display filtering, and no tests for the new rendering logic.

### Suggestions

- The new `buildConfigLines` function is ~200 lines with repetitive patterns for each section. Consider extracting a small helper that takes a list of key-label pairs and a source map, reducing duplication and making it trivial to add new sections later.
- No unit tests were added for `buildConfigLines`. A table-driven test feeding various `d.config` maps (nil values, empty slices, nested maps) would catch regressions in the rendering logic.
- The removal of `truncateRunes` in `formatConfigValue` means arbitrarily long JSON blobs or lists render without bounds. If a repo override contains a large nested object, it could break the TUI layout. Consider a max-width or ellipsis for inline values.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Re-review: the falsy-value filter in Agent Configs / Repo Overrides is still unchanged despite prior feedback and a MEDIUM comment from @Muriano. Several LOW findings (dead formatConfigValue, no truncation, unreachable empty branch, %v for localDetected, missing header) also remain unaddressed. No new issues introduced by this iteration.

### Issues

⚠️ **cli/internal/tui/dashboard.go:1145** — STILL UNFIXED: the filter `if vs == "" || vs == "0" || vs == "false" { continue }` in the Agent Configs section (and the identical one in Repo Overrides below) silently hides values that are often meaningful — an explicit `enabled: false`, `pr_draft: false`, `timeout: 0`, `retries: 0`, etc. This defeats the stated goal of the PR (‘show all fields’). Also flagged by @Muriano. Skip only truly empty strings, or drop the filter entirely and display every key.
⚠️ **cli/internal/tui/dashboard.go:1166** — Same falsy-value filter duplicated in the Repo Overrides block; must be fixed together with the Agent Configs one.
🟡 **cli/internal/tui/dashboard.go:1215** — `formatConfigValue` is still dead code — `buildConfigLines` uses its own `fmtVal` closure and no other caller remains. Remove it or route `fmtVal` through it to prevent two formatters from drifting.
🟡 **cli/internal/tui/dashboard.go:1000** — No viewport-based truncation anywhere. Long paths (`issue_prompt`, `implement_prompt`), a large `discovery_orgs` list, or JSON-marshaled maps will wrap/overflow on narrow terminals. Consider truncating at render time or marking lines collapsible.
🟡 **cli/internal/tui/dashboard.go:752** — `buildConfigLines` now always appends at least the Server section when `d.config != nil`, so `if len(lines) == 0` in `renderConfig` is unreachable. Drop the dead branch or guard `buildConfigLines` to return nil on a truly empty payload.
🟡 **cli/internal/tui/dashboard.go:1115** — Local Directories block still uses raw `%v` for `localDetected[name]`. If a detected entry becomes a map/slice, output will render in Go's native `map[k:v]` form instead of the JSON shape used elsewhere. Prefer `fmtVal(localDetected[name])`.
🟡 **cli/internal/tui/dashboard.go:747** — The old `  Configuration` header and separator were removed and not replaced. Confirm the parent tab chrome still labels this pane; otherwise the Config tab becomes headerless and inconsistent with the other tabs.

### Suggestions

- Address the MEDIUM falsy-filter issue (also raised by @Muriano) — it is the highest-impact change remaining.
- Once the filter is fixed, delete `formatConfigValue` or collapse it and `fmtVal` into a single helper.
- Add a render-time width cap (using lipgloss/runewidth) to avoid overflow on narrow terminals for long prompt paths and JSON maps.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Re-review: the previously-flagged MEDIUM falsy-value filter (`vs == "0" || vs == "false"`) was fixed in both Agent Configs and Repo Overrides — the filter now only skips empty strings, so explicit `false`/`0` values are shown. Remaining items are the low-severity issues from the prior review that the author did not address.

### Issues

🟡 **cli/internal/tui/dashboard.go:1215** — STILL UNFIXED: `formatConfigValue` remains dead code — `buildConfigLines` uses its own `fmtVal` closure and no caller of `formatConfigValue` remains in the new rendering path. The function was even touched in this diff (truncateRunes calls removed) without a caller, which increases the drift risk between the two formatters. Remove `formatConfigValue` or route `fmtVal` through it.
🟡 **cli/internal/tui/dashboard.go:752** — STILL UNFIXED: `buildConfigLines` always appends at least the Server section header + 3 kv + blank line whenever `d.config != nil` (already guaranteed upstream), so the `if len(lines) == 0 { return "No configuration loaded." }` branch in `renderConfig` is unreachable. Either drop the dead branch or guard `buildConfigLines` to return nil when the payload is truly empty.
🟡 **cli/internal/tui/dashboard.go:1115** — STILL UNFIXED: the Local Directories block renders detected entries with raw `%v` (`fmt.Sprintf("      %s → %v", name, localDetected[name])`). If a detected entry ever becomes a map/slice, it will render in Go's native `map[k:v]` form rather than the JSON shape used everywhere else. Use `fmtVal(localDetected[name])` for consistency.
🟡 **cli/internal/tui/dashboard.go:747** — STILL UNFIXED: the previous `  Configuration` header + separator line were removed and not replaced. If the parent tab chrome does not already label this pane, the Config tab is now headerless and inconsistent with the other tabs. Confirm visually or re-add a section-level header.
🟡 **cli/internal/tui/dashboard.go:1000** — STILL UNFIXED: no viewport-aware truncation. Long values (`issue_prompt`, `implement_prompt` paths, big `discovery_orgs` lists, JSON-marshaled maps inside agent_configs/repo_overrides) will wrap or overflow on narrow terminals — and now that the falsy-value filter is gone and more fields are displayed, the risk of overflow is higher. Consider truncating at render time using the pane width.
🟡 **cli/internal/tui/dashboard.go:760** — `maxVisible := height` (was `height - 2`) together with the removal of the header/separator means the scrollable region now assumes two more rows of vertical budget than before. If the parent layout still reserves those two rows for tab chrome elsewhere, the bottom of the Config content can be clipped. Verify that `height` passed to `renderConfig` already excludes tab chrome, otherwise restore the `-2` budget or recompute based on the actual available rows.

### Suggestions

- Delete `formatConfigValue` now that `buildConfigLines` no longer calls it; keeping two near-identical formatters invites future drift.
- Replace the raw `%v` rendering of `localDetected[name]` with `fmtVal(...)` so nested maps/slices are shown as JSON like elsewhere.
- Remove the unreachable `if len(lines) == 0` branch in `renderConfig`, or make `buildConfigLines` return nil on a genuinely empty config.
- Add width-aware truncation (reuse `truncateRunes` with the pane width) for value cells to prevent wrapping on narrow terminals.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

El issue MEDIUM anterior (filtro de valores falsy con '0'/'false') ha sido corregido — ahora solo se omiten strings vacíos. El resto del diff es una reescritura limpia de la UI del Config tab: secciones estructuradas, expansión de maps/slices, y exposición de campos adicionales del daemon. No se detectan issues nuevos.

### Suggestions

- En `fmtVal`, `json.Marshal` ignora el error silenciosamente (`b, _ := json.Marshal(val)`). Aunque es solo para display, un fallback explícito (e.g., `fmt.Sprintf("%v", val)`) en caso de error sería más robusto.
- La función `formatConfigValue` (línea 1212) ya no trunca valores pero tampoco se usa en `buildConfigLines` (que tiene su propio `fmtVal`). Considerar eliminarla si ya no tiene callers, o unificar ambas funciones para evitar duplicación.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: previous HIGH findings about the falsy-value filter (lines 1155/1175) have been addressed — the filter is now correctly limited to empty strings only. Remaining concerns are minor and mostly carried over from the prior review.

### Issues

⚠️ **cli/internal/tui/dashboard.go:1215** — formatConfigValue no longer truncates slices/maps. If any other caller in the file still relies on the prior 60-rune cap, long values may now wrap/overflow the UI. Confirm all remaining callers of formatConfigValue tolerate untruncated output, or keep truncation there and let buildConfigLines use its own fmtVal.
🟡 **cli/internal/tui/dashboard.go:751** — The `len(lines) == 0` branch returning 'No configuration loaded.' is effectively unreachable because buildConfigLines always appends Server section headers when d.config != nil. Either drop the branch or return a distinct message ('Empty configuration.').
🟡 **cli/internal/tui/dashboard.go:935** — strSlice only matches []any; a []string placed directly into d.config (e.g. future in-process path bypassing JSON round-trip) will silently render empty. Consider handling []string too.
🟡 **cli/internal/tui/dashboard.go:1083** — Inconsistent unit presentation: 'Log retention' appends ' days' while 'Max days' (retention_days) renders the raw number without a unit. Align formatting between the two closely-related fields.
🟡 **cli/internal/tui/dashboard.go:913** — keyStyle and boldStyle are re-instantiated on every call to buildConfigLines. Cheap, but they belong at package scope alongside the other style vars for consistency.

### Suggestions

- Grep callers of formatConfigValue and either restore per-caller truncation or drop the function if buildConfigLines is the only consumer.
- Normalize retention fields: either always append ' days' or never, for consistent presentation.
- Consider handling []string (and possibly []int) in strSlice/fmtVal to future-proof against non-JSON config sources.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: the HIGH falsy-value filter findings have been addressed — only empty strings are now skipped. Remaining concerns are minor: a possibly-unreachable empty branch, and the previously-flagged `formatConfigValue` truncation change whose other callers should be confirmed. Overall this iteration is a solid improvement.

### Issues

⚠️ **cli/internal/tui/dashboard.go:1215** — formatConfigValue no longer truncates slices/maps to 60 runes. Confirm every remaining caller of formatConfigValue (outside buildConfigLines, which uses its own fmtVal) tolerates untruncated, potentially multi-line JSON — otherwise other UI surfaces may wrap or overflow. Safer to keep truncation in formatConfigValue and rely on fmtVal for the Config tab.
🟡 **cli/internal/tui/dashboard.go:751** — The `if len(lines) == 0` branch returning 'No configuration loaded.' is effectively unreachable: when d.config != nil, buildConfigLines always appends at least the Server section. Drop the branch or use a distinct message like 'Empty configuration.' so it's not misleading.
🟡 **cli/internal/tui/dashboard.go:935** — strSlice only matches `[]any`; a future in-process path that bypasses JSON round-trip and places `[]string` directly in d.config would silently render empty. Consider handling `[]string` too or documenting the JSON-decoded invariant.
🟡 **cli/internal/tui/dashboard.go:913** — keyStyle and boldStyle are re-instantiated on every call to buildConfigLines. Move them to package scope alongside the other style vars for consistency and to avoid per-render allocation.
🟡 **cli/internal/tui/dashboard.go:1083** — Inconsistent unit rendering: 'Log retention' appends ' days' only when non-empty, but 'Max days' (retention_days) is rendered raw without a unit. Unify the presentation (either both show ' days' or neither).
🟡 **cli/internal/tui/dashboard.go:966** — section() computes padding from the raw label then wraps with headerStyle.Render. If headerStyle ever gains a background color, the fill dashes and label would render identically but trailing styling edges may differ. Minor cosmetic risk — consider padding then rendering the full line once.

### Suggestions

- Verify all remaining callers of formatConfigValue visually (e.g. any log/status/tooltip context) since the 60-rune truncation was removed.
- Consider rendering 0/false values explicitly (as the author now does) with a visual hint for defaults vs. explicit settings if this becomes confusing later.

---

sergiotejon:

El issue origen #331 queda abierto para que Heimdallm regenere la implementación contra el estado actual de `main`.

Issue #310, opened 2026-04-23
feat(nats): task 11: rate limiter integration with NATS


Muriano:
Integrate rate limiter with NATS consumer acknowledge rate. Either replace token-bucket with NATS consumer rate limiting, or keep as middleware that workers call before GitHub API calls.

Blocked by: task 5
Part of #298

PR #338, opened 2026-04-24
feat(nats): rate limiter integration with NATS workers (#310)

Muriano:
## Summary

- Add `Pipeline.Limiter()` accessor to expose the shared rate limiter
- Add `Acquire()` calls in NATS worker handlers before GitHub API calls:
  - **reviewHandler**: `TierRepo` before `GetPR()`
  - **publishHandler**: `TierRepo` before `SubmitReview()`
  - **stateHandler**: `TierWatch` before `CheckItem()`
- Limiter read under `cfgMu` to handle config reloads safely
- NATS `MaxAckPending` controls concurrency, rate limiter controls API call rate — they complement each other

**Closes:** #310

## Test plan

- [ ] Smoke test: workers respect rate limit under load

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -525,6 +525,14 @@ func main() {
 	// existing review pipeline. This replaces the goroutine-per-PR
 	// pattern that Tier 2 used to use.
 	reviewHandler := func(ctx context.Context, msg bus.PRReviewMsg) {
+		cfgMu.Lock()
+		limiter := pipe.Limiter()
+		cfgMu.Unlock()
+		if err := limiter.Acquire(ctx, scheduler.TierRepo); err != nil {
+			slog.Warn("review-worker: rate limit cancelled", "err", err)
+			return
+		}
+
 		pr, err := ghClient.GetPR(msg.Repo, msg.Number)
 		if err != nil {
 			slog.Error("review-worker: fetch PR from GitHub",
@@ -616,6 +624,13 @@ func main() {
 			Summary:  rev.Summary,
 			Issues:   issues,
 			Severity: rev.Severity,
+		}
+
+		cfgMu.Lock()
+		limiter := pipe.Limiter()
+		cfgMu.Unlock()
+		if err := limiter.Acquire(ctx, scheduler.TierRepo); err != nil {
+			return fmt.Errorf("rate limit cancelled: %w", err)
 		}
 
 		ghID, ghState, err := ghClient.SubmitReview(
@@ -868,6 +883,13 @@ func main() {
 		} else {
 			slog.Warn("state-handler: KV get failed, using zero LastSeen",
 				"key", key, "err", err)
+		}
+
+		cfgMu.Lock()
+		limiter := pipe.Limiter()
+		cfgMu.Unlock()
+		if err := limiter.Acquire(ctx, scheduler.TierWatch); err != nil {
+			return false, fmt.Errorf("rate limit cancelled: %w", err)
 		}
 
 		changed, snap, err := adapter.CheckItem(ctx, item)
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -528,8 +528,10 @@ func main() {
 		cfgMu.Lock()
 		limiter := pipe.Limiter()
 		cfgMu.Unlock()
+		// Acquire returns only ctx.Err() (shutdown). On cancellation the
+		// message is acked without processing — acceptable because the
+		// daemon is shutting down and the PR will be re-detected next startup.
 		if err := limiter.Acquire(ctx, scheduler.TierRepo); err != nil {
-			slog.Warn("review-worker: rate limit cancelled", "err", err)
 			return
 		}
 
@@ -867,6 +869,15 @@ func main() {
 	// Consumes state check requests, calls GitHub API, updates KV backoff.
 	// Reuses the existing CheckItem/HandleChange logic from tier2Adapter.
 	stateHandler := func(ctx context.Context, msg bus.StateCheckMsg) (bool, error) {
+		// Rate limit before any GitHub API call. TierWatch (50ms) matches
+		// the old Tier 3 priority — state checks are lightweight and high-priority.
+		cfgMu.Lock()
+		limiter := pipe.Limiter()
+		cfgMu.Unlock()
+		if err := limiter.Acquire(ctx, scheduler.TierWatch); err != nil {
+			return false, fmt.Errorf("rate limit cancelled: %w", err)
+		}
+
 		item := &scheduler.WatchItem{
 			Type:     msg.Type,
 			Repo:     msg.Repo,
@@ -883,13 +894,6 @@ func main() {
 		} else {
 			slog.Warn("state-handler: KV get failed, using zero LastSeen",
 				"key", key, "err", err)
-		}
-
-		cfgMu.Lock()
-		limiter := pipe.Limiter()
-		cfgMu.Unlock()
-		if err := limiter.Acquire(ctx, scheduler.TierWatch); err != nil {
-			return false, fmt.Errorf("rate limit cancelled: %w", err)
 		}
 
 		changed, snap, err := adapter.CheckItem(ctx, item)
--- a/daemon/internal/scheduler/pipeline.go
+++ b/daemon/internal/scheduler/pipeline.go
@@ -248,3 +248,8 @@ func (p *Pipeline) Stop() {
 func (p *Pipeline) Queue() *WatchQueue {
 	return p.queue
 }
+
+// Limiter returns the shared rate limiter for external callers (e.g. NATS workers).
+func (p *Pipeline) Limiter() *RateLimiter {
+	return p.limiter
+}
--- a/docs/superpowers/specs/2026-04-24-rate-limiter-nats-design.md
+++ b/docs/superpowers/specs/2026-04-24-rate-limiter-nats-design.md
@@ -12,30 +12,29 @@ NATS `MaxAckPending` controls concurrency (how many workers run simultaneously).
 
 ## Changes
 
-### 1. Expose rate limiter to workers
+### 1. Expose rate limiter via Pipeline.Limiter()
 
-The `RateLimiter` is currently created inside `Pipeline.NewPipeline()`. For workers to use it, create a shared limiter in main.go and pass it to both the pipeline and the workers.
+Add a `Limiter() *RateLimiter` accessor to `Pipeline` (same pattern as `Queue()`). Workers access the shared limiter via `pipe.Limiter()` under `cfgMu` to handle config reloads.
 
 ### 2. Add Acquire to worker handlers
 
-In each worker handler closure in main.go, call `limiter.Acquire(ctx, scheduler.TierRepo)` before the GitHub API call:
+In each worker handler closure in main.go, call `limiter.Acquire()` at the top before any GitHub API call:
 
-- **reviewHandler**: before `ghClient.GetPR()`
-- **publishHandler**: before `ghClient.SubmitReview()`
-- **stateHandler**: before `adapter.CheckItem()`
+- **reviewHandler**: `TierRepo` (200ms) before `ghClient.GetPR()`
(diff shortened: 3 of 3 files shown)


sergiotejon:
## 🟡 LOW Issue

The cfgMu.Lock/Unlock pattern to read pipe.Limiter() is repeated three times. If pipe itself can be swapped on config reload, this guards the pointer read, but Pipeline.Limiter() just returns a field — consider caching the limiter reference once at worker registration time, or documenting why the lock is needed on each invocation. If pipe is never swapped, the lock is unnecessary overhead on every message.

**Location:** `daemon/cmd/heimdallm/main.go` line 528

---

sergiotejon:
## 🟡 LOW Issue

In the state handler, the Acquire call is placed AFTER an earlier block that already logged an error with 'key'/'err' (line ~883). The diff shows the Acquire inserted between that error log and the adapter.CheckItem call. Verify the preceding block's error path doesn't already return — if it falls through on error, rate limiting still happens, but the flow is now harder to follow. Consider moving Acquire to the top of the handler like reviewHandler does for consistency.

**Location:** `daemon/cmd/heimdallm/main.go` line 885

---

sergiotejon:
## ⚠️ MEDIUM Issue

reviewHandler returns silently (no error) on limiter.Acquire failure, while publishHandler and stateHandler return the error wrapped. Inconsistent behavior: if reviewHandler is a NATS message handler, returning without error will ACK the message and drop the PR review silently. Context cancellation during shutdown is fine, but any other Acquire failure would cause silent message loss. Should Nak/return error to trigger redelivery.

**Location:** `daemon/cmd/heimdallm/main.go` line 530

---

sergiotejon:
## 🟡 LOW Issue

Design doc says 'Create the limiter in main.go before buildPipeline, pass it to the pipeline config. Workers use the same instance.' but the implementation instead exposes Limiter() via a getter on Pipeline. The doc and code diverge — either update the doc to reflect the getter approach or refactor to match the doc's shared-instance design.

**Location:** `docs/superpowers/specs/2026-04-24-rate-limiter-nats-design.md` line 23

---

sergiotejon (approved):

New changes add rate limiter Acquire calls to three NATS worker handlers (review, publish, state). The pattern is consistent but has a minor concurrency concern with cfgMu usage and a logic issue in the state handler where Acquire happens after a pre-check that may have already logged an error.

### Suggestions

- Unify error handling across the three worker handlers — either all return errors (for NATS redelivery) or all log-and-drop, with a clear rationale documented.
- Cache the limiter reference at handler construction time instead of re-reading pipe.Limiter() under cfgMu on every message, unless pipe is actually swappable at runtime.
- Update the design doc to match the actual implementation (getter on Pipeline vs. shared instance passed to config), or refactor the code to match the doc.
- Add a unit/integration test that verifies Acquire is called before each GitHub API call in the three worker paths.

---

sergiotejon:
## 🟡 LOW Issue

Limiter() getter has no nil-guard. If a Pipeline is constructed without a limiter (e.g. in tests or a future code path), callers will nil-panic on Acquire. Consider documenting that limiter is always non-nil post-NewPipeline, or guard the getter.

**Location:** `daemon/internal/scheduler/pipeline.go` line 253

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

The PR correctly wires the existing token-bucket rate limiter into the three NATS worker handlers (review, publish, state) that make GitHub API calls, using the existing TierRepo/TierWatch priorities. The integration is small and pragmatic, but there are real inconsistencies in error handling across handlers and a design-vs-implementation divergence worth addressing.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:530** — reviewHandler returns silently (nil) on limiter.Acquire failure while publishHandler and stateHandler return a wrapped error. If this is a JetStream consumer, a nil return typically ACKs the message, so a non-shutdown Acquire error would silently drop the PR review. Return an error (or explicitly Nak) so redelivery occurs; only swallow ctx.Err() on shutdown if that is the intended behavior, and make it consistent across all three handlers.
🟡 **daemon/cmd/heimdallm/main.go:528** — The cfgMu.Lock/Unlock around pipe.Limiter() is repeated in all three handlers. If pipe can be swapped on config reload, this guards the pointer read — but the pattern is undocumented and easy to get wrong. If pipe is never swapped, the locking is pure overhead per message. Either cache the *RateLimiter at handler registration (limiters generally shouldn't be swapped mid-flight) or add a comment explaining the invariant.
🟡 **daemon/cmd/heimdallm/main.go:885** — In stateHandler, Acquire is inserted after an earlier error-log block (logs key/err around line ~883) rather than at the top of the handler like reviewHandler. Verify the preceding block does not fall through on error in a way that makes rate limiting run after an already-failed state. Move Acquire to the top for consistency and easier reasoning.
🟡 **docs/superpowers/specs/2026-04-24-rate-limiter-nats-design.md:23** — Design doc says to construct the limiter in main.go before buildPipeline and pass it into the pipeline config, but the implementation instead adds a Limiter() getter on Pipeline. Update the doc to match the getter approach or refactor the code to the shared-instance design the doc describes.
🟡 **daemon/internal/scheduler/pipeline.go:253** — Pipeline.Limiter() has no nil guard. If a Pipeline is ever constructed without a limiter (tests, future code path), callers will nil-panic on Acquire. Either document the post-NewPipeline invariant on the getter or guard against nil.
⚠️ **daemon/cmd/heimdallm/main.go:891** — stateHandler uses scheduler.TierWatch while the design doc explicitly states 'Use TierRepo priority for all workers — they're equivalent to Tier 2 processing.' Either the doc or the code is wrong; reconcile the intended tier for the state/watch worker.

### Suggestions

- Unify the Acquire failure behavior across the three handlers: distinguish ctx.Err() (shutdown, swallow or return cleanly) from other errors (return/Nak to trigger redelivery), and apply the same pattern everywhere.
- Cache the *RateLimiter at worker registration time (or pass it in explicitly) instead of re-locking cfgMu on every message; rate limiters rarely need to be hot-swapped.
- Reconcile TierRepo vs TierWatch for the state worker between the design doc and the implementation.
- Move the Acquire call to the top of stateHandler for symmetry with reviewHandler, and add a short comment on Pipeline.Limiter() documenting that it is always non-nil after NewPipeline.
- Consider adding at least one test that exercises a handler under a saturated limiter to lock in the intended ACK/Nak semantics.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

The PR correctly wires the existing token-bucket rate limiter into three NATS worker handlers, sharing the pipeline's limiter via a new getter. The approach is sound, but there are error-handling inconsistencies and a spec/implementation mismatch on tier priority that should be addressed before merge.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:530** — reviewHandler silently drops the message on rate-limit cancellation (logs + bare return), while publishHandler and stateHandler propagate errors. If the NATS framework auto-acks on a nil-error return, a cancelled review is permanently lost rather than redelivered. At minimum this should match the other handlers' error-propagation pattern so the bus layer can NAK/redeliver.
⚠️ **daemon/cmd/heimdallm/main.go:885** — stateHandler uses scheduler.TierWatch (50 ms wait, highest priority) but the design doc specifies TierRepo for all workers ('they're equivalent to Tier 2 processing'). TierWatch is 4x more aggressive at acquiring tokens, which could starve the pipeline's own Tier 2 polling under pressure. If this is intentional (state checks should mirror Tier 3), update the spec; otherwise use TierRepo.
🟡 **daemon/cmd/heimdallm/main.go:528** — During a config-reload race, a handler can capture the old pipeline's limiter (via pipe.Limiter() under cfgMu) just before the pipeline is stopped. The old limiter's hourly Refill goroutine is now dead, so if all tokens are consumed Acquire blocks until context cancellation — potentially stalling the NATS handler for the full ack-wait timeout. The window is narrow but worth documenting or mitigating (e.g. a channel-close sentinel in the limiter on Stop).
🟡 **daemon/cmd/heimdallm/main.go:628** — The Acquire call in the publish path sits between result assembly and SubmitReview. If the limiter blocks for a significant time under load, the review result (computed earlier in the handler) grows increasingly stale. Consider whether the existing NATS ack deadline is compatible with potentially long Acquire waits here.

### Suggestions

- Extract the repeated cfgMu.Lock(); limiter := pipe.Limiter(); cfgMu.Unlock() pattern into a small closure (e.g. `acquireLimiter := func() *scheduler.RateLimiter { cfgMu.Lock(); defer cfgMu.Unlock(); return pipe.Limiter() }`) to reduce duplication across the three handler sites.
- Add a brief comment in main.go near the handler registrations explaining that NATS MaxAckPending controls concurrency while the limiter controls API call rate — the design doc captures this but the code doesn't, and the doc may drift.
- Consider whether stateHandler's adapter.CheckItem always makes a GitHub API call. If it sometimes short-circuits (e.g. local cache hit), acquiring a rate-limit token unconditionally wastes budget — moving Acquire closer to the actual API call or inside the adapter would be more precise.

---

Muriano:
## Respuesta a los comentarios de review

Fixes en `cb98654`.

### ✅ Arreglados

**reviewHandler silently drops on Acquire failure** (@sergiotejon 🔴, @vbuenog ⚠️)
→ Documentado: `Acquire` solo retorna `ctx.Err()` (shutdown). En shutdown, el ack sin procesar es correcto — el PR se re-detecta en el siguiente arranque. Añadido comentario explicativo.

**Acquire position in stateHandler** (@sergiotejon 🟡, @vbuenog 🟡)
→ Movido al top del handler, antes del KV Get, para consistencia con reviewHandler.

**TierWatch vs TierRepo discrepancia** (@sergiotejon ⚠️, @vbuenog 🟡)
→ TierWatch es **intencional** — state checks son lightweight (1 API call) y high-priority, igual que el Tier 3 original que usaba TierWatch. Spec actualizado para reflejar esto.

**Doc diverge del code** (@sergiotejon 🟡, @vbuenog 🟡)
→ Spec reescrito para reflejar la implementación real (Limiter() getter, TierWatch para state).

### 📝 No blockers

**cfgMu repeated pattern** — 3 sitios, extraer helper añade indirección para 3 líneas. No vale la pena.

**Limiter() nil guard** — `NewPipeline` siempre crea el limiter. No hay path donde sea nil post-construction.

**Stale limiter after reload** — Ventana mínima, y `Acquire` con ctx cancelled retorna inmediatamente. El refill goroutine del pipeline viejo se detiene con `pipe.Stop()`, pero los tokens existentes en el channel siguen disponibles.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Author addressed the previous medium finding by documenting that Acquire only returns ctx.Err() on shutdown, making the silent ack in reviewHandler correct behavior. The stateHandler Acquire was moved to the top of the handler, and the spec doc was rewritten to match the implementation. Remaining low-severity items (cfgMu repetition, nil-guard on Limiter()) were acknowledged with reasonable justifications. No new issues in the updated diff.

### Suggestions

- Consider adding a brief doc comment on Pipeline.Limiter() stating that the returned limiter is always non-nil after NewPipeline, to preempt future misuse.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Author addressed all previous findings — reviewHandler's silent-ack path is now documented as shutdown-only (Acquire returns only ctx.Err()), stateHandler's Acquire is at the top of the handler, TierWatch rationale is documented in both code and spec, and the design doc now matches the Limiter() getter implementation. Remaining 'no blockers' (cfgMu repetition, nil guard) have reasonable justifications. No new issues introduced.

### Suggestions

- Consider a brief comment on Pipeline.Limiter() stating the post-NewPipeline non-nil invariant, so future callers don't need to re-derive it.
- If config reload can swap the pipeline, document once (e.g. on Pipeline.Limiter or near the handler registrations) that callers must read it under cfgMu per-message — this makes the repeated pattern intentional rather than cargo-culted.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previous findings have been addressed. The explanatory comment on reviewHandler's silent return clarifies that Acquire only returns ctx.Err() (shutdown), making the ack-without-processing acceptable. TierWatch for stateHandler is confirmed intentional and the spec is updated. The stateHandler Acquire was moved to the top for consistency. No new issues found in the current diff.

---

PR #339, opened 2026-04-24
fix: issue re-triage loop + locked PR publish retry (#322)

Muriano:
## Summary

Fixes two infinite-loop bugs observed in production logs:

**Bug 1: Issue #144 re-triaged every poll cycle**
- `FetchComments` called the pulls review-comments API (`/pulls/N/comments`) which returns 404 for issues (not PRs)
- The 404 killed the marker scanner, falling through to dedup checks that couldn't catch the re-triage
- **Fix:** Treat 404 from review-comments as empty (no review comments on issues is expected)

**Bug 2: Review #681 PublishPending retried 422 indefinitely**
- `PublishPending()` treated all GitHub errors as transient, retrying every poll cycle
- 422 "lock prevents review" is permanent — the PR's conversation is locked
- **Fix:** Treat 4xx (except 429 rate limit) as permanent failures, mark review as orphaned

**Closes:** #322

## Test plan

- [ ] `go test ./internal/github/ ./internal/pipeline/ -count=1` — tests pass
- [ ] Smoke test: issue #144 no longer re-triages every cycle, review #681 stops retrying

### DIFF
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -513,14 +513,20 @@ func (c *Client) FetchComments(repo string, number int) ([]Comment, error) {
 
 	r1 := <-reviewCh
 	r2 := <-issueCh
-	if r1.err != nil {
+	// Review comments (pulls API) return 404 for issues — this is expected.
+	// Treat as empty rather than fatal so issue marker scanning works.
+	if r1.err != nil && !strings.Contains(r1.err.Error(), "status 404") {
 		return nil, r1.err
 	}
 	if r2.err != nil {
 		return nil, r2.err
 	}
 
-	all := append(r1.comments, r2.comments...)
+	var all []Comment
+	if r1.err == nil {
+		all = append(all, r1.comments...)
+	}
+	all = append(all, r2.comments...)
 	sort.Slice(all, func(i, j int) bool {
 		return all[i].CreatedAt.Before(all[j].CreatedAt)
 	})
--- a/daemon/internal/pipeline/pipeline.go
+++ b/daemon/internal/pipeline/pipeline.go
@@ -475,6 +475,15 @@ func (p *Pipeline) PublishPending() {
 			SeverityToEvent(rev.Severity, len(issues)),
 		)
 		if err != nil {
+			errStr := err.Error()
+			// 422 "lock prevents review" and other 4xx (except 429 rate limit)
+			// are permanent — mark as orphaned to stop infinite retry.
+			if strings.Contains(errStr, "status 4") && !strings.Contains(errStr, "status 429") {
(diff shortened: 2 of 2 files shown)


sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

The PR fixes two real bugs (issue re-triage loop caused by 404 on pulls API for issues, and infinite retry on permanent publish failures) but relies on fragile string-based HTTP status detection in both files. The existing codebase already uses typed errors (PermanentSubmitError + markOrphanIfPermanent) for the publish-retry path, making the pipeline.go change here redundant and inconsistent with the established pattern.

### Issues

🔴 **daemon/internal/pipeline/pipeline.go:480** — String-based 4xx detection (`strings.Contains(errStr, "status 4")`) is fragile and overly broad — it could match any error message containing the substring "status 4" (e.g. "status 4 retries attempted"). The main branch already has a typed `PermanentSubmitError` and a `markOrphanIfPermanent` helper (see commit 9002737 / PR #326) that handles this same case robustly. This string-matching approach conflicts with and is strictly inferior to the existing typed-error pattern. Use `markOrphanIfPermanent` instead.
⚠️ **daemon/internal/pipeline/pipeline.go:484** — Store error silently discarded: `_ = p.store.MarkReviewPublished(...)`. If MarkReviewPublished fails, the review stays unpublished and will be retried on the next tick — the exact infinite-retry problem this PR is trying to fix. Should at minimum log a warning (the existing markOrphanIfPermanent helper already does this).
⚠️ **daemon/internal/github/client.go:517** — String-based 404 detection (`strings.Contains(r1.err.Error(), "status 404")`) is fragile — it depends on the exact error format string produced elsewhere in the codebase. If that format changes, this silently breaks. Consider introducing a typed error (e.g. `NotFoundError` or a status-code accessor) so the check is compile-time safe.
🟡 **daemon/internal/github/client.go:529** — The function docstring (line 529-530) says FetchComments is for PRs only, but the new 404 handling implies it's now intentionally called for issues too. The docstring should be updated to reflect this expanded contract.

### Suggestions

- Rebase onto main and use the existing `markOrphanIfPermanent` / `PermanentSubmitError` typed-error pattern (from PR #326) instead of string-matching HTTP status codes in PublishPending. This eliminates the fragile string matching and keeps both publish paths (Run and PublishPending) in lockstep.
- Introduce a typed error for HTTP status codes in the GitHub client (e.g. `type HTTPError struct { StatusCode int }`) so both the 404 check in FetchComments and any future status-based branching can use `errors.As` instead of string matching.
- If string matching must be kept temporarily, use a regex like `status 4\d\d` instead of `status 4` to avoid matching unrelated substrings, and do not discard the MarkReviewPublished error with `_ =`.

---

PR #340, opened 2026-04-24
fix: issue re-triage loop + locked PR publish retry (#322)

Muriano:
## Summary

Cherry-picks two bugfix PRs from main into main-nats:

**#296: fix(daemon): stop issue-triage cost-runaway**
- FetchIssueCommentsOnly avoids calling the PR review-comments API for issues (404)
- Issue triage in-flight claims + circuit breaker
- Prevents the re-triage-every-cycle loop on issue #144

**#326: fix(daemon): mark review orphan on permanent submit failure**
- Typed PermanentSubmitError with StatusCode
- markOrphanIfPermanent helper for 4xx (except 429)
- Stops infinite retry of locked/deleted PRs (review #681)

Cherry-picked from main with conflict resolution (NATS bus block in main.go, removed TimelineEvent tests that depend on code not yet in main-nats).

**Closes:** #322

## Test plan

- [ ] Smoke test: issue #144 no longer re-triages, review #681 stops retrying

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -131,6 +131,13 @@ func main() {
 		slog.Info("startup: cleared stale inflight rows", "count", n)
 	}
 
+	// Mirror of the PR-side sweep above for issue-triage claims (#292).
+	if n, err := s.ClearStaleIssueTriageInFlight(30 * time.Minute); err != nil {
+		slog.Warn("startup: clear stale issue triage inflight failed", "err", err)
+	} else if n > 0 {
+		slog.Info("startup: cleared stale issue triage inflight rows", "count", n)
+	}
+
 	// ── NATS event bus ──────────────────────────────────────────────────
 	eventBus := bus.New(bus.Config{
 		DataDir:              filepath.Join(dataDir(), "nats"),
@@ -216,11 +223,19 @@ func main() {
 	}
 	p.SetCircuitBreakerLimits(&cbLimits)
 
+	// Issue-side circuit-breaker caps (theburrowhub/heimdallm#292) — mirrors
+	// the PR-side defenses against runaway triage loops.
+	issueCBLimits := store.IssueCircuitBreakerLimits{
+		PerIssue24h: cfg.CircuitBreaker.PerIssue24h,
+		PerRepoHr:   cfg.CircuitBreaker.PerIssueRepoHr,
+	}
+
 	// GitExec drives the auto_implement flow (#27): branch, commit, push, PR.
 	// Wired unconditionally — the pipeline guards against running git ops on
 	// an issue that is classified as review_only, so this dep is harmless
 	// when auto_implement is not in use.
 	issuePipe := issuepipeline.New(s, ghClient, exec, issuepipeline.NewGitExec(), broker, &notifyWithSSE{notifier: notifier})
+	issuePipe.SetCircuitBreakerLimits(&issueCBLimits)
 
 	// Resolve bot login for re-review / re-triage context filtering.
 	if login, err := ghClient.AuthenticatedUser(); err == nil {
--- a/daemon/internal/config/circuit_breaker.go
+++ b/daemon/internal/config/circuit_breaker.go
@@ -3,21 +3,38 @@ package config
 // CircuitBreakerConfig caps the number of reviews per PR and per repo to
 // prevent cost-runaway loops. The defaults are conservative — users with
 // high-volume workflows must explicitly raise them. See
-// theburrowhub/heimdallm#243 for the incident that prompted these caps.
+// theburrowhub/heimdallm#243 for the incident that prompted these caps,
+// and theburrowhub/heimdallm#292 for the issue-side extension.
+//
+// Field semantics: applyDefaults() treats 0 as "unset" and substitutes
+// the documented default. There is currently no way to express
+// "unlimited" through TOML — set the cap high (e.g. 99999) if you need
+// near-unbounded behaviour. The PR-side and issue-side per-repo axes
+// are independent: a busy repo running both PR review and issue triage
+// can consume PerRepoHr PR reviews AND PerIssueRepoHr triages per hour.
 type CircuitBreakerConfig struct {
-	// PerPR24h caps reviews on the same PR over any 24-hour window.
-	// 0 = unlimited. Default 3.
+	// PerPR24h caps PR reviews on the same PR over any 24-hour window.
+	// Default 3; set to 0 to apply the default.
 	PerPR24h int `toml:"per_pr_24h"`
-	// PerRepoHr caps reviews on the same repo over any 1-hour window.
-	// 0 = unlimited. Default 20.
+	// PerRepoHr caps PR reviews on the same repo over any 1-hour window.
+	// Default 20; set to 0 to apply the default.
 	PerRepoHr int `toml:"per_repo_hr"`
+	// PerIssue24h caps issue triages on the same issue over any 24-hour
+	// window. Default 3 (same as the PR cap); set to 0 to apply the default.
+	PerIssue24h int `toml:"per_issue_24h"`
+	// PerIssueRepoHr caps issue triages on the same repo over any 1-hour
+	// window. Default 10 — tighter than the PR cap because each triage is
+	// a full-context Claude run; set to 0 to apply the default.
+	PerIssueRepoHr int `toml:"per_issue_repo_hr"`
 }
 
 // DefaultCircuitBreakerConfig returns the safe defaults applied when the
 // [circuit_breaker] TOML section is missing or zero-valued.
 func DefaultCircuitBreakerConfig() CircuitBreakerConfig {
 	return CircuitBreakerConfig{
-		PerPR24h:  3,
-		PerRepoHr: 20,
+		PerPR24h:       3,
+		PerRepoHr:      20,
+		PerIssue24h:    3,
+		PerIssueRepoHr: 10,
 	}
 }
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -653,6 +653,12 @@ func (c *Config) applyDefaults() {
 	if c.CircuitBreaker.PerRepoHr == 0 {
 		c.CircuitBreaker.PerRepoHr = 20
 	}
+	if c.CircuitBreaker.PerIssue24h == 0 {
+		c.CircuitBreaker.PerIssue24h = 3
+	}
+	if c.CircuitBreaker.PerIssueRepoHr == 0 {
+		c.CircuitBreaker.PerIssueRepoHr = 10
+	}
 }
 
 // applyEnvOverrides applies HEIMDALLM_* environment variable overrides.
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -527,6 +527,16 @@ func (c *Client) FetchComments(repo string, number int) ([]Comment, error) {
 	return all, nil
 }
 
+// FetchIssueCommentsOnly retrieves the issue comments for an issue or PR
+// number WITHOUT also calling /pulls/:n/comments. Callers that operate on
+// issue numbers MUST use this method: /pulls/:n/comments always 404s on
+// issues, and FetchComments treats that as a hard error. See
+// theburrowhub/heimdallm#292 — the 404 cascade broke the marker-scan
+// idempotency check and produced a re-triage loop.
+func (c *Client) FetchIssueCommentsOnly(repo string, number int) ([]Comment, error) {
+	return c.fetchIssueComments(repo, number)
+}
+
 func (c *Client) fetchReviewComments(repo string, number int) ([]Comment, error) {
 	path := fmt.Sprintf("/repos/%s/pulls/%d/comments", repo, number)
 	resp, err := c.do("GET", path, "application/vnd.github+json")
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -192,6 +192,49 @@ func (c *Client) fetchByQualifier(username, qualifier string, repos []string) ([
 	return result.Items, nil
 }
 
+// PermanentSubmitError signals that SubmitReview hit a state from
+// which no retry will recover — e.g. the PR's conversation is locked,
+// the PR has been deleted, or the repo was archived. Callers should
+// mark the local review row as orphaned (via the same sentinel as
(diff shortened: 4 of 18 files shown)


sergiotejon:
## ⚠️ MEDIUM Issue

Error classification relies on substring matching ('status 4', 'status 429'). This is fragile: any future change to error formatting elsewhere will silently break the guard, and '429' could coincidentally appear in other error text. Prefer a typed error (e.g., a github.HTTPError with a StatusCode field) and switch on the code, or at minimum use a more specific prefix like 'status 4xx' boundaries (e.g., regex ^status 4\d{2}$ after extracting).

**Location:** `daemon/internal/pipeline/pipeline.go` line 481

---

sergiotejon:
## 🟡 LOW Issue

Using MarkReviewPublished with publishedID=-1 to represent 'orphaned/permanent failure' overloads the semantics of a success API. Consider a dedicated MarkReviewOrphaned/MarkReviewFailed method so queries and metrics can distinguish successfully-published reviews from permanently-failed ones.

**Location:** `daemon/internal/pipeline/pipeline.go` line 484

---

sergiotejon:
## 🟡 LOW Issue

404 detection uses strings.Contains(err.Error(), 'status 404'). Same fragility as above — a typed error with StatusCode would be more reliable and avoids false positives if 404 appears in an unrelated message body.

**Location:** `daemon/internal/github/client.go` line 516

---

sergiotejon (approved):

Two targeted fixes: tolerate 404 from the pulls review-comments endpoint when fetching comments for issues, and break the infinite retry loop on permanent publish failures (4xx except 429) by marking reviews orphaned. Logic is sound; a couple of robustness concerns around string-based error matching.

### Suggestions

- Introduce a typed HTTP error (StatusCode int) from the github client and switch on codes instead of substring matching in both the 404 tolerance and the 4xx permanent-failure guard.
- Consider treating 403 separately — some 403s (abuse/secondary rate limit) are transient and retryable, unlike other 4xx; grouping them as 'permanent' may drop reviews that would have succeeded later.
- Add a dedicated store method (e.g., MarkReviewOrphaned) rather than reusing MarkReviewPublished with -1 to keep published vs. permanently-failed reviews distinguishable.
- Add a unit test for FetchComments where the review-comments endpoint returns 404 to lock in the new behavior, and one for PublishPending where a 422 causes the review to be marked orphaned (not retried).

---

sergiotejon:
## 🟡 LOW Issue

Minor: the pre-change code used a single append; the new code allocates 'all' as nil and appends twice. Functionally fine, but you could preserve capacity by 'all := make([]Comment, 0, len(r1.comments)+len(r2.comments))' to avoid an extra growth.

**Location:** `daemon/internal/github/client.go` line 525

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Both changes in this PR are already superseded by better implementations on main. The 404 FetchComments fix is solved by FetchIssueCommentsOnly (issue #292, PR #296), and the locked-PR publish retry is solved by typed PermanentSubmitError + markOrphanIfPermanent (issue #325, PR #326). The string-matching approach here is fragile and inferior to what already shipped. This PR should be closed rather than merged.

### Issues

🔴 **daemon/internal/pipeline/pipeline.go:481** — String-based HTTP status classification ('status 4' / 'status 429') is fragile and already superseded on main by a typed PermanentSubmitError with a StatusCode field and a shared markOrphanIfPermanent helper (PR #326). Merging this would regress the existing typed-error approach. The substring 'status 4' can also false-positive on unrelated error text (e.g. 'failed to update status 4 times').
🔴 **daemon/internal/pipeline/pipeline.go:478** — This diff targets PublishPending at a code shape that no longer exists on main — main already has markOrphanIfPermanent(rev.ID, err, 'retry publish') at this location. This will produce merge conflicts and, if force-merged, would duplicate or regress the orphan-marking logic.
⚠️ **daemon/internal/github/client.go:516** — The 404 tolerance for FetchComments is already solved on main via FetchIssueCommentsOnly (line 569), which avoids calling the PR review endpoint entirely for issues. That approach is strictly better: it eliminates the 404 rather than catching it. Merging this string-match workaround would leave dead code or a confusing alternate path.
🟡 **daemon/internal/github/client.go:516** — strings.Contains(err.Error(), 'status 404') is brittle — any change to the error format in the github client would silently break this guard. The main branch avoids this entirely by not calling the PR endpoint for issues.
🟡 **daemon/internal/pipeline/pipeline.go:484** — MarkReviewPublished with publishedID=-1 to represent 'orphaned' overloads a success API. While main uses the same sentinel convention, the pipeline.go changes here bypass the shared markOrphanIfPermanent helper that centralizes this pattern, creating a drift risk between the Run and PublishPending paths.

### Suggestions

- Close this PR — both fixes already shipped through #296 (FetchIssueCommentsOnly) and #326 (typed PermanentSubmitError + markOrphanIfPermanent). The branch is stale relative to main.
- If any edge cases from the original issue (#322) remain unaddressed after #296 and #326, open a new PR based on current main rather than rebasing this one.
- If the 404-tolerance approach in FetchComments is still desired as defense-in-depth alongside FetchIssueCommentsOnly, implement it using a typed error (e.g. check resp.StatusCode directly) rather than string matching on the error message.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

The fixes address real bugs: FetchComments no longer fails for issues (pulls review API returns 404), and PublishPending no longer retries forever on permanent 4xx failures like the 422 'lock prevents review' case. Logic is correct, but error classification via substring matching on err.Error() is fragile and should be replaced with a typed HTTP error carrying a StatusCode. Additionally, reusing MarkReviewPublished(-1, "") to represent an orphaned/failed state overloads a success-path API and will confuse downstream queries/metrics.

### Issues

⚠️ **daemon/internal/pipeline/pipeline.go:481** — Error classification by strings.Contains(err.Error(), "status 4") and "status 429" is fragile. 'status 4' matches 'status 4xx' but also any future error text containing that substring, and '429' can appear in unrelated messages. A 5xx wrapped string that happens to contain 'status 4' elsewhere would be misclassified. Introduce a typed error (e.g. github.HTTPError{StatusCode int}) and switch on the numeric code; treat 429 and 5xx as retryable, 4xx (except 429) as permanent.
🟡 **daemon/internal/pipeline/pipeline.go:484** — Overloading MarkReviewPublished with publishedID=-1 and empty URL to signal a permanent failure conflates success and failure states in the store. Downstream queries and metrics cannot distinguish 'published' from 'orphaned'. Add an explicit MarkReviewOrphaned/MarkReviewFailed(reviewID, reason, at) method so state is first-class and auditable.
🟡 **daemon/internal/github/client.go:516** — 404 detection via strings.Contains(err.Error(), "status 404") has the same fragility as the pipeline change — any error message body echoing '404' (e.g. a URL containing 404 or a nested message) could trigger a false positive. Prefer a typed HTTPError and compare StatusCode == http.StatusNotFound.
🟡 **daemon/internal/github/client.go:525** — Minor: the rewritten append path allocates nil then grows twice. Preallocate with capacity to avoid an extra slice growth when both channels return comments: all := make([]Comment, 0, len(r1.comments)+len(r2.comments)) (guard r1 when its err is non-nil).
🟡 **daemon/internal/pipeline/pipeline.go:478** — No test coverage is added for either new branch (the 404-tolerant FetchComments path or the permanent-failure orphaning path). These are exactly the kinds of error-classification branches that regress silently; add unit tests that inject a fake GitHub client/error to cover 404, 422, 429, and 500 cases.

### Suggestions

- Introduce a typed github.HTTPError{StatusCode int, Body string} at the HTTP boundary and have callers switch on StatusCode; remove all strings.Contains(err.Error(), "status …") call sites in one pass.
- Add a dedicated store method (e.g. MarkReviewOrphaned) and a corresponding status column/enum value so orphaned reviews are queryable and observable, and add a metric/log counter for permanent publish failures.
- Log the HTTP status code and response body (truncated) when orphaning a review to aid post-hoc debugging — right now only the wrapped err is logged.
- Consider a max-retry/age cap on the retryable branch as a defense-in-depth against a 429 or 5xx that never clears, so PublishPending cannot be starved by one stuck review.

---

Muriano:
## Respuesta

La segunda review de sergiotejon menciona que main ya tiene fixes mejores via #296 y #326 (typed errors, FetchIssueCommentsOnly, markOrphanIfPermanent). Esto es correcto — esos PRs se mergearon en main después de que main-nats divergió.

Las implementaciones en main son superiores (typed HTTPError vs string matching). Cuando Task 12 reconcilie main-nats con main, estos fixes se sobreescribirán por las versiones typed.

Esta PR es un fix interim para que main-nats no sufra los loops mientras completamos el epic. Los string matches son feos pero funcionales y temporales.

Sobre los LOW/suggestions (typed errors, MarkReviewOrphaned, preallocate): todos válidos pero ya resueltos en main. No tiene sentido implementarlos aquí para que se sobreescriban en el merge.

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review finds the previous MEDIUM concern (fragile string-based error classification) is resolved via the new typed *PermanentSubmitError and the dedicated FetchIssueCommentsOnly method (which removes the need for 404-string sniffing). New tests cover locked-PR (positive), 5xx (negative), 422-without-lock (negative), and body-not-truncated cases, as well as the issue-side circuit breaker and in-flight claim dedup. Remaining concerns are minor.

### Issues

🟡 **daemon/internal/github/client.go:235** — classifyPermanentSubmit422 still relies on substring matching against GitHub's response body ("lock prevents review", "pull request is locked"). GitHub's error copy is reasonably stable and tests cover both match/no-match cases, but a future copy change would silently disable the orphan path. Consider also matching on the structured errors[].code / resource where GitHub provides them, so the substring is a fallback rather than the sole signal.
🟡 **daemon/cmd/heimdallm/main.go:134** — ClearStaleIssueTriageInFlight runs only at startup (mirroring the PR-side sweep). Combined with the deliberate design choice that the in-flight claim is retained across circuit-breaker trips (documented in TestIntegration_IssueCircuitBreakerTripsAfterCap), a long-running daemon can hold a tripped-breaker claim indefinitely: the issue effectively stays locked out until the next restart even after the 24h window elapses. Consider a periodic sweep or an explicit release on breaker trip scoped to the breaker-reason rows.
🟡 **daemon/internal/issues/integration_test.go:352** — countingNotifier implements Notify(_, _ string), but the pipeline's Notifier interface is declared elsewhere (via notifyWithSSE in main.go) with its own signature. If that interface ever takes a third arg or a context, this test would silently fail to compile only under the integration tag. Low-risk, but worth aligning on the canonical Notifier interface rather than a test-local shape.

### Suggestions

- Consider a runtime (not just startup) sweep of stale issue-triage in-flight rows so held-claim-across-breaker-trip does not require a daemon restart to recover after the 24h window lapses.
- When/if the main-nats branch reconciles with main (#296/#326), double-check that PermanentSubmitError's classification rules and the FetchIssueCommentsOnly contract port cleanly, since the typed-error versions on main may have different reason taxonomies.
- If GitHub exposes a stable errors[].code for the locked-PR case, prefer that over the message substring (keep the substring as a fallback).

---

sergiotejon:
## 🟡 LOW Issue

PermanentSubmitError.Body carries the FULL (untruncated) response body intentionally, but the code path uses io.LimitReader(resp.Body, maxBodyBytes) upstream — so 'full' is bounded by maxBodyBytes, not unbounded. The comment ('FULL response (not safe-truncated)') is slightly misleading; clarify that it is the full LimitReader-bounded body (vs. the 200-byte maxErrBodyLen cutoff) to avoid future confusion.

**Location:** `daemon/internal/github/client.go` line 287

---

sergiotejon:
## ⚠️ MEDIUM Issue

Field-semantics doc states '0 = apply the default' — this is a breaking behavioural change from the previous 'PerPR24h: 0 = unlimited' contract documented on the same struct before this PR. Existing operators who set per_pr_24h = 0 in TOML to disable the cap will silently get the default of 3 after upgrade. Consider a release note / CHANGELOG entry, or preserve the prior 'unlimited' semantics and use -1 (or a separate bool) for default-substitution.

**Location:** `daemon/internal/config/circuit_breaker.go` line 20

---

sergiotejon:
## 🟡 LOW Issue

countingNotifier.Notify signature (two string args) must exactly match the Notifier interface; if the interface evolves this test will fail in a confusing way. Minor — consider embedding a no-op base or using a mock generator for resilience.

**Location:** `daemon/internal/issues/integration_test.go` line 244

---

sergiotejon (approved):

This PR backports the typed PermanentSubmitError and FetchIssueCommentsOnly approach from main into main-nats, which directly addresses the previous medium and low findings about fragile substring matching. The new code uses typed errors (*PermanentSubmitError with StatusCode/Reason fields) and a dedicated endpoint method instead of string matching. Circuit breaker extension to issue-side is well-structured with reasonable defaults and proper test coverage (concurrency, cap enforcement, notification suppression). Tests are thorough and lock in the specific regressions cited. A couple of minor observations on the new code only.

### Suggestions

- Add a CHANGELOG/release note about the circuit-breaker zero-value semantics change (0 now means 'apply default' rather than 'unlimited'), since this is a user-visible config behaviour change.
- Consider extracting the lock-detection substring list ('lock prevents review', 'pull request is locked') into named constants so it's obvious where to add new permanent-failure signatures when GitHub changes wording.
- The classifyPermanentSubmit422 helper is a good seam for future expansion (archived repo, deleted PR); add a brief TODO/roadmap note referencing likely future cases so the intent is preserved.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Thoroughly reworked PR that addresses all 5 previous findings. String-based error matching replaced with typed PermanentSubmitError, FetchComments patching replaced with dedicated FetchIssueCommentsOnly, and circuit breaker infrastructure properly wired. Test coverage is comprehensive. No significant new issues found.

### Issues

🟡 **daemon/internal/github/client.go:280** — Comment says Body carries 'the FULL response (not safe-truncated)' but respBody was already read through io.LimitReader(resp.Body, maxBodyBytes), so it IS bounded — just not by the tighter maxErrBodyLen used in the generic error path. Minor documentation inaccuracy; no behavioral impact.

### Suggestions

- Consider adding a test case for classifyPermanentSubmit422 with an empty body to document behavior (currently returns false, which is correct, but worth locking in).
- The circuit breaker config doc change from '0 = unlimited' to 'set to 0 to apply the default' is a documentation correction (the code always defaulted), but it may surprise users who relied on the old (incorrect) docs. A CHANGELOG entry would help.

---

Issue #311, opened 2026-04-23
feat(nats): task 12: remove old scheduler and SSE broker


Muriano:
Remove scheduler/tier1.go, tier2.go, tier3.go, queue.go, pipeline.go. Remove sse/broker.go. Update main.go — replace tier adapter with NATS publishers + consumers. Keep ratelimit.go if used as middleware.

Blocked by: tasks 6, 8, 9, 10, 11
Part of #298

PR #342, opened 2026-04-24
feat(nats): remove old scheduler — standalone pollers replace Pipeline (#311)

Muriano:
## Summary

Remove the scheduler Pipeline orchestrator and replace with standalone polling goroutines in main.go.

**Deleted 10 files** (-1,443 lines):
- `scheduler/pipeline.go` — Pipeline orchestrator
- `scheduler/tier1.go`, `tier2.go`, `tier3.go` — Tier runners
- `scheduler/queue.go` — WatchQueue heap (replaced by NATS KV)
- All 5 corresponding test files

**Created 1 file** (+134 lines):
- `scheduler/types.go` — Exported types still referenced by main.go and bus package

**Modified main.go** (net +353 lines):
- `startPollers()` replaces `buildPipeline` + `pipe.Start()`
- Standalone goroutines: discovery poller, NATS bridge, PR/issue poller
- Config reload: cancel old poller contexts → start new ones
- Workers reference shared `limiter` directly

**Kept**: `ratelimit.go`, `scheduler.go`, SSE broker (too many dependents)

**Net deletion**: 956 lines removed

**Closes:** #311

## Test plan

- [ ] `go test ./... -count=1` — full suite passes (17 packages)
- [ ] `go test ./... -race -count=1` — no races
- [ ] Smoke test: pollers start, discovery → NATS → bridge → PR poll → workers
- [ ] Config reload: pollers stop and restart cleanly


sergiotejon:
## ⚠️ MEDIUM Issue

runTier2 spawns an inner goroutine to read reposChan but only exits via ctx.Done(). On reload, startPollers creates a new reposChan in the outer closure... actually reposChan is captured from the outer main scope (single shared channel across reloads). This means old runTier2 goroutines that have exited via ctx.Done() are fine, but the shared reposChan across reload cycles can cause the bridge from a new cycle to race with a lingering reader from the previous cycle until it observes ctx.Done(). Consider creating reposChan inside startPollers so each cycle has its own channel.

**Location:** `daemon/cmd/heimdallm/main.go` line 1765

---

sergiotejon:
## 🟡 LOW Issue

The inner goroutine in runTier2 that reads from reposChan is never tracked by the parent WaitGroup. On shutdown, startPollers' wg.Wait() returns when runTier2 returns, but the inner reader goroutine may still be running briefly. Minor leak window, but it can also drop a repo update that was published just before cancellation (written to channel but never assigned to repos).

**Location:** `daemon/cmd/heimdallm/main.go` line 1765

---

sergiotejon:
## ⚠️ MEDIUM Issue

Tier 3 (watch queue / per-item polling) has been removed entirely along with queue.go, tier3.go, and related tests. The PR description says 'standalone pollers replace Pipeline' but doesn't explicitly address dropping the watch tier. If Tier 3 was responsible for fast per-item status checks, removing it may regress behavior. The StateCheckMsg handler still exists and uses TierWatch rate limit bucket, so state checks are presumably now driven by NATS messages rather than the WatchQueue — verify this is intentional and documented.

**Location:** `daemon/cmd/heimdallm/main.go`

---

sergiotejon:
## 🟡 LOW Issue

sendDiscoveryRepos is called once before the ticker loop begins, but this initial call happens unconditionally on every startPollers invocation — including reloads where coldStart=false. The original Pipeline's Tier1 may have respected coldStart for discovery too; verify this doesn't cause an unwanted immediate discovery API call on every /reload.

**Location:** `daemon/cmd/heimdallm/main.go` line 1670

---

sergiotejon:
## 🟡 LOW Issue

In the discovery ticker loop, limiter.Acquire is called before sendDiscoveryRepos, but sendDiscoveryRepos itself does not call limiter.Acquire for the disc.Discovered() call. The initial pre-ticker sendDiscoveryRepos call also skips Acquire. Inconsistent rate-limit accounting compared to the ticker path.

**Location:** `daemon/cmd/heimdallm/main.go` line 1690

---

sergiotejon:
## 🟡 LOW Issue

runTier2 uses a hardcoded 2-second sleep waiting for Tier 1 to send the first batch. This is fragile — if discovery takes longer (network latency, large orgs), the first tick runs with empty repos. Consider waiting for the first reposChan value with a timeout instead of a blind sleep.

**Location:** `daemon/cmd/heimdallm/main.go` line 1765

---

sergiotejon:
## 🟡 LOW Issue

In runTier2's processTick, adapter.FetchPRsToReview() does not take a context. If the daemon is shutting down mid-fetch, the call can't be cancelled. Pre-existing issue carried over from the old Tier 2, but worth flagging.

**Location:** `daemon/cmd/heimdallm/main.go` line 1800

---

sergiotejon (approved):

Refactor replaces the 3-tier Pipeline orchestrator with standalone poller goroutines inlined in main.go. The change simplifies the architecture by removing the WatchQueue/Tier 3 watch layer and consolidating rate limiter ownership in main. The reload and shutdown paths are correctly updated to swap cancel/wg pairs atomically. However, the inlined runTier2 has a goroutine leak on reload, Tier 3 (watch) functionality appears to be dropped entirely without clear justification in the diff, and there are minor concurrency concerns around the repos slice.

**8 issue(s) found** — see individual comments above for details.

### Suggestions

- Move reposChan declaration inside startPollers so each reload cycle gets a fresh channel and there's no cross-cycle coupling.
- Track runTier2's inner reposChan-reader goroutine with a sync.WaitGroup so shutdown is fully deterministic.
- Document in the PR description (and a code comment) the intentional removal of Tier 3 / WatchQueue and how state checks are now handled end-to-end via NATS StateCheckMsg.
- Replace the 2s sleep in runTier2 with a bounded wait for the first reposChan value, falling back to empty repos after a timeout.
- Consider extracting runTier2, sendDiscoveryRepos, bridgeDiscovery, and runBridgeConsumer out of main.go into a new package (e.g. internal/pollers) — main.go is now ~1800 lines and mixes wiring with business logic.
- Add a test covering the reload path: start pollers, trigger reload, verify old goroutines exited and new ones are running against the new config.

---

sergiotejon:
## 🟡 LOW Issue

limiter is declared at function scope and shared across reloads. This is correct (rate budget persists across config reloads), but the comment 'Shared rate limiter (was Pipeline.limiter)' could be clearer that it intentionally outlives reload cycles.

**Location:** `daemon/cmd/heimdallm/main.go` line 513

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

PR removes the Pipeline orchestrator and Tier 3 watch queue, inlining standalone polling goroutines into main.go. The refactor is coherent and preserves rate-limiter semantics across reloads, but introduces real concurrency concerns: a shared reposChan is captured from outer scope across reload cycles, an inner goroutine in runTier2 is not tracked by the WaitGroup, and Tier 3 functionality is dropped without an explicit migration note. Several smaller rate-limit accounting and cold-start inconsistencies also slipped in.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:489** — reposChan is created once in main() outer scope (line ~489 area) and captured by every startPollers invocation. On /reload, the new runTier2 and new bridge share the same channel with any lingering goroutines from the previous cycle that haven't yet observed ctx.Done(). This can cause a new cycle's bridge publish to be consumed by an old reader (lost update) or race with the new reader. Move `reposChan := make(chan []string, 1)` inside startPollers so each reload cycle gets its own channel.
🟡 **daemon/cmd/heimdallm/main.go:1765** — The inner goroutine inside runTier2 that reads from reposChan is not registered with the parent WaitGroup. startPollers' wg.Wait() returns when runTier2 returns, but this reader may still be running briefly and can consume a repo update that was just published — the update is written to the channel but never assigned to `repos` because runTier2 has already exited. Track it via wg.Add/wg.Done, or have runTier2 itself own the read loop inside its main select.
⚠️ **daemon/cmd/heimdallm/main.go:0** — Tier 3 (watch queue + per-item polling) is deleted entirely along with queue.go/queue_test.go/tier3.go. The PR description only says 'standalone pollers replace Pipeline' and does not explicitly address the dropped tier. The StateCheckMsg handler still uses TierWatch rate-limit bucket, implying state checks are now driven purely by NATS messages. This behavioral change deserves an explicit note in the PR description / changelog, and verification that whatever previously enqueued WatchItems now produces equivalent StateCheckMsgs (otherwise fast per-item status polling is lost).
🟡 **daemon/cmd/heimdallm/main.go:1670** — sendDiscoveryRepos is called unconditionally before the ticker loop on every startPollers invocation, including reloads where coldStart=false. The Tier 2 path respects coldStart, but discovery does not — every /reload triggers an immediate discovery pass and publish. If /reload is called frequently (e.g. UI config PATCH), this amplifies API usage. Gate the pre-ticker call on coldStart, or document why discovery intentionally differs from Tier 2.
🟡 **daemon/cmd/heimdallm/main.go:1690** — Rate-limit accounting is inconsistent for discovery: limiter.Acquire(TierDiscovery) is called before sendDiscoveryRepos inside the ticker branch, but the initial pre-ticker sendDiscoveryRepos call skips Acquire, and sendDiscoveryRepos itself never calls Acquire for disc.Discovered() (which may do API work). Either push Acquire inside sendDiscoveryRepos or add it to the initial call path so the budget reflects reality.
🟡 **daemon/cmd/heimdallm/main.go:1765** — runTier2 uses a hardcoded `time.After(2 * time.Second)` to wait for Tier 1's first batch. If discovery takes longer (large orgs, network latency, NATS consumer setup delay), the first tick runs with empty `repos` and processTick returns early — delaying first poll by the full PollInterval. Prefer waiting on the first reposChan value with a bounded timeout (e.g. select on reposChan and time.After), so the cold-start tick runs as soon as repos are actually available.
🟡 **daemon/cmd/heimdallm/main.go:1800** — adapter.FetchPRsToReview() does not accept a context. During shutdown mid-fetch, the call cannot be cancelled and wg.Wait() in the defer or reload path may block on an in-flight GitHub call. Pre-existing, but the refactor would have been a natural place to plumb context through.
🟡 **daemon/cmd/heimdallm/main.go:513** — The `limiter` variable intentionally outlives reload cycles (budget must persist across reloads), but the comment 'Shared rate limiter (was Pipeline.limiter)' doesn't state this invariant. A future refactor could easily move the declaration inside startPollers and silently break the rate budget. Expand the comment to state: limiter MUST be declared outside startPollers so reload does not reset the hourly budget.
🟡 **daemon/cmd/heimdallm/main.go:510** — The discoveryInterval and pollInterval are snapshotted once at startPollers entry under cfgMu, but tier1ConfigFn and tier2ConfigFn re-read cfg under cfgMu on every invocation. This means interval changes via /reload are only picked up at reload time (correct, because /reload calls startPollers again) but org/repo list changes are picked up per-tick. Worth a one-line comment making this split explicit so the reader doesn't wonder why intervals use the snapshot pattern while configs use the closure pattern.

### Suggestions

- Move `reposChan := make(chan []string, 1)` inside startPollers so each reload cycle owns its own channel and cannot race with lingering goroutines from the previous cycle.
- Track the inner reposChan reader goroutine in runTier2 with the parent WaitGroup (or fold its work into runTier2's main select) so shutdown fully drains it and no repo update is silently dropped.
- Add a PR-description paragraph (and a top-of-file comment) explicitly documenting that Tier 3 / WatchQueue was removed in favor of NATS-driven StateCheckMsg, with a pointer to where the equivalent behavior now lives.
- Replace the 2-second blind sleep in runTier2 with `select { case <-reposChan: ...; case <-time.After(timeout): ...; case <-ctx.Done(): return }` so cold-start polls run as soon as the first repo batch arrives.
- Either gate the pre-ticker sendDiscoveryRepos call on coldStart or route all discovery API calls through limiter.Acquire(TierDiscovery) for consistent budget accounting.
- Expand the `limiter` comment to state that it is intentionally declared outside startPollers so the hourly API budget persists across /reload cycles — otherwise a future refactor could silently reset it.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Replaces the Pipeline orchestrator with standalone poller goroutines managed directly in main.go. The refactor is structurally sound and the reload/shutdown logic is correctly ported. However, there is an untracked goroutine race on reload, queue.go deletion likely breaks tier2.go/tier3.go compilation (unless handled in truncated diff), and test coverage regresses with no replacements for deleted tests.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:810** — runTier2 spawns an inner goroutine to consume from reposChan that is NOT tracked by the WaitGroup. When pollerCancel() + pollerWg.Wait() complete during reload, this goroutine may still be running briefly, racing with the new Tier 2's goroutine for reads from the shared reposChan. Fix: use a nested WaitGroup inside runTier2 and defer wg.Wait() before returning, or move the consumer goroutine into startPollers where it can be tracked.
🔴 **daemon/internal/scheduler/queue.go:0** — queue.go (WatchQueue, WatchItem) is deleted, but tier2.go (line 67: WatchQueue *WatchQueue, line 150: deps.WatchQueue.Push) and tier3.go (line 35/39: *WatchItem, line 45: *WatchQueue) still reference these types. Unless tier2.go and tier3.go are also modified in the truncated portion of this diff, this PR will not compile.
⚠️ **daemon/cmd/heimdallm/main.go:539** — reposChan is allocated once outside startPollers and shared across reload cycles. On reload, the old bridge may have left a buffered message (cap=1) that the new Tier 2 consumes as stale data. Consider draining the channel after oldWg.Wait() returns, or create a fresh channel per startPollers invocation.
⚠️ **daemon/cmd/heimdallm/main.go:577** — The initial sendDiscoveryRepos call (cold-start path) bypasses the rate limiter — only the ticker path calls limiter.Acquire before sendDiscoveryRepos. The old RunTier1 had the same pattern, but it's worth a deliberate comment since the function also makes GitHub API calls through disc.Discovered().
⚠️ **daemon/internal/scheduler/bridge_test.go:0** — bridge_test.go and queue_test.go are deleted with no replacement tests for the new sendDiscoveryRepos, bridgeDiscovery, runBridgeConsumer, or runTier2 functions in main.go. These are core polling paths. Consider at minimum a TestRunTier2 with a fake adapter to cover the cold-start and tick paths.
⚠️ **daemon/cmd/heimdallm/main.go:1651** — sendDiscoveryRepos largely duplicates the existing sendRepos in scheduler/tier1.go. After this PR, RunTier1, Tier1Deps (tier1.go) and RunTier2, Tier2Deps (tier2.go) become dead code since Pipeline was their only caller. Either delete the now-unused tier files or refactor to reuse them.
🟡 **daemon/cmd/heimdallm/main.go:1702** — bridgeDiscovery's exponential backoff (1s → 30s) never resets to 1s after a successful runBridgeConsumer session. If the consumer runs for hours then fails, the first retry still uses the last backoff value. Reset backoff to 1s at the top of the for-loop body, before runBridgeConsumer.
🟡 **daemon/cmd/heimdallm/main.go:1209** — Reload path: cfg is swapped under cfgMu, then startPollers runs and the new cancel/wg are written back under cfgMu in a second critical section. Between these two lock acquisitions, a concurrent shutdown defer could read the stale pollerCancel/pollerWg. This window is narrow (reloadMu serialises reloads) but the defer-shutdown path only takes cfgMu, not reloadMu.

### Suggestions

- Track the inner reposChan-consumer goroutine inside runTier2 with a local WaitGroup and defer its Wait before returning, so the caller's wg.Wait() only completes after all child goroutines have exited.
- Delete or update tier1.go, tier2.go, and tier3.go to remove the now-unused RunTier1/RunTier2/RunTier3 functions and their dependency on the deleted WatchQueue type. This avoids dead code and ensures compilation.
- Add integration tests for the new standalone poller functions — at minimum a test that verifies runTier2 fires processTick on cold start, and a test for bridgeDiscovery forwarding messages to reposChan.
- Consider creating reposChan inside startPollers instead of sharing it across reload cycles, which eliminates the stale-buffer and goroutine-race concerns in one change.

---

PR #345, opened 2026-04-24
epic: embed NATS in backend — event-driven pipeline replacing tier-based polling (#298)

### LINKED ISSUE #286: bug: OS notifications stopped — SSE connection pool exhausted by zombie connections
## Problem

The user reports no OS notifications for days despite the daemon actively reviewing PRs and issues.

## Diagnosis

1. The daemon logs show \`notify\` entries (reviews started/completed) — it IS working
2. OS notifications are handled by the Flutter app, which receives SSE events from \`GET /events\` and calls \`local_notifier\`
3. Testing the SSE endpoint returns: **\`too many SSE connections\`**
4. The broker has a hard limit of 10 connections (\`maxSubscribers = 10\` in \`sse/broker.go:32\`)

The SSE connection pool is exhausted. Likely causes:
- Zombie connections from previous app instances (app crash, rebuild, etc.) that were never cleaned up
- The broker doesn't detect or evict dead connections
- Each app rebuild/restart may create a new SSE connection without closing the old one

## Impact

- No OS notifications for PR reviews, issue triage, or auto-implement events
- The user has no real-time awareness of daemon activity
- Silent failure — no error shown in the app

## Expected Fix

1. **Connection cleanup**: The broker should detect dead/stale SSE connections and evict them (heartbeat/ping mechanism)
2. **Connection reuse**: The Flutter app should close its SSE connection before creating a new one on hot restart / rebuild
3. **Graceful degradation**: If SSE connection fails, the app should show a warning (e.g., "Live updates unavailable") instead of silently failing

## Location

- \`daemon/internal/sse/broker.go\` — subscriber management, \`maxSubscribers = 10\`
- \`daemon/internal/server/handlers.go\` — \`handleSSE\` handler
- \`flutter_app/lib/features/dashboard/dashboard_providers.dart\` — SSE stream provider

### LINKED ISSUE #298: epic: embed NATS in backend — event-driven pipeline replacing tier-based polling

Muriano:
## Epic: Embed NATS in backend — event-driven pipeline replacing tier-based polling (#298)

This PR merges `main-nats` into `main`. **All code has been reviewed and approved** across 20+ individual PRs.

## Architecture change

```
Before: Tier 1 → channel → Tier 2 → goroutines (unbounded) → Pipeline.Run()
After:  Pollers → core NATS pub/sub → Workers (semaphore-bounded)
        Watch state in SQLite (durable, survives restart)
        SSE events via NATS fan-out (no subscriber limit)
```

### NATS migration (13 tasks)
- Embedded NATS server (core only, no JetStream — 33MB RSS)
- 5 NATS workers: ReviewWorker, PublishWorker, TriageWorker, ImplementWorker, StateWorker
- 5 NATS publishers: discovery, PR review, PR publish, issue triage/implement, state check
- SQLite-backed WatchStore replaces in-memory WatchQueue (durable backoff state)
- SSE bridge: broker → NATS → unlimited fan-out (fixes #286)
- Rate limiter shared between pollers and workers
- Old scheduler Pipeline removed — standalone polling goroutines
- 9 integration tests covering full message flow

### JetStream removal
- Removed JetStream entirely after discovering 500-750MB memory footprint
- Kept core NATS only (~20MB) for SSE event fan-out
- SQLite replaces JetStream streams/consumers/KV

### Bugfixes included
- #296: Stop issue-triage cost-runaway (FetchIssueCommentsOnly)
- #322: Issue re-triage loop + locked PR publish retry
- #324: Silence PR SHA-skip side effects + respect re-request review
- #326: Mark review orphan on permanent submit failure
- #352: Enroll all processed items in watch_state
- Gradual backfill of open items into watch_state (no startup blocking)
- SQLite deadlock fix (QueryRow + LEFT JOIN instead of open iterator)

### Stats
- ~3,000 lines added, ~1,500 removed
- New packages: `bus/`, `worker/`
- Removed: `scheduler/pipeline.go`, `tier1-3.go`, `queue.go`, `streams.go`

## Closes

Closes #298


sergiotejon:
## 🟡 LOW Issue

SSE→NATS bridge goroutine has no shutdown signal tied to eventBus.Stop() or main context. If broker.Subscribe() channel is closed on shutdown it exits, but publishes to eventBus.Conn() after Stop() may log spurious errors. Consider binding to a context.

**Location:** `daemon/cmd/heimdallm/main.go` line 172

---

sergiotejon:
## 🟡 LOW Issue

eventBus.Start uses context.Background() rather than a cancellable context; if NATS blocks during startup there's no way to abort. os.Exit(1) on failure bypasses the deferred eventBus.Stop() from earlier resources (broker etc.), but since this is startup it's acceptable — still worth noting for graceful init ordering.

**Location:** `daemon/cmd/heimdallm/main.go` line 163

---

sergiotejon:
## ⚠️ MEDIUM Issue

Significant duplication between triageHandler and implementHandler (~60 lines identical config resolution + RunOptions construction). Extract a shared helper like buildIssueRunOptions(msg, mode) to reduce drift risk — one will inevitably get a fix the other misses.

**Location:** `daemon/cmd/heimdallm/main.go` line 560

---

sergiotejon:
## ⚠️ MEDIUM Issue

In the reload path, pollerCancel/pollerWg are read under cfgMu, then written back under cfgMu after startPollers. Between oldCancel() and the second Lock(), a concurrent reader (e.g. shutdown defer) could observe the stale oldCancel/oldWg pair and call cancel()/Wait() on an already-stopped set. Low risk due to reloadMu serialization of reloads, but shutdown is not serialized with reload.

**Location:** `daemon/cmd/heimdallm/main.go` line 1207

---

sergiotejon:
## ⚠️ MEDIUM Issue

reviewHandler: if ctx is cancelled during limiter.Acquire, the message is silently dropped (returned without error). Comment says 'acked without processing' — verify NATS worker wrapper treats a bare return as ack; if so, PR reviews in-flight at shutdown are lost and rely on next-startup re-detection. Confirm this matches the documented behavior.

**Location:** `daemon/cmd/heimdallm/main.go` line 485

---

sergiotejon:
## ⚠️ MEDIUM Issue

publishHandler's 4xx/5xx detection uses string matching on err.Error() ('status 4', 'status 429'). Fragile — any wording change in the GitHub client error formatting silently flips retry semantics. Prefer a typed error or status code field.

**Location:** `daemon/cmd/heimdallm/main.go` line 520

---

sergiotejon:
## 🟡 LOW Issue

State poller and all workers use independent context.WithCancel(context.Background()) with separate defer cancels. On reload these are NOT restarted, so they continue using the original cfg snapshot pattern (they read cfg under cfgMu each iteration, which is fine) — but the inconsistency with pollers (which ARE restarted on reload) deserves a comment explaining why workers don't need reload.

**Location:** `daemon/cmd/heimdallm/main.go` line 700

---

sergiotejon:
## ⚠️ MEDIUM Issue

stateHandler key format uses '.' separator with a comment 'NATS KV doesn't allow ":"'. Ensure this key scheme is consistent with how keys are written elsewhere (Enroll, ScanReady) — a mismatch would silently return zero LastSeen and defeat the dedup check.

**Location:** `daemon/cmd/heimdallm/main.go` line 750

---

sergiotejon (approved):

Large architectural refactor replacing tier-based polling with an embedded NATS event-driven pipeline. Overall structure is sound with good separation of workers, proper context cancellation, and preserved semantics (cold start, reload serialization). However, there is significant code duplication between the triage and implement workers, potential race conditions around pollerCancel/pollerWg access, and the SSE-to-NATS bridge has unbounded goroutine behavior on broker cap exhaustion.

**8 issue(s) found** — see individual comments above for details.

### Suggestions

- Extract the shared issue pipeline options builder used by triageHandler and implementHandler to a helper to reduce duplication and drift risk.
- Replace string-matching on GitHub error messages in publishHandler with a typed error or explicit HTTP status code check.
- Add a comment or assertion that the KV key format ('type.githubID') is the single source of truth, referenced by Enroll/Scan/Get paths.
- Consider consolidating the many per-worker context.WithCancel pairs into a single shutdown context to simplify ordering guarantees.
- Document (or guard) the shutdown-vs-reload ordering for pollerCancel/pollerWg access, since shutdown's defer reads them without reloadMu.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Large architectural migration from tier-based polling to NATS event-driven workers. The overall structure is sound, but there are shutdown-ordering risks, fragile error classification, significant code duplication, and a potential channel leak in the SSE bridge.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:637** — Worker goroutines (review, publish, triage, implement, state) are cancelled via context but never waited on. Defers run LIFO so `eventBus.Stop()` (registered earliest) runs last — but without a WaitGroup.Wait() on each worker, in-flight handlers may still be using the NATS connection when the bus shuts down, causing panics or lost messages. The pollers correctly use a WaitGroup; apply the same pattern to all workers.
🔴 **daemon/cmd/heimdallm/main.go:699** — HTTP error classification in publishHandler uses fragile string matching (`strings.Contains(errStr, "status 4")`). This matches any error whose message happens to contain the substring 'status 4' anywhere (e.g. 'processed status 4 items', a wrapped inner error, etc.). Prefer a typed error or at least a tighter pattern like a regex anchored to the HTTP client's actual format (e.g. `regexp.MustCompile(`HTTP status 4\d{2}`)`).
⚠️ **daemon/cmd/heimdallm/main.go:734** — triageHandler and implementHandler are ~70 lines each and nearly identical — they differ only in `ghIssue.Mode`. This violates DRY and means any future config-resolution bug must be fixed in two places. Extract a shared `runIssueWorker(ctx, msg, mode)` helper.
⚠️ **daemon/cmd/heimdallm/main.go:163** — The SSE→NATS bridge goroutine subscribes to the broker (`bridgeCh`) but never unsubscribes or drains on shutdown. When the bridge goroutine exits (e.g. eventBus.Stop() closes NATS), the broker still holds a reference to bridgeCh and may block trying to send to it, stalling all other SSE subscribers.
⚠️ **daemon/cmd/heimdallm/main.go:519** — `reposChan` is created once outside `startPollers` but consumed by goroutines launched inside it. On config reload, the old pollers are cancelled and new ones start — but a stale message buffered in the channel (cap 1) from the previous cycle could be read by the new bridge consumer, causing the new Tier 2 to operate on an outdated repo list until the next discovery tick.
🟡 **daemon/cmd/heimdallm/main.go:599** — reviewHandler silently swallows `limiter.Acquire` errors (returns without logging). While the comment says this is acceptable during shutdown, adding a debug-level log would aid troubleshooting if acquire fails for unexpected reasons (e.g. a bug in the limiter).
🟡 **daemon/cmd/heimdallm/main.go:527** — The rate-limiter refill goroutine uses a 1-hour ticker, so the first refill won't fire until 1 hour after start. If the daemon starts with a partially-depleted budget (e.g. quick restart after heavy use), it could hit the limit early. Consider calling `limiter.Refill()` once at startup or using a time.Timer that fires at the next clock-hour boundary.

### Suggestions

- Add a WaitGroup (or errgroup) for all NATS workers and Wait() on it before calling eventBus.Stop(), mirroring the pollerWg pattern already used for the discovery/tier2 goroutines.
- Extract the shared triage/implement handler body into a single helper like `runIssuePipeline(ctx context.Context, msg bus.IssueMsg, mode config.IssueMode)` to eliminate the ~70-line duplication.
- Replace the string-based HTTP status classification with a typed error returned by the GitHub client (e.g. `type GitHubAPIError struct { StatusCode int }`) so publish-worker can switch on the status code directly.
- Drain `reposChan` inside `startPollers` before launching the new bridge consumer, or create a fresh channel on each reload cycle, to prevent stale repo lists from leaking across reloads.
- Register an unsubscribe/cleanup for `bridgeCh` (tied to eventBus lifecycle or a dedicated context) so the broker doesn't block on a dead subscriber during shutdown.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Large architectural migration from tier-based polling to NATS event-driven workers. The overall structure is sound, but there are shutdown-ordering risks, fragile error classification, significant code duplication, and a potential channel leak in the SSE bridge.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:637** — Worker goroutines (review, publish, triage, implement, state) are cancelled via context but never waited on. Defers run LIFO so `eventBus.Stop()` (registered earliest) runs last — but without a WaitGroup.Wait() on each worker, in-flight handlers may still be using the NATS connection when the bus shuts down, causing panics or lost messages. The pollers correctly use a WaitGroup; apply the same pattern to all workers.
🔴 **daemon/cmd/heimdallm/main.go:699** — HTTP error classification in publishHandler uses fragile string matching (`strings.Contains(errStr, "status 4")`). This matches any error whose message happens to contain the substring 'status 4' anywhere (e.g. 'processed status 4 items', a wrapped inner error, etc.). Prefer a typed error or at least a tighter pattern like a regex anchored to the HTTP client's actual format (e.g. `regexp.MustCompile(`HTTP status 4\d{2}`)`).
⚠️ **daemon/cmd/heimdallm/main.go:734** — triageHandler and implementHandler are ~70 lines each and nearly identical — they differ only in `ghIssue.Mode`. This violates DRY and means any future config-resolution bug must be fixed in two places. Extract a shared `runIssueWorker(ctx, msg, mode)` helper.
⚠️ **daemon/cmd/heimdallm/main.go:163** — The SSE→NATS bridge goroutine subscribes to the broker (`bridgeCh`) but never unsubscribes or drains on shutdown. When the bridge goroutine exits (e.g. eventBus.Stop() closes NATS), the broker still holds a reference to bridgeCh and may block trying to send to it, stalling all other SSE subscribers.
⚠️ **daemon/cmd/heimdallm/main.go:519** — `reposChan` is created once outside `startPollers` but consumed by goroutines launched inside it. On config reload, the old pollers are cancelled and new ones start — but a stale message buffered in the channel (cap 1) from the previous cycle could be read by the new bridge consumer, causing the new Tier 2 to operate on an outdated repo list until the next discovery tick.
🟡 **daemon/cmd/heimdallm/main.go:599** — reviewHandler silently swallows `limiter.Acquire` errors (returns without logging). While the comment says this is acceptable during shutdown, adding a debug-level log would aid troubleshooting if acquire fails for unexpected reasons (e.g. a bug in the limiter).
🟡 **daemon/cmd/heimdallm/main.go:527** — The rate-limiter refill goroutine uses a 1-hour ticker, so the first refill won't fire until 1 hour after start. If the daemon starts with a partially-depleted budget (e.g. quick restart after heavy use), it could hit the limit early. Consider calling `limiter.Refill()` once at startup or using a time.Timer that fires at the next clock-hour boundary.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Large, well-structured migration from tier-based polling to a NATS-backed event pipeline. Core architecture is sound (JetStream streams, KV-backed watch, dedicated workers per subject), reload lifecycle is handled carefully, and shutdown ordering is mostly correct. However, main.go has grown to ~1800 lines with significant duplication (triage vs implement handlers) and several fragile patterns that should be addressed before merge: error-string matching for retry semantics, un-contexted bridge goroutine, workers not restarted on reload, and a reload/shutdown race on pollerCancel.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:520** — publishHandler classifies permanent vs transient GitHub errors by substring-matching err.Error() for 'status 4' and 'status 429'. Any rewording in the GitHub client (e.g., 'HTTP 404' vs 'status 404') silently flips retry semantics — transient errors get acked and lost, or permanent errors get retried forever. Use a typed error with a StatusCode field, or expose an IsPermanent() method on the client's error type.
⚠️ **daemon/cmd/heimdallm/main.go:560** — triageHandler and implementHandler duplicate ~60 lines of config resolution and RunOptions construction, differing only in IssueMode. Extract buildIssueRunOptions(msg, mode) — otherwise one will drift when a field is added to the other (e.g., a new PR option added only to triage).
⚠️ **daemon/cmd/heimdallm/main.go:1207** — Reload path reads pollerCancel/pollerWg under cfgMu, cancels+waits outside the lock, then starts new pollers and writes back. The shutdown defer also reads these under cfgMu. If shutdown fires while reload is between oldCancel() and the second write, the defer observes the stale (already-cancelled) pair, Wait() returns immediately, and the newly-started poller set leaks with no cancel signal. reloadMu only serializes concurrent reloads, not reload vs shutdown.
⚠️ **daemon/cmd/heimdallm/main.go:485** — reviewHandler returns silently when limiter.Acquire fails on ctx cancellation. If the worker wrapper treats a nil return as ack, in-flight review requests are lost at shutdown and rely on next-startup re-detection. Confirm the worker wraps NAK-on-shutdown, or propagate an error so the message is redelivered.
⚠️ **daemon/cmd/heimdallm/main.go:750** — stateHandler constructs KV key as fmt.Sprintf('%s.%d', msg.Type, msg.GithubID) with a comment that NATS KV disallows ':'. If Enroll/ScanReady/EvictStale use a different separator or include Repo/Number in the key, Get() silently returns an error, LastSeen stays zero, and CheckItem's dedup check is defeated — every state check registers as changed. Centralize key construction in one helper in the bus package.
🟡 **daemon/cmd/heimdallm/main.go:152** — SSE→NATS bridge goroutine is never explicitly cancelled. It exits only when broker.Subscribe()'s channel closes. If eventBus.Stop() runs before broker is torn down, Publish() calls log spurious warnings and may race with JetStream teardown. Bind the loop to main context or to eventBus lifecycle.
🟡 **daemon/cmd/heimdallm/main.go:163** — eventBus.Start uses context.Background() — unabortable if NATS init hangs. Use a startup context with a reasonable timeout.
🟡 **daemon/cmd/heimdallm/main.go:475** — reviewWorker, publishW, triageW, implementW, stateW, and the state-poller are each given their own context.WithCancel(context.Background()) and are NOT torn down and restarted on reload (unlike pollers). This is likely intentional (workers read cfg under cfgMu per iteration), but the asymmetry with pollers is surprising and deserves a comment explaining why — otherwise the next contributor may 'fix' it.
🟡 **daemon/cmd/heimdallm/main.go:528** — Rate limiter Refill is hardcoded to a 1-hour ticker and not restarted on reload. If in the future the 4500/hour budget becomes configurable, this ticker won't pick up the change. Minor — flag as a known limitation.
🟡 **daemon/cmd/heimdallm/main.go:157** — When broker.Subscribe() returns nil (subscriber cap reached), the bridge is silently disabled with only a warning. Any client relying on SSE events from pre-NATS publishers will silently miss events. At minimum surface this as a startup metric or fail-fast — the 'interim bridge' comment suggests this will be removed soon, but until Task 12 lands this is the only path for those events.
🟡 **daemon/cmd/heimdallm/main.go:326** — runReview signature changed from returning nothing to returning *store.Review. All existing early-return paths now return nil, which matches the new contract, but any future contributor adding a return path must remember to return the review. Consider naming the return value (rev *store.Review) so zero-value returns are explicit.

### Suggestions

- Extract the two issue handlers (triage/implement) into one parameterized handler factory — eliminates ~60 lines of duplication and fixes a drift risk.
- Move GitHub error classification (permanent vs transient) into the gh client package as a typed error or IsPermanent()/StatusCode() method, so publishHandler doesn't depend on error-message wording.
- Centralize NATS KV key construction (watch bucket) in the bus package so stateHandler, Enroll, ScanReady, and EvictStale cannot drift on separator/format.
- main.go is approaching 1800 lines. Consider extracting the worker wiring (reviewHandler, publishHandler, triageHandler, implementHandler, stateHandler) into daemon/internal/worker/wiring.go or similar — main.go should orchestrate, not embed 5 full handler bodies.
- Serialize reload vs shutdown: have the shutdown defer also acquire reloadMu, or use an atomic.Pointer for pollerCancel/pollerWg so the swap is a single atomic operation.
- Add a test for reload under load: concurrent /reload + in-flight worker messages + pending state checks, to catch the goroutine-leak race and any KV key mismatch.
- Clarify in a comment near the worker blocks why workers are not restarted on reload (they read cfg per-iteration under cfgMu) — prevents a well-meaning future 'symmetry' refactor.

---

sergiotejon:
## ⚠️ MEDIUM Issue

Unresolved: triageHandler and implementHandler still contain ~60 lines of duplicated config resolution and RunOptions construction. Extract a shared buildIssueRunOptions(msg, mode) helper to prevent drift.

**Location:** `daemon/cmd/heimdallm/main.go` line 560

---

sergiotejon:
## ⚠️ MEDIUM Issue

Unresolved: reload path reads pollerCancel/pollerWg under cfgMu, releases, then writes back after startPollers. Shutdown defer is not serialized with reloadMu, so a concurrent shutdown could observe stale pointers. Consider holding cfgMu across the swap or having shutdown also take reloadMu.

**Location:** `daemon/cmd/heimdallm/main.go` line 1207

---

sergiotejon:
## ⚠️ MEDIUM Issue

Unresolved: publishHandler still classifies GitHub errors via strings.Contains(err.Error(), "status 4"/"status 429"). Brittle against any change in client error formatting; prefer a typed error exposing the HTTP status code.

**Location:** `daemon/cmd/heimdallm/main.go` line 520

---

sergiotejon:
## ⚠️ MEDIUM Issue

Unresolved: reviewHandler silently returns on limiter.Acquire ctx cancellation. Confirm and document that the worker wrapper acks on nil return; otherwise in-flight PRs at shutdown are lost without redelivery guarantee beyond next-startup re-detection.

**Location:** `daemon/cmd/heimdallm/main.go` line 485

---

sergiotejon:
## ⚠️ MEDIUM Issue

Unresolved: stateHandler key format "{type}.{githubID}" must match the write path in Enroll/ScanReady. A silent mismatch yields zero LastSeen and defeats dedup. Add a shared key helper or a test asserting round-trip.

**Location:** `daemon/cmd/heimdallm/main.go` line 750

---

sergiotejon:
## 🟡 LOW Issue

Unresolved: SSE→NATS bridge goroutine is not bound to a context; after eventBus.Stop() any further Publish calls may log spurious errors. Bind to a shutdown context.

**Location:** `daemon/cmd/heimdallm/main.go` line 172

---

sergiotejon:
## 🟡 LOW Issue

Unresolved: eventBus.Start uses context.Background(); a hanging NATS startup cannot be aborted. Acceptable at startup but worth a timeout.

**Location:** `daemon/cmd/heimdallm/main.go` line 163

---

sergiotejon:
## 🟡 LOW Issue

Unresolved: workers and state poller use independent context.Background()-rooted cancels and are not restarted on reload, unlike pollers. Add a comment explaining why they don't need reload, or unify lifecycle.

**Location:** `daemon/cmd/heimdallm/main.go` line 700

---

sergiotejon (approved):

Re-review of the NATS embed PR. The diff shown is identical to the previously reviewed version — no changes have been made in response to the prior feedback. All previously flagged issues remain present in the code.

**8 issue(s) found** — see individual comments above for details.

### Suggestions

- Extract buildIssueRunOptions(ghIssue, mode, cfg) helper shared by triage and implement workers to remove ~60 lines of duplication.
- Introduce a typed GitHub error (e.g. *gh.HTTPError with StatusCode int) and switch publishHandler to errors.As rather than substring matching.
- Centralize the watch KV key format in a single helper (e.g. bus.WatchKey(type, githubID)) used by Enroll, ScanReady, Get, and EvictStale to prevent silent mismatches.
- Either serialize shutdown with reloadMu or hold cfgMu across the full oldCancel/Wait/startPollers/swap sequence to eliminate the stale-pointer race.
- Bind the SSE bridge goroutine and eventBus.Start to a cancellable context derived from main's lifecycle, and document worker lifecycle vs poller lifecycle on reload.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of PR #345. The diff is unchanged from the prior review — none of the previously flagged findings have been addressed in code. All medium-severity issues remain: brittle error string matching in publishHandler, duplicated config/RunOptions construction in triage/implement handlers, reload-vs-shutdown race on pollerCancel/pollerWg, silent ack on limiter cancellation in reviewHandler, and ad-hoc KV key construction in stateHandler. Low-severity items (SSE bridge context, eventBus.Start context, worker lifecycle comment) also remain unaddressed.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:520** — Still unresolved: publishHandler classifies GitHub errors via strings.Contains(err.Error(), "status 4") and "status 429". Any rewording in the GitHub client (e.g. 'HTTP 404') silently flips retry semantics — transient errors get acked and lost, or permanent errors get retried forever. Expose a typed error with a StatusCode field (or IsPermanent() method) on the client and branch on that instead of substrings.
⚠️ **daemon/cmd/heimdallm/main.go:560** — Still unresolved: triageHandler and implementHandler duplicate ~60 lines of config resolution and RunOptions construction, differing only in IssueMode and a couple of log prefixes. Extract buildIssueRunOptions(msg, mode) so a future change to one (e.g., new PR option) cannot silently diverge.
⚠️ **daemon/cmd/heimdallm/main.go:1207** — Still unresolved: reload path reads pollerCancel/pollerWg under cfgMu, drops the lock, cancels+waits, then re-acquires to write the new pair. Shutdown defer reads the same pair under cfgMu but is NOT serialized with reloadMu. If shutdown fires after oldCancel() but before the second write, defer sees the already-cancelled pair, Wait() returns immediately, and the newly-started poller set leaks. Either take reloadMu in the shutdown defer, or hold cfgMu across the swap (with startPollers releasing it internally before any blocking call).
⚠️ **daemon/cmd/heimdallm/main.go:485** — Still unresolved: reviewHandler returns silently when limiter.Acquire fails on ctx cancellation. Comment claims the message is 'acked without processing' — this needs to be verified against worker.NewReviewWorker's wrapper, and either documented at the call site or changed to propagate an error so NATS redelivers. Relying on next-startup re-detection is a behavior contract that should be explicit.
⚠️ **daemon/cmd/heimdallm/main.go:750** — Still unresolved: stateHandler constructs the KV key inline as fmt.Sprintf("%s.%d", msg.Type, msg.GithubID). If Enroll/ScanReady/EvictStale in the bus package use a different scheme (different separator, or include Repo/Number), Get() returns an error, LastSeen stays zero, and CheckItem's dedup check is defeated — every state check registers as changed. Centralize key construction in one helper in the bus package (e.g. bus.WatchKey(type, githubID)) and add a round-trip test.
🟡 **daemon/cmd/heimdallm/main.go:172** — Still unresolved: SSE→NATS bridge goroutine is not bound to a context and exits only when broker.Subscribe()'s channel closes. If eventBus.Stop() runs before the broker is torn down, Publish() logs spurious warnings and may race with JetStream teardown. Bind the loop to a shutdown context or eventBus lifecycle.
🟡 **daemon/cmd/heimdallm/main.go:163** — Still unresolved: eventBus.Start uses context.Background(); a hanging NATS init cannot be aborted. Use a startup context with a reasonable timeout (e.g. 30s).
🟡 **daemon/cmd/heimdallm/main.go:700** — Still unresolved: workers and state poller use independent context.WithCancel(context.Background()) and are NOT restarted on reload, unlike pollers. Likely intentional (workers read cfg under cfgMu per iteration) but the asymmetry is surprising. Add a one-line comment explaining why reload is safe without restarting them, so the next contributor doesn't 'fix' it.

### Suggestions

- Move the inline KV key format into a bus.WatchKey(type, githubID) helper used by both Enroll/ScanReady/Get; add a unit test asserting round-trip through the KV so any future separator change fails loudly.
- Introduce a bus.PermanentError (or similar) that wraps GitHub client errors with a StatusCode, and have publishHandler check errors.As rather than strings.Contains.
- Take reloadMu in the shutdown defer (or hold cfgMu across the full reload swap) so shutdown and reload cannot race on pollerCancel/pollerWg.
- Extract buildIssueRunOptions(ctx, msg, mode) used by both issue workers — currently the only real difference is ghIssue.Mode and a log prefix.

---

sergiotejon:
## ⚠️ MEDIUM Issue

Still unresolved: triageHandler and implementHandler duplicate ~60 lines of config resolution and RunOptions construction. Extract a shared buildIssueRunOptions(msg, mode) helper to prevent drift between the two modes.

**Location:** `daemon/cmd/heimdallm/main.go` line 560

---

sergiotejon:
## ⚠️ MEDIUM Issue

Still unresolved: reload path reads pollerCancel/pollerWg under cfgMu, releases the lock around oldCancel()/oldWg.Wait() and startPollers, then writes new values back. Shutdown defer is not serialized via reloadMu, so a concurrent shutdown can observe stale pointers and either double-stop or leak the post-reload set. Hold cfgMu across the swap, or have the shutdown defer also acquire reloadMu.

**Location:** `daemon/cmd/heimdallm/main.go` line 1207

---

sergiotejon:
## ⚠️ MEDIUM Issue

Still unresolved: publishHandler classifies GitHub errors with strings.Contains(err.Error(),"status 4"/"status 429"). Brittle — any reformatting in the GitHub client silently flips retry semantics (permanent ack vs nak). Prefer a typed error exposing HTTP status code.

**Location:** `daemon/cmd/heimdallm/main.go` line 520

---

sergiotejon:
## ⚠️ MEDIUM Issue

Still unresolved: reviewHandler returns silently on limiter.Acquire ctx cancellation. Relies on worker wrapper's ack-on-nil-return semantics; if that ever changes, in-flight PRs at shutdown would be lost without redelivery. Add an explicit comment and/or a test pinning the wrapper contract.

**Location:** `daemon/cmd/heimdallm/main.go` line 485

---

sergiotejon:
## ⚠️ MEDIUM Issue

Still unresolved: stateHandler builds key as fmt.Sprintf("%s.%d", type, githubID). Must exactly match the writer in WatchKV.Enroll/ScanReady — a silent mismatch yields zero LastSeen and defeats dedup. Introduce a shared key helper (e.g. bus.WatchKey) used by all three sites, or add a round-trip test.

**Location:** `daemon/cmd/heimdallm/main.go` line 750

---

sergiotejon:
## 🟡 LOW Issue

Still unresolved: SSE→NATS bridge goroutine is not bound to a shutdown context. After eventBus.Stop(), Publish calls from the still-draining broker channel will log spurious errors. Bind to a cancellable context or check a done signal before publishing.

**Location:** `daemon/cmd/heimdallm/main.go` line 172

---

sergiotejon:
## 🟡 LOW Issue

Still unresolved: eventBus.Start uses context.Background(); a hung NATS startup cannot be aborted. Use context.WithTimeout for a bounded startup window.

**Location:** `daemon/cmd/heimdallm/main.go` line 163

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Re-review of the NATS embed PR. The diff shown is unchanged from the prior review on the key problem areas (triage/implement duplication, reload/shutdown race on pollerCancel/pollerWg, string-based GitHub error classification, silent ack on limiter cancellation, stateHandler key format, SSE bridge lifecycle, eventBus.Start context, worker lifecycle comment). All previously flagged findings remain applicable; none appear addressed in this diff.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:560** — Still unresolved: triageHandler and implementHandler duplicate ~60 lines of config resolution and RunOptions construction. Extract a shared buildIssueRunOptions(msg, mode) helper to prevent drift between the two modes.
⚠️ **daemon/cmd/heimdallm/main.go:1207** — Still unresolved: reload path reads pollerCancel/pollerWg under cfgMu, releases the lock around oldCancel()/oldWg.Wait() and startPollers, then writes new values back. Shutdown defer is not serialized via reloadMu, so a concurrent shutdown can observe stale pointers and either double-stop or leak the post-reload set. Hold cfgMu across the swap, or have the shutdown defer also acquire reloadMu.
⚠️ **daemon/cmd/heimdallm/main.go:520** — Still unresolved: publishHandler classifies GitHub errors with strings.Contains(err.Error(),"status 4"/"status 429"). Brittle — any reformatting in the GitHub client silently flips retry semantics (permanent ack vs nak). Prefer a typed error exposing HTTP status code.
⚠️ **daemon/cmd/heimdallm/main.go:485** — Still unresolved: reviewHandler returns silently on limiter.Acquire ctx cancellation. Relies on worker wrapper's ack-on-nil-return semantics; if that ever changes, in-flight PRs at shutdown would be lost without redelivery. Add an explicit comment and/or a test pinning the wrapper contract.
⚠️ **daemon/cmd/heimdallm/main.go:750** — Still unresolved: stateHandler builds key as fmt.Sprintf("%s.%d", type, githubID). Must exactly match the writer in WatchKV.Enroll/ScanReady — a silent mismatch yields zero LastSeen and defeats dedup. Introduce a shared key helper (e.g. bus.WatchKey) used by all three sites, or add a round-trip test.
🟡 **daemon/cmd/heimdallm/main.go:172** — Still unresolved: SSE→NATS bridge goroutine is not bound to a shutdown context. After eventBus.Stop(), Publish calls from the still-draining broker channel will log spurious errors. Bind to a cancellable context or check a done signal before publishing.
🟡 **daemon/cmd/heimdallm/main.go:163** — Still unresolved: eventBus.Start uses context.Background(); a hung NATS startup cannot be aborted. Use context.WithTimeout for a bounded startup window.
🟡 **daemon/cmd/heimdallm/main.go:700** — Still unresolved: state poller and all four workers use independent context.Background()-rooted cancels and are not restarted on reload, unlike pollers. Add a comment explaining why they don't need reload (they read cfg under cfgMu per-iteration), or unify lifecycle with the poller set.

---

sergiotejon:
## 🟡 LOW Issue

Still unresolved: state poller and all four workers use independent context.Background()-rooted cancels and are not restarted on reload, unlike pollers. Add a comment explaining why they don't need reload (they read cfg under cfgMu per-iteration), or unify lifecycle with the poller set.

**Location:** `daemon/cmd/heimdallm/main.go` line 700

---

sergiotejon (approved):

Re-review of the NATS embed PR. The diff shown is unchanged from the prior review on the key problem areas (triage/implement duplication, reload/shutdown race on pollerCancel/pollerWg, string-based GitHub error classification, silent ack on limiter cancellation, stateHandler key format, SSE bridge lifecycle, eventBus.Start context, worker lifecycle comment). All previously flagged findings remain applicable; none appear addressed in this diff.

**8 issue(s) found** — see individual comments above for details.

### Suggestions

- Extract buildIssueRunOptions(msg bus.IssueMsg, mode config.IssueMode) (*gh.Issue, issuepipeline.RunOptions, error) and call it from both triageHandler and implementHandler.
- Introduce a bus.WatchKey(type string, githubID int64) string helper and use it in Enroll, ScanReady, and stateHandler so the key format can never drift.
- Have the GitHub client return a typed *HTTPError{Status int} so publishHandler can switch on status code rather than substring-match err.Error().
- Either serialize shutdown with reloadMu or hold cfgMu across the full stop→start swap in the reload path to eliminate the pollerCancel/pollerWg race.
- Add a brief comment block above the worker/state-poller goroutines documenting that they intentionally survive reload (unlike pollers) because they read cfg snapshots per message.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of NATS migration PR. The diff shown matches the previous review — all prior findings remain unaddressed per the discussion thread. No new changes detected to evaluate.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:520** — publishHandler still classifies GitHub errors via strings.Contains(err.Error(), "status 4") / "status 429". Brittle — any rewording in the GitHub client (e.g. 'HTTP 404') silently flips retry semantics: transient errors get acked and lost, or permanent errors get retried forever. Expose a typed error with StatusCode (or IsPermanent()) on the client and branch on that.
⚠️ **daemon/cmd/heimdallm/main.go:560** — triageHandler and implementHandler still duplicate ~60 lines of config resolution and RunOptions construction, differing only in IssueMode and log prefixes. Extract buildIssueRunOptions(msg, mode) so future changes can't silently diverge between modes.
⚠️ **daemon/cmd/heimdallm/main.go:1207** — Reload path reads pollerCancel/pollerWg under cfgMu, drops the lock around oldCancel()/oldWg.Wait() and startPollers, then re-acquires to write the new pair. Shutdown defer reads the same pair under cfgMu but is NOT serialized with reloadMu. If shutdown fires mid-reload, defer may see the already-cancelled pair (Wait returns immediately) and leak the newly-started set. Either take reloadMu in the shutdown defer, or hold cfgMu across the swap.
⚠️ **daemon/cmd/heimdallm/main.go:485** — reviewHandler returns silently when limiter.Acquire fails on ctx cancellation. Comment claims 'acked without processing' — this relies on worker.NewReviewWorker's wrapper treating nil return as ack. Pin this contract with a test or explicit comment at the wrapper; otherwise a future worker refactor could silently lose in-flight PRs at shutdown.
⚠️ **daemon/cmd/heimdallm/main.go:750** — stateHandler constructs the KV key inline as fmt.Sprintf("%s.%d", msg.Type, msg.GithubID). If Enroll/ScanReady/EvictStale in the bus package use a different scheme, Get() returns an error, LastSeen stays zero, and CheckItem's dedup check is defeated — every state check registers as changed. Centralize key construction in the bus package (e.g. bus.WatchKey) and add a round-trip test.
🟡 **daemon/cmd/heimdallm/main.go:172** — SSE→NATS bridge goroutine is not bound to a shutdown context; it exits only when broker.Subscribe() closes. If eventBus.Stop() runs before broker teardown, Publish() logs spurious warnings and may race with JetStream teardown. Bind the loop to a shutdown context or the eventBus lifecycle.
🟡 **daemon/cmd/heimdallm/main.go:163** — eventBus.Start uses context.Background(); a hanging NATS init cannot be aborted. Use a startup context with a reasonable timeout (e.g. 30s).
🟡 **daemon/cmd/heimdallm/main.go:700** — Workers and state poller use independent context.WithCancel(context.Background()) and are NOT restarted on reload, unlike pollers. Likely intentional (they read cfg under cfgMu per iteration) but the asymmetry is surprising. Add a one-line comment explaining why reload is safe without restarting them.

### Suggestions

- Introduce typed GitHub errors with a StatusCode field to replace string-match retry classification in publishHandler.
- Extract buildIssueRunOptions(msg, mode) used by both triage and implement workers.
- Either hold cfgMu across the reload swap or acquire reloadMu in the shutdown defer to serialize reload vs shutdown.
- Add a bus.WatchKey(type, githubID) helper used by Enroll/ScanReady/stateHandler, plus a round-trip test.
- Add a comment near the worker/state-poller startup explaining why they don't need reload restart.

---

PR #350, opened 2026-04-24
refactor(bus): drop JetStream — core NATS + SQLite (750MB → 34MB)

Muriano:
## Summary

Remove JetStream from the embedded NATS server. Keep core NATS only for SSE event fan-out. Replace JetStream with SQLite + Go primitives.

### Memory

| Before | After |
|--------|-------|
| 750 MB RSS | **34 MB RSS** |

### What was removed
- JetStream server mode (`JetStream: true`, `StoreDir`, memory/store caps)
- 3 JetStream streams (HEIMDALLM_WORK, DISCOVERY, EVENTS)
- 6 durable consumers
- NATS KV bucket (HEIMDALLM_WATCH)
- `bus/streams.go` (deleted)
- All `jetstream` package imports

### What replaced it
| JetStream feature | Replacement |
|---|---|
| Durable streams/consumers | Core NATS pub/sub (fire-and-forget, pollers re-detect) |
| KV bucket (watch state) | SQLite `watch_state` table |
| MaxAckPending (backpressure) | Go counting semaphore (`chan struct{}`) |
| Nats-Msg-Id dedup | Store-level dedup (PRAlreadyReviewed, etc.) |
| NakWithDelay (retry) | Poller re-enqueues on next cycle |

### What stayed
- Embedded nats-server (core NATS only — ~20MB base)
- SSE bridge: `conn.Publish` / `conn.ChanSubscribe` — unchanged
- All subjects and message types — unchanged
- All worker handler logic — unchanged

### Stats
- -1,645 lines / +906 lines (net -739)
- 24 files changed

## Test plan

- [ ] `go test ./... -count=1` — all 14 packages pass
- [ ] Daemon RSS: ~34MB (measured)
- [ ] Smoke test: pollers, workers, SSE all functional


vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Solid refactor replacing JetStream with core NATS + SQLite watch store, dramatically reducing memory footprint (750MB → 34MB). The change simplifies the bus layer and removes durable storage requirements. However, this is a significant architectural shift with notable tradeoffs around message durability and at-most-once delivery that should be carefully considered.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:587** — Loss of message durability: Core NATS provides at-most-once delivery with no persistence. If the daemon crashes or restarts with messages in flight, work will be lost. The previous JetStream implementation guaranteed durability (as demonstrated by the removed TestIntegration_Durability test). Ensure this tradeoff is acceptable for all message types (PR reviews, triage, implement, publish) and document the recovery semantics (e.g., will a missed PR review be retried via polling?).
🔴 **daemon/internal/bus/bus.go:22** — Removed JetStream deduplication. The previous work stream had a 2-minute dedup window via MsgID, and discovery stream had 5 minutes. With core NATS, duplicate publishes will be delivered multiple times. Verify workers are idempotent or handle duplicates explicitly (e.g., via the SQLite watch store or DB-level idempotency).
🔴 **daemon/cmd/heimdallm/main.go:1004** — Loss of NAK/retry semantics. PublishWorker tests previously verified messages were re-delivered on handler error (TestIntegration_PRPublishFlow_NakOnError was deleted). With core NATS, a handler returning an error simply drops the message. Confirm this is intentional — transient errors (network, rate limits) previously got automatic retries via JetStream; now they are lost unless workers implement their own retry logic.
⚠️ **daemon/internal/bus/bus_test.go:38** — TestBus_MaxConcurrentWorkers_Clamp asserts clamping of 0 → 1 but the diff in bus.go does not show any clamping logic in New() or MaxConcurrentWorkers(). Verify the clamp is actually implemented somewhere (possibly in a part of the diff that is truncated or in another file); otherwise this test will fail.
⚠️ **daemon/cmd/heimdallm/main.go:593** — bridgeDiscovery no longer has retry/backoff logic. The previous implementation retried with exponential backoff on consumer errors. ChanSubscribe failure now causes immediate exit with just a log. For a long-running daemon this is brittle — if the subscribe transiently fails at startup the bridge silently dies. Consider retrying or making subscribe failure fatal.
🟡 **daemon/cmd/heimdallm/main.go:598** — Channel buffer size of 8 for discovery messages is unexplained. If a burst exceeds 8 pending messages while the bridge is blocked writing to out, core NATS will drop messages (SlowConsumer). Consider documenting the rationale or adding slow-consumer telemetry.
🟡 **daemon/cmd/heimdallm/main.go:149** — filepath import may now be unused since DataDir is gone. If no other usage remains, remove the import to avoid a build failure.
🟡 **daemon/internal/bus/integration_test.go:42** — time.Sleep(100ms) before publishing is fragile — it relies on subscription being attached quickly. Prefer conn.Flush() after subscribe to guarantee the server has registered the subscription before publishing.
🟡 **daemon/internal/bus/bus.go:76** — The Drain() call is preserved, but the surrounding comment about JetStream acks was removed. Drain() is still important for core NATS to flush pending publishes — the change is fine, just verify no pending messages are lost on Stop().

### Suggestions

- Document the durability tradeoff in the PR description and/or code comments: what happens to in-flight work on daemon restart? Is there a reconciliation path (e.g., Tier 1 polling re-enqueues work)?
- Add an integration test covering handler-error behavior with core NATS, analogous to the removed NakOnError test, so the new retry contract is explicit.
- Consider whether workers need application-level deduplication now that JetStream's MsgID-based dedup is gone (particularly for publish/triage/implement where double-execution could be harmful).
- Verify filepath and any other imports in main.go are still used after removing DataDir.
- Replace time.Sleep(100ms) in integration tests with conn.Flush() or a readiness signal to reduce flakiness.
- Harden bridgeDiscovery with a retry loop (as the old implementation had) so a transient subscribe error doesn't permanently disable repo discovery.

---

Issue #352, opened 2026-04-24
bug: issues never update state after triage — not enrolled in watch_state


Muriano:
## Problem

Issues triaged or implemented by Heimdallm never update their `state` in the local store. They stay as `open` forever, even after being closed on GitHub.

Example: `freepik-company/fc-vertexai-api#141` was closed on GitHub at 2026-04-23T13:41:34Z but the local store still shows `state=open`.

## Root Cause

The watch_state table (which replaced the old WatchQueue) only gets entries for PRs — the `reviewHandler` in main.go calls `watchStore.Enroll(ctx, "pr", ...)` after review. But the `triageHandler` and `implementHandler` do NOT enroll issues.

Without an entry in watch_state, the state poller never publishes a StateCheckMsg for the issue, so the state worker never checks its GitHub state, and the store never updates.

## Fix

Add `watchStore.Enroll(ctx, "issue", ...)` to the triageHandler and implementHandler closures in main.go, mirroring what reviewHandler does for PRs.

PR #353, opened 2026-04-24
fix: enroll all processed items in watch_state (#352)

Muriano:
## Summary

All processed items (PRs and issues) now get enrolled in watch_state so the state poller detects closures/merges.

Previously only the NATS review worker enrolled PRs. Issues were never enrolled — their state stayed `open` forever in the UI.

**Added enrollment in:**
- triageHandler (issues after triage)
- implementHandler (issues after implement)
- ProcessPR (PRs from Tier 2 path)
- HandleChange (PRs from state check re-review)

**Closes:** #352

## Test plan

- [ ] After deploy: closed issues update in the Activity tab


vbuenog (changes requested):
## 🤖 Heimdallm AI Review

The PR migrates from embedded JetStream (persistent streams, consumers, KV) to core NATS pub/sub plus a SQLite-backed WatchStore, and ensures items are enrolled for state watching in the triage/implement/tier2 paths. The memory footprint win is real, but switching from JetStream to core NATS silently discards at-least-once delivery semantics (dedup, redelivery on NAK, server-side retry, durable replay after restart). A few enrollment call sites swallow errors or use the wrong context.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2228** — tier2Adapter.ProcessPR ignores the error returned by watchStore.Enroll and uses context.Background() instead of the passed ctx. Other enrollment call sites (review-worker, triage-worker, implement-worker) log a warning on failure and pass ctx — be consistent so failures are visible and enroll respects cancellation.
⚠️ **daemon/cmd/heimdallm/main.go:2540** — tier2Adapter.HandleChange has the same issue: error from watchStore.Enroll is dropped and context.Background() is used instead of ctx.
🟡 **daemon/cmd/heimdallm/main.go:2228** — The `if a.watchStore != nil` guard in ProcessPR/HandleChange suggests watchStore may be nil in some flows. main.go always constructs it, so either the guard is dead code (remove it) or there is an untested path (tests?) that passes nil — which will silently skip enrollment and mask bugs.
🔴 **daemon/internal/bus/bus.go:96** — Bus.MaxConcurrentWorkers() simply returns b.cfg.MaxConcurrentWorkers, but TestBus_MaxConcurrentWorkers_Clamp expects 0 to be clamped to 1. The clamp isn't visible in this diff — verify it exists in New() or the getter, otherwise the test will fail or the semaphore-size 0 will deadlock workers.
⚠️ **daemon/cmd/heimdallm/main.go:1712** — bridgeDiscovery now uses ChanSubscribe with a buffer of 8. Core NATS drops messages on slow consumers (no redelivery, no backpressure to publisher). If Tier-1 discovery publishes faster than Tier-2 drains reposChan, repo batches will be silently lost. The previous JetStream pull consumer provided redelivery; the new path does not. Consider logging the slow-consumer callback and sizing the buffer to the expected burst.
🟡 **daemon/internal/bus/bus.go:36** — Bus.Start now takes `_ context.Context`. If embedded server startup ever becomes cancellable or WatchStore init is re-added, this silently ignores cancellation. Minor, but worth preserving the parameter name for future use or documenting why it's unused.
⚠️ **daemon/internal/bus/integration_test.go:0** — Durability and NAK-on-error integration tests were removed. These are not just obsolete — they documented guarantees the system no longer provides (crash-recovery replay, retry on handler failure). Either add a test that asserts the new (weaker) behavior explicitly, or document the semantic change in the bus package comment so future readers don't assume at-least-once delivery.
🟡 **daemon/internal/bus/integration_test.go:35** — Several integration tests rely on `time.Sleep(100ms)` to let the subscription attach. Reducing from 200ms to 100ms makes an already-flaky pattern more flaky on loaded CI. Prefer waiting for a subscription-ready signal (e.g., conn.Flush after Subscribe) instead of a sleep.
🔴 **daemon/cmd/heimdallm/main.go:670** — Handler error semantics changed: with JetStream, a handler returning an error triggered NAK + redelivery. With core NATS the message is simply dropped. Verify that review/publish/triage/implement handlers no longer depend on retry-on-error behavior — any transient-failure path now needs explicit retry logic or will be lost silently.

### Suggestions

- Audit every worker handler for paths that previously relied on NAK-based redelivery (network flaps, GitHub 5xx, DB contention). Add explicit retry/backoff where needed, since core NATS will not redeliver.
- Replace context.Background() with the caller's ctx in the two tier2Adapter.Enroll calls, and log any returned error the way the other call sites do.
- Confirm (and test) the MaxConcurrentWorkers=0 → 1 clamp lives in bus.New or MaxConcurrentWorkers(); otherwise TestBus_MaxConcurrentWorkers_Clamp is asserting behavior that doesn't exist.
- Register a slow-consumer error handler on conn.SetErrorHandler (or at Subscribe time) so dropped discovery messages are logged instead of silently lost.
- Add a package-level comment in bus/bus.go stating explicitly that delivery is best-effort and non-durable, since the old code guaranteed the opposite and call sites may still assume that.
- Remove the `if a.watchStore != nil` guard in tier2Adapter if the field is always set in production — defensive nil checks hide wiring bugs.

---

Muriano:
Wrong branch — recreating from clean main-nats.

PR #354, opened 2026-04-24
fix: enroll all processed items in watch_state (#352)

Muriano:
## Summary

All processed items (PRs and issues) now get enrolled in watch_state so the state poller detects closures/merges.

Previously only the NATS review worker enrolled PRs. Issues were never enrolled — their state stayed `open` forever in the UI.

**Added enrollment in:**
- triageHandler (issues after triage)
- implementHandler (issues after implement)
- ProcessPR (PRs from Tier 2 path)
- HandleChange (PRs from state check re-review)

**Closes:** #352

## Test plan

- [ ] After deploy: closed issues update in the Activity tab

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -512,6 +512,7 @@ func main() {
 		login:                &cachedLogin,
 		runReview:            runReview,
 		publishPub:           publishPub,
+		watchStore:           watchStore,
 		lastSkippedUpdatedAt: make(map[int64]time.Time),
 	}
 
@@ -836,6 +837,12 @@ func main() {
 			slog.Error("triage-worker: pipeline run failed",
 				"repo", msg.Repo, "number", msg.Number, "err", err)
 		}
+
+		// Enroll for state watching so closed/resolved issues update in the UI.
+		if err := watchStore.Enroll(ctx, "issue", msg.Repo, msg.Number, msg.GithubID); err != nil {
+			slog.Warn("triage-worker: failed to enroll watch",
+				"repo", msg.Repo, "number", msg.Number, "err", err)
+		}
 	}
 
 	triageW := worker.NewTriageWorker(conn, maxWorkers, triageHandler)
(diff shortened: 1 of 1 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

The change correctly enrolls processed items (issues from triage/implement workers and PRs from tier2 adapter) into watch_state so UI can reflect state changes. Implementation is straightforward, but there are inconsistencies in error handling and context usage between the new enrollment sites.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2228** — Enroll error is silently discarded in ProcessPR, but the worker handlers above log the error via slog.Warn. Inconsistent handling — at minimum log the failure so silent enrollment drops are observable.
⚠️ **daemon/cmd/heimdallm/main.go:2539** — Same issue in HandleChange: Enroll return value is ignored. Should mirror the Warn-level logging used in the triage/implement handlers for consistency and diagnosability.
⚠️ **daemon/cmd/heimdallm/main.go:2228** — Uses context.Background() instead of the passed-in ctx for Enroll in ProcessPR/HandleChange. This detaches the DB call from request cancellation/timeouts and differs from the worker handlers which correctly pass ctx. If intentional (to ensure enrollment completes after ctx cancellation), add a comment; otherwise use ctx.
🟡 **daemon/cmd/heimdallm/main.go:840** — Enrollment runs unconditionally after pipeline execution, including when pipelineRun returned an error. Consider whether failed triage/implement runs should still enroll for watching — may be desired (to retry or observe state), but worth being explicit.

### Suggestions

- Log the Enroll error in ProcessPR and HandleChange to match the worker handlers' slog.Warn pattern.
- Use the provided ctx instead of context.Background() in ProcessPR/HandleChange enroll calls, unless there is a deliberate reason to decouple — if so, document it.
- Consider extracting the 'enroll after processing' logic into a small helper to avoid four near-identical call sites drifting over time.
- Add a brief test/assertion that confirms enrollment occurs for both issue and PR paths to lock in the fix.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: previous medium findings (missing error logging in ProcessPR/HandleChange, use of context.Background instead of ctx) have all been addressed — Enroll calls now log via slog.Warn and use the passed-in ctx consistent with the worker handlers. Only the low-severity design question about enrolling after failed pipeline runs remains unchanged.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:840** — Enrollment still runs unconditionally after pipelineRun errors in both triage and implement handlers (also line 924). Likely intentional so failed items still get state tracking, but a one-line comment explaining the choice would help future readers.

### Suggestions

- Add a brief comment above the unconditional Enroll call in the triage/implement handlers clarifying that enrollment is intentional even on pipeline failure (state watching is independent of pipeline success).

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: author addressed prior low-severity feedback by adding clear comments explaining that enrollment runs even after pipeline failure because state tracking is independent of pipeline success. The watchStore is properly plumbed into tier2Adapter with nil-guards on the PR paths. No new issues introduced.

### Suggestions

- Consider whether the triage/implement worker enrollment paths should also nil-guard watchStore for consistency with the tier2Adapter paths, though in main.go it's always constructed so the guard is not strictly needed.

---

PR #355, opened 2026-04-24
fix: startup scan enrolls open PRs/issues in watch_state

Muriano:
## Summary

On daemon startup, scans all PRs and issues with `state=open` in the store and enrolls them in `watch_state` using `INSERT OR IGNORE` — items already enrolled are not touched.

This backfills items processed before the NATS migration that were never enrolled in watch_state, so the state poller detects when they close on GitHub and updates the UI.

### Changes
- `WatchStore.EnrollIfAbsent()` — new method, `INSERT OR IGNORE` (doesn't reset backoff on existing items)
- `enrollOpenItems()` in main.go — startup scan, runs after watch store init

## Test plan

- [ ] On startup with stale open items: log shows `"enrolled open items into watch_state" count=N`
- [ ] After 30s (state poller tick): stale items get checked and updated to closed

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -2777,17 +2777,25 @@ func enrollOpenItems(s *store.Store, ws *bus.WatchStore) (int, error) {
 	if err != nil {
 		return 0, fmt.Errorf("query open prs: %w", err)
 	}
-	defer rows.Close()
 	for rows.Next() {
 		var ghID int64
 		var repo string
 		var number int
 		if err := rows.Scan(&ghID, &repo, &number); err != nil {
+			slog.Warn("startup-enroll: scan PR failed", "err", err)
 			continue
 		}
-		if added, err := ws.EnrollIfAbsent(ctx, "pr", repo, number, ghID); err == nil && added {
+		added, err := ws.EnrollIfAbsent(ctx, "pr", repo, number, ghID)
+		if err != nil {
+			slog.Warn("startup-enroll: enroll PR failed", "repo", repo, "number", number, "err", err)
+			continue
+		}
+		if added {
 			enrolled++
 		}
+	}
+	if err := rows.Err(); err != nil {
+		slog.Warn("startup-enroll: PR iteration error", "err", err)
 	}
 	rows.Close()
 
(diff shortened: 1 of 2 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

Adds a startup backfill that enrolls open PRs/issues into watch_state via a new EnrollIfAbsent method using INSERT OR IGNORE. Logic is straightforward and correct, but scan errors are silently swallowed and the duplicate defer on rows is unnecessary.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:2784** — rows.Scan errors are silently ignored with `continue`. A scan failure here could mask data corruption or schema drift; at minimum log the error so operators can investigate.
⚠️ **daemon/cmd/heimdallm/main.go:2789** — Errors from EnrollIfAbsent are discarded (`err == nil && added`). If enrollment fails for a real reason (DB error, constraint issue), there's no visibility. Log or aggregate these errors instead of dropping them.
🟡 **daemon/cmd/heimdallm/main.go:2792** — `defer rows.Close()` is followed by an explicit `rows.Close()` call. The explicit close is fine (to release the connection before the next query) but the defer becomes a no-op second close. Minor, but consider dropping the defer or the explicit call for clarity.
🟡 **daemon/cmd/heimdallm/main.go:2785** — `rows.Err()` is never checked after the loop for either query; iteration errors (e.g., driver-level failures mid-scan) will be silently lost.
🟡 **daemon/cmd/heimdallm/main.go:2771** — Uses `context.Background()` instead of a shutdown-aware context. If the server is signaled during startup backfill on a large dataset, enrollment can't be cancelled cleanly.
🟡 **daemon/internal/bus/watch.go:87** — EnrollIfAbsent duplicates most of Enroll's logic. Consider having Enroll delegate to EnrollIfAbsent (or share a helper) to keep INSERT semantics consistent if defaults change later.

### Suggestions

- Log scan and EnrollIfAbsent errors instead of silently continuing, so operators can detect backfill problems.
- Check rows.Err() after each loop to surface iteration-level failures.
- Remove the redundant defer rows.Close() now that rows.Close() is called explicitly before the next query.
- Consider refactoring Enroll to share the insert logic with EnrollIfAbsent to avoid drift between the two code paths.
- Thread the main shutdown context through enrollOpenItems rather than using context.Background().

---

PR #356, opened 2026-04-24
fix: replace blocking startup enroll with gradual per-tick backfill

Muriano:
## Summary

The startup `enrollOpenItems` scan from #355 blocked the daemon with 368 open PRs, preventing it from starting.

**Fix:** Replace with `enrollOneOpenItem` called once per state-poller tick (every 30s). Enrolls 1 item per tick until all open items are in watch_state. With 370 items, full backfill completes in ~3 hours — acceptable since the pollers re-detect active items much sooner.

## Test plan

- [ ] Daemon starts instantly (no blocking)
- [ ] `watch_state` gradually populates over time

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -156,15 +156,6 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Enroll any open PRs/issues not yet in watch_state so the state
-	// poller picks them up. Covers items processed before the NATS
-	// migration that were never enrolled.
-	if enrolled, err := enrollOpenItems(s, watchStore); err != nil {
-		slog.Warn("startup: enroll open items failed", "err", err)
-	} else if enrolled > 0 {
-		slog.Info("startup: enrolled open items into watch_state", "count", enrolled)
-	}
-
 	broker := sse.NewBroker()
 	broker.Start()
 
@@ -965,6 +956,10 @@ func main() {
 			case <-statePollerCtx.Done():
 				return
 			case <-ticker.C:
+				// Gradually enroll one open item not yet in watch_state per tick.
+				// Backfills items from before the NATS migration without blocking startup.
+				enrollOneOpenItem(s, watchStore)
+
 				if evicted, err := watchStore.EvictStale(statePollerCtx); err != nil {
 					slog.Warn("state-poller: evict failed", "err", err)
 				} else if evicted > 0 {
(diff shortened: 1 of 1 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

The refactor correctly removes the blocking startup enrollment in favor of gradual per-tick backfill, but it regresses observability (errors now silently swallowed) and is inefficient: it performs full table scans of open PRs and issues on every 30s tick forever, even after backfill is complete.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2773** — Query errors are silently swallowed with `continue` — the previous implementation logged them via slog.Warn. If the DB query fails persistently the poller will appear healthy while backfill silently never progresses. At minimum log at Warn/Debug.
🟡 **daemon/cmd/heimdallm/main.go:2782** — Scan errors are silently swallowed. A malformed row will be skipped without any signal, making debugging difficult. Previous code logged scan failures.
⚠️ **daemon/cmd/heimdallm/main.go:2788** — EnrollIfAbsent errors cause an immediate `return` without logging, unlike the previous version which logged and continued. One transient error aborts this tick's backfill attempt and is invisible to operators.
⚠️ **daemon/cmd/heimdallm/main.go:2763** — Every 30s tick re-queries all open PRs and all open issues in full, iterating until it finds an unenrolled one. Once backfill completes, this is a permanent recurring full-table scan with no termination condition or early exit — consider a single SQL query with LEFT JOIN watch_state ... WHERE ws.id IS NULL LIMIT 1, or a one-shot completion flag.
🟡 **daemon/cmd/heimdallm/main.go:2763** — Backfill rate is 1 item per 30s (~120/hour). For repos with thousands of open items this takes many hours to converge, during which those items are not being state-polled. Consider enrolling N per tick or draining until first EnrollIfAbsent that reports already-present.
🟡 **daemon/cmd/heimdallm/main.go:2795** — rows.Err() is no longer checked after iteration; truncated result sets due to driver errors will be indistinguishable from a clean scan.
🟡 **daemon/cmd/heimdallm/main.go:2764** — context.Background() is used instead of the statePollerCtx available in the caller — backfill enrollments cannot be cancelled on shutdown and may outlive the poller. Pass ctx in as a parameter.
🟡 **daemon/cmd/heimdallm/main.go:959** — enrollOneOpenItem runs synchronously on the state-poller tick before EvictStale/other work. If the DB is slow, backfill scanning can delay the poller. Minor, but worth noting.

### Suggestions

- Replace the two full scans with a single indexed query that LEFT JOINs watch_state and returns only unenrolled items (LIMIT N), so completed backfill has O(1) cost.
- Restore Warn-level logging for query/scan/enroll errors; silent failure modes were a real regression from the prior implementation.
- Accept a context.Context parameter (pass statePollerCtx) so backfill respects shutdown.
- Emit an Info log once when the first tick finds nothing to enroll, so operators can confirm backfill is complete.
- Use defer rows.Close() inside the loop body scope (e.g. an inner func) instead of repeating rows.Close() before every return path.

---

PR #357, opened 2026-04-24
fix: auto-dismiss legacy items with empty repo from watch_state

Muriano:
## Summary

Legacy PRs with empty `repo` field were being enrolled into watch_state by the backfill process. The state worker tried to check them via GitHub API, got 404, and logged warnings every 30s indefinitely.

**Fixes:**
- `enrollOpenItems` query: `AND p.repo != ''` — skip items without repo
- State handler: auto-delete items with empty repo from watch_state
- State handler: remove items that return 404 from GitHub (deleted repos, revoked access)
- Fixed `enrollOneOpenItem` → `enrollOpenItems` function name mismatch

## Test plan

- [ ] No more 404 spam in logs for legacy items

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -988,7 +988,9 @@ func main() {
 		// Auto-dismiss legacy items with missing data — they can never be checked.
 		if msg.Repo == "" {
 			key := fmt.Sprintf("%s.%d", msg.Type, msg.GithubID)
-			watchStore.Delete(ctx, key)
+			if err := watchStore.Delete(ctx, key); err != nil {
+				slog.Warn("state-handler: failed to delete legacy item", "key", key, "err", err)
+			}
 			slog.Info("state-handler: auto-dismissed legacy item with empty repo",
 				"type", msg.Type, "number", msg.Number, "github_id", msg.GithubID)
 			return false, nil
@@ -1022,7 +1024,9 @@ func main() {
 			// 404 means the repo/PR was deleted or we don't have access.
 			// Remove from watch to stop retrying.
 			if strings.Contains(err.Error(), "status 404") {
-				watchStore.Delete(ctx, key)
+				if delErr := watchStore.Delete(ctx, key); delErr != nil {
+					slog.Warn("state-handler: failed to delete unreachable item", "key", key, "err", delErr)
+				}
 				slog.Info("state-handler: removed unreachable item from watch",
 					"type", msg.Type, "repo", msg.Repo, "number", msg.Number)
 				return false, nil
@@ -2806,9 +2810,13 @@ func enrollOpenItems(s *store.Store, ws *bus.WatchStore) {
 		for rows.Next() {
 			var it item
 			if err := rows.Scan(&it.ghID, &it.repo, &it.number); err != nil {
+				slog.Warn("state-poller: backfill scan failed", "type", q.typ, "err", err)
 				continue
 			}
 			batch = append(batch, it)
+		}
+		if err := rows.Err(); err != nil {
+			slog.Warn("state-poller: backfill iteration error", "type", q.typ, "err", err)
 		}
 		rows.Close() // close before writing to avoid SQLite lock contention
 
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -958,7 +958,7 @@ func main() {
 			case <-ticker.C:
 				// Gradually enroll one open item not yet in watch_state per tick.
 				// Backfills items from before the NATS migration without blocking startup.
-				enrollOneOpenItem(s, watchStore)
+				enrollOpenItems(s, watchStore)
 
 				if evicted, err := watchStore.EvictStale(statePollerCtx); err != nil {
 					slog.Warn("state-poller: evict failed", "err", err)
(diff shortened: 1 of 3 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

The PR adds sensible auto-dismissal of legacy items with empty repo, handles 404 errors to stop retrying unreachable items, and batches backfill enrollment up to 10 items per tick. The changes are well-targeted, but there are a few concerns around error handling robustness, context usage, and a missing rows.Err() check.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:992** — watchStore.Delete error is ignored in the auto-dismiss path. If Delete fails, the item will be re-enqueued on the next tick and the 'auto-dismissed' log message will be misleading. Consider logging or returning the error.
⚠️ **daemon/cmd/heimdallm/main.go:1023** — Using strings.Contains(err.Error(), "status 404") is fragile — depends on exact error message format. If the GitHub client's error formatting changes, this silently stops working and items keep retrying forever. Prefer a typed error / errors.As with a sentinel or HTTPError type exposing StatusCode.
🟡 **daemon/cmd/heimdallm/main.go:1025** — Delete error ignored here too; same issue as the legacy-item path.
🟡 **daemon/cmd/heimdallm/main.go:2791** — context.Background() is used for enrollment, bypassing statePollerCtx. If the poller is cancelled during shutdown, Enroll calls will continue. Pass the poller's context through to enrollOpenItems.
🟡 **daemon/cmd/heimdallm/main.go:2808** — No rows.Err() check after the rows.Next() loop — iteration errors (e.g. connection drop mid-scan) are silently swallowed, potentially causing partial batches without any warning.
🟡 **daemon/cmd/heimdallm/main.go:2810** — Individual Scan errors are silently 'continue'd without logging. A malformed row will be invisibly skipped every tick.
🟡 **daemon/cmd/heimdallm/main.go:2819** — Inconsistent failure handling: a single Enroll failure aborts the remaining batch with 'return' (skipping the other type too). Consider continuing within the batch but breaking the outer type loop, or documenting the intent — the current behavior means one transient failure blocks issue backfill when it occurs during PR backfill.
🟡 **daemon/cmd/heimdallm/main.go:989** — Existing items already enrolled with empty repo will hit the auto-dismiss path every time they're dispatched. Consider also filtering empty-repo rows when dispatching state checks, so they're cleaned up once rather than each tick. Minor — self-healing works either way.

### Suggestions

- Replace the 404 string match with a typed error check (e.g. a custom HTTPError or errors.Is/As) so the cleanup path is robust to error-message changes.
- Thread statePollerCtx through to enrollOpenItems instead of context.Background(), so enrollment respects shutdown.
- Check rows.Err() after the Next loop, and log Scan failures instead of silently continuing.
- Log (or at least count) Delete errors in both auto-dismiss branches so silent KV failures are observable.
- Consider a one-shot cleanup query for existing watch_state rows with empty repo, rather than relying solely on dispatch-time dismissal.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: Author addressed 5 of 8 previous findings (Delete error logging in both paths, rows.Err() check, Scan error logging, empty-repo query filter). Three issues remain: 404 detection via string matching, context.Background() usage, and abort-on-single-failure behavior.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:1025** — Still using strings.Contains(err.Error(), "status 404") to detect 404s. Fragile — depends on exact error message format from the GitHub client. If formatting changes (e.g. to 'HTTP 404' or wrapping), detection silently fails and items retry forever. Prefer errors.As with a typed HTTP error exposing StatusCode.
🟡 **daemon/cmd/heimdallm/main.go:2791** — enrollOpenItems still uses context.Background() instead of accepting/propagating statePollerCtx. Enroll calls will continue during shutdown cancellation. Pass the poller's context through.
🟡 **daemon/cmd/heimdallm/main.go:2825** — A single Enroll failure still uses 'return' which aborts the outer type loop too — one transient failure during PR backfill blocks issue backfill this tick. Consider 'break' to continue with the next type, or document intent.

### Suggestions

- Introduce a typed HTTP error (e.g. github.HTTPError with StatusCode field) in the GitHub client and switch the 404 check to errors.As for robustness.
- Thread statePollerCtx into enrollOpenItems so enrollment respects shutdown cancellation.
- Replace the final 'return' in the Enroll failure path with 'break' so the issue type is still attempted when a PR enroll fails transiently.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Most prior feedback is addressed — delete errors are now logged, rows.Err() and per-row Scan errors are surfaced, and the enrollment queries now filter empty-repo rows to prevent new legacy enrollments. Two non-trivial items remain: the fragile strings.Contains("status 404") check and using context.Background() instead of the poller's context. A minor loop-control concern also persists.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:1024** — Still using strings.Contains(err.Error(), "status 404") to detect 404s. This is fragile — any change to the GitHub client's error formatting silently disables the cleanup and items keep retrying forever. Prefer errors.As against a typed HTTPError/APIError exposing StatusCode.
🟡 **daemon/cmd/heimdallm/main.go:2791** — enrollOpenItems still uses context.Background() instead of accepting the poller's context. During shutdown, in-flight Enroll calls won't be cancelled. Pass statePollerCtx through as a parameter.
🟡 **daemon/cmd/heimdallm/main.go:2822** — On the first Enroll failure the function returns, skipping not only the rest of the current batch but also the other type (e.g. a transient PR-enroll failure prevents issue backfill this tick). Consider `break` to skip only the current type, or `continue` within the batch to tolerate a single transient failure.
🟡 **daemon/cmd/heimdallm/main.go:989** — The backfill queries now exclude empty-repo rows, but items already enrolled with empty repo still hit the auto-dismiss branch once per dispatch. Self-healing, but if the number is nontrivial, a one-shot cleanup query at startup (DELETE FROM watch_state WHERE repo = '') would cut log noise. Minor.

### Suggestions

- Replace strings.Contains(err.Error(), "status 404") with errors.As against the GitHub client's typed error exposing StatusCode, so the cleanup doesn't silently break on a message-format change.
- Thread statePollerCtx into enrollOpenItems so Enroll calls honor shutdown cancellation.
- Change `return` inside the Enroll-failure branch to `break` (or per-item `continue`) so one transient failure on PRs doesn't block issues for the same tick.

---

PR #358, opened 2026-04-24
fix(flutter): include agent configs in PATCH /config diff

### LINKED ISSUE #285: bug: --no-session-persistence toggle reverts to enabled on its own
## Problem

The \`--no-session-persistence\` flag on the Claude agent profile keeps re-enabling itself after the user explicitly disables it. The user turns it off, and when they check again later, it's back on.

This is particularly harmful because \`claude-standup\` (Claude's new activity recapitulation tool) cannot read session history when \`--no-session-persistence\` is active. Heimdallm's reviews become invisible to standup reports.

## Screenshot

The agent config shows \`--no-session-persistence\` as OFF (grey toggle), but it reverts to ON without user action.

## Suspected Cause

Likely the same class of bug as #284 (store overrides TOML config). The agent profile in SQLite may have \`no_session_persistence = true\` persisted from a previous save, and any config reload or agent upsert cycle resets the toggle to the stored value, discarding the user's change.

Alternatively, the Flutter UI save for agent profiles may not be correctly sending the \`false\` value for boolean toggles — similar to the \`null\` erasure bug fixed in #144 where absent fields were interpreted as defaults rather than explicit values.

## Expected Behavior

When the user disables \`--no-session-persistence\`, it stays disabled permanently until the user explicitly re-enables it.

## Impact

- Claude sessions from Heimdallm reviews are not persisted
- \`claude-standup\` cannot recapitulate Heimdallm's activity
- The user has no visibility into what Claude did during reviews
- Repeated manual intervention required with no lasting effect

Muriano:
## Summary

- `_computeGlobalDiff()` was missing `agentConfigs` entirely — changes to per-agent CLI settings (`no_session_persistence`, `model`, `effort`, `permission_mode`, etc.) were never included in the PATCH payload sent to the daemon
- The TOML file kept stale values and the UI reverted on next config reload, causing the `--no-session-persistence` toggle to re-enable itself

Closes #285

## Test plan
- [x] `flutter test` — 121/121 passed
- [x] `go test ./...` — all packages pass
- [ ] Manual: toggle `--no-session-persistence` off → verify TOML updates → reload config → confirm toggle stays off


vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Adds per-agent diff computation for the PATCH /config payload so agent configuration changes are actually transmitted. The logic correctly unions old/new agent keys, falls back to default CLIAgentConfig for missing entries, and emits only changed fields. Implementation is straightforward and aligns with the surrounding diff pattern, though list equality and null-coercion deserve attention.

### Issues

🔴 **flutter_app/lib/features/config/config_providers.dart:141** — `o.extraFlags != n.extraFlags` compares List<String> by reference identity in Dart, not element-wise. Two lists with identical contents will be reported as different, producing spurious diffs on every PATCH. Use a deep equality check (e.g. ListEquality().equals or listEquals from package:flutter/foundation.dart).
⚠️ **flutter_app/lib/features/config/config_providers.dart:142** — `ad['prompt'] = n.promptId ?? ''` coerces null to empty string, which loses the distinction between 'unset' and 'explicitly empty'. If the server treats '' as a valid prompt id or differently from null, this could overwrite a cleared prompt incorrectly. Confirm server semantics; otherwise send `n.promptId` directly.
⚠️ **flutter_app/lib/features/config/config_providers.dart:129** — When an agent exists only in `old` (i.e., removed in `updated`), substituting a default `CLIAgentConfig()` will diff old values against defaults rather than signaling deletion. If agent removal is a supported operation, the server won't see it as removed.

### Suggestions

- Replace the List reference comparison with `listEquals(o.extraFlags, n.extraFlags)` from package:flutter/foundation.dart to avoid false-positive diffs.
- Consider extracting the per-agent field comparison into a helper (e.g. `_diffAgent(old, new)`) for readability and to keep this in sync with any future fields added to CLIAgentConfig.
- Add a unit test covering: no-op diff (identical configs), single-field change, added agent, and removed agent — to lock in the expected PATCH payload shape.
- Clarify whether `promptId == null` should be sent as `null` vs `''`; align with the backend contract.

---

sergiotejon:
## ⚠️ MEDIUM Issue

promptId null is coerced to empty string ('') in the diff payload. If the backend distinguishes null (unset) from '' (explicitly cleared), this may produce incorrect server-side state. Confirm backend semantics or send null explicitly.

**Location:** `flutter_app/lib/features/config/config_providers.dart` line 144

---

sergiotejon:
## ⚠️ MEDIUM Issue

extraFlags comparison uses != which is reference equality for List in Dart. Two lists with identical contents will be treated as different, producing spurious diffs on every save. Use listEquals or DeepCollectionEquality.

**Location:** `flutter_app/lib/features/config/config_providers.dart` line 132

---

sergiotejon (approved):

New diff scope is limited to agent config diffing in config_providers.dart. Logic is correct but has minor semantic concerns around null prompt coercion and reference-equality on extraFlags list.

### Suggestions

- Replace `o.extraFlags != n.extraFlags` with `!listEquals(o.extraFlags, n.extraFlags)` from package:flutter/foundation.dart.
- Clarify backend contract for null vs empty prompt and align serialization accordingly.
- Consider explicitly representing agent additions/removals rather than synthesizing defaults.

---

sergiotejon:
## 🟡 LOW Issue

Using `const CLIAgentConfig()` as a default for added/removed agents means deletions are diffed as field-by-field resets to defaults rather than an explicit removal. If the backend supports agent removal, this won't communicate intent; if not, added agents with default-equal fields produce an empty diff and are silently skipped.

**Location:** `flutter_app/lib/features/config/config_providers.dart` line 130

---

Muriano:
## Review Response

### 1. `extraFlags` reference equality — No aplica

`extraFlags` es `String`, no `List<String>` (ver `config_model.dart:7`). El operador `!=` en Dart compara Strings por valor, no por referencia. No hay bug aquí.

### 2. `promptId` null → `''` coercion — No aplica

Enviar `null` en el PATCH payload causaría HTTP 400: el handler `handlePatchConfig` (`handlers.go:554`) llama a `config.ContainsNull(patch)` y rechaza explícitamente cualquier valor null con el mensaje *"null values not allowed in PATCH — use DELETE to remove fields"*.

Además, el Go struct `CLIAgentConfig` usa `string` (no `*string`), por lo que `""` y "ausente" son equivalentes (zero value). El `?? ''` es la forma correcta de representar "unset" en este contrato.

### 3. Agent removal con default — Válido, fuera de scope

Los agent names (`claude`, `gemini`, `codex`) son fijos — no existe operación de "eliminar agente" ni en la UI ni en el backend. Si un agent name aparece solo en `old` y no en `updated`, el diff contra defaults produce un map vacío que se descarta (`if (ad.isNotEmpty)`), que es exactamente lo correcto. No se necesita señalizar eliminación porque no es una operación soportada.

vbuenog (approved):
## 🤖 Heimdallm AI Review

The diff adds agent config diffing to PATCH /config payload. Implementation is correct given the backend contract (no nulls allowed, fixed agent names, extraFlags is String). Author has already addressed prior review comments with valid technical justifications.

### Suggestions

- Consider extracting the per-field comparison into a method on CLIAgentConfig (e.g., diffTo(other)) to keep config_providers.dart focused on orchestration and make the field list easier to maintain as new agent settings are added.
- If agent removal ever becomes a supported operation in the future, this logic will need revisiting since unknown agents currently diff against defaults and may be silently dropped.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: author's responses to prior findings are technically sound. extraFlags is a String (value-compared correctly by !=), the backend rejects null in PATCH payloads making '' the correct sentinel for unset promptId, and agent names are a fixed set so default-substitution producing an empty diff is correct behavior. No new changes since last review and no remaining issues.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Targeted fix that adds agent configs to the global config diff. Logic is straightforward and the author's responses to prior bot review comments are technically sound (extraFlags is a String per the model, null is rejected by the backend PATCH handler, and agent names are fixed so removal-via-default is a non-issue). The main residual concern is maintainability: the per-field comparison block duplicates the field list of CLIAgentConfig and will silently go stale when fields are added.

### Issues

⚠️ **flutter_app/lib/features/config/config_providers.dart:129** — Per-field diff list duplicates the CLIAgentConfig schema. Any new field added to CLIAgentConfig will silently be omitted from the PATCH diff (the existing 'agent configs missing from diff' bug this PR fixes is exactly this failure mode at the parent-object level). Consider implementing a `toPatchMap`/`diff` method on CLIAgentConfig itself, or at minimum adding a code comment on the model class warning that this loop must be updated in lockstep.
🟡 **flutter_app/lib/features/config/config_providers.dart:144** — The `promptId ?? ''` coercion encodes a non-obvious backend contract (null is rejected by handlePatchConfig; '' is the canonical 'unset' marker since the Go struct uses string, not *string). This is correct per the author's response, but the rationale is invisible at the call site. A one-line comment referencing the backend's null-rejection rule would prevent a future contributor from 'fixing' this to send null.
🟡 **flutter_app/lib/features/config/config_providers.dart:130** — Falling back to `const CLIAgentConfig()` for missing keys is safe today only because the agent set is fixed (claude/gemini/codex). If a fourth agent is ever added on one side of the comparison without the other, the diff will silently emit field-by-field default values instead of signaling addition. Worth a brief comment near `allAgentNames` documenting the closed-set assumption.
⚠️ **flutter_app/lib/features/config/config_providers.dart:0** — No test coverage is added for this diff path despite the bug being a regression in serialization behavior. A unit test asserting that mutating each CLIAgentConfig field produces the expected PATCH payload would lock in the contract and catch future field additions that forget to update this loop.

### Suggestions

- Extract the per-agent diff into `CLIAgentConfig.diffTo(other)` (or a top-level helper) so the field list lives next to the model definition and is easier to keep in sync.
- Add a unit test for `_computeGlobalDiff` covering: (a) agent field changed → appears under ai.agents.<name>, (b) no agent changes → no `agents` key emitted, (c) promptId cleared → empty string sent.
- Add a short comment above `if (o.promptId != n.promptId) ad['prompt'] = n.promptId ?? '';` explaining the backend contract (null rejected, '' = unset) so the coercion isn't 'cleaned up' later.
- Consider whether the JSON key should be `prompt` or `prompt_id` — the model field is `promptId` but the wire key here is `prompt`. Verify this matches the Go struct's JSON tag to avoid a silent no-op on the server.

---

PR #360, opened 2026-04-25
fix(daemon): use GitHub ID for PR in-flight claim on cold-start (closes #359)

### LINKED ISSUE #359: fix(pipeline): duplicate PR reviews on cold-start — in-flight lock not persisted
## Problem

After a container restart, **all open PRs that have never been reviewed** are picked up simultaneously by Tier 2. If any review takes longer than `poll_interval` (currently 1m), the next poll cycle picks up the same PR again because:

1. The **in-flight claim is memory-only** — lost on restart
2. The review has not been stored in the DB yet (still running)
3. `PRAlreadyReviewed()` finds no prior review in the store → returns false
4. SHA dedup inside `pipeline.Run()` also finds no prior review → proceeds

**Real-world impact (2026-04-25):** `overmind-swarm/mobile#262` received 2 duplicate APPROVED reviews within 21 seconds of each other (review IDs `4175387269` and `4175387499`). 7 total reviews were published in the first 2 minutes after a cold start, when only 5 were needed.

## Root cause

The in-flight claim (`runReview: in-flight claim skipped (defenses still apply)`) is stored in memory only. On cold start the claim map is empty, so concurrent reviews for the same PR are not blocked.

### 1. Persist in-flight claims in the SQLite store (primary fix)

Write a row to `heimdallm.db` when a review starts (keyed on `pr_id + head_sha`), delete it on completion or failure. On startup, check the store before starting any review. This survives restarts.

### 2. Mutex/lock per PR during review execution

Before calling the AI CLI, acquire a per-PR lock (in-memory is fine as a secondary defense). If another goroutine is already reviewing the same PR, skip. This prevents the intra-process race where Tier 2 cycle N+1 starts while cycle N's review is still running.

### 3. Pre-review "claimed" marker in the store

Similar to #1 but lighter: write a `claimed_at` timestamp to the PR record before starting the review. `PRAlreadyReviewed()` checks this timestamp and skips if `claimed_at` is recent (e.g., within `execution_timeout`). On completion, `claimed_at` is cleared and replaced by the actual review record.

## Recommendation

Implement **#1 + #2**: persistent claim for cold-start resilience, in-memory mutex for intra-process races. #3 is an alternative to #1 if the schema change is too invasive.

## Workaround

Increase `poll_interval` to a value higher than the longest expected review time (e.g., 3-5m). This reduces the probability of overlap but does not eliminate it for very slow reviews.

sergiotejon:
## Summary
- On cold-start, `stored` PR is nil (not yet upserted), so the in-flight claim was skipped — allowing duplicate reviews on the same SHA
- Fall back to `pr.ID` (GitHub-assigned ID) when `stored` is nil, removing the `stored != nil` precondition from the claim guard
- Reuses the existing `reviews_in_flight` table and `ClaimInFlightReview` method — same pattern as the issue pipeline (#292)

## Impact
Prevents duplicate PR reviews after container restarts. Real-world case: `mobile#262` received 2 APPROVED reviews within 21 seconds on 2026-04-25.

Closes #359

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -324,12 +324,26 @@ func main() {
 			slog.Warn("runReview: GetPRByGithubID failed, proceeding without persistent claim",
 				"pr_id", pr.ID, "repo", pr.Repo, "err", err)
 		}
-		// stored may be nil either way — downstream Claim guard handles that.
-		var claimed bool
+		// Resolve the claim key: prefer the internal store ID when available,
+		// fall back to the GitHub-assigned ID on cold-start (PR not yet
+		// upserted). Both are stable int64 identifiers; the composite PK
+		// (pr_id, head_sha) stays unique either way because GitHub IDs are
+		// globally unique and internal IDs are repo-scoped sequences that
+		// never collide with GitHub's range.
+		//
+		// This fixes the cold-start duplicate-review bug where stored==nil
+		// caused the claim to be skipped entirely (#359).
 		var claimPRID int64
+		if stored != nil {
+			claimPRID = stored.ID
+		} else {
+			claimPRID = pr.ID // GitHub-assigned ID — always available
+		}
+
+		var claimed bool
 		var claimSHA string
-		if stored != nil && pr.Head.SHA != "" {
-			ok, err := s.ClaimInFlightReview(stored.ID, pr.Head.SHA)
+		if pr.Head.SHA != "" {
+			ok, err := s.ClaimInFlightReview(claimPRID, pr.Head.SHA)
 			if err != nil {
 				slog.Warn("runReview: claim inflight failed, proceeding", "err", err)
 			} else if !ok {
@@ -338,35 +352,11 @@ func main() {
 				return
 			} else {
 				claimed = true
-				claimPRID = stored.ID
 				claimSHA = pr.Head.SHA
 			}
 		} else {
-			// Defensive log: the claim guard was short-circuited. Surfaces
-			// the wiring regression theburrowhub/heimdallm#264 (empty SHA)
-			// and the early-stage "PR not yet upserted" path; downstream
-			// defenses (fail-closed SHA in pipeline.Run, circuit breaker,
-			// PublishedAt grace) still cap cost in both cases.
-			//
-			// The reason string is computed from the actual predicates
-			// rather than an else-branch nil check, so a future edit to
-			// the outer guard doesn't silently mislead operators — the
-			// log stays truthful no matter what combination of conditions
-			// steered us into this branch.
-			var reason string
-			switch {
-			case stored == nil:
-				reason = "stored PR not found"
-			case pr.Head.SHA == "":
-				reason = "empty Head.SHA from caller"
-			default:
-				// Unreachable under today's guard (stored != nil && SHA != "")
-				// but kept so a future added clause still yields a
-				// non-misleading message.
-				reason = "claim precondition failed"
-			}
 			slog.Info("runReview: in-flight claim skipped (defenses still apply)",
-				"pr", pr.Number, "repo", pr.Repo, "reason", reason)
+				"pr", pr.Number, "repo", pr.Repo, "reason", "empty Head.SHA from caller")
 		}
 		defer func() {
 			if claimed {
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"crypto/rand"
-	"database/sql"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -298,10 +297,9 @@ func main() {
 		// gated by a stale in-flight row from a prior HEAD. See
 		// theburrowhub/heimdallm#243.
 		//
-		// For early-stage PRs that have not yet been upserted, OR for PRs
-		// where the HEAD SHA is not yet known, skip the claim — the
-		// downstream SHA dedup in pipeline.Run (already fail-closed per
-		// Task 1) handles those paths.
+		// For PRs where the HEAD SHA is not yet known, skip the claim —
+		// the downstream SHA dedup in pipeline.Run (already fail-closed per
+		// Task 1) handles that path.
 		//
 		// On Claim error (transient SQLite blip, disk pressure), we log and
 		// proceed fail-open. This is safe because the downstream defenses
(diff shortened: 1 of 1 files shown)


Muriano (changes requested):
## 🤖 Heimdallm AI Review

The fix correctly addresses the cold-start path where stored==nil caused the claim to be skipped entirely (#359). However, it introduces a new dedup gap: the claim key uses GitHub's global ID on cold-start but the internal SQLite auto-increment ID on warm path. Since these are different values for the same PR, a race between a cold-start claim and a post-upsert retry can produce two successful claims for the same (PR, SHA) pair, allowing a duplicate review. The downstream defenses (circuit breaker, SHA guard, PublishedAt grace) mitigate the blast radius, but the claim mechanism itself is broken for that window. The inline comment asserting 'never collide' addresses cross-PR collisions but misses this same-PR dual-key problem.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:335** — Claim key mixes two ID namespaces for the same PR — defeats dedup on cold-to-warm race. When stored==nil (cold-start), claimPRID = pr.ID (GitHub global ID, e.g. 1987654321). When stored!=nil (warm), claimPRID = stored.ID (SQLite auto-increment, e.g. 42). If event 1 claims with (1987654321, sha) and then event 2 arrives after the PR is upserted, it claims with (42, sha) — both succeed because the composite keys are different. The comment at lines 327-336 argues ranges 'never collide' but that addresses cross-PR uniqueness, not same-PR identity. Fix: always use pr.ID (GitHub ID) as the claim key in both paths, since it is always available and stable. Alternatively, use stored.GithubID when stored != nil (same value as pr.ID).

---

Muriano:
## 🔴 HIGH Issue

Claim key mixes two ID namespaces for the same PR — defeats dedup on cold-to-warm race. When stored==nil (cold-start), claimPRID = pr.ID (GitHub global ID, e.g. 1987654321). When stored!=nil (warm), claimPRID = stored.ID (SQLite auto-increment, e.g. 42). If event 1 claims with (1987654321, sha) and then event 2 arrives after the PR is upserted, it claims with (42, sha) — both succeed because the composite keys are different. The comment at lines 327-336 argues ranges 'never collide' but that addresses cross-PR uniqueness, not same-PR identity. Fix: always use pr.ID (GitHub ID) as the claim key in both paths, since it is always available and stable. Alternatively, use stored.GithubID when stored != nil (same value as pr.ID).

**Location:** `daemon/cmd/heimdallm/main.go` line 335

---

Muriano (changes requested):

The fix correctly addresses the cold-start path where stored==nil caused the claim to be skipped entirely (#359). However, it introduces a new dedup gap: the claim key uses GitHub's global ID on cold-start but the internal SQLite auto-increment ID on warm path. Since these are different values for the same PR, a race between a cold-start claim and a post-upsert retry can produce two successful claims for the same (PR, SHA) pair, allowing a duplicate review. The downstream defenses (circuit breaker, SHA guard, PublishedAt grace) mitigate the blast radius, but the claim mechanism itself is broken for that window. The inline comment asserting 'never collide' addresses cross-PR collisions but misses this same-PR dual-key problem.

### Suggestions

- Simplify by always using pr.ID (GitHub ID) as the claim key — it is always available, globally unique, and identical across cold-start and warm paths. This eliminates the dual-namespace problem entirely and removes the if/else branching for claimPRID.
- Add a test that simulates the cold-to-warm race: claim with stored==nil, then upsert the PR, then attempt a second claim for the same GitHub PR+SHA with stored!=nil. The second claim should fail (currently it would succeed).

---

Muriano:
## ⚠️ MEDIUM Issue

Mixed ID namespace in reviews_in_flight.pr_id: The claim table's pr_id column has always stored internal DB row IDs (autoincrement, starting at 1). This change introduces GitHub-assigned IDs into the same column on cold-start (stored==nil). Two concrete failure modes: (1) If two events for the same PR race and one resolves stored!=nil (warm) while the other doesn't (cold), they produce different claim keys (internal ID vs GitHub ID) for the same (PR, SHA), defeating deduplication. (2) The trigger-review path (SetTriggerReviewFn, ~line 1347) always uses pr.ID (internal DB ID); a concurrent trigger + cold-start poll for the same PR would also use mismatched keys. Fix: either upsert the PR before claiming (so stored is always available) or add a dedicated github_pr_id column to reviews_in_flight and use a composite key that includes the ID type.

**Location:** `daemon/cmd/heimdallm/main.go` line 340

---

Muriano:
## 🟡 LOW Issue

Comment claims 'GitHub IDs are globally unique and internal IDs are repo-scoped sequences that never collide with GitHub's range' — this is not guaranteed. Internal IDs are SQLite AUTOINCREMENT (1, 2, 3…) and GitHub PR node IDs can also be small integers (especially PR.ID which is the REST API numeric ID, not the GraphQL node_id). A PR with GitHub ID 5 could theoretically collide with internal row ID 5 belonging to a different PR, causing a false claim match that silently drops a legitimate review. The collision is unlikely in practice but the comment presents it as a design invariant when it is not.

**Location:** `daemon/cmd/heimdallm/main.go` line 335

---

Muriano (approved):

The PR fixes a real cold-start bug (#359) where stored==nil caused the in-flight claim to be skipped entirely. The approach of falling back to the GitHub ID is pragmatically better than no claim at all, but it mixes two distinct ID spaces (internal DB autoincrement vs GitHub global IDs) in the same reviews_in_flight.pr_id column, which can cause deduplication failures in concurrent warm+cold-start races and conflicts with the trigger-review path that always uses internal IDs.

### Suggestions

- Consider upserting the PR record before attempting the claim, so stored.ID is always available and the claim namespace stays consistent with internal IDs. This would eliminate the mixed-namespace issue entirely.
- If the upsert-first approach is too invasive, consider prefixing or partitioning the claim key to distinguish ID sources (e.g., store a string key like 'gh:12345' vs 'db:42') to make the two spaces explicitly non-overlapping.
- The removed diagnostic logging (stored==nil vs empty SHA vs unreachable) was valuable for production debugging. Consider preserving a condensed version that logs which ID source was used for the claim (e.g., 'claim_id_source=github' vs 'claim_id_source=internal').

---

Muriano (approved):

The PR correctly unifies the poll-path claim key to always use GitHub IDs, removing the cold-start vs warm-start mismatch within that path. However, the trigger-review path (SetTriggerReviewFn, line 1347) still calls ClaimInFlightReview with the internal SQLite autoincrement ID (pr.ID from s.GetPR), while the poll path now writes GitHub IDs into the same reviews_in_flight.pr_id column. This cross-path namespace mismatch means a triggered review and a poll-path review for the same (PR, SHA) will use different claim keys and both succeed, defeating deduplication.

### Suggestions

- Consider adding a comment or constant to make explicit that reviews_in_flight.pr_id now stores GitHub IDs, so future callers of ClaimInFlightReview use the correct namespace.
- A migration or cleanup query to convert any existing rows from internal IDs to GitHub IDs would prevent stale claims from the old namespace blocking new reviews after upgrade.

---

Muriano:
## ⚠️ MEDIUM Issue

Trigger-review path still uses internal DB ID for ClaimInFlightReview (pr.ID from s.GetPR at line 1300) while the poll path now uses GitHub ID (pr.ID from the GitHub PR struct). Both write to the same reviews_in_flight.pr_id column, so a concurrent trigger + poll for the same PR produces two different claim keys — (internal_id, sha) vs (github_id, sha) — and both succeed. Fix: use pr.GithubID (or equivalent GitHub-sourced ID) in the trigger path as well, or normalize both paths to the same ID source.

**Location:** `daemon/cmd/heimdallm/main.go` line 1347

---

PR #363, opened 2026-04-25
fix(issues): break re-triage loop by detecting bot own comments (closes #362)

### LINKED ISSUE #362: fix(issues): triage re-processing loop — claim key invalidated by bot own comment
## Problem

The issue in-flight claim uses `(issue_id, updated_at)` as composite key. When the bot posts a triage comment, GitHub bumps `updated_at`, producing a **new claim key** on the next poll cycle. The previous claim no longer matches, so the issue is triaged again — creating an infinite loop.

**Real-world impact (2026-04-25):** `overmind-swarm/dev#317` (label `heimdallm-triage`) received 3 duplicate triage comments in 3 minutes before manual intervention (label removal + container stop).

This is the same root cause as #346 but now confirmed with the persistent claim mechanism from #292 — the claim protects against concurrent execution within the same `updated_at` but **not** against sequential re-processing after the bot's own comment bumps the timestamp.

## Root cause

1. Issue has `heimdallm-triage` label → `review_only` action
2. Bot posts triage comment → GitHub bumps `updated_at` to T+1
3. Next poll (1m): claim key is now `(issue_id, T+1)` — different from the previous `(issue_id, T)` → claim succeeds
4. Bot posts another comment → `updated_at` bumps to T+2 → repeat

## Proposed fix

After a successful triage (`posted=true`), check whether the most recent comment on the issue was authored by the bot itself. If so, skip reprocessing. This can be implemented as:

1. **In `alreadyProcessed()` (fetcher.go):** query the issue's latest comment author. If it matches the bot login and the comment was posted after the last known triage, return `true`.

2. **Alternative — store-based:** after posting a triage comment, record `(issue_id, comment_id)` in the store. On the next cycle, check if the latest activity is the bot's own comment — if so, skip.

Either approach breaks the feedback loop without changing the claim key semantics.

## Related

- #346 — original report of `default_action = "review_only"` loop (same root cause, different trigger)
- #292 — persistent in-flight claim (protects concurrency, not sequential re-processing)

sergiotejon:
## Summary
- The 30s `RecomputeGrace` window is shorter than the 1m `poll_interval`, so after the bot posts a triage comment, the `updated_at` bump exceeds the grace on the next cycle — causing infinite re-triage
- Add bot-comment detection in `alreadyProcessed()`: if the most recent comment is from the bot itself, skip reprocessing
- A human comment after the bot still triggers re-triage as expected
- Wire `SetBotLogin` on the Fetcher from `main.go`

## Tests
- `TestFetcher_BotCommentSkipsReprocess` — bot's last comment prevents reprocessing
- `TestFetcher_HumanCommentAfterBotAllowsReprocess` — human reply after bot allows reprocessing

## Impact
Prevents duplicate triage comments. Real-world case: `overmind-swarm/dev#317` received 3 duplicate triages in 3 minutes on 2026-04-25.

Closes #362

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -260,7 +260,9 @@ func main() {
 	issuePipe.SetCircuitBreakerLimits(&issueCBLimits)
 
 	// Resolve bot login for re-review / re-triage context filtering.
+	var resolvedBotLogin string
 	if login, err := ghClient.AuthenticatedUser(); err == nil {
+		resolvedBotLogin = login
 		p.SetBotLogin(login)
 		issuePipe.SetBotLogin(login)
 		slog.Info("bot login resolved", "login", login)
@@ -268,6 +270,7 @@ func main() {
 		slog.Warn("could not resolve bot login for re-review context", "err", err)
 	}
 	issueFetcher := issuepipeline.NewFetcher(ghClient, ghClient, s, issuePipe)
+	issueFetcher.SetBotLogin(resolvedBotLogin) // break re-triage loop (#362)
 	srv := server.New(s, broker, p, apiToken)
 	srv.SetNATSConn(eventBus.Conn())
 	srv.SetConfigPath(cfgPath)
--- a/daemon/internal/issues/fetcher.go
+++ b/daemon/internal/issues/fetcher.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strings"
 	"time"
 
 	"github.com/heimdallm/daemon/internal/config"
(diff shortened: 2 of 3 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, focused fix for the re-triage loop (#362). The approach is correct: detecting the bot's own comments as the most recent activity and skipping reprocessing. Case-insensitive comparison, empty-login guard, and two well-structured tests (bot-skips + human-after-bot-allows) cover the key paths. No issues found.

### Suggestions

- Consider passing botLogin as a constructor parameter to NewFetcher instead of a separate SetBotLogin setter — it would make the dependency explicit and avoid a partially-initialized Fetcher. That said, SetPublisher follows the same setter pattern, so this is consistent with the codebase today.
- The bot-comment check assumes GitHub returns comments in chronological order (oldest first). This is the default API behavior, but a brief code comment noting this assumption would help future maintainers.

---

PR #364, opened 2026-04-25
fix(issues): bot-comment check must allow mode transitions

sergiotejon:
## Summary
- The bot-comment dedup from #363 was too aggressive: it blocked **all** reprocessing when the last comment was from the bot, even when the issue mode changed (e.g. `heimdallm-triage` → `heimdallm-develop`)
- Now only skips when `latest.ActionTaken == issue.Mode` — a mode change always allows reprocessing
- Added `TestFetcher_ModeChangeBypassesBotCommentCheck` test

## Impact
Without this fix, swapping a label from `heimdallm-triage` to `heimdallm-develop` on an already-triaged issue would silently do nothing.

Refs: #362

### DIFF
--- a/daemon/internal/issues/fetcher.go
+++ b/daemon/internal/issues/fetcher.go
@@ -213,12 +213,17 @@ func (f *Fetcher) alreadyProcessed(issue *github.Issue) (bool, string, error) {
 	// dedup gates so a retry marker can override all of them. The API call
 	// is skipped when the comment fetcher is nil (legacy callers / tests
 	// that pre-date marker support).
+	// cachedComments holds the comments fetched during marker scanning,
+	// reused later for the bot-comment check to avoid a second API call.
+	var cachedComments []github.Comment
+
 	if f.comments != nil {
 		comments, cmErr := f.comments.FetchIssueCommentsOnly(issue.Repo, issue.Number)
 		if cmErr != nil {
 			slog.Warn("issues fetcher: marker scan failed, falling through to dedup checks",
 				"repo", issue.Repo, "number", issue.Number, "err", cmErr)
 		} else {
+			cachedComments = comments
 			switch ScanMarkers(comments) {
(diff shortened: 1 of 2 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-scoped fix that moves the bot-comment dedup check after the LatestIssueReview lookup and adds a mode-match guard (`latest.ActionTaken == string(issue.Mode)`). This correctly allows reprocessing when a label/mode change occurs while still preventing the self-triggered re-triage loop (#362). The `cachedComments` hoisting avoids a second API call. `latest` is guaranteed non-nil at the new check location (guarded by the ErrNoRows early-return at line 230). Tests cover same-mode skip, mode-change bypass, and human-comment-after-bot scenarios. No issues found.

### Suggestions

- Consider adding a brief log line (slog.Debug) when the bot-comment check allows reprocessing due to a mode change — it would help trace mode-transition events in production without needing to correlate multiple log entries.
- The existing TestFetcher_BotCommentSkipsReprocess now implicitly depends on fixture() returning IssueModeReviewOnly matching the ActionTaken. A small inline comment noting this coupling would prevent future test fragility if fixture() defaults change.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-scoped fix. Moves the bot-comment dedup check after the mode/ActionTaken lookup and adds a mode-match condition so that label-driven mode transitions (e.g. review_only → develop) are not blocked by the bot's own prior comment. The cachedComments variable is safely scoped and only populated on success. The `latest` variable is guaranteed non-nil at the new check site due to prior early returns. Tests cover both the same-mode skip and the mode-change bypass. No issues found.

### Suggestions

- Consider adding a brief log line (slog.Debug) when the bot-comment check is bypassed due to mode change — it would help trace why an issue was reprocessed in production logs without adding noise at Info level.

---

PR #366, opened 2026-04-27
[codex] fix PR review dedup for mismatched GitHub IDs

### LINKED ISSUE #351: bug: PR github_id mismatch between Search API and Pulls API causes re-review loop
## Problem

PR #337 is re-processed every poll cycle (~1 min) despite having been successfully reviewed. The pipeline catches it via SHA dedup (`skipping re-review, HEAD SHA unchanged`) but the PR should never reach the worker.

## Root Cause

The GitHub Search Issues API returns a different `id` than the Pulls API for the same PR:

- **Search API** (used by `FetchPRsToReview`): `github_id = 4321703389`
- **Pulls API** (used by `pipeline.Run` → `UpsertPR`): `github_id = 3578062677`

`PRAlreadyReviewed(githubID, updatedAt)` looks up the PR by `github_id`. Since the Search API id (4321703389) doesn't match the stored Pulls API id (3578062677), the lookup returns nil → "not reviewed" → published to NATS → worker runs → pipeline catches it by SHA.

## Impact

- 2 unnecessary GitHub API calls per tick per affected PR (GetPR + HeadSHA check)
- Noisy logs (`review-worker: processing` + `skipping re-review` every minute)
- No AI token waste (SHA guard prevents actual review execution)

## Fix

The `FetchPRsToReview` method (Search API path) should resolve the correct PR id from the Pulls API, or `PRAlreadyReviewed` should also check by `(repo, number)` as a fallback when the github_id lookup fails.

ivanmunozruiz:
## Summary

Fixes #351.

- Adds a store lookup by stable PR identity `(repo, number)`.
- Uses that lookup as a scheduler-side fallback when `github_id` lookup misses.
- Keeps the existing `github_id` fast path for normal Pulls API / Tier 3 flows.
- Adds regression coverage for the Search API vs Pulls API ID mismatch and duplicate-row preference.

## Root cause

Tier 2 checks `PRAlreadyReviewed` with the ID returned by GitHub Search Issues API, but stored PR rows are written from Pulls API data. Those IDs can differ for the same PR, causing the scheduler to publish review work that the downstream SHA guard later skips.

## Validation

- `git diff --check` passes.
- `make test-docker GO_TEST_ARGS="-run TestPRAlreadyReviewed -count=1 ./internal/pipeline"` could not run locally because Docker returned `500 Internal Server Error` on the Docker socket. I did not run host `go test`, per repo instructions.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -1813,7 +1813,7 @@ func runTier2(
 				if _, ok := monitoredSet[pr.Repo]; !ok {
 					continue
 				}
-				if adapter.PRAlreadyReviewed(pr.ID, pr.UpdatedAt) {
+				if adapter.PRAlreadyReviewed(pr.ID, pr.Repo, pr.Number, pr.UpdatedAt) {
 					continue
 				}
 				if err := prPublisher.PublishPRReview(ctx, pr.Repo, pr.Number, pr.ID, pr.HeadSHA); err != nil {
@@ -1908,17 +1908,17 @@ func upsertDiscoveredRepos(c *config.Config, prs []*gh.PullRequest) []string {
 // ── tier2Adapter bridges main.go's concrete types to Pipeline interfaces ──
 
 type tier2Adapter struct {
-	ghClient  *gh.Client
-	ghToken   string
-	pipeline  *pipeline.Pipeline
-	issuePipe *issuepipeline.Pipeline
-	fetcher   *issuepipeline.Fetcher
-	store     *store.Store
-	broker    *sse.Broker
-	cfgMu     *sync.Mutex
-	cfg       **config.Config
-	loginMu   *sync.Mutex
-	login     *string
+	ghClient   *gh.Client
+	ghToken    string
+	pipeline   *pipeline.Pipeline
+	issuePipe  *issuepipeline.Pipeline
+	fetcher    *issuepipeline.Fetcher
+	store      *store.Store
+	broker     *sse.Broker
+	cfgMu      *sync.Mutex
+	cfg        **config.Config
+	loginMu    *sync.Mutex
+	login      *string
 	runReview  func(pr *gh.PullRequest, aiCfg config.RepoAI) *store.Review
 	publishPub *bus.PRPublishPublisher
 	watchStore *bus.WatchStore
@@ -2360,8 +2360,15 @@ func (a *tier2Adapter) PromoteReady(ctx context.Context, repos []string) (int, e
 }
 
 // PRAlreadyReviewed implements scheduler.Tier2Store.
-func (a *tier2Adapter) PRAlreadyReviewed(githubID int64, updatedAt time.Time) bool {
+func (a *tier2Adapter) PRAlreadyReviewed(githubID int64, repo string, number int, updatedAt time.Time) bool {
 	existing, _ := a.store.GetPRByGithubID(githubID)
+	if existing == nil && repo != "" && number > 0 {
+		existing, _ = a.store.GetPRByRepoNumber(repo, number)
+		if existing != nil {
+			slog.Debug("pr dedup: matched stored PR by repo/number after github_id miss",
+				"repo", repo, "pr", number, "incoming_github_id", githubID, "stored_github_id", existing.GithubID)
+		}
+	}
 	if existing == nil {
 		return false
 	}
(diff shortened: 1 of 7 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Adds a stable repo/number fallback for PR dedup when GitHub's Search Issues API and Pulls API return different global IDs for the same PR. The fix is well-scoped: extends the Tier2Store interface, threads repo/number through both Tier 2 and Tier 3 call sites, adds a new store query that prefers the reviewed row when duplicates exist, and includes targeted tests covering the mismatch scenario, the duplicate-row tiebreak, and the existing grace-window/legacy-row paths. Test adapter in pipeline/testutil_test.go is kept in sync with the production adapter as the comment requires.

### Suggestions

- Consider adding an index on prs(repo, number) if one doesn't already exist, since GetPRByRepoNumber will now be called on every Tier 2 poll for PRs whose github_id misses.
- The duplicate-row situation that GetPRByRepoNumber's ORDER BY guards against suggests a latent data integrity issue — worth a follow-up to add a unique constraint on (repo, number) once existing duplicates are cleaned up, so the Search vs Pulls API mismatch produces an upsert collision rather than a second row.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-tested fix for a real dedup bug caused by GitHub's Search Issues API vs Pulls API returning different global IDs for the same PR. Adds a repo/number fallback lookup, with proper ordering to prefer the reviewed row. Good test coverage including the exact scenario from issue #351. One minor performance suggestion but no blocking issues.

### Suggestions

- Consider adding a composite index on (repo, number) to support the new GetPRByRepoNumber query efficiently: CREATE INDEX IF NOT EXISTS idx_prs_repo_number ON prs(repo, number). Currently only idx_prs_repo exists. The correlated subquery in the ORDER BY clause joins reviews per matching row, so without a covering index the fallback path scans more than necessary. For small stores this is fine, but it's cheap insurance.
- The testAdapter in testutil_test.go intentionally omits the slog.Debug call present in the production tier2Adapter. This is reasonable for tests, but the comment 'Keep the two in sync' (line 29) might mislead future contributors into thinking the logic must be identical. Consider a brief note that the debug log is intentionally omitted in the test copy.

---

Muriano:
## ⚠️ MEDIUM Issue

GetPRByRepoNumber queries WHERE repo = ? AND number = ? with a correlated subquery against reviews, but there is no index on prs(repo, number). The existing schema only has idx_prs_repo on (repo) alone. This fallback fires on every Tier 2 poll cycle for any PR whose github_id changed between APIs, causing a partial table scan + correlated subquery each time. Add CREATE INDEX IF NOT EXISTS idx_prs_repo_number ON prs(repo, number) to the migration block in store.go.

**Location:** `daemon/internal/store/prs.go` line 78

---

Muriano (approved):

Clean, well-scoped fix for the Search Issues API vs Pulls API github_id mismatch (issue #351). Adds a repo/number fallback in PRAlreadyReviewed with correct tiebreaker logic preferring reviewed rows. Tests cover the new path including the duplicate-row scenario. One missing database index should be addressed before merge.

### Suggestions

- Consider logging (at warn level, not just debug) when the github_id mismatch is first detected — this signals a data inconsistency that operators may want to track, and slog.Debug may be filtered out in production configurations.
- The silently discarded error from GetPRByRepoNumber (existing, _ = ...) follows the pre-existing pattern for GetPRByGithubID, but a future cleanup pass could at least log store errors at debug level to aid troubleshooting.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

The previous MEDIUM issue (missing index on prs(repo, number)) has been addressed — the composite index is now created in store.go. The overall change is clean: the fallback lookup by repo/number is well-guarded (repo != '' && number > 0), the SQL query correctly prefers rows with review history, tests cover the new dedup path including the duplicate-row scenario, and both the production adapter and test adapter are kept in sync. No new issues found.

### Suggestions

- The testAdapter in testutil_test.go mirrors production PRAlreadyReviewed but omits the slog.Debug line present in main.go. This is fine for tests, but the comment 'Keep the two in sync' could note the intentional logging omission to avoid future confusion.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

The previous review's medium-severity finding about the missing index on prs(repo, number) has been addressed: store.go now creates idx_prs_repo_number in the migration block. The fallback logic in PRAlreadyReviewed correctly handles the GitHub Search vs Pulls API ID mismatch, with good test coverage including the new TestPRAlreadyReviewed_FallsBackToRepoNumberWhenGithubIDDiffers and TestPR_GetByRepoNumberPrefersReviewedRow tests. The signature change is propagated consistently across the scheduler interface, main.go adapter, test adapter, and all call sites. The ORDER BY clause in GetPRByRepoNumber sensibly prefers rows with prior review history to dedup against real prior work. No new issues introduced.

### Suggestions

- Consider extracting the duplicated PRAlreadyReviewed fallback logic between main.go's tier2Adapter and pipeline's testAdapter into a shared helper to prevent future drift, as the comment in testutil_test.go already warns about.
- The db.Exec call for the new index ignores its error like the surrounding lines, which is consistent but worth a follow-up to add proper error handling for migration failures across the board.

---

PR #370, opened 2026-04-27
fix: make discovery interval follow poll interval

### LINKED ISSUE #282: bug: Tier 1 discovery interval hardcoded to 15m — ignores poll_interval config
## Problem

The Tier 1 (repo discovery) interval is hardcoded to 15 minutes in \`scheduler/pipeline.go:54-55\`:

\`\`\`go
if cfg.DiscoveryInterval <= 0 {
    cfg.DiscoveryInterval = 15 * time.Minute
}
\`\`\`

There is no config field to override this. The user's \`poll_interval\` setting (e.g. \`1m\`) only affects Tier 2 (PR/issue polling) and Tier 3 (watch cycle), not Tier 1.

This means newly created repos or repos that gain their first PR take up to 15 minutes to be discovered, regardless of how aggressively the user has configured polling.

## Expected

Either:
- \`poll_interval\` should also govern discovery (simplest)
- Or add a separate \`discovery_interval\` config field in \`[github]\` (more flexible)

## Location

\`daemon/internal/scheduler/pipeline.go\` line 54-55

ivanmunozruiz:
## Summary

Fixes #282.

- Make Tier 1 discovery use `poll_interval` when `discovery_interval` is not explicitly configured.
- Preserve explicit `discovery_interval` as an override for deployments that want a separate discovery cadence.
- Stop materializing a default `15m` discovery interval in config defaults so env/store `poll_interval` changes can drive the effective discovery cadence.
- Add parser coverage for fallback/default/explicit interval behavior and negative discovery intervals.

## Behavior Change

Deployments with `discovery_topic` but no explicit `discovery_interval` now refresh discovery at the effective `poll_interval` cadence instead of the previous implicit `15m` cadence. With default config this means `5m`; if `poll_interval` is set lower, discovery follows that lower cadence too. Set `discovery_interval` explicitly to keep discovery slower, for example when Search API budget is tight across many `discovery_orgs`.

## Validation

- `git diff --check` passed
- `make test-docker GO_TEST_ARGS="-run TestParseDiscoveryInterval ./cmd/heimdallm"` blocked locally by Docker socket 500 before the container started: `request returned 500 Internal Server Error for API route .../_ping`

Go tests were not run on host per repo policy; CI should exercise the Docker/Go suite.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -527,7 +527,10 @@ func main() {
 
 		// Tier 1: Discovery — publishes to NATS
 		cfgMu.Lock()
-		discoveryInterval := parseDiscoveryInterval(cfg.GitHub.DiscoveryInterval)
+		discoveryInterval := parseDiscoveryInterval(
+			cfg.GitHub.DiscoveryInterval,
+			cfg.GitHub.PollInterval,
+		)
 		cfgMu.Unlock()
 		wg.Add(1)
 		go func() {
@@ -1652,13 +1655,14 @@ func parsePollInterval(s string) time.Duration {
 	return d
 }
 
-// parseDiscoveryInterval falls back to 15m when the value is empty or invalid.
+// parseDiscoveryInterval falls back to pollInterval when the discovery-specific
+// value is empty or invalid.
 // Config.Validate rejects invalid durations before we reach here, so the
-// fallback only covers the unset-in-TOML-but-topic-defaulted case.
-func parseDiscoveryInterval(s string) time.Duration {
-	d, err := time.ParseDuration(s)
+// fallback normally covers the unset discovery_interval case.
+func parseDiscoveryInterval(discoveryInterval, pollInterval string) time.Duration {
+	d, err := time.ParseDuration(discoveryInterval)
 	if err != nil || d <= 0 {
-		return 15 * time.Minute
+		return parsePollInterval(pollInterval)
 	}
 	return d
 }
--- a/daemon/cmd/heimdallm/main_interval_test.go
+++ b/daemon/cmd/heimdallm/main_interval_test.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"testing"
+	"time"
+)
+
+func TestParseDiscoveryIntervalFallsBackToPollInterval(t *testing.T) {
+	got := parseDiscoveryInterval("", "1m")
+	if got != time.Minute {
+		t.Fatalf("parseDiscoveryInterval(empty, 1m) = %v, want 1m", got)
+	}
+}
+
+func TestParseDiscoveryIntervalPreservesExplicitInterval(t *testing.T) {
+	got := parseDiscoveryInterval("30m", "1m")
+	if got != 30*time.Minute {
+		t.Fatalf("parseDiscoveryInterval(30m, 1m) = %v, want 30m", got)
+	}
+}
+
+func TestParseDiscoveryIntervalUsesPollDefaultWhenBothInvalid(t *testing.T) {
+	got := parseDiscoveryInterval("", "nope")
+	if got != 5*time.Minute {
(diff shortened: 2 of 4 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, focused change that replaces a hardcoded 15m discovery interval fallback with a dynamic fallback to the configured poll interval. The fallback chain is sound (discovery → poll → 5m default), thread safety is preserved under cfgMu, tests cover the three key scenarios, and the config comment/doc updates are accurate.

### Suggestions

- Consider adding a test case for parseDiscoveryInterval with a negative duration (e.g., "-5m", "1m") to explicitly verify the d <= 0 guard delegates to parsePollInterval.
- A brief TOML config comment or CHANGELOG note mentioning the behavior change (discovery no longer defaults to 15m independently) could help users who relied on the previous default.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Small, well-scoped change that makes discovery_interval default to poll_interval at runtime instead of hard-coded 15m. Logic is correct, tests cover the key cases, and comments/docs are updated consistently. One minor concern about the rationale for removing the stricter discovery cadence given the Search API rate limit noted in the previous comment.

### Issues

🟡 **daemon/internal/config/config.go:64** — The original comment cited a stricter Search API rate limit (30 req/min) as the reason for an independent 15m default. With discovery now defaulting to poll_interval (default 5m), users relying on defaults could push the Search API harder than before. Worth confirming this is intentional and that doc/release notes call out the behavior change so existing deployments aren't surprised.
🟡 **daemon/cmd/heimdallm/main_interval_test.go:23** — TestParseDiscoveryIntervalUsesPollDefaultWhenBothInvalid hard-codes the expected 5m, which couples this test to parsePollInterval's internal default. If that default ever changes, this test fails for an unrelated reason. Consider referencing the same constant/helper or asserting via parsePollInterval("nope") to keep the tests decoupled.

### Suggestions

- Add a test (or extend an existing one) verifying that when both discovery_interval and poll_interval are explicitly set, the explicit poll_interval value (not the 5m default) is used as the fallback — this guards the realistic config path.
- Update any user-facing docs/CHANGELOG to flag that discovery now follows poll_interval by default; previous deployments inherited 15m implicitly.
- Consider whether parseDiscoveryInterval should warn/log when falling back, so operators notice that discovery is running at poll cadence.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Small, focused change making discovery interval fall back to poll interval instead of a hardcoded 15m. Logic is correct, tests cover the main cases, and documentation is updated consistently.

### Suggestions

- Consider adding a test for the case where discovery_interval is explicitly invalid (non-empty but unparseable) to confirm fallback behavior, though Config.Validate should reject these upstream.
- The doc comment on parseDiscoveryInterval still says 'Config.Validate rejects invalid durations before we reach here' — worth verifying Validate also covers the discovery_interval field now that the default is gone.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: the author addressed the previous low-severity test coupling finding by using parsePollInterval("nope") as the expected value, decoupling the test from the hard-coded 5m default. The Search API rate-limit concern around the default behavior change remains worth confirming in docs/release notes, but the doc comment in config.go has been updated to describe the new behavior. No new issues introduced.

### Suggestions

- Consider noting the default-behavior change (discovery now follows poll_interval, default 5m, vs previous 15m) in release notes so operators relying on defaults aren't surprised by increased Search API usage.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-scoped change. The fallback from a hardcoded 15m to pollInterval is correct and the test coverage is solid across all branches (empty, explicit, negative, both-invalid). No issues found on re-review.

### Suggestions

- Consider a table-driven test in main_interval_test.go to reduce repetition across the four cases — purely cosmetic, not blocking.

---

PR #371, opened 2026-04-27
docs: explain macOS TCC permission dialogs

Muriano:
## Summary

- Adds a Troubleshooting section to README documenting why macOS shows permission dialogs for Music/Downloads/Pictures when `HEIMDALLM_LOCAL_DIR_BASE` is set to `$HOME`
- Explains the fix: scope the env var to specific code directories instead of `~`

## Test plan
- [x] Docs-only change, no code modified

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,6 +507,34 @@ Alternatively, **release-please** (via `release.yml`) automates this: it reads c
 
 ---
 
+## Troubleshooting
+
+### macOS asks for permission to access Music, Downloads, or other folders
+
+On macOS, the daemon's `local_dir_base` feature calls `os.Stat()` on
+candidate paths for each monitored repository (to auto-detect local
+clones). If `HEIMDALLM_LOCAL_DIR_BASE` is set to a broad path like `$HOME`,
+the stat probes can touch TCC-protected directories (`~/Music`,
+`~/Downloads`, `~/Pictures`, etc.), causing macOS to show permission
+dialogs — even though Heimdallm has no interest in those folders.
+
+**Fix:** set `HEIMDALLM_LOCAL_DIR_BASE` to the specific directory (or
+directories) where your repos actually live:
+
+```bash
+# Good — scoped to your actual code directories
+export HEIMDALLM_LOCAL_DIR_BASE=~/projects,~/work
+
+# Bad — triggers TCC prompts for every protected subfolder of $HOME
+export HEIMDALLM_LOCAL_DIR_BASE=~
+```
+
+The native file picker (`Browse` button in repo settings) may also
+trigger similar prompts when macOS enumerates recent/default locations.
+This is standard `NSOpenPanel` behavior and not specific to Heimdallm.
+
+---
+
 ## License
 
 MIT


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Documentation-only PR adding a Troubleshooting section to the README explaining macOS TCC permission dialogs caused by the daemon's local_dir_base stat probes. Content is accurate, well-scoped, and provides clear good/bad examples. No code changes.

### Suggestions

- Consider noting that comma-separated values in HEIMDALLM_LOCAL_DIR_BASE require the parser to support that format — confirm the daemon actually splits on commas, otherwise the example may mislead users.
- Optionally mention the macOS Settings path (System Settings → Privacy & Security → Files and Folders) where users can review/revoke granted permissions.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Documentation-only PR adding a helpful Troubleshooting section that explains macOS TCC permission dialogs caused by broad HEIMDALLM_LOCAL_DIR_BASE values. Content is clear and well-scoped, but the bash example has a tilde-expansion subtlety that could mislead users.

### Issues

⚠️ **README.md:522** — The example `export HEIMDALLM_LOCAL_DIR_BASE=~/projects,~/work` will not expand the second `~` in bash — tilde expansion only happens at the start of a word (or in specific assignment contexts after unquoted `:`), not after a comma. The value will literally become `/Users/<you>/projects,~/work`, and the daemon will try to stat a path containing a literal `~`. Either use absolute paths in the example (`/Users/you/projects,/Users/you/work`), use `$HOME` (`$HOME/projects,$HOME/work`), or document whichever separator/expansion the daemon actually supports. Also worth confirming the env var is comma-separated (vs. `:` like PATH) — the README should match the parser.
🟡 **README.md:516** — Claim that `os.Stat()` is what triggers TCC prompts is plausible but worth double-checking against the actual code path that auto-detects local clones — if it uses `os.ReadDir`, `filepath.Walk`, or opens the directory, the explanation should match the real call to avoid future doc drift. A short cross-reference (file/function name) would make the doc resilient to refactors.
🟡 **README.md:530** — The note about the native file picker triggering prompts is useful but vague. Consider linking to Apple's NSOpenPanel/TCC docs or naming the specific protected resources (Files and Folders, Full Disk Access) so users know exactly which permission to grant or deny in System Settings → Privacy & Security.

### Suggestions

- Replace `~` in the multi-path example with `$HOME` or absolute paths so the example is copy-pasteable and actually works in bash/zsh.
- Confirm the path separator the daemon expects (comma vs. colon vs. OS-specific `os.PathListSeparator`) and document it explicitly — comma is unusual on Unix and may surprise users.
- Consider adding a one-line pointer to the Go code that performs the auto-detect stat (file:function) so the troubleshooting note stays anchored if the implementation moves.
- Optional: mention that users can also grant Heimdallm Full Disk Access in System Settings as an alternative to narrowing the env var, for users who legitimately need broad scanning.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Documentation-only change adding a Troubleshooting section to README.md explaining macOS TCC permission dialogs caused by HEIMDALLM_LOCAL_DIR_BASE stat probes. The explanation is clear, technically accurate, and provides actionable guidance with good/bad examples.

### Suggestions

- Consider adding a brief note about whether users can verify which directories are being probed (e.g., via logs or debug output) to help diagnose the issue.
- Could mention the specific macOS versions where TCC behavior applies, though this is minor since TCC has been standard for several releases.

---

PR #372, opened 2026-04-27
fix: restore docker go test gate

vbuenog:
## Summary
- update the pinned Go Docker test image to Go 1.25 to match daemon/go.mod
- update the Linux verifier Go toolchain and related docs
- harden .dockerignore so local secrets/runtime config stay out of Docker build contexts

## Verification
- make test-docker GO_TEST_ARGS="-run TestDoesNotExist ./..."
- make test-docker

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,22 @@
-# Docker context exclusions for Dockerfile.linux-verify
+# Docker context exclusions for repo-root builds
 .git
+.codex
+
+# Local secrets and runtime config
+.env
+.env.*
+!.env.example
+docker/.env
+docker/.env.*
+!docker/.env.example
+docker/config
+*.pem
+*.key
+*.crt
+*.p12
+*.pfx
+
+# Build outputs
 flutter_app/build
 daemon/bin
 dist
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -41,7 +41,7 @@ Hardening details are inlined in the Makefile target. Summary:
 
 | Concern | Guarantee |
 |---|---|
-| Image | Official `golang:1.21-alpine`, **pinned by SHA256 digest** (not a mutable tag) |
+| Image | Official `golang:1.25-alpine`, **pinned by SHA256 digest** (not a mutable tag) |
 | Host filesystem access | Single mount: repo at `/src`, **read-only** (`:ro`) |
 | Build cache | Redirected to `/tmp/heimdallm-gocache/` — does not touch `~/.cache` or `~/go` |
 | Privileges | Runs as invoking user (`--user $(id -u):$(id -g)`) — no root in the container |
--- a/Dockerfile.linux-verify
+++ b/Dockerfile.linux-verify
@@ -1,6 +1,6 @@
 # Dockerfile.linux-verify
 #
-# Replicates the CI environment (ubuntu-22.04, Go 1.21, Flutter 3.x stable)
+# Replicates the CI environment (ubuntu-22.04, Go 1.25, Flutter 3.x stable)
 # and runs the full verification pipeline:
 #
 #   daemon tests → flutter pub get → flutter test →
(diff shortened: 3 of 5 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Straightforward Go version bump (1.21 → 1.25) across Docker, Makefile, and docs, plus hardened .dockerignore to exclude secrets from the build context. Changes are consistent across files.

### Issues

⚠️ **Dockerfile.linux-verify:27** — The Go tarball is downloaded over HTTPS but its SHA256 checksum is never verified before extraction. A compromised mirror or MITM could inject a tampered toolchain. Consider adding a checksum verification step (e.g., `echo '<sha256>  /tmp/go.tar.gz' | sha256sum -c -`). This is pre-existing but worth addressing alongside the version bump.
🟡 **Makefile:66** — The SHA256 digest for golang:1.25-alpine should be independently verified against Docker Hub (e.g., `docker pull golang:1.25-alpine && docker inspect --format='{{.RepoDigests}}'`). An incorrect or stale digest will cause silent pull failures or pin to a wrong image.
🟡 **Dockerfile.linux-verify:27** — The comment says 'matches daemon/go.mod' but daemon/go.mod is not included in this diff. Confirm that go.mod already specifies Go 1.25; otherwise the version claim is misleading and the build may behave unexpectedly.

### Suggestions

- Add SHA256 checksum verification for the Go tarball download in Dockerfile.linux-verify to match the digest-pinning discipline already applied to the Docker image in the Makefile.
- Consider adding a smoke-test step in CI that runs `make test-docker` to validate the new image digest and Go version actually work end-to-end before merge.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

PR bumps Go toolchain from 1.21 to 1.25 across Dockerfile, Makefile (with new pinned digest), AGENTS.md, and README, and expands .dockerignore to exclude secrets and local env files. Changes are consistent and the digest pin is preserved. Previous findings concerned unrelated Dart files not present in this diff.

### Suggestions

- Verify the new golang:1.25-alpine SHA256 digest (f6751d82...) matches the official multi-arch index on Docker Hub for reproducibility across arm64/amd64 hosts.
- Confirm daemon/go.mod's go directive aligns with 1.25 (or is ≤1.25) so the pinned image actually compiles the module without toolchain auto-download.
- Consider adding docker/secrets/ or similar broad secret directories to .dockerignore alongside the *.pem/*.key patterns for defense-in-depth.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Restores the docker go test gate by bumping the pinned golang image and toolchain references from 1.21 to 1.25, aligning Dockerfile, Makefile, AGENTS.md and README. Also expands .dockerignore to exclude secrets and local env files from build context. Changes are consistent and the image remains pinned by SHA256 digest.

### Suggestions

- Consider verifying the new golang:1.25-alpine SHA256 digest against the official registry once more before merge, since digest typos silently break reproducibility.
- The .dockerignore now references docker/config and docker/.env paths — confirm those directories exist or are intentional placeholders to avoid confusion.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Straightforward Go toolchain version bump from 1.21 to 1.25 across Dockerfile.linux-verify, Makefile docker test image (with new pinned SHA256 digest), and accompanying docs. Also adds sensible .dockerignore exclusions for secrets, env files, certificates, and build outputs. Changes look consistent and the dockerignore negation patterns (e.g. !.env.example after .env.*) are ordered correctly. No correctness issues spotted in the diff itself; main residual risk is whether other references to Go 1.21 still exist elsewhere in the repo (CI workflows, other Dockerfiles, tool-versions files) that weren't touched here.

### Issues

🟡 **.dockerignore:0** — Excluding `.git` from the Docker build context will break any build step that relies on git metadata (e.g. `git describe` for version stamping, VCS info embedded by `go build`). Worth confirming Dockerfile.linux-verify and any other root-context build don't need .git — if they do, this exclusion needs to be scoped or removed.
🟡 **Makefile:66** — The GO_DOCKER_IMAGE digest is pinned but there's no automated check that it matches the Go version in daemon/go.mod. If go.mod is later bumped (e.g. to 1.26) and this line is missed, `make test-docker` will silently run on an older toolchain — which is exactly the regression this PR is fixing. Consider a Makefile guard or CI check that compares the toolchain version in the image against go.mod.
🟡 **Dockerfile.linux-verify:27** — Go tarball URL/version is hard-coded twice (here and in the Makefile image digest). They can drift independently. A single source of truth (build arg, or a script that derives both from go.mod) would prevent the next 1.21→1.25-style mismatch.

### Suggestions

- Run `grep -rn '1\.21' --include='*.md' --include='Dockerfile*' --include='Makefile' --include='*.yml' --include='*.yaml'` in the repo to confirm no other references to the old Go version were missed (CI workflows, contributor docs, asdf/tool-versions, etc.).
- Verify the new SHA256 digest `f6751d823c…` actually corresponds to `golang:1.25-alpine` on Docker Hub before merging.
- Consider adding a one-line CI assertion that `go version` inside the docker test image matches the `go` directive in `daemon/go.mod`, so this gate doesn't silently break again.

---

PR #373, opened 2026-04-27
fix: refresh topic discovery before publishing repos

vbuenog:
## Summary
- refresh topic-based repository discovery before Tier 1 publishes repo lists
- keep using the previous discovery cache when a refresh fails
- stop including cached discovered repos when discovery_topic is disabled
- add regression coverage for refresh, disabled discovery, and refresh-failure fallback

## Verification
- make test-docker GO_DOCKER_IMAGE=golang:1.25-alpine@sha256:f6751d823c26342f9506c03797d2527668d095b0a15f1862cddb4d927a7a4ced GO_TEST_ARGS="-run TestSendDiscoveryRepos ./cmd/heimdallm"
- make test-docker GO_DOCKER_IMAGE=golang:1.25-alpine@sha256:f6751d823c26342f9506c03797d2527668d095b0a15f1862cddb4d927a7a4ced

Note: this branch is based on main and uses the Go image override until #372 lands, because main's default make test-docker image is still Go 1.21.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -560,9 +560,6 @@ func main() {
 				case <-ctx.Done():
 					return
 				case <-ticker.C:
-					if err := limiter.Acquire(ctx, scheduler.TierDiscovery); err != nil {
-						return
-					}
 					sendDiscoveryRepos(ctx, discoverySvc, limiter, repoPublisher, tier1ConfigFn)
 				}
 			}
@@ -585,7 +582,11 @@ func main() {
 			tier2ConfigFn := func() []string {
 				cfgMu.Lock()
 				defer cfgMu.Unlock()
-				return discovery.MergeRepos(cfg.GitHub.Repositories, discoverySvc.Discovered(), cfg.GitHub.NonMonitored)
+				var discovered []string
+				if cfg.GitHub.DiscoveryTopic != "" {
+					discovered = discoverySvc.Discovered()
+				}
+				return discovery.MergeRepos(cfg.GitHub.Repositories, discovered, cfg.GitHub.NonMonitored)
 			}
 			runTier2(ctx, adapter, limiter, prReviewPublisher, tier2ConfigFn, reposChan, pollInterval, coldStart)
 		}()
@@ -1699,36 +1700,21 @@ func sendDiscoveryRepos(
 	configFn func() scheduler.Tier1Config,
 ) {
 	cfg := configFn()
-	discovered := disc.Discovered()
-
-	// Merge static + discovered, exclude non-monitored
-	nonMon := make(map[string]struct{}, len(cfg.NonMonitored))
-	for _, r := range cfg.NonMonitored {
-		nonMon[r] = struct{}{}
-	}
-	seen := make(map[string]struct{})
-	var repos []string
-	for _, r := range cfg.StaticRepos {
-		if _, skip := nonMon[r]; skip {
-			continue
+	var discovered []string
+	if cfg.DiscoveryTopic != "" {
+		if limiter != nil {
+			if err := limiter.Acquire(ctx, scheduler.TierDiscovery); err != nil {
+				slog.Warn("tier1: acquire discovery rate-limit token failed", "err", err)
+				return
+			}
 		}
-		if _, dup := seen[r]; dup {
-			continue
+		if err := disc.Refresh(cfg.DiscoveryTopic, cfg.DiscoveryOrgs); err != nil {
+			slog.Warn("tier1: discovery refresh failed, using cached discovered repos", "err", err)
 		}
-		seen[r] = struct{}{}
-		repos = append(repos, r)
-	}
-	for _, r := range discovered {
-		if _, skip := nonMon[r]; skip {
-			continue
-		}
-		if _, dup := seen[r]; dup {
-			continue
-		}
-		seen[r] = struct{}{}
-		repos = append(repos, r)
+		discovered = disc.Discovered()
 	}
 
+	repos := discovery.MergeRepos(cfg.StaticRepos, discovered, cfg.NonMonitored)
 	slog.Info("tier1: discovery complete", "repos", len(repos))
 	if err := pub.PublishRepos(ctx, repos); err != nil {
 		slog.Error("tier1: publish repos failed", "err", err)
--- a/daemon/cmd/heimdallm/main_discovery_tier1_test.go
+++ b/daemon/cmd/heimdallm/main_discovery_tier1_test.go
@@ -1,0 +1,143 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"reflect"
+	"testing"
+
+	"github.com/heimdallm/daemon/internal/scheduler"
+)
+
+type fakeTier1Discovery struct {
+	refreshCalls int
+	refreshTopic string
+	refreshOrgs  []string
+	refreshRepos []string
+	refreshErr   error
+	repos        []string
+}
+
+func (f *fakeTier1Discovery) Discovered() []string {
+	if len(f.repos) == 0 {
+		return nil
+	}
+	out := make([]string, len(f.repos))
+	copy(out, f.repos)
+	return out
+}
+
+func (f *fakeTier1Discovery) Refresh(topic string, orgs []string) error {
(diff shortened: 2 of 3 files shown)


sergiotejon:
## 🟡 LOW Issue

When Refresh fails, disc.Discovered() is still called and may return stale cached repos. This is intentional per the log message, but if Refresh partially mutates internal state on error (e.g. clears cache), behavior depends on the Discovery implementation contract. Consider documenting on the Tier1Discovery.Refresh interface whether Discovered() returns prior cache on error.

**Location:** `daemon/cmd/heimdallm/main.go` line 1716

---

sergiotejon:
## 🟡 LOW Issue

Rate limit token is consumed before Refresh even if Refresh fails immediately; on persistent failures this still burns tokens each tick. Acceptable since the limiter exists to throttle external calls regardless of outcome, but worth noting.

**Location:** `daemon/cmd/heimdallm/main.go` line 1717

---

sergiotejon:
## 🟡 LOW Issue

Test asserts published repo order as [org/static, org/discovered], which couples the test to discovery.MergeRepos ordering semantics. If MergeRepos ordering ever changes (e.g., to sorted), this will break. Consider asserting set-equality unless ordering is part of the contract.

**Location:** `daemon/cmd/heimdallm/main_discovery_tier1_test.go` line 78

---

sergiotejon:
## 🟡 LOW Issue

TestSendDiscoveryReposPublishesCachedReposWhenRefreshFails relies on the limiter having capacity 1; if Acquire fails the function returns early and the test would falsely pass on refreshCalls=0. Consider also asserting pub.calls==1 to ensure publish actually happened.

**Location:** `daemon/cmd/heimdallm/main_discovery_tier1_test.go` line 140

---

sergiotejon (approved):

Refactors tier1 discovery to refresh topic data before each publish cycle, properly gates discovery on DiscoveryTopic config, and centralizes repo merging via discovery.MergeRepos. The rate limiter Acquire was correctly moved inside sendDiscoveryRepos so it only runs when discovery is actually performed. Good test coverage for the new behavior including the topic-disabled and refresh-failure paths.

### Suggestions

- Add a test covering the case where limiter.Acquire fails (e.g., context cancelled) to verify the early return and Warn log path.
- Consider documenting on the Tier1Discovery interface what Discovered() should return after a failed Refresh (stale cache vs empty).
- The tier2ConfigFn now gates discovered repos on DiscoveryTopic != ""; consider extracting this gating logic into a small helper to keep tier1 and tier2 behavior consistent and avoid drift.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

PR correctly fixes the bug where discovery topic changes weren't being refreshed before publishing repos. The refactor consolidates merge logic via discovery.MergeRepos and adds proper test coverage with three scenarios (refresh success, topic disabled, refresh failure with cached fallback). Rate limiter is now correctly scoped to discovery operations only, and the topic-disabled path properly skips both refresh and discovered repos. No confirmed bugs or security issues found.

### Suggestions

- Document on Tier1Discovery.Refresh interface whether Discovered() returns prior cache on error, to make the contract explicit for future implementations.
- In TestSendDiscoveryReposPublishesCachedReposWhenRefreshFails, also assert pub.calls==1 to ensure publish actually happened and the test isn't falsely passing due to early return.
- Consider asserting set-equality rather than slice ordering in the published repos tests, unless MergeRepos ordering is part of its documented contract.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid refactor that fixes a real correctness bug — discovery topic refresh now happens before each publish, and tier2 no longer leaks stale discovered repos when DiscoveryTopic is unset. Code is cleaner (delegates to discovery.MergeRepos), error handling is reasonable (cached repos used on refresh failure with a warn log), and tests cover the three main paths. A few minor observations worth addressing.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:1716** — Tier1Discovery.Refresh contract is undocumented re: Discovered() behavior on error. The fallback to cached repos is intentional per the log, but it depends on implementation not clearing cache on failure. Add a doc comment on the interface clarifying that Discovered() must continue to return the last successful result when Refresh returns an error.
🟡 **daemon/cmd/heimdallm/main_discovery_tier1_test.go:140** — TestSendDiscoveryReposPublishesCachedReposWhenRefreshFails does not assert pub.calls==1. If a future change caused early-return before publish (e.g., limiter exhaustion), refreshCalls=1 would still hold but no publish occurred — the test would still pass. Add assertion that publish was actually invoked.
🟡 **daemon/cmd/heimdallm/main_discovery_tier1_test.go:78** — Asserting exact published-repo ordering ([org/static, org/discovered]) couples the test to discovery.MergeRepos current ordering. If MergeRepos semantics change (e.g. sorted), the test breaks even though behavior is still correct. Prefer set-equality or document ordering as part of MergeRepos contract.
🟡 **daemon/cmd/heimdallm/main.go:585** — tier2ConfigFn calls discoverySvc.Discovered() while holding cfgMu. If the discovery service ever takes its own lock that overlaps with the config mutex elsewhere, this introduces a lock-order dependency. Minor — consider snapshotting config inside the lock and calling Discovered() outside.
🟡 **daemon/cmd/heimdallm/main.go:1710** — Rate-limit token is consumed before the Refresh call; on persistent Refresh failures, tokens are still burned each tick. As reviewer noted, this is acceptable for an external-call throttle but consider whether discovery failures should not count against the budget — e.g., release on error or acquire after the call. Low priority, just a design call worth recording.

### Suggestions

- Document on the Tier1Discovery.Refresh interface that Discovered() must return the last successful snapshot when Refresh returns an error.
- In the failure-path test, also assert pub.calls == 1 to guard against silent regressions where publish is skipped.
- Replace exact-slice equality in the success-path test with a set/membership check, or pin MergeRepos ordering as a documented contract.
- Consider extracting the refresh+publish flow into a smaller helper if more tiers grow this pattern, but not necessary now.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured PR that moves discovery refresh into the publish cycle for data freshness, deduplicates merge logic via MergeRepos, and gates discovery behind topic configuration. Core logic is sound with good test coverage for the three main scenarios.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:560** — Behavioral change in goroutine shutdown: the old limiter.Acquire failure (including context cancellation) caused `return` which immediately exited the ticker goroutine. Now it only returns from sendDiscoveryRepos, and the goroutine relies on `case <-ctx.Done()` on the next select iteration. This also means a non-context Acquire timeout (e.g. pool exhaustion after 500ms) no longer kills the goroutine — it just skips one cycle. The new behavior is arguably more resilient, but the silent change to the shutdown path warrants a review comment or explicit design note.
🟡 **daemon/cmd/heimdallm/main.go:1709** — Double-logging on refresh failure: discovery.Service.Refresh already logs `slog.Warn("discovery: refresh failed, keeping previous cache", ...)` when it returns an error with no repos. The new sendDiscoveryRepos then logs a second warning for the same failure event. Consider removing one of the two log lines or demoting the inner one to Debug to avoid noisy duplicate warnings in operator logs.
🟡 **daemon/cmd/heimdallm/main_discovery_tier1_test.go:1** — No test for the limiter.Acquire failure path (lines 1704-1708). When Acquire returns an error, the function returns early without publishing — this is an important guard that should have test coverage. A test with a zero-token limiter and cancelled context would verify that repos are NOT published when rate-limiting fails.
🟡 **daemon/cmd/heimdallm/main_discovery_tier1_test.go:1** — No test for the `limiter == nil` branch (line 1703). All three tests pass a non-nil NewRateLimiter(1). A test confirming that passing nil limiter still calls Refresh and publishes would document this defensive path.

### Suggestions

- Add a test case for Acquire failure (zero-token limiter + cancelled context) to verify that sendDiscoveryRepos does NOT call Refresh or PublishRepos when rate limiting fails.
- Consider removing the warning log in sendDiscoveryRepos (line 1709) since Service.Refresh already logs a similar warning internally — or demote the inner log to Debug to reduce duplicate noise.
- The tier1.go file in scheduler still has the old sendRepos function with inline merge logic (not using MergeRepos). Consider updating it too for consistency, or document why the two paths differ.

---

PR #374, opened 2026-04-27
fix: keep TOML repo lists above store state

### LINKED ISSUE #284: bug: store layer overrides TOML config for repositories/non_monitored — wrong precedence
## Problem

The current config precedence is TOML < env < store (SQLite). This means the store layer can silently override the user's explicit TOML configuration for \`repositories\` and \`non_monitored\`.

In practice, \`upsertDiscoveredRepos\` / \`processDiscoveredRepos\` write to the store's \`repositories\` and \`non_monitored\` keys on every poll cycle. These store values take precedence over TOML, so a user who explicitly adds a repo to \`repositories\` in TOML can have it moved to \`non_monitored\` by the auto-discovery process without any user action.

**Concrete case:** \`freepik-company/fc-vertexai-api\` was in TOML \`repositories\` with explicit \`issue_tracking.enabled = true\`, shown as MONITORED in the UI with all LEDs active, but the daemon was not processing its issues because the store had it in \`non_monitored\`.

## Root Cause

The store was designed for web UI settings (PUT /config), where the operator intentionally overrides TOML from the browser. But \`repositories\` and \`non_monitored\` are also written by the auto-discovery polling — an automated process that should never override explicit user config.

## Correct Precedence

For \`repositories\` / \`non_monitored\` specifically:

**store < TOML < env**

The TOML file is the user's source of truth for which repos to monitor. The store should only hold repos discovered at runtime that are NOT in TOML. Environment variables should override everything (deployment overrides).

## Deeper Problem: Duplication

\`repositories\` and \`non_monitored\` live in both TOML and SQLite with merge semantics that are unclear and buggy. The store should hold runtime-only data (reviews, dismissed flags, agent profiles, first-seen timestamps). Config that the user explicitly manages (which repos to monitor) should live exclusively in TOML.

## Impact

- Repos explicitly configured in TOML can silently stop being monitored
- Issue tracking stops working for repos without active PRs (discovered by #281)
- The Flutter UI shows the repo as monitored (reads TOML/config) while the daemon ignores it (reads store)
- The user has no visibility into why a repo stopped being processed

## Suggested Fix

1. Stop writing \`repositories\` and \`non_monitored\` to the store from auto-discovery
2. Auto-discovered repos that aren't in TOML should be written to TOML (via the PATCH /config endpoint from #144) — not to the store
3. Or: keep discovery writes in the store but make TOML win over store for these two keys (invert precedence for repo lists only)

## Related

- #144 — config round-trip data loss (same class of problem: config source-of-truth confusion)
- #281 — repos drop out of issue tracking when they have no active PRs

vbuenog:
## Summary
- fix repository/non_monitored store merge semantics so explicit TOML/env repo choices win over SQLite runtime state
- keep store-only discovered repos as lower-priority additions for compatibility
- stop legacy PUT /config from persisting repositories into SQLite; PATCH /config continues to write repo lists to TOML
- document repo-list source-of-truth precedence

Fixes #284

## Verification
- make test-docker GO_DOCKER_IMAGE=golang:1.25-alpine@sha256:f6751d823c26342f9506c03797d2527668d095b0a15f1862cddb4d927a7a4ced GO_TEST_ARGS="./internal/config/... ./internal/server/..."
- make test-docker GO_DOCKER_IMAGE=golang:1.25-alpine@sha256:f6751d823c26342f9506c03797d2527668d095b0a15f1862cddb4d927a7a4ced

Note: this branch is based on main and uses the Go image override until #372 lands, because main's default make test-docker image is still Go 1.21.

### DIFF
--- a/daemon/internal/config/store.go
+++ b/daemon/internal/config/store.go
@@ -13,11 +13,11 @@ type StoreLister interface {
 	ListConfigs() (map[string]string, error)
 }
 
-// MergeStoreLayer is the full "apply the store layer on top of TOML+env"
-// operation: fetch rows, apply them atomically, re-validate. Callers that
-// need each step separately can still use ListConfigs + ApplyStore + Validate
-// directly; this helper exists so main.go's bootstrap and reload paths can
-// share one code path and so tests can drive the whole flow with a fake.
+// MergeStoreLayer fetches rows, applies them atomically, then re-validates.
+// Most store-backed keys are operator overrides from the legacy PUT /config
+// path and therefore still sit above TOML+env. Repository lists are different:
+// the poller also writes them as runtime discovery state, so ApplyStore merges
+// those rows below explicit TOML/env values.
 //
 // Returns the first error encountered. On error the receiver is untouched
 // (ApplyStore is atomic and Validate is a read-only check), so the caller is
@@ -38,7 +38,12 @@ func (c *Config) MergeStoreLayer(s StoreLister) error {
 
 // ApplyStore merges runtime-overridable config values written by the
 // PUT /config handler on top of whatever is already in the Config (TOML +
-// env vars). Precedence is TOML < env < store, so this step runs last.
+// env vars). Most keys use TOML < env < store precedence.
+//
+// Repository lists are the exception. Auto-discovery also writes
+// repositories/non_monitored rows, so those store rows are treated as
+// lower-priority runtime additions: store-only repos are kept, but explicit
+// TOML/env list entries win conflicts.
 //
 // The handler stores string values bare and everything else as JSON, so the
 // decoding here is symmetric to handlers.go:handlePutConfig.
@@ -61,6 +66,10 @@ func (c *Config) MergeStoreLayer(s StoreLister) error {
 // leak through to the receiver even when a later row fails.
 func (c *Config) ApplyStore(rows map[string]string) error {
 	shadow := *c
+	var storeRepos []string
+	var storeNonMonitored []string
+	var sawStoreRepos bool
+	var sawStoreNonMonitored bool
 	for key, raw := range rows {
 		switch key {
 		case "poll_interval":
@@ -76,13 +85,15 @@ func (c *Config) ApplyStore(rows map[string]string) error {
 			if err := json.Unmarshal([]byte(raw), &repos); err != nil {
 				return fmt.Errorf("config: apply store key %q: %w", key, err)
 			}
-			shadow.GitHub.Repositories = repos
+			storeRepos = repos
+			sawStoreRepos = true
 		case "non_monitored":
 			var nm []string
 			if err := json.Unmarshal([]byte(raw), &nm); err != nil {
 				return fmt.Errorf("config: apply store key %q: %w", key, err)
 			}
-			shadow.GitHub.NonMonitored = nm
+			storeNonMonitored = nm
+			sawStoreNonMonitored = true
 		case "repo_first_seen":
 			// Auxiliary data read directly from the store by the HTTP
 			// config handler (to render NEW badges) — not applied to the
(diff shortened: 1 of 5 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Refactors store layer so repo lists merge as lower-priority runtime state, with explicit TOML/env entries winning conflicts. Logic is well-scoped, atomic via shadow copy, and well-tested across the new precedence cases. No blocking issues found.

### Suggestions

- Asymmetric env handling: mergeStoreRepoLists only short-circuits store repo additions when HEIMDALLM_REPOSITORIES is set, but there is no equivalent gate for HEIMDALLM_NON_MONITORED. If the project supports that env var, store-only non_monitored entries will still be appended on top of the env-defined list — consider mirroring the behavior or documenting why non_monitored is treated as additive.
- store_test.go uses fmt.Sprintf("%v", ...) for slice comparison; reflect.DeepEqual would be more idiomatic and catch nil-vs-empty-slice differences.
- mergeStoreRepoLists could be a method on *Config for consistency with ApplyStore/MergeStoreLayer, though the free function is fine as a private helper.
- Consider a test asserting deterministic ordering when both TOML and store contribute repositories (current tests cover it implicitly, but an explicit ordering assertion would prevent regressions if iteration logic changes).

---

sergiotejon:
## 🟡 LOW Issue

Asymmetry with HEIMDALLM_NON_MONITORED: env-provided repositories short-circuit store-only monitored additions, but there is no equivalent guard for an env-provided non_monitored list. If HEIMDALLM_NON_MONITORED exists as a counterpart env var (mirroring HEIMDALLM_REPOSITORIES), store non_monitored rows still append below it, which is consistent with the 'lower priority runtime additions' rule but may merit a brief comment or a parallel guard depending on intended semantics.

**Location:** `daemon/internal/config/store.go` line 165

---

sergiotejon (approved):

PR cleanly inverts the precedence of repository lists so explicit TOML/env values win over store-backed runtime discovery state. The merge logic in mergeStoreRepoLists is well-scoped, atomicity of ApplyStore is preserved (mutations applied to shadow), and tests cover the key precedence scenarios including env override, TOML repos vs store non_monitored, and TOML non_monitored vs store repositories. Documentation and handler key classification are updated consistently.

### Suggestions

- Use reflect.DeepEqual or slices.Equal for slice comparisons in tests instead of fmt.Sprintf("%v", ...) — it's more idiomatic and avoids false positives/negatives from formatting quirks.
- Consider recording on Config (e.g., a private field set during applyEnvOverrides) whether HEIMDALLM_REPOSITORIES was explicit, rather than re-reading the env in ApplyStore; this keeps the store layer's behavior reproducible from its inputs alone.
- Add a test for the case where the same repo appears in both storeRepos and storeNonMonitored to lock in the documented 'within the store layer, non_monitored wins' behavior.
- The doc comment on ApplyStore now spans store.go and the test file; consider whether the duplicated precedence explanation in store_test.go can be shortened to avoid drift.

---

sergiotejon:
## 🟡 LOW Issue

mergeStoreRepoLists reads HEIMDALLM_REPOSITORIES directly via csvEnv at merge time. This couples ApplyStore (otherwise a pure function over rows + receiver) to process environment, which can surprise tests/callers that set env after config load. Consider capturing the 'env set' bit on the Config during applyEnvOverrides instead, so ApplyStore stays deterministic w.r.t. its inputs.

**Location:** `daemon/internal/config/store.go` line 155

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid fix that correctly inverts repo-list precedence so explicit TOML/env entries win over runtime-discovered store rows. Logic is well-tested and the docs/handler updates are consistent. Two real concerns: ApplyStore now reads process env directly (breaking its previous purity) and HEIMDALLM_NON_MONITORED has no symmetric guard.

### Issues

⚠️ **daemon/internal/config/store.go:167** — mergeStoreRepoLists calls csvEnv("HEIMDALLM_REPOSITORIES") at merge time, making ApplyStore non-deterministic with respect to its arguments. ApplyStore was previously a pure function over (rows, receiver); now it depends on os.Getenv at call time, which can surprise reload paths and tests that mutate env after load. Capture the 'env-set' flag on Config during applyEnvOverrides (e.g., a private bool field) and consult that here instead.
⚠️ **daemon/internal/config/store.go:168** — Asymmetric guard: HEIMDALLM_REPOSITORIES short-circuits store-only monitored additions, but there is no equivalent for HEIMDALLM_NON_MONITORED (which exists in the env loader). If an operator pins non_monitored via env, stale store non_monitored rows still get appended. Either add a parallel guard or document explicitly that env non_monitored is additive while env repositories is authoritative.
🟡 **daemon/internal/config/store.go:88** — Subtle semantic change worth a release note: previously a store 'repositories' row fully replaced TOML repositories (last-write-wins). After this change it only adds store-only entries on top. Operators who deleted a repo from TOML expecting the daemon to drop it (while a stale store row still listed it) will see different effective behavior. The docs paragraph mentions the precedence but not the migration impact for existing SQLite rows from older daemons.
🟡 **daemon/internal/config/store.go:155** — Minor: when sawStoreRepos is false the storeRepos slice is nil and the second range loop is a no-op, but the early return on env-set still executes and is harmless dead work. Consider gating the env check + monitored-merge block on `sawStoreRepos` to keep intent obvious.
🟡 **daemon/internal/config/store_test.go:41** — Tests compare slices via fmt.Sprintf("%v", ...). reflect.DeepEqual is the idiomatic choice here and avoids accidental matches on stringification edge cases (e.g., slices containing values whose %v formatting collides). Low priority but worth aligning with the rest of the test suite.

### Suggestions

- Persist an envRepositoriesSet bool on Config during applyEnvOverrides and read it from mergeStoreRepoLists, restoring ApplyStore's purity and making the env-precedence rule discoverable from the struct.
- Mirror the HEIMDALLM_REPOSITORIES guard for HEIMDALLM_NON_MONITORED if env-provided non_monitored is meant to be authoritative; otherwise leave a one-line comment near the guard explaining the deliberate asymmetry.
- Add a brief CHANGELOG/release note that store 'repositories' rows from older daemons no longer override TOML — operators with stale rows may want to clean configs table during upgrade.
- Switch the new test assertions to reflect.DeepEqual for slice comparison consistency.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid fix that correctly inverts repo-list precedence so TOML/env entries win over runtime-discovery store rows. Logic is mostly correct with good test coverage, but there are a couple of medium concerns around coupling and an INVARIANT comment that the new append-based merge silently contradicts.

### Issues

⚠️ **daemon/internal/config/store.go:170** — mergeStoreRepoLists reads HEIMDALLM_REPOSITORIES directly from the environment via csvEnv(), duplicating knowledge of the env-var name that lives in applyEnvOverrides. If that var is ever renamed or wrapped, this check silently diverges. Consider adding a bool field to Config (e.g. envReposExplicit) set during applyEnvOverrides so the merge function doesn't need to re-interrogate the environment.
⚠️ **daemon/internal/config/store.go:62** — The INVARIANT comment (lines 54-61) states 'every case below *replaces the whole field* (slice/struct/string assignment) rather than mutating it in place'. The new mergeStoreRepoLists call uses append on shadow slices, which is not a whole-field replacement. The atomicity guarantee still holds (append extends the shadow's slice header without affecting the original's length), but the documented invariant is now inaccurate. Update the comment to reflect the new append-based merge pattern.
🟡 **daemon/internal/config/store.go:181** — The third guard in the storeRepos loop (storeNonMonitoredSet check) is unreachable. Any repo in storeNonMonitoredSet that was skipped from nonMonitoredSet was skipped because it's in repoSet — which is already caught by check 1. Any repo added to nonMonitoredSet is caught by check 2. The check is harmless but dead code; consider removing it or adding a comment explaining the defensive intent.
🟡 **daemon/internal/config/store_test.go:40** — Tests compare slices via fmt.Sprintf("%v") != fmt.Sprintf("%v"). slices.Equal (Go 1.21+) or reflect.DeepEqual would be more idiomatic, avoid string serialization overhead, and give clearer semantics. This pattern is used in all new and updated test functions.
🟡 **daemon/internal/config/store_test.go:1** — Missing test for the documented behavior 'a repo in both store lists is not monitored' (store.go comment at line 155). There's no test where the same repo appears in both storeRepos and storeNonMonitored without any TOML involvement, verifying it lands only in NonMonitored.

### Suggestions

- Add a Config-level flag (e.g. envReposExplicit bool) set during applyEnvOverrides, so mergeStoreRepoLists can check the flag instead of re-reading the environment. This removes the duplicated env-var name and makes the dependency explicit.
- Update the INVARIANT comment block on ApplyStore to note that mergeStoreRepoLists uses append (not whole-field replacement) on shadow slices, and explain why atomicity still holds under Go's slice-header semantics.
- Add a test case where the same repo appears in both store repositories and store non_monitored (no TOML entries) to verify the 'not monitored' contract documented in the comment.
- Consider using slices.Equal for test assertions instead of fmt.Sprintf comparison — it's clearer and avoids a serialization round-trip.

---

PR #378, opened 2026-04-28
fix: prevent duplicate SSE notification streams

### LINKED ISSUE #377: bug: OS notifications repeat many times for the same event
## Problema

Cuando llega una notificación de heimdallm (notificación del sistema operativo), el mismo evento se dispara repetidamente y satura el centro de notificaciones.

## Pasos para reproducir

1. Tener heimdallm en ejecución con notificaciones del SO habilitadas.
2. Esperar a que se produzca un evento que genere notificación (p. ej. un review terminado, un PR creado, etc.).
3. Observar cómo la misma notificación aparece varias veces seguidas en lugar de una sola vez.

## Comportamiento esperado

Una notificación por evento. Si el mismo evento se reemite (reconexiones SSE, retry, etc.) debería deduplicarse.

## Comportamiento observado

La misma notificación se repite varias veces.

## Notas

- Pendiente de capturar logs del daemon y del cliente cuando ocurra de nuevo, para identificar si el origen está en:
  - Reconexiones SSE que reentregan eventos ya procesados.
  - Múltiples consumidores activos en distintas pestañas/instancias del cliente.
  - Lógica de `notify` del daemon disparando varias veces por el mismo cambio de estado.
- Relacionado (cerrados, scope distinto): #243 (amplificación multi-instancia), #322 (spam tras re-review), #286 (SSE pool exhausted).

ivanmunozruiz:
## Summary

Fixes #377.

- Makes `SseClient.connect()` idempotent while a stream is already active.
- Tracks the active connection and pending reconnect timer so an error followed by `onDone` cannot open parallel `/events` streams.
- Adds regression coverage for duplicate `connect()` calls and the error-then-close reconnect path.

## Why

The OS notification spam can happen when the Flutter SSE client accumulates multiple active `/events` connections. Each connection receives the same server fan-out event and pushes it into the same client stream, so a single review event can trigger repeated notifications.

## Validation

- `flutter test test/core/sse_client_test.dart`
- `flutter analyze`
- `make build-web`

Note: `flutter_app/pubspec.lock` had a pre-existing local change and is intentionally not included in this PR.

### DIFF
--- a/flutter_app/lib/core/api/sse_client.dart
+++ b/flutter_app/lib/core/api/sse_client.dart
@@ -9,19 +9,29 @@ class SseEvent {
   const SseEvent({required this.type, required this.data});
 }
 
+/// SSE client that maintains at most one active HTTP stream and one pending
+/// reconnect timer per instance.
 class SseClient {
   final String path;
   final http.Client _httpClient;
   final PlatformServices _platform;
+  final Duration _errorReconnectDelay;
+  final Duration _doneReconnectDelay;
   StreamController<SseEvent>? _controller;
   StreamSubscription<String>? _subscription;
+  Timer? _reconnectTimer;
+  bool _connecting = false;
 
   SseClient({
     required PlatformServices platform,
     this.path = '/events',
     http.Client? httpClient,
-  })  : _platform = platform,
-        _httpClient = httpClient ?? http.Client();
+    Duration errorReconnectDelay = const Duration(seconds: 5),
+    Duration doneReconnectDelay = const Duration(seconds: 3),
+  }) : _platform = platform,
+       _httpClient = httpClient ?? http.Client(),
+       _errorReconnectDelay = errorReconnectDelay,
+       _doneReconnectDelay = doneReconnectDelay;
 
   /// Parses SSE wire format into events. Static for testability.
   static List<SseEvent> parseEvents(String raw) {
@@ -47,6 +57,10 @@ class SseClient {
 
   /// Returns a stream of SSE events from the daemon.
   Stream<SseEvent> connect() {
+    final existing = _controller;
+    if (existing != null && !existing.isClosed) {
+      return existing.stream;
+    }
     _controller = StreamController<SseEvent>.broadcast(
       onCancel: () => disconnect(),
     );
@@ -57,6 +71,16 @@ class SseClient {
   static const _maxBufferSize = 1024 * 1024; // 1 MB
 
   void _startListening() async {
+    final controller = _controller;
+    if (controller == null ||
+        controller.isClosed ||
+        _connecting ||
+        _subscription != null) {
+      return;
+    }
+    _reconnectTimer?.cancel();
+    _reconnectTimer = null;
+    _connecting = true;
     try {
       final request = http.Request(
         'GET',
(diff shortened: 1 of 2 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Solid fix for duplicate SSE streams: adds reuse of active controller, single-flight connect via `_connecting` and `_subscription` guards, and a coalesced reconnect timer with injectable delays. Tests cover dedup and the error+done double-reconnect race. No blocking issues found.

### Suggestions

- Consider exponential backoff (with jitter and a cap) for repeated reconnect failures instead of fixed 3s/5s delays, to avoid hammering the daemon during prolonged outages.
- On the buffer-overflow path, `_scheduleReconnect(Duration.zero)` will immediately retry against the same likely-misbehaving stream; a small delay or counter could prevent a tight reconnect loop.
- `_startListening` swallows the catch-block exception via `_controller?.addError(e)` but never reschedules a reconnect on initial-request failure (e.g. transient DNS/network error before `response`); callers see an error but no retry, unlike the post-connect path.
- Minor: `_connecting` is a plain bool mutated across async gaps; safe in Dart's single-threaded isolate model but a brief comment documenting the invariant would help future readers.

---

sergiotejon:
## 🟡 LOW Issue

After `await _httpClient.send(request)`, if the controller was closed during the await, the response stream is abandoned without being drained or cancelled. This can leak the underlying socket/connection. Consider listening and immediately cancelling, or ensuring the response body is discarded.

**Location:** `flutter_app/lib/core/api/sse_client.dart` line 94

---

sergiotejon:
## ⚠️ MEDIUM Issue

In the catch block, `_controller?.addError(e)` propagates the error to consumers but does not schedule a reconnect. Previously, transient send() failures would also not reconnect, but now with the explicit reconnect machinery it would be more consistent to schedule a reconnect here as well so a transient network failure during initial connect doesn't permanently surface an error to subscribers.

**Location:** `flutter_app/lib/core/api/sse_client.dart` line 132

---

sergiotejon:
## ⚠️ MEDIUM Issue

`connect()` reuses an existing non-closed controller, but does not call `_startListening()` if there's no active subscription/connecting attempt. If a previous `_startListening` failed in the catch block (which only addError's), subsequent `connect()` calls will return a stream that never produces events. Consider calling `_startListening()` defensively when reusing the controller.

**Location:** `flutter_app/lib/core/api/sse_client.dart` line 55

---

sergiotejon:
## 🟡 LOW Issue

No exponential backoff or max retry cap on reconnects. Under sustained server outage, the client will reconnect every 5 seconds indefinitely, which can flood logs and the network. Consider adding backoff with jitter.

**Location:** `flutter_app/lib/core/api/sse_client.dart` line 153

---

sergiotejon:
## 🟡 LOW Issue

The reconnect test asserts `requests == 2` after both an error and onDone fire. This validates only-one-reconnect behavior, but it relies on a 50ms wall-clock delay which can be flaky on slow CI. Consider using `fakeAsync` from package:fake_async for deterministic timing.

**Location:** `flutter_app/test/core/sse_client_test.dart` line 175

---

sergiotejon (approved):

Solid fix for duplicate SSE notification streams. The PR correctly introduces guards against concurrent connections via `_connecting` flag, `_subscription` null-check, and a centralized `_scheduleReconnect` that prevents overlapping timers. Reusing the existing controller in `connect()` avoids opening duplicate streams when multiple consumers subscribe. Configurable reconnect delays improve testability, and new tests cover both the deduplication and the error+done race condition. Code is clean and the disconnect path properly cancels timers and resets state.

### Suggestions

- Add exponential backoff with jitter to `_scheduleReconnect` to avoid hammering the server during outages.
- Schedule a reconnect (instead of only addError) in the outer try/catch so initial connection failures recover automatically.
- Consider invoking `_startListening()` from `connect()` when reusing an existing controller to recover from prior failed attempts.
- Use `package:fake_async` for the timing-sensitive reconnect tests to reduce CI flakiness.
- Document the reconnect behavior (delays, no-backoff, dedup semantics) on the `SseClient` class doc comment.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid fix for duplicate SSE streams. The approach—reusing an open controller, guarding with a _connecting flag, and replacing uncancellable Future.delayed with a Timer—is correct and well-tested. Two pre-existing concerns raised by the AI reviewer are valid and worth addressing, plus one additional issue.

### Issues

⚠️ **flutter_app/lib/core/api/sse_client.dart:94** — After `await _httpClient.send(request)`, if the controller was closed during the await, the method returns without draining or cancelling the response stream. The underlying HTTP socket/connection will leak. Drain or cancel the response body before returning (e.g. `response.stream.drain().catchError((_) {});`).
⚠️ **flutter_app/lib/core/api/sse_client.dart:132** — The catch block sets `_connecting = false` and propagates the error but never calls `_scheduleReconnect`. A transient failure during `_httpClient.send()` (DNS timeout, socket error) will leave the client permanently disconnected with no automatic recovery, which is inconsistent with the onError/onDone reconnect behavior. Add `_scheduleReconnect(_errorReconnectDelay)` after `addError`.
🟡 **flutter_app/lib/core/api/sse_client.dart:107** — Buffer-overflow reconnect uses `Duration.zero`, which means an immediate retry. If a malicious or malformed server keeps exceeding the buffer, this creates a tight infinite reconnect loop with no backoff. Consider using `_errorReconnectDelay` (or a dedicated delay) instead of `Duration.zero`.
🟡 **flutter_app/lib/core/api/sse_client.dart:113** — The `onError` handler cancels `_subscription` explicitly, but `cancelOnError` defaults to `false` so after the cancel call, the `onDone` callback may still fire (implementation-dependent), which could schedule a second reconnect. Since `_scheduleReconnect` guards against an already-active timer this is likely safe, but nulling `_subscription` before cancel would be more defensive. Verify this doesn't cause a double-reconnect in practice.

### Suggestions

- In the catch block (line ~132), add `_scheduleReconnect(_errorReconnectDelay);` so transient send() failures trigger automatic recovery, matching the onError/onDone behavior.
- After the early-return on line 94 (controller closed during await), drain the response stream to avoid leaking the HTTP connection: `response.stream.drain<void>().catchError((_) {});`.
- Replace `Duration.zero` on the buffer-overflow reconnect (line 107) with `_errorReconnectDelay` to prevent a tight retry loop against a malicious stream.
- Consider adding a test for the catch-block reconnect path (e.g. mock client that throws on send) to verify recovery from transient connect failures.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid fix for the duplicate-stream bug with good guards (_connecting flag, single _reconnectTimer, controller reuse) and meaningful test coverage. A few edge cases remain around the initial-connect failure path and resource cleanup if the controller closes mid-await.

### Issues

🟡 **flutter_app/lib/core/api/sse_client.dart:94** — After `await _httpClient.send(request)`, if the controller was closed during the await the response stream is returned but never listened to or cancelled, leaking the underlying socket. Drain or cancel the response stream when bailing out.
⚠️ **flutter_app/lib/core/api/sse_client.dart:132** — The catch block calls `_controller?.addError(e)` but does not schedule a reconnect. Now that explicit reconnect machinery exists, a transient failure during the initial `send()` permanently surfaces an error to subscribers instead of recovering. Schedule a reconnect here for consistency with onError/onDone.
⚠️ **flutter_app/lib/core/api/sse_client.dart:55** — `connect()` reuses an existing non-closed controller without verifying a listener/timer is active. If a prior `_startListening()` threw in the catch block (which only addErrors and does not reschedule), later `connect()` calls return a live controller that will never emit events. Defensively call `_startListening()` (which is already idempotent via its own guards) when reusing the controller.
🟡 **flutter_app/lib/core/api/sse_client.dart:153** — Reconnect uses fixed delays with no exponential backoff or jitter. Under a sustained outage the client will hammer the server every 3-5s indefinitely.
🟡 **flutter_app/test/core/sse_client_test.dart:175** — The 'only one reconnect' test depends on real `Future.delayed` timing (10ms reconnect + 50ms wait), which is prone to flakiness on slow CI. Consider `package:fake_async` for deterministic scheduling.

### Suggestions

- In the outer catch, mirror the onError path: cancel any partial subscription, then `_scheduleReconnect(_errorReconnectDelay)` instead of (or in addition to) `addError`, so transient initial-connect failures recover automatically.
- When `_controller` is closed/null after the awaited `send()` returns, call `response.stream.listen((_) {}).cancel()` (or similar) to release the HTTP socket rather than dropping the response on the floor.
- Add exponential backoff with jitter and a sensible cap to `_scheduleReconnect`, resetting the backoff on a successful read.
- Convert the timing-sensitive reconnect tests to `fakeAsync` to eliminate wall-clock flakiness.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review: previous medium issues (drain on disconnect-during-connect, reconnect on catch) are addressed in the new diff. The connect() defensive issue is effectively mitigated because the catch block now calls _scheduleReconnect. Remaining items are low-severity suggestions (backoff, fakeAsync) that don't block merge.

### Suggestions

- Consider adding exponential backoff with jitter and a max retry cap to avoid tight reconnect loops during sustained outages.
- Consider using package:fake_async to make timing-sensitive reconnect tests deterministic on slow CI.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All four previous findings have been addressed. The response stream is now drained when disconnected mid-connect, the catch block schedules a reconnect, buffer overflow uses _errorReconnectDelay instead of Duration.zero, and _cancelSubscription nulls _subscription before cancelling to prevent double-reconnect from onDone. New test coverage is thorough and exercises each fix. No new issues found.

---

PR #380, opened 2026-04-28
fix: avoid duplicate review publish retries

ivanmunozruiz:
## Summary

- Defer publish retry enqueues for freshly-created unpublished reviews so `PublishPending` cannot race the pipeline's initial GitHub publish.
- Make the publish worker skip too-fresh review rows; if the row is still unpublished later, `PublishPending` will enqueue it again.
- Add cmd-layer coverage for the retry-age guard and pending publish filtering.

## Root Cause

PR `freepik-company/ai-platform-iqs-v3#437` was reviewed twice on the same head SHA by the same account one second apart. Local logs show a single local review row (`review_id=310`) was saved, then the pipeline published it directly while the publish retry path also picked up the same `github_review_id=0` row before `MarkReviewPublished` landed.

This is not two local daemons; it is a race between initial publish and retry publish.

## Validation

- `docker run --rm --user "$(id -u):$(id -g)" -v "$(pwd):/src:ro" -w /src/daemon -e HOME=/tmp/home -e GOCACHE=/tmp/go-build -e GOMODCACHE=/tmp/gomod golang:1.25-alpine@sha256:f6751d823c26342f9506c03797d2527668d095b0a15f1862cddb4d927a7a4ced sh -c 'go vet ./... && go test -count=1 ./cmd/heimdallm'`

Note: `make test-docker GO_TEST_ARGS="-run 'TestReviewReadyForPublishRetry|TestTier2AdapterPublishPendingDefersFreshReviews' ./cmd/heimdallm"` currently fails before tests because the persistent `/tmp/heimdallm-gocache/gomod` mount does not resolve base modules (`github.com/BurntSushi/toml`, `modernc.org/sqlite`, `github.com/go-chi/chi/v5`). The same pinned image passes with an in-container module cache.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -38,7 +38,9 @@ import (
 	"github.com/nats-io/nats.go"
 )
 
-// Let the initial pipeline publish finish before retry workers consider the row.
+// reviewPublishRetryMinAge only gates retry paths. The pipeline still submits
+// immediately after InsertReview; after this window, PublishPending will
+// re-enqueue rows that remain unpublished.
 const reviewPublishRetryMinAge = 2 * time.Minute
 
 func main() {
@@ -684,9 +686,9 @@ func main() {
 			return nil // idempotent — ack
 		}
 		if !reviewReadyForPublishRetry(rev, time.Now().UTC()) {
-			slog.Info("publish-worker: review still in initial publish window, skipping",
+			slog.Debug("publish-worker: review still in initial publish window, skipping",
 				"review_id", msg.ReviewID, "created_at", rev.CreatedAt)
-			return nil // PublishPending will re-enqueue if it remains unpublished.
+			return nil // Ack early messages; PublishPending re-enqueues aged unpublished rows.
 		}
 
 		pr, err := s.GetPR(rev.PRID)
@@ -2256,9 +2258,6 @@ func (a *tier2Adapter) PublishPending() {
 }
 
 func (a *tier2Adapter) publishPending(now time.Time) {
-	if a.publishPub == nil {
-		return
-	}
 	reviews, err := a.store.ListUnpublishedReviews()
 	if err != nil || len(reviews) == 0 {
 		return
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -38,6 +38,9 @@ import (
 	"github.com/nats-io/nats.go"
 )
 
+// Let the initial pipeline publish finish before retry workers consider the row.
+const reviewPublishRetryMinAge = 2 * time.Minute
+
 func main() {
 	if len(os.Args) > 1 {
 		switch os.Args[1] {
@@ -679,6 +682,11 @@ func main() {
 			slog.Info("publish-worker: already published, skipping",
 				"review_id", msg.ReviewID, "github_review_id", rev.GitHubReviewID)
 			return nil // idempotent — ack
+		}
+		if !reviewReadyForPublishRetry(rev, time.Now().UTC()) {
+			slog.Info("publish-worker: review still in initial publish window, skipping",
+				"review_id", msg.ReviewID, "created_at", rev.CreatedAt)
+			return nil // PublishPending will re-enqueue if it remains unpublished.
 		}
 
 		pr, err := s.GetPR(rev.PRID)
@@ -2244,15 +2252,35 @@ func (a *tier2Adapter) ProcessPR(ctx context.Context, pr scheduler.Tier2PR) erro
 
 // PublishPending implements scheduler.Tier2PRProcessor.
 func (a *tier2Adapter) PublishPending() {
+	a.publishPending(time.Now().UTC())
+}
+
+func (a *tier2Adapter) publishPending(now time.Time) {
+	if a.publishPub == nil {
+		return
+	}
 	reviews, err := a.store.ListUnpublishedReviews()
 	if err != nil || len(reviews) == 0 {
 		return
 	}
 	for _, rev := range reviews {
+		if !reviewReadyForPublishRetry(rev, now) {
+			continue
+		}
 		if err := a.publishPub.PublishPRPublish(context.Background(), rev.ID); err != nil {
 			slog.Warn("publish-pending: enqueue failed", "review_id", rev.ID, "err", err)
 		}
 	}
+}
+
+func reviewReadyForPublishRetry(rev *store.Review, now time.Time) bool {
+	if rev == nil || rev.GitHubReviewID != 0 {
+		return false
+	}
+	if rev.CreatedAt.IsZero() {
+		return true
+	}
+	return !now.Before(rev.CreatedAt.Add(reviewPublishRetryMinAge))
 }
 
 // ProcessRepo implements scheduler.Tier2IssueProcessor.
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -38,11 +38,6 @@ import (
 	"github.com/nats-io/nats.go"
 )
 
-// reviewPublishRetryMinAge only gates retry paths. The pipeline still submits
-// immediately after InsertReview; after this window, PublishPending will
-// re-enqueue rows that remain unpublished.
-const reviewPublishRetryMinAge = 2 * time.Minute
-
 func main() {
 	if len(os.Args) > 1 {
 		switch os.Args[1] {
@@ -685,11 +680,6 @@ func main() {
 				"review_id", msg.ReviewID, "github_review_id", rev.GitHubReviewID)
 			return nil // idempotent — ack
 		}
-		if !reviewReadyForPublishRetry(rev, time.Now().UTC()) {
-			slog.Debug("publish-worker: review still in initial publish window, skipping",
-				"review_id", msg.ReviewID, "created_at", rev.CreatedAt)
-			return nil // Ack early messages; PublishPending re-enqueues aged unpublished rows.
-		}
 
 		pr, err := s.GetPR(rev.PRID)
 		if err != nil {
@@ -703,6 +693,15 @@ func main() {
 				"review_id", msg.ReviewID)
 			_ = s.MarkReviewPublished(rev.ID, -1, "", time.Now().UTC())
 			return nil // permanent — ack
+		}
+		inFlight, err := s.ReviewInFlight(pr.GithubID, rev.HeadSHA)
+		if err != nil {
+			return fmt.Errorf("check in-flight review: %w", err)
+		}
+		if inFlight {
+			slog.Debug("publish-worker: initial publish still in flight, skipping retry",
+				"review_id", msg.ReviewID, "github_id", pr.GithubID, "head_sha", rev.HeadSHA)
+			return nil // PublishPending re-enqueues if the row remains unpublished after release.
 		}
 
 		// Rebuild ReviewResult from stored JSON
@@ -2254,16 +2253,21 @@ func (a *tier2Adapter) ProcessPR(ctx context.Context, pr scheduler.Tier2PR) erro
 
 // PublishPending implements scheduler.Tier2PRProcessor.
 func (a *tier2Adapter) PublishPending() {
-	a.publishPending(time.Now().UTC())
+	a.publishPending()
 }
 
-func (a *tier2Adapter) publishPending(now time.Time) {
+func (a *tier2Adapter) publishPending() {
 	reviews, err := a.store.ListUnpublishedReviews()
 	if err != nil || len(reviews) == 0 {
 		return
 	}
 	for _, rev := range reviews {
-		if !reviewReadyForPublishRetry(rev, now) {
+		ready, err := a.reviewReadyForPublishRetry(rev)
+		if err != nil {
+			slog.Warn("publish-pending: in-flight check failed", "review_id", rev.ID, "err", err)
+			continue
+		}
+		if !ready {
 			continue
 		}
 		if err := a.publishPub.PublishPRPublish(context.Background(), rev.ID); err != nil {
@@ -2272,14 +2276,19 @@ func (a *tier2Adapter) publishPending(now time.Time) {
 	}
 }
 
-func reviewReadyForPublishRetry(rev *store.Review, now time.Time) bool {
+func (a *tier2Adapter) reviewReadyForPublishRetry(rev *store.Review) (bool, error) {
 	if rev == nil || rev.GitHubReviewID != 0 {
-		return false
+		return false, nil
 	}
-	if rev.CreatedAt.IsZero() {
-		return true
+	pr, err := a.store.GetPR(rev.PRID)
+	if err != nil {
+		return false, err
 	}
-	return !now.Before(rev.CreatedAt.Add(reviewPublishRetryMinAge))
+	inFlight, err := a.store.ReviewInFlight(pr.GithubID, rev.HeadSHA)
+	if err != nil {
+		return false, err
+	}
+	return !inFlight, nil
 }
 
 // ProcessRepo implements scheduler.Tier2IssueProcessor.
--- a/daemon/cmd/heimdallm/main_test.go
+++ b/daemon/cmd/heimdallm/main_test.go
@@ -115,7 +115,7 @@ func TestTier2AdapterPublishPendingDefersFreshReviews(t *testing.T) {
 	s := newMemStore(t)
 	now := time.Date(2026, 4, 28, 12, 0, 0, 0, time.UTC)
 	oldReviewID := seedPRWithReview(t, s, 101, now.Add(-reviewPublishRetryMinAge-time.Second))
-	freshReviewID := seedPRWithReview(t, s, 102, now.Add(-reviewPublishRetryMinAge+time.Second))
+	seedPRWithReview(t, s, 102, now.Add(-reviewPublishRetryMinAge+time.Second))
 
 	conn := newInProcessNATS(t)
 	ch := make(chan *nats.Msg, 2)
@@ -150,13 +150,12 @@ func TestTier2AdapterPublishPendingDefersFreshReviews(t *testing.T) {
 		t.Fatal("old pending review was not enqueued")
 	}
 
+	// publishPending publishes synchronously; the short wait catches stray
+	// buffered messages without making this negative assertion expensive.
 	select {
 	case msg := <-ch:
 		var got bus.PRPublishMsg
 		_ = bus.Decode(msg.Data, &got)
-		if got.ReviewID == freshReviewID {
-			t.Fatalf("fresh review %d was enqueued during initial publish window", freshReviewID)
-		}
 		t.Fatalf("unexpected extra publish message: %+v", got)
 	case <-time.After(150 * time.Millisecond):
 	}
--- a/daemon/cmd/heimdallm/main_test.go
+++ b/daemon/cmd/heimdallm/main_test.go
@@ -94,28 +94,68 @@ func seedPRWithReview(t *testing.T, s *store.Store, githubID int64, createdAt ti
 	return revID
 }
 
-func TestReviewReadyForPublishRetry(t *testing.T) {
+func TestTier2AdapterReviewReadyForPublishRetry(t *testing.T) {
+	s := newMemStore(t)
 	now := time.Date(2026, 4, 28, 12, 0, 0, 0, time.UTC)
+	readyID := seedPRWithReview(t, s, 101, now)
+	inFlightID := seedPRWithReview(t, s, 102, now)
+	if claimed, err := s.ClaimInFlightReview(102, "abc123"); err != nil {
+		t.Fatalf("claim in-flight review: %v", err)
+	} else if !claimed {
+		t.Fatal("expected in-flight claim to succeed")
+	}
+	a := &tier2Adapter{store: s}
 
-	if reviewReadyForPublishRetry(&store.Review{GitHubReviewID: 123, CreatedAt: now.Add(-time.Hour)}, now) {
+	readyRev, err := s.GetReview(readyID)
+	if err != nil {
+		t.Fatalf("get ready review: %v", err)
+	}
+	ready, err := a.reviewReadyForPublishRetry(readyRev)
+	if err != nil {
+		t.Fatalf("reviewReadyForPublishRetry ready: %v", err)
+	}
+	if !ready {
+		t.Fatal("unpublished review with no in-flight claim should be ready")
+	}
+
+	inFlightRev, err := s.GetReview(inFlightID)
+	if err != nil {
+		t.Fatalf("get in-flight review: %v", err)
+	}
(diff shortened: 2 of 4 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Targeted fix that introduces a 2-minute grace window before retry workers re-enqueue unpublished reviews, preventing duplicate publish attempts. Logic is small, well-tested with both unit and in-process NATS integration tests, and handles the legacy zero CreatedAt case. No correctness, security, or architectural concerns found.

### Suggestions

- Consider making reviewPublishRetryMinAge configurable (env var or config) so operators can tune the window without recompiling.
- The second select in TestTier2AdapterPublishPendingDefersFreshReviews waits 150ms to assert no extra message arrives; this is a small timing-based heuristic that could occasionally be flaky under heavy CI load — consider increasing or documenting the rationale.
- Minor: the comment 'PublishPending will re-enqueue if it remains unpublished' is slightly misleading since PublishPending now also defers fresh reviews; clarify that retry will occur once the row ages past the window.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid, well-scoped fix. Adds a 2-minute grace window before retry workers (both the publish-worker handler and PublishPending sweep) reconsider an unpublished review, preventing duplicate publish attempts racing the initial pipeline. Helper is pure and well-tested, including boundary and legacy-zero-CreatedAt cases. Tests use an in-process NATS server which is a nice pattern.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:41** — The 2-minute window is a hard-coded magic constant. If the initial publish path can legitimately take longer (slow GitHub API, backoff, queue lag), retries are still suppressed beyond 2m only because PublishPending re-enqueues — but the publish-worker handler will keep ack'ing within the window without action. Consider making this configurable or at least documenting the upper bound of the initial publish path that justifies 2m.
🟡 **daemon/cmd/heimdallm/main.go:686** — The publish-worker handler returns nil (ack) when inside the initial publish window, relying on PublishPending to re-enqueue later. If PublishPending is disabled, misconfigured, or its ticker is paused, an unpublished review whose first publish attempt failed could be silently dropped for 2 minutes with no immediate retry. Worth a brief comment confirming PublishPending is always running, or alternatively NAK with delay rather than ack.
🟡 **daemon/cmd/heimdallm/main.go:2257** — The added `if a.publishPub == nil { return }` guard is new behavior not present before the refactor. It's defensive and harmless, but worth confirming this state is actually reachable in production — if not, the guard is dead code that masks a misconfiguration that should panic at startup instead.
🟡 **daemon/cmd/heimdallm/main_test.go:145** — TestTier2AdapterPublishPendingDefersFreshReviews relies on a 150ms timeout to assert no second message arrives. Under a loaded CI runner this could either flake (false positive on slow scheduling — though here absence of message is the pass condition, so it's safe) or, conversely, allow a slowly-arriving unwanted message to slip past. The pass-on-timeout direction is fine; just note that increasing the timeout adds confidence at the cost of test runtime.

### Suggestions

- Extract reviewPublishRetryMinAge to config or at least a documented tunable, since the right value depends on observed publish latency.
- Consider logging at debug rather than info in the publish-worker skip branch — it will fire frequently in steady state and could be noisy.
- In TestTier2AdapterPublishPendingDefersFreshReviews, the second select branch's fallthrough t.Fatalf is reached for any unexpected message, but the prior `if got.ReviewID == freshReviewID` check is redundant with the generic fatal — simplify to a single t.Fatalf describing the unexpected publish.
- The unrelated whitespace/alignment change in TestResolveImplementPrompt_DefaultFallbackWhenAgentMissing should ideally be in a separate commit to keep the fix's diff focused.

---

sergiotejon:
## ⚠️ MEDIUM Issue

The publish-worker skip path returns nil (ack) and relies on PublishPending to re-enqueue. If the periodic PublishPending scheduler is disabled, misconfigured, or stops running, fresh reviews skipped here will never be retried. Consider NAKing with a delay (or scheduling a delayed redelivery) instead of silently acking, so retry doesn't depend solely on the polling loop.

**Location:** `daemon/cmd/heimdallm/main.go` line 686

---

sergiotejon:
## 🟡 LOW Issue

`reviewPublishRetryMinAge` is a hardcoded 2-minute constant. The initial pipeline publish duration may vary across environments (slow CI, large diffs, GitHub API latency). Consider making this configurable via env/config so operators can tune it without recompiling.

**Location:** `daemon/cmd/heimdallm/main.go` line 2275

---

sergiotejon:
## 🟡 LOW Issue

Treating `CreatedAt.IsZero()` as immediately retryable is reasonable for legacy rows, but worth a brief comment in code (not just the test) explaining the intent, since future readers may assume zero-time should be treated conservatively.

**Location:** `daemon/cmd/heimdallm/main.go` line 2270

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Adds a 2-minute initial publish window to prevent the publish-pending retry worker and publish worker from racing the initial pipeline publish for a review. Logic is well-isolated in `reviewReadyForPublishRetry`, with solid unit and integration-style tests using an in-process NATS server.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:686** — The publish-worker skip path returns nil (ack) and relies on PublishPending to re-enqueue. If the periodic PublishPending scheduler is disabled, misconfigured, or stops running, fresh reviews skipped here will never be retried. Consider NAKing with a delay (or scheduling a delayed redelivery) instead of silently acking, so retry doesn't depend solely on the polling loop.
🟡 **daemon/cmd/heimdallm/main.go:2275** — `reviewPublishRetryMinAge` is a hardcoded 2-minute constant. The initial pipeline publish duration may vary across environments (slow CI, large diffs, GitHub API latency). Consider making this configurable via env/config so operators can tune it without recompiling.
🟡 **daemon/cmd/heimdallm/main.go:2270** — Treating `CreatedAt.IsZero()` as immediately retryable is reasonable for legacy rows, but worth a brief comment in code (not just the test) explaining the intent, since future readers may assume zero-time should be treated conservatively.
🟡 **daemon/cmd/heimdallm/main_test.go:162** — The 'unexpected extra publish message' assertion uses a 150ms timeout, which can be flaky under loaded CI runners. Consider a longer timeout or a deterministic synchronization point (e.g., publish a sentinel message after `publishPending` and assert the fresh review's message never arrives before the sentinel).

---

sergiotejon:
## 🟡 LOW Issue

The 'unexpected extra publish message' assertion uses a 150ms timeout, which can be flaky under loaded CI runners. Consider a longer timeout or a deterministic synchronization point (e.g., publish a sentinel message after `publishPending` and assert the fresh review's message never arrives before the sentinel).

**Location:** `daemon/cmd/heimdallm/main_test.go` line 162

---

sergiotejon (approved):

Adds a 2-minute initial publish window to prevent the publish-pending retry worker and publish worker from racing the initial pipeline publish for a review. Logic is well-isolated in `reviewReadyForPublishRetry`, with solid unit and integration-style tests using an in-process NATS server.

### Suggestions

- Add a test covering the publish-worker handler skip path (review still in initial publish window) to mirror the PublishPending coverage.
- Consider logging at debug rather than info in the publish-worker skip branch to avoid log noise when many fresh reviews flow through.
- Document the interaction between `reviewPublishRetryMinAge` and the PublishPending scheduler interval so operators understand the effective retry latency floor.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean fix to prevent duplicate review publish retries by gating on a 2-minute age window. Logic is correct, tests are thorough (unit + integration with in-process NATS), and the time-injectable publishPending design enables deterministic testing. A few minor observability and robustness notes.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:688** — The publish worker silently acks (return nil) messages for fresh reviews, relying entirely on the periodic PublishPending sweep to re-enqueue them. If the sweep is delayed or the scheduler stalls, these reviews sit in limbo beyond the 2-minute window with no alert. Consider using a NATS NAK with delay instead, so the message stays in the queue and is redelivered after the retry window, rather than depending on a separate sweep to rediscover it.
🟡 **daemon/cmd/heimdallm/main.go:2262** — The nil publishPub guard returns silently with no log line. If the adapter is misconfigured (publishPub never set), PublishPending becomes a permanent silent no-op. A single slog.Warn on the first occurrence would make this much easier to diagnose in production.
🟡 **daemon/cmd/heimdallm/main.go:2275** — reviewReadyForPublishRetry checks rev.GitHubReviewID != 0, but in the publishPending path the input comes from ListUnpublishedReviews which already filters WHERE github_review_id=0. The check is harmless (defense-in-depth) but may mislead future readers into thinking published reviews can appear in that list. A brief comment clarifying this is intentional for the publish-worker caller would help.

### Suggestions

- Consider using NATS NAK-with-delay (or requeue with a backoff) in the publish worker instead of acking fresh reviews and relying on the sweep, so that the retry is self-contained within the message lifecycle.
- Add a sync.Once-guarded warning log when publishPub is nil inside publishPending, to surface misconfiguration without spamming logs.
- The diff appears partial — the bus package (SubjPRPublish, PRPublishMsg, Decode, NewPRPublishPublisher) and the publishPub struct field addition are not included. Ensure all new types and the struct change are reviewed as part of this PR.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

No code changes since the last review round. The core logic is sound and tests are well-structured. The previously flagged ack-vs-NAK concern in the publish worker remains the most notable unaddressed item; the other low-severity items from the prior round are minor and acceptable as-is.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:690** — Still unaddressed from prior round (flagged by both reviewers): the publish-worker acks fresh reviews and relies entirely on the periodic PublishPending sweep to retry them. A NATS NAK with delay would keep retry responsibility in the message queue rather than coupling it to the sweep scheduler's availability.

### Suggestions

- Address the ack-vs-NAK feedback or add a comment explaining why silent ack plus sweep-based retry is the preferred design (e.g., backpressure characteristics, simpler retry semantics).
- The 150ms negative-assertion timeout in TestTier2AdapterPublishPendingDefersFreshReviews is still potentially flaky on slow CI; the sentinel-message approach suggested in prior discussion would make it deterministic.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Adds a 2-minute retry guard to avoid duplicate publish attempts; logic is correct, well-tested, and the prior review feedback is already captured as open suggestions.

### Suggestions

- Consider NAKing with delay instead of acking skipped early messages so retries don't depend solely on PublishPending polling
- Make reviewPublishRetryMinAge configurable via env/config for environments with variable pipeline latency
- Add an inline comment explaining why CreatedAt.IsZero() is treated as immediately retryable (legacy rows)
- Replace the 150ms negative-assertion timeout with a deterministic sentinel to reduce CI flakiness

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Re-review: the diff is unchanged from the previous review. None of the prior findings have been addressed in code — the ack-on-skip behavior, hardcoded retry constant, missing zero-time comment, and tight 150ms negative-assertion timeout all remain as previously flagged.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:688** — Publish-worker still acks (returns nil) when the review is in the initial publish window, making retry entirely dependent on the PublishPending poller. If that scheduler is disabled or stalls, skipped reviews will never be republished. Prefer a NAK-with-delay or delayed redelivery so retry isn't solely poll-driven.
🟡 **daemon/cmd/heimdallm/main.go:42** — reviewPublishRetryMinAge remains a hardcoded 2-minute constant. Pipeline publish latency varies by environment (CI load, diff size, GitHub API); make this configurable via env/config to avoid recompiles for tuning.
🟡 **daemon/cmd/heimdallm/main.go:2279** — The CreatedAt.IsZero() branch treating legacy rows as immediately retryable still lacks an inline comment explaining intent. Add a short doc comment so future readers don't mistake it for a conservative default.
🟡 **daemon/cmd/heimdallm/main_test.go:160** — The negative assertion still uses a 150ms timeout, which can be flaky on loaded CI runners. Increase the window or use a deterministic sentinel-message synchronization to make the 'no extra publish' assertion robust.

---

sergiotejon:
## ⚠️ MEDIUM Issue

Publish-worker still acks (returns nil) when the review is in the initial publish window, making retry entirely dependent on the PublishPending poller. If that scheduler is disabled or stalls, skipped reviews will never be republished. Prefer a NAK-with-delay or delayed redelivery so retry isn't solely poll-driven.

**Location:** `daemon/cmd/heimdallm/main.go` line 688

---

sergiotejon:
## 🟡 LOW Issue

reviewPublishRetryMinAge remains a hardcoded 2-minute constant. Pipeline publish latency varies by environment (CI load, diff size, GitHub API); make this configurable via env/config to avoid recompiles for tuning.

**Location:** `daemon/cmd/heimdallm/main.go` line 42

---

sergiotejon:
## 🟡 LOW Issue

The CreatedAt.IsZero() branch treating legacy rows as immediately retryable still lacks an inline comment explaining intent. Add a short doc comment so future readers don't mistake it for a conservative default.

**Location:** `daemon/cmd/heimdallm/main.go` line 2279

---

sergiotejon:
## 🟡 LOW Issue

The negative assertion still uses a 150ms timeout, which can be flaky on loaded CI runners. Increase the window or use a deterministic sentinel-message synchronization to make the 'no extra publish' assertion robust.

**Location:** `daemon/cmd/heimdallm/main_test.go` line 160

---

sergiotejon (approved):

Re-review: the diff is unchanged from the previous review. None of the prior findings have been addressed in code — the ack-on-skip behavior, hardcoded retry constant, missing zero-time comment, and tight 150ms negative-assertion timeout all remain as previously flagged.

### Suggestions

- Replace the early-ack in the publish worker with a NAK + delay (or JetStream AckWait/redelivery) so retries don't rely solely on PublishPending polling.
- Expose reviewPublishRetryMinAge via configuration (env var or config struct) with the current 2m as default.
- Add an inline comment on the CreatedAt.IsZero() branch documenting that legacy/zero-time rows are intentionally treated as retryable.
- Harden the negative test assertion with a sentinel publish or a longer timeout to reduce CI flakiness.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of duplicate review publish retry guard. The diff is unchanged from the prior review — same constant, same ack-on-skip behavior, same 150ms negative-assertion timeout, and still no inline comment on the IsZero() legacy branch. All previously raised LOW findings remain applicable; the bot's MEDIUM concern about silent-ack-plus-poller-dependency is also still valid as a design tradeoff worth confirming.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:688** — Publish-worker acks (returns nil) when the review is inside the initial publish window, making retry entirely dependent on the PublishPending poller. If that scheduler is disabled, misconfigured, or stalls, skipped reviews will not be retried by the message bus. Consider NAK-with-delay or delayed redelivery so retry isn't solely poll-driven; at minimum, document the invariant that PublishPending is always running.
🟡 **daemon/cmd/heimdallm/main.go:41** — reviewPublishRetryMinAge is a hardcoded 2-minute constant. Pipeline publish latency varies by environment (CI load, diff size, GitHub API backoff). Consider making it configurable via env/config so operators can tune without recompiling.
🟡 **daemon/cmd/heimdallm/main.go:2279** — The CreatedAt.IsZero() branch treats legacy rows as immediately retryable, but this intent is only documented in the test. Add a short inline comment so future readers don't assume zero-time should be treated conservatively.
🟡 **daemon/cmd/heimdallm/main_test.go:160** — The 'unexpected extra publish message' assertion uses a 150ms timeout. Pass-on-timeout is the safe direction (a slow CI runner won't fail the test), but a slowly-arriving unwanted message could slip past unnoticed. Consider a deterministic sentinel-message synchronization or a longer timeout.

### Suggestions

- Promote reviewPublishRetryMinAge to a configurable value (env var or config field) with the current 2m as default.
- Either NAK-with-delay in the publish-worker skip path or add an inline comment asserting PublishPending is always scheduled in production.
- Add an inline comment on the CreatedAt.IsZero() branch in reviewReadyForPublishRetry explaining it's a legacy-row escape hatch, not a conservative default.
- Replace the 150ms negative-assertion timeout with a sentinel-message sync so the 'no extra publish' check is deterministic under CI load.

---

sergiotejon:
## ⚠️ MEDIUM Issue

Publish-worker still returns nil (acks) when ReviewInFlight is true, making retry solely dependent on the PublishPending poller. If the poller is disabled or stalls, the skipped review will never be republished. Prefer NAK-with-delay or scheduled redelivery so retry isn't poll-only.

**Location:** `daemon/cmd/heimdallm/main.go` line 703

---

sergiotejon:
## ⚠️ MEDIUM Issue

reviewReadyForPublishRetry only checks GitHubReviewID != 0 and in-flight status; it no longer enforces the prior CreatedAt minimum-age guard visible in earlier review context. If the min-age check was intentionally removed, document why; if not, the publish-pending poller may now race with freshly-inserted reviews before the initial publish attempt enqueues.

**Location:** `daemon/cmd/heimdallm/main.go` line 2274

---

sergiotejon:
## 🟡 LOW Issue

Negative assertion uses a 150ms timeout, which is flaky under CI load. Use a deterministic sentinel publish (e.g., publish a known marker after publishPending and assert it arrives without preceding extras) instead of a wall-clock window.

**Location:** `daemon/cmd/heimdallm/main_test.go` line 196

---

sergiotejon:
## 🟡 LOW Issue

ReviewInFlight uses COUNT(*) where SELECT 1 ... LIMIT 1 with sql.ErrNoRows handling would be marginally cheaper and clearer for a boolean existence check. Minor style nit.

**Location:** `daemon/internal/store/inflight.go` line 36

---

sergiotejon (approved):

Re-review of duplicate review publish retry fix. The author has refactored to use ReviewInFlight checks both in the publish-worker and PublishPending poller, with reasonable test coverage including in-process NATS. Previous findings about reviewPublishRetryMinAge and CreatedAt.IsZero() comment appear unaddressed in this diff (those lines are not shown changed), but since this re-review focuses on new changes, the main concerns remain around the ack-on-skip retry strategy and test timeout flakiness.

### Suggestions

- Consider returning an error or using msg.NakWithDelay (or equivalent JetStream redelivery) instead of acking when the review is in-flight, so retries don't depend on the poll-driven PublishPending scheduler.
- Make reviewPublishRetryMinAge configurable via env/config (still applies if the constant remains in the file).
- Replace the 150ms negative-assertion window with a deterministic sentinel message or synchronous flush check to harden the test on slow CI.
- Use EXISTS/SELECT 1 LIMIT 1 in ReviewInFlight instead of COUNT(*) for a cleaner existence check.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Simpler in-flight gate (no time-based retry window) is cleaner than the prior version; tests cover the ready/in-flight/published paths well. The publish-worker still acks on the in-flight skip path, leaving retry dependent on the PublishPending poller, and the negative test still uses a tight 150ms timeout.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:703** — Publish-worker still returns nil (acks) when ReviewInFlight is true. Retry is therefore entirely dependent on the PublishPending poller; if that scheduler is disabled, misconfigured, or stalls, skipped reviews will never be republished by the bus. Prefer NAK-with-delay / scheduled redelivery, or at minimum document the invariant that PublishPending must always be running.
🟡 **daemon/cmd/heimdallm/main_test.go:295** — The 'unexpected extra publish message' negative assertion still uses a 150ms timeout. Pass-on-timeout means a slow CI runner won't false-fail, but a slowly-arriving unwanted message could slip past unnoticed. Consider a deterministic sentinel-message synchronization (publish a sentinel after publishPending and assert the in-flight review's message never arrives before it).

### Suggestions

- Document on the publish-worker skip path (or at the PublishPending call site) that PublishPending is the sole retry mechanism for the in-flight skip case, so an operator disabling the scheduler understands the consequence.
- Replace the 150ms negative-assertion window with a sentinel-message handshake to make the 'no duplicate enqueue' assertion robust under load.

---

Muriano:
## ⚠️ MEDIUM Issue

Publish-worker returns nil (acks) when ReviewInFlight is true, making retry solely dependent on the PublishPending poller. If the poller stalls, is disabled, or misconfigured, the skipped review will never be republished. Recommend NAK-with-delay or a scheduled redelivery so retry doesn't rely on a single polling loop. Impact: silent loss of retry under poller failure; fix: NAK with backoff or schedule a delayed redelivery on the same subject.

**Location:** `daemon/cmd/heimdallm/main.go` line 703

---

Muriano (approved):

PR adds an in-flight check to prevent duplicate publish retries in both the publish-worker and the PublishPending poller. The change is correct, well-tested (in-process NATS, store-level, and adapter-level tests), and the ReviewInFlight helper is small and focused. The main concern carried over from prior review remains: the publish-worker silently acks when a review is in-flight, so retry is fully dependent on the PublishPending poller — if that scheduler stalls or is disabled, the skipped review will never be republished.

### Suggestions

- Make reviewPublishRetryMinAge (and related timing constants) configurable via env/config so operators can tune without recompiling.
- Add an inline comment near the CreatedAt.IsZero() branch explaining that zero-time legacy rows are treated as immediately retryable (intent is non-obvious).
- Replace the 150ms wall-clock negative assertion in TestTier2AdapterPublishPendingDefersInFlightReviews with a deterministic sentinel-message synchronization to avoid CI flakiness.
- ReviewInFlight could use SELECT 1 ... LIMIT 1 with sql.ErrNoRows instead of COUNT(*) for a clearer boolean existence check (minor).

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean implementation of in-flight deduplication for publish retries. The new ReviewInFlight store method, the guard in the NATS publish worker, and the publishPending sweep filter all work together correctly. Tests are comprehensive, covering claim/release lifecycle, the readyForPublishRetry predicate, and the end-to-end NATS integration. The previous NAK-vs-ACK concern appears to be a deliberate design choice documented by the inline comment. Only minor observations remain.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:2256** — The PublishPending → publishPending indirection adds no value. Both methods live on the same unexported type and the tests are in the same package (package main), so they can call PublishPending() directly. Consider dropping the private wrapper and keeping the logic in the public method.
🟡 **daemon/cmd/heimdallm/main.go:2268** — publishPending issues 2 extra DB queries per unpublished review (GetPR + ReviewInFlight). This is fine for typical volumes, but if a backlog builds up (e.g. after a long GitHub outage) the N+1 pattern could cause noticeable sweep latency. Consider a single joined query if this becomes a concern — not blocking.

### Suggestions

- Collapse PublishPending/publishPending into a single method — the test-only motivation doesn't hold since both are in the same package.
- The previous NAK-with-delay suggestion (flagged twice) is clearly a deliberate design choice per the comment at line 705. If the team is happy coupling retry responsibility to the sweep scheduler, no further action needed — just ensure the sweep interval is short enough to cover the ack-and-wait window.

---

PR #381, opened 2026-04-28
fix: stop circuit breaker requeue loop

ivanmunozruiz:
## What changed

- Suppress review enqueue when the PR circuit breaker is already tripped, so a review-requested PR does not emit the same breaker banner every poll cycle.
- Deduplicate the breaker SSE event per PR/update/reason/head SHA.
- Change the PR-side breaker from whole-PR counting to PR + HEAD SHA counting. A repeated re-request on the same commit can still trip the safety cap, but a developer's follow-up commit gets a fresh allowance.
- Keep the per-repo hourly cap unchanged.

## Why

The observed loop on `freepik-company/ai-platform-iqs-v3#443` was caused by Tier 2 repeatedly enqueueing the same PR after the breaker had already rejected it. The breaker error was correct as a safety mechanism, but the poller treated the PR as reviewable again on the next tick, producing repeated banners and notifications.

The previous `per_pr_24h` semantics were also too blunt: after three reviews in 24h, a legitimate fix commit could be blocked. The cap now protects the failure mode we care about: repeated reviews of the same HEAD.

## Validation

- `docker run --rm --user "$(id -u):$(id -g)" -v "$(pwd):/src:ro" -w /src/daemon -e HOME=/tmp/home -e GOCACHE=/tmp/go-build -e GOMODCACHE=/tmp/gomod golang:1.25-alpine@sha256:f6751d823c26342f9506c03797d2527668d095b0a15f1862cddb4d927a7a4ced sh -c 'go vet ./... && go test -count=1 ./...'`

Note: `make test-docker GO_TEST_ARGS="-count=1 ./cmd/heimdallm ./internal/store ./internal/pipeline"` still fails before tests with missing module resolution from the mounted `/tmp/heimdallm-gocache` `GOMODCACHE`; the same pinned image passes with an in-container `GOMODCACHE`.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -491,6 +491,7 @@ func main() {
 		publishPub:           publishPub,
 		watchStore:           watchStore,
 		lastSkippedUpdatedAt: make(map[int64]time.Time),
+		lastBreakerTrips:     make(map[breakerTripKey]breakerTripDedup),
 	}
 
 	repoPublisher := bus.NewRepoPublisher(conn)
@@ -1827,7 +1828,7 @@ func runTier2(
 				if _, ok := monitoredSet[pr.Repo]; !ok {
 					continue
 				}
-				if adapter.PRAlreadyReviewed(pr.ID, pr.Repo, pr.Number, pr.UpdatedAt) {
+				if adapter.PRAlreadyReviewed(pr.ID, pr.Repo, pr.Number, pr.UpdatedAt, pr.HeadSHA) {
 					continue
 				}
 				if err := prPublisher.PublishPRReview(ctx, pr.Repo, pr.Number, pr.ID, pr.HeadSHA); err != nil {
@@ -1937,12 +1938,24 @@ type tier2Adapter struct {
 	publishPub *bus.PRPublishPublisher
 	watchStore *bus.WatchStore
 
-	// skipMu protects lastSkippedUpdatedAt, which deduplicates review_skipped
-	// SSE events across consecutive poll cycles for the same (PR ID, updated_at)
-	// pair. Entries are pruned at the end of each FetchPRsToReview cycle so the
-	// map stays bounded to the current set of review-requested PRs.
+	// skipMu protects the lightweight SSE dedup caches below.
 	skipMu               sync.Mutex
 	lastSkippedUpdatedAt map[int64]time.Time
+	lastBreakerTrips     map[breakerTripKey]breakerTripDedup
+}
+
+const breakerTripDedupTTL = 24 * time.Hour
+
+type breakerTripKey struct {
+	Repo    string
+	Number  int
+	HeadSHA string
+	Reason  string
+}
+
+type breakerTripDedup struct {
+	UpdatedAt time.Time
+	EmittedAt time.Time
 }
 
 // discoveryStore is the subset of *store.Store that processDiscoveredRepos
@@ -2204,6 +2217,7 @@ func (a *tier2Adapter) FetchPRsToReview() ([]scheduler.Tier2PR, error) {
 		}
 	}
 	a.skipMu.Unlock()
+	a.pruneBreakerTripDedup(time.Now())
 
 	return out, nil
 }
@@ -2401,7 +2415,7 @@ func (a *tier2Adapter) PromoteReady(ctx context.Context, repos []string) (int, e
 }
 
 // PRAlreadyReviewed implements scheduler.Tier2Store.
-func (a *tier2Adapter) PRAlreadyReviewed(githubID int64, repo string, number int, updatedAt time.Time) bool {
+func (a *tier2Adapter) PRAlreadyReviewed(githubID int64, repo string, number int, updatedAt time.Time, headSHA string) bool {
 	existing, _ := a.store.GetPRByGithubID(githubID)
 	if existing == nil && repo != "" && number > 0 {
 		existing, _ = a.store.GetPRByRepoNumber(repo, number)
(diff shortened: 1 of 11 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Solid fix for the circuit-breaker requeue loop. Adds HEAD-SHA-scoped breaker counts (allowing follow-up commits) and emits a deduplicated SSE event so the breaker no longer re-pushes the banner each poll. Tests cover trip-suppression, new-HEAD-SHA bypass, and store-level counts. A couple of minor concerns around dedup-map lifecycle and test timing, but nothing blocking.

### Suggestions

- lastBreakerTrips entries are pruned only by FetchPRsToReview's seenKeys; entries added via the tier3 HandleChange path for a PR not present in the next tier2 batch will be evicted and may re-emit the SSE on the following tick. Consider keying eviction off both code paths or adding a TTL.
- TestPRAlreadyReviewedCircuitBreakerSuppressesRepeatedEnqueue relies on a 100ms drain window after broker.Subscribe; under load this could be flaky. Prefer a synchronous assertion (e.g., counting events with a deterministic wait or checking a captured slice) over time-based draining.
- CheckCircuitBreaker now performs an extra SQLite query per PRAlreadyReviewed call. For large monitored sets this runs on every poll cycle; consider scoping the breaker check to PRs that fail the freshness window only (already the case) and confirming the head_sha index exists on reviews.
- Minor: prIdentityKey uses '%s#%d'; if a repo name ever contains '#' the key could collide with another repo/number pair. Extremely unlikely with GitHub repo names but worth a comment.

---

sergiotejon:
## 🟡 LOW Issue

circuitBreakerBlocksPR guards against a.cfgMu/a.cfg being nil, but a.skipMu (used in clearCircuitBreakerDedup and publishCircuitBreakerTrippedOnce) is a value-type sync.Mutex on the adapter struct and is always usable; however, the new lastBreakerTrips map is only initialized at construction in main() and lazily in publishCircuitBreakerTrippedOnce. clearCircuitBreakerDedup calls delete on a possibly-nil map (safe in Go) but FetchPRsToReview also iterates it without nil check (also safe). Worth confirming all adapter construction sites initialize the map to avoid surprises in future tests.

**Location:** `daemon/cmd/heimdallm/main.go` line 2456

---

sergiotejon:
## ⚠️ MEDIUM Issue

The testAdapter.PRAlreadyReviewed signature was updated to accept headSHA, but per the comment ('Keep the two in sync') it does NOT replicate the new circuit-breaker suppression branch added to the production adapter. The test copy now diverges from production behavior in a way the comment explicitly warns against. Either port the breaker logic or update the comment to scope what 'in sync' means.

**Location:** `daemon/internal/pipeline/testutil_test.go` line 30

---

sergiotejon:
## ⚠️ MEDIUM Issue

Pruning of lastBreakerTrips uses seenKeys built from the current FetchPRsToReview batch, but PRAlreadyReviewed is also called from HandleChange (tier3 path) with PRs that may not appear in the tier2 fetch batch. Entries inserted via the tier3 path could be pruned on the next tier2 cycle, causing a duplicate SSE emission. Consider either keying off PR existence in the store or only pruning entries older than some TTL.

**Location:** `daemon/cmd/heimdallm/main.go` line 2216

---

sergiotejon:
## 🟡 LOW Issue

alreadyEmitted check uses '!updatedAt.After(prev.UpdatedAt)' — if updatedAt equals prev.UpdatedAt the event is suppressed (correct), but if a PR's updatedAt moves backward (clock skew or GitHub API quirk) with same reason+SHA, it will also be suppressed silently. Minor; current behavior is probably fine but worth a comment.

**Location:** `daemon/cmd/heimdallm/main.go` line 2495

---

sergiotejon:
## 🟡 LOW Issue

When headSHA is empty the code falls back to the whole-PR count but still uses the same PerPR24h cap. The doc comment explains this, but the reason string ('per-PR cap reached') differs from the SHA-scoped one ('per-PR HEAD cap reached'). The dedup key in publishCircuitBreakerTrippedOnce includes headSHA—an empty headSHA followed by a non-empty one (or vice versa) for the same PR will emit duplicate SSE events with different reasons. Probably acceptable but worth noting.

**Location:** `daemon/internal/store/circuitbreaker.go` line 75

---

sergiotejon (approved):

Solid fix for the circuit breaker requeue loop. The core change—suppressing re-enqueue when the breaker is tripped, scoping the per-PR cap to HEAD SHA, and deduplicating SSE events—is well-reasoned and well-tested. A few concerns around nil-safety, testutil drift, and a potential map-growth edge case.

### Suggestions

- Add a test exercising PRAlreadyReviewed via the HandleChange/tier3 path to confirm the breaker dedup map interacts correctly with both fetch sources.
- Consider extracting the SSE dedup key to include reason as well so transitions between per-PR and per-PR-HEAD reasons emit once each rather than being collapsed/double-emitted ambiguously.
- Update the 'Keep the two in sync' comment in testutil_test.go to clarify which behaviors are intentionally simplified vs. mirrored, or port the breaker branch to keep parity.
- A short comment on why lastBreakerTrips is keyed by repo#number rather than GitHub ID (consistency with prIdentityKey) would help future readers.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured fix that correctly shifts the circuit breaker from per-PR to per-HEAD-SHA granularity and suppresses the requeue loop by treating a tripped breaker as 'already reviewed'. The SSE dedup mechanism is sound and thread-safe. Two medium issues around test-adapter divergence and tier3-vs-tier2 pruning mismatch, plus a couple of minor concerns.

### Issues

⚠️ **daemon/internal/pipeline/testutil_test.go:30** — testAdapter.PRAlreadyReviewed now accepts headSHA but does NOT replicate the new circuitBreakerBlocksPR branch added to the production tier2Adapter. The comment on line 27 says 'Keep the two in sync', but the production version now has a meaningful new code path (circuit breaker suppression) that this test copy lacks entirely. The reloop regression tests exercise this copy, so they cannot catch breaker-related dedup regressions. Either port a minimal version of the breaker check into testAdapter or update the sync-comment to explicitly scope what must match.
⚠️ **daemon/cmd/heimdallm/main.go:2216** — lastBreakerTrips is pruned using seenKeys built from the tier2 FetchPRsToReview batch, but PRAlreadyReviewed is also called from HandleChange (tier3 path) for PRs that may not appear in the tier2 fetch. Entries inserted via the tier3 path will be pruned on the next tier2 cycle, causing a duplicate SSE emission the next time the tier3 check fires. Consider adding a time-based TTL or pruning only entries older than a configurable window instead of relying solely on the tier2 seen-set.
🟡 **daemon/cmd/heimdallm/main.go:2456** — circuitBreakerBlocksPR has extensive nil guards (a == nil, a.store == nil, a.cfgMu == nil, a.cfg == nil, *a.cfg == nil) for states that cannot occur in production — the adapter is always fully constructed in main() before use. These won't cause bugs but add noise. The nil-receiver check on clearCircuitBreakerDedup (line 2478) is similarly unnecessary since it's only called from circuitBreakerBlocksPR which already checked a != nil.
🟡 **daemon/internal/store/circuitbreaker.go:75** — When headSHA is empty, the fallback whole-PR count uses the same PerPR24h cap but emits a different reason string ('per-PR cap' vs 'per-PR HEAD cap'). Since publishCircuitBreakerTrippedOnce deduplicates by (headSHA, reason), transitioning from empty headSHA to non-empty (or vice versa) for the same PR will emit two SSE events with different reasons. Edge case but worth a comment.
🟡 **daemon/cmd/heimdallm/main_test.go:336** — TestPRAlreadyReviewedCircuitBreakerAllowsNewHeadSHA creates an adapter with broker: sse.NewBroker() but never calls broker.Start(). If any code path in this test unexpectedly tried to publish an SSE event, it could behave differently than production. Not currently a bug since the breaker doesn't trip in this test, but fragile if the test is extended.

### Suggestions

- Port the circuit-breaker suppression logic into testAdapter (or extract the shared logic into a helper function both adapters can call) so the reloop regression tests cover the new breaker-in-PRAlreadyReviewed path.
- Replace the seenKeys-based pruning of lastBreakerTrips with a TTL-based approach (e.g., prune entries older than 1 hour) so that tier3-originated entries are not incorrectly evicted by the tier2 cycle.
- Consider extracting the shared nil-guard + config-read pattern from circuitBreakerBlocksPR into a small helper that returns (CircuitBreakerLimits, bool) to reduce the nil-check boilerplate.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Targeted fix that scopes the circuit breaker to HEAD SHA and short-circuits the tier2/tier3 enqueue loop when the breaker is tripped, with dedup to avoid SSE banner spam. Logic is sound and well-tested. Two existing reviewer comments (tier3 vs tier2 pruning interaction, testAdapter divergence) reflect real concerns; a couple of smaller robustness notes around map init and empty-HeadSHA semantics.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2216** — lastBreakerTrips is pruned in FetchPRsToReview using seenKeys built only from the tier2 batch, but PRAlreadyReviewed is also invoked from HandleChange (tier3 WatchItem path). A trip recorded via tier3 for a PR that is not in the current tier2 fetch will be pruned next cycle, causing the SSE 'circuit_breaker_tripped' banner to re-emit on the next tier3 tick. Either prune by store existence/TTL, or have tier3 contribute its key into the seen set.
⚠️ **daemon/internal/pipeline/testutil_test.go:30** — testAdapter.PRAlreadyReviewed accepts headSHA but does not replicate the new circuit-breaker suppression branch. The doc comment 'Keep the two in sync' now overstates parity — port the breaker block or scope the comment to the dedup-by-PublishedAt portion only.
🟡 **daemon/cmd/heimdallm/main.go:2456** — publishCircuitBreakerTrippedOnce lazily initializes lastBreakerTrips, but clearCircuitBreakerDedup and the FetchPRsToReview prune loop both touch the map under skipMu without lazy init. Deletes/iteration over a nil map are safe in Go, but if a future construction site forgets to initialize the map and only the clear path runs, dedup is silently disabled. Initialize unconditionally in the adapter constructor (already done in main but not enforced) or lazy-init in clearCircuitBreakerDedup as well.
🟡 **daemon/internal/store/circuitbreaker.go:75** — When headSHA is empty the breaker falls back to the whole-PR count but emits 'per-PR cap reached' while the SHA-scoped path emits 'per-PR HEAD cap reached'. The adapter's dedup key includes headSHA, so an empty→non-empty (or vice-versa) transition for the same PR yields two SSE events with different reasons. Either normalize the reason string or include 'reason kind' in the dedup key explicitly.
🟡 **daemon/cmd/heimdallm/main.go:2495** — alreadyEmitted uses '!updatedAt.After(prev.UpdatedAt)', which also suppresses cases where updatedAt moves backward (clock skew, GitHub API quirk) with same reason+SHA. Behavior is fine, but worth a one-line comment explaining the intentional inequality.

### Suggestions

- Address the tier3 pruning interaction: either include tier3 keys in seenKeys before pruning, or change pruning to a TTL/last-seen timestamp model so HandleChange-only PRs aren't garbage-collected mid-cycle.
- Either port the breaker suppression into testAdapter.PRAlreadyReviewed or tighten the 'Keep the two in sync' comment to scope what parity actually means after this change.
- Consider initializing lastBreakerTrips inside the adapter constructor unconditionally (helper) rather than relying on each call site in main()/tests to remember.
- Add a brief comment near the empty-headSHA fallback in CheckCircuitBreaker explaining when it can occur (legacy rows / pre-resolution) and that the cap is intentionally shared with the SHA-scoped axis.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of fix #381 (circuit breaker requeue loop). The author addressed the prior MEDIUM finding about TTL-based pruning by switching from seenKeys-based pruning to a 24h TTL prune (pruneBreakerTripDedup), which correctly handles tier3-path entries. The testutil divergence is now explicitly documented in the updated comment, scoping 'in sync' to the persisted freshness portion only. Tests are added for the new circuit-breaker suppression path, HEAD-SHA scoping, and TTL pruning. Logic is sound and the fix targets the root cause (persistent GitHub review-requested results re-enqueueing tripped PRs). No new blocking issues.

### Suggestions

- Consider adding a short comment near the alreadyEmitted check in publishCircuitBreakerTrippedOnce explaining that non-monotonic updated_at intentionally suppresses duplicates (clock-skew tolerance), matching the rationale already in the inline comment.
- The empty-headSHA fallback path in CheckCircuitBreaker emits a different reason string than the SHA-scoped path; if the same PR transitions from empty to non-empty headSHA within the dedup TTL, two SSE banners may fire. Low risk in practice but worth a one-line note in publishCircuitBreakerTrippedOnce.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of fix #381 (circuit breaker requeue loop). The author addressed the prior MEDIUM finding about TTL-based pruning by switching from seenKeys-based pruning to a 24h TTL prune (pruneBreakerTripDedup), which correctly handles tier3-path entries. The testutil divergence is now explicitly documented in the updated comment, scoping 'in sync' to the persisted freshness portion only. Tests are added for the new circuit-breaker suppression path, HEAD-SHA scoping, and TTL pruning. Logic is sound and the fix targets the root cause (persistent GitHub review-requested results re-enqueueing tripped PRs). No new blocking issues.

---

sergiotejon:
## 🟡 LOW Issue

Empty-vs-non-empty headSHA for the same PR will still produce two distinct dedup keys with different reason strings ('per-PR cap reached' vs 'per-PR HEAD cap reached'), potentially yielding duplicate SSE banners during a transient empty-SHA window. Behavior is documented but could be hardened by normalizing the reason or dedup key.

**Location:** `daemon/internal/store/circuitbreaker.go` line 75

---

sergiotejon (approved):

Author addressed the main previous findings: pruning is now TTL-based (24h) instead of batch-keyed, alreadyEmitted has an explanatory comment about non-monotonic updated_at, and the testutil comment was updated to scope what 'in sync' means. The empty-headSHA reason-string concern remains but is documented. New tests cover the breaker dedup, TTL pruning, and HEAD SHA allowance. Overall the fix looks solid.

### Suggestions

- Consider adding a brief test that exercises the cmd-layer adapter through HandleChange (tier3 path) to confirm TTL-based pruning keeps tier3-inserted dedup entries alive across tier2 cycles.
- Optionally collapse the dedup key's Reason field (or strip headSHA-vs-empty distinction) so a transient missing-SHA does not produce a second banner for the same logical trip.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: author addressed both MEDIUM findings (TTL-based pruning replaces tier2-batch-scoped pruning; testutil comment scoped to freshness-only parity) and added the requested clarifying comment about non-monotonic updated_at suppression. Two LOW findings remain unchanged: inconsistent lazy-init of lastBreakerTrips, and divergent reason strings between SHA and whole-PR breaker paths.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:2456** — publishCircuitBreakerTrippedOnce lazy-inits lastBreakerTrips, but clearCircuitBreakerDedup and pruneBreakerTripDedup do not. Deletes/iteration over a nil map are safe today, but a future adapter construction site that forgets to initialize the map (e.g. a new test) would silently disable dedup if only the clear or prune path runs first. Either lazy-init in all three sites or rely solely on construction-time init and remove the lazy branch in publishCircuitBreakerTrippedOnce so the invariant is consistent.
🟡 **daemon/internal/store/circuitbreaker.go:75** — When headSHA is empty the breaker falls back to the whole-PR count but emits 'per-PR cap reached' while the SHA-scoped path emits 'per-PR HEAD cap reached'. publishCircuitBreakerTrippedOnce's dedup key includes both headSHA and reason, so a PR that flips between empty and non-empty headSHA can emit two banners with different reasons for what is effectively the same trip. Either normalize the reason text or include a kind discriminator explicitly in the dedup key.

### Suggestions

- Consider asserting at adapter construction that lastBreakerTrips is non-nil (or remove the lazy init in publishCircuitBreakerTrippedOnce) to keep the invariant single-sourced.
- Add a brief regression test exercising HandleChange (tier3) → PRAlreadyReviewed → breaker trip path to lock in that TTL-based pruning preserves the tier3-recorded entry across a tier2 fetch cycle.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Previous MEDIUM findings (testutil sync comment, seen-set pruning) are both addressed. TTL-based dedup pruning and the updated test-adapter comment properly scope the behavior divergence. The new circuit-breaker-blocks-PR flow, SSE dedup, and HEAD-SHA-scoped counting are well-structured and well-tested. No significant new issues.

### Suggestions

- TestPRAlreadyReviewedCircuitBreakerAllowsNewHeadSHA initializes the adapter without lastBreakerTrips (relying on lazy init in publishCircuitBreakerTrippedOnce). Consider initializing it like the other circuit breaker test does for consistency and resilience if the test is later extended to a tripping path.
- Minor: clearCircuitBreakerDedup iterates all map entries to find matching (repo, number). If the dedup map ever grows large, a secondary index or composite key restructure would help — not needed at current scale but worth a mental note.

---

PR #382, opened 2026-04-30
fix: filter PR auto-discovery by discovery_orgs

sergiotejon:
## Summary
- `FetchPRsToReview` searches all of GitHub for review-requested PRs (no `repo:` qualifier). `upsertDiscoveredRepos` was blindly adopting every repo it found, letting repos from orgs outside `discovery_orgs` (e.g. `freepik-company/fc-py-cogito`) get auto-adopted.
- This caused endless 403 errors every minute (issue fetch + promote) and wasted API rate limit.
- Now both `upsertDiscoveredRepos` and the PR pipeline loop check the repo's org prefix against `DiscoveryOrgs` (case-insensitive). When the list is empty, behaviour is unchanged (backward compat).

## Test plan
3 new tests in `main_discover_test.go`:
- `TestUpsertDiscoveredRepos_FiltersOutsideDiscoveryOrgs`
- `TestUpsertDiscoveredRepos_NoOrgFilterWhenDiscoveryOrgsEmpty`
- `TestUpsertDiscoveredRepos_OrgFilterCaseInsensitive`

Full suite passes: `make test-docker GO_DOCKER_IMAGE="golang:1.25-alpine"`

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -1900,6 +1900,15 @@ func upsertDiscoveredRepos(c *config.Config, prs []*gh.PullRequest) []string {
 		known[r] = struct{}{}
 	}
 
+	// Build an org allowlist from DiscoveryOrgs. When set, repos whose org
+	// prefix is not in the list are silently skipped — prevents the PR
+	// review-requested search (which spans all of GitHub) from auto-adopting
+	// repos outside the operator's intended organisations.
+	allowedOrgs := make(map[string]struct{}, len(c.GitHub.DiscoveryOrgs))
+	for _, o := range c.GitHub.DiscoveryOrgs {
+		allowedOrgs[strings.ToLower(o)] = struct{}{}
+	}
+
 	enable := c.GitHub.AutoEnablePRForDiscovery()
 	added := []string{}
 	for _, pr := range prs {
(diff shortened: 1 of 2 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

PR correctly filters PR auto-discovery by configured discovery_orgs, mirroring filter in both upsertDiscoveredRepos and tier2Adapter.FetchPRsToReview. Implementation is straightforward, case-insensitive, backward compatible (empty list = accept all), and well-tested for the upsert path.

### Suggestions

- Consider extracting the duplicated allowlist-building and org-extraction logic into a small helper (e.g., config.GitHub.IsOrgAllowed(repo string) bool) to avoid drift between the two filter sites.
- Add a unit test for the tier2Adapter.FetchPRsToReview org filter analogous to the upsert tests, since that path is currently untested in this PR.
- The skip in FetchPRsToReview is silent; consider a slog.Debug similar to the upsert path for parity and easier production debugging.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Adds an org allowlist filter to PR auto-discovery and the tier2 PR fetch path, gated by DiscoveryOrgs. Logic is straightforward, case-insensitive, and well-tested for the upsert path. Minor duplication and a missing test for the tier2 adapter filter.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:2151** — Org-allowlist construction is duplicated between upsertDiscoveredRepos and FetchPRsToReview. Consider extracting a small helper (e.g. buildOrgAllowlist + isOrgAllowed) to keep the two filters in sync as the rule evolves.
🟡 **daemon/cmd/heimdallm/main.go:2168** — The tier2 adapter filter silently drops PRs with no debug log, unlike the upsert path which logs at debug level. Inconsistent observability — adding a parallel slog.Debug would aid troubleshooting when operators wonder why a PR was skipped.
🟡 **daemon/cmd/heimdallm/main.go:1925** — Org parsing uses strings.IndexByte(pr.Repo, '/') and treats malformed repos (no slash, or leading slash) as org="", which then fails the allowlist check and is silently skipped. Acceptable, but worth a Warn rather than Debug since a malformed pr.Repo upstream likely indicates a bug elsewhere.
⚠️ **daemon/cmd/heimdallm/main_discover_test.go:96** — No test coverage for the new filter inside FetchPRsToReview (tier2Adapter). The upsert path is covered three ways, but the parallel logic in the adapter has no regression test.

### Suggestions

- Extract the allowlist build + org-extraction into a shared helper to remove the duplicated block across both call sites.
- Add a test for tier2Adapter.FetchPRsToReview verifying PRs from disallowed orgs are filtered out when DiscoveryOrgs is set.
- Consider normalizing DiscoveryOrgs once at config load time rather than lowercasing on every discovery cycle.

---

PR #384, opened 2026-04-30
fix: use time.Now() in circuit breaker tests to stay within 24h window

### LINKED ISSUE #383: [Bug] TestPRAlreadyReviewedCircuitBreakerSuppressesRepeatedEnqueue fails on Go 1.25
## Summary

`TestPRAlreadyReviewedCircuitBreakerSuppressesRepeatedEnqueue` in `daemon/cmd/heimdallm/main_test.go:208` fails consistently on Go 1.25.

## Reproduction

```bash
make test-docker GO_DOCKER_IMAGE="golang:1.25-alpine" \
  GO_TEST_ARGS="-run TestPRAlreadyReviewedCircuitBreakerSuppressesRepeatedEnqueue ./cmd/heimdallm/..."
```

Output:
```
--- FAIL: TestPRAlreadyReviewedCircuitBreakerSuppressesRepeatedEnqueue (0.01s)
    main_test.go:263: circuit breaker trip should suppress enqueue
```

## Why CI on main appears green

The Makefile `GO_DOCKER_IMAGE` is pinned to `golang:1.21-alpine` which cannot compile the module (`go.mod requires go >= 1.25.0`). The CI workflow on `macos-14` runners uses `setup-go` natively and does run Go 1.25, but this test was likely introduced after the last successful CI run or has a race/ordering issue that only surfaces in the Docker environment.

Reproduces on `main` at `b2ddae8` — confirmed not introduced by any branch.

## Likely cause

Introduced or broken by #381 (fix: stop circuit breaker requeue loop). The `tier2Adapter` constructed in the test may be missing a newly required field from that PR.

## Impact

Blocks the "Daemon (Go)" CI check on all PRs (e.g. #382). Does not affect runtime behaviour.

sergiotejon:
## Summary
- `TestPRAlreadyReviewedCircuitBreakerSuppressesRepeatedEnqueue` and `TestPRAlreadyReviewedCircuitBreakerAllowsNewHeadSHA` used a hardcoded date (`2026-04-28`) for review timestamps. Once real time moved past 2026-04-29, the reviews fell outside the 24h window that `CheckCircuitBreaker` queries with `time.Now().Add(-24h)`, causing the count to return 0 and the breaker to never trip.
- Fix: replace the hardcoded `time.Date(...)` with `time.Now().UTC().Truncate(time.Second)`.

Closes #383

## Test plan
```bash
make test-docker GO_DOCKER_IMAGE="golang:1.25-alpine"
```
Full suite passes (all 15 packages).

### DIFF
--- a/daemon/cmd/heimdallm/main_test.go
+++ b/daemon/cmd/heimdallm/main_test.go
@@ -207,7 +207,7 @@ func TestTier2AdapterPublishPendingDefersInFlightReviews(t *testing.T) {
 
 func TestPRAlreadyReviewedCircuitBreakerSuppressesRepeatedEnqueue(t *testing.T) {
 	s := newMemStore(t)
-	now := time.Date(2026, 4, 28, 14, 30, 0, 0, time.UTC)
+	now := time.Now().UTC().Truncate(time.Second)
 	prID, err := s.UpsertPR(&store.PR{
 		GithubID:  1001,
 		Repo:      "org/repo",
@@ -318,7 +318,7 @@ func TestBreakerTripDedupPrunesByTTL(t *testing.T) {
 
 func TestPRAlreadyReviewedCircuitBreakerAllowsNewHeadSHA(t *testing.T) {
 	s := newMemStore(t)
-	now := time.Date(2026, 4, 28, 14, 30, 0, 0, time.UTC)
+	now := time.Now().UTC().Truncate(time.Second)
 	prID, err := s.UpsertPR(&store.PR{
 		GithubID:  1002,
 		Repo:      "org/repo",


Muriano (approved):
## 🤖 Heimdallm AI Review

Reemplaza fechas hardcodeadas por time.Now() en dos tests del circuit breaker para evitar que expiren al salir de la ventana de 24h. Cambio correcto y mínimo.

### Suggestions

- Considerar extraer un helper compartido (e.g., testNow()) si más tests adoptan este patrón, para mantener consistencia.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Replaces hardcoded date with time.Now() in two circuit breaker tests to prevent test failures once the 24-hour window from the original fixed date elapses. The fix is correct and minimal, but introduces non-determinism into tests that may have implicit time-based assumptions.

### Issues

🟡 **daemon/cmd/heimdallm/main_test.go:210** — Using time.Now() makes tests non-deterministic. If any downstream logic in the circuit breaker depends on absolute time boundaries (e.g., day rollovers, TTL edges), tests could become flaky depending on when they run. A better fix would be to address the root cause: why does the production code couple to wall-clock time within a 24h window? Consider injecting a clock or extending the window in test setup rather than chasing time.Now().
🟡 **daemon/cmd/heimdallm/main_test.go:210** — The PR title states the original test was failing because the hardcoded date fell outside a 24h window relative to time.Now() used elsewhere in the system under test. This suggests the underlying code mixes injected/fixed times with time.Now() calls, which is a design smell worth addressing rather than papering over in tests.

### Suggestions

- Consider injecting a clock interface (e.g., clockwork or a simple func() time.Time) into the circuit breaker so tests can use a fixed deterministic time without coupling to wall-clock.
- If a clock injection refactor is out of scope, add a comment in the test explaining why time.Now() is required (the 24h window constraint) so future maintainers understand the intent.
- Verify that Truncate(time.Second) is actually needed — if not, remove it to keep the test minimal.

---

PR #385, opened 2026-04-30
feat: gate PR reviews on requested_reviewers check, remove 2-min grace

sergiotejon:
## Summary

- Adds `GetPRHeadInfo` method that returns HEAD SHA + `requested_reviewers` from the same `GET /pulls/{n}` call (zero extra API cost)
- Gates PR review enqueue on bot being present in `requested_reviewers` — eliminates ghost enqueues from Search API replication lag
- Removes the 2-minute `PublishedAt` grace window from `PRAlreadyReviewed` — a push + re-request-review within 2 min is now picked up immediately
- `PRAlreadyReviewed` retains: dismissed check + circuit breaker
- Fail-open on API errors (existing behavior preserved)

## Motivation

The 2-min grace suppressed ghost re-enqueues from the Search API lag, but also blocked legitimate re-review requests made within 2 min of the last review. Checking `requested_reviewers` directly is the correct guard — it's authoritative and doesn't have a blind window.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -2212,25 +2212,37 @@ func (a *tier2Adapter) FetchPRsToReview() ([]scheduler.Tier2PR, error) {
 		delete(a.lastSkippedUpdatedAt, pr.ID)
 		a.skipMu.Unlock()
 
-		// Resolve the HEAD SHA so the persistent in-flight claim (#258) can
-		// key on (pr_id, head_sha) downstream. The Search Issues API does
-		// NOT populate head.sha, so this is an extra /pulls/N call per PR
-		// that cleared the review guards — bounded by the small number of
-		// review-requested PRs per cycle. See theburrowhub/heimdallm#264
-		// for the bug this closes: before this lookup the SHA was empty,
-		// and runReview silently skipped the claim guard on every tick.
+		// Resolve the HEAD SHA and confirm the bot is still a pending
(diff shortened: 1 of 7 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Replaces 2-minute PublishedAt grace window with an upstream requested_reviewers check on the Pulls API, sharing the same GET /pulls/{n} call used for HEAD SHA resolution. Implementation is clean: PRHeadInfo type is well-scoped, getPR helper avoids duplication, case-insensitive matching is correct, tests cover the new behavior and legacy paths. Trade-off (slow review + rapid re-request no longer suppressed by time grace) is documented and intentional. No confirmed bugs.

### Suggestions

- daemon/internal/pipeline/pipeline_reloop_test.go: TestPRAlreadyReviewed_FallsBackToRepoNumberWhenGithubIDDiffers now discards the result with `_ =` and asserts nothing — the test name still claims to verify fallback resolution but no longer does. Either assert the resolved PR row directly via the store, or rename/remove the test to reflect that it only checks no-panic.
- daemon/cmd/heimdallm/main.go: When botLogin is empty the requested_reviewers ghost-filter is silently skipped. Consider logging a warning at startup (or once) if botLogin is unresolved, since the new defense relies on it being set.
- daemon/cmd/heimdallm/main.go: On shaErr the code falls through to `headSHA := info.HeadSHA` (empty) but the bot-membership check is also skipped — worth a brief comment noting this is an intentional fail-open since info is zero-valued in that branch.

---

PR #387, opened 2026-05-01
fix(cli): make j/k/arrows scroll Stats and Config tabs directly

sergiotejon:
## Summary

- Stats and Config tabs used cursor-as-selected-row but rendered without row highlight, so j/k/arrow presses appeared to do nothing
- Switches these read-only tabs to scroll-offset mode: cursor = first visible line, each j/k press scrolls viewport by one line
- `clampCursor` caps at `total - visible` for scroll-offset tabs
- pgdn/pgup continue to work as before (page-sized jumps)

### DIFF
--- a/cli/internal/tui/dashboard.go
+++ b/cli/internal/tui/dashboard.go
@@ -544,10 +544,37 @@ func (d *Dashboard) scrollDetailUp() {
 	}
 }
 
+// isScrollOffsetTab reports whether the active tab uses the cursor as a
+// scroll offset (first visible line) rather than a selected-row index.
+// Read-only informational tabs (Stats, Config) have no selectable rows,
+// so j/k/arrows scroll the viewport directly instead of moving a highlight.
+func (d *Dashboard) isScrollOffsetTab() bool {
+	return d.activeTab == tabStats || d.activeTab == tabConfig
+}
+
 func (d *Dashboard) clampCursor() {
 	max := d.tabItemCount()
-	if max > 0 && d.cursor >= max {
-		d.cursor = max - 1
+	if d.isScrollOffsetTab() {
+		// cursor is a scroll offset: clamp so the last line is visible
+		// at the bottom of the viewport.
+		visible := d.contentHeight()
+		if d.activeTab == tabConfig {
+			visible -= 2 // header + separator
+		}
+		if visible < 1 {
+			visible = 1
+		}
+		upper := max - visible
+		if upper < 0 {
+			upper = 0
+		}
+		if d.cursor > upper {
+			d.cursor = upper
+		}
+	} else {
+		if max > 0 && d.cursor >= max {
+			d.cursor = max - 1
+		}
 	}
 	if d.cursor < 0 {
 		d.cursor = 0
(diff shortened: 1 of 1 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

Sensible UX fix: Stats/Config tabs now scroll directly via cursor-as-offset semantics. Logic is correct, but the scroll-clamp math is duplicated in three places and may drift between clampCursor and the renderers.

### Issues

⚠️ **cli/internal/tui/dashboard.go:555** — clampCursor subtracts 2 from visible only for tabConfig, but renderStats may also reserve header lines (its maxVisible calculation isn't shown to use the full height). If renderStats reserves any rows for a header, clampCursor will let cursor advance past what renderStats can actually display, causing a visual 'dead zone' at the bottom where j does nothing. Verify the height accounting matches between clampCursor and renderStats.
🟡 **cli/internal/tui/dashboard.go:844** — The start/end clamp block (start = d.cursor; clamp to len(lines)-maxVisible; clamp to 0; end = start+maxVisible; clamp to len(lines)) is duplicated verbatim in renderConfig and renderStats, and a third visible-line calculation lives in clampCursor. Extract a small helper (e.g. scrollWindow(offset, total, visible) (start, end int)) so the three sites can't drift.
🟡 **cli/internal/tui/dashboard.go:547** — isScrollOffsetTab encodes tab semantics as a list of tab constants. As more read-only tabs are added this will silently miss them. Consider attaching the scroll-mode to the tab definition itself (e.g. a field on the tab descriptor) rather than a hard-coded enum check, or at minimum add a comment near the tab constants pointing at this function.
🟡 **cli/internal/tui/dashboard.go:549** — If visibleRange is no longer used by Stats/Config and isn't called elsewhere, it should be removed; if it is still used by other tabs, fine. Worth a quick grep to avoid leaving dead code.

### Suggestions

- Extract a shared scrollWindow(offset, total, visible) helper and use it in renderConfig, renderStats, and clampCursor so the three computations cannot diverge.
- Add a unit test that asserts pressing j at the last visible line of Stats/Config advances the viewport by one line and that further j presses past EOF are no-ops (the bug this PR fixes).
- Confirm Home/End / PageUp / PageDown keys (if any) still behave correctly under the new offset semantics — a cursor that used to be a row index may now under/overshoot.

---

Muriano (approved):

Small, focused UX fix that switches Stats/Config tabs to treat the cursor as a scroll offset. Logic is correct and bounds are clamped defensively. Minor duplication and a small inconsistency between clamp and render for the Stats header.

### Suggestions

- Extract the `start = cursor; clamp to [0, len-maxVisible]; end = start+maxVisible` block into a shared helper (e.g. `scrollRange(cursor, total, maxVisible)`) since renderConfig and renderStats now duplicate it verbatim.
- Consider adding a unit test for clampCursor covering both tab modes (selectable vs scroll-offset) to lock in the new branch behavior.
- The magic `-= 2` for the Config header could be a named constant (e.g. configHeaderLines) to make the intent self-documenting and keep clamp/render in sync.

---

Muriano:
## ⚠️ MEDIUM Issue

Inconsistent visible-area calculation between clampCursor and renderStats: clampCursor only subtracts 2 (header+separator) when activeTab == tabConfig, but renderStats also reserves lines for its own header (the diff shows `maxVisible < 1` guard implying a prior subtraction). If renderStats reserves header lines, clampCursor will allow a cursor value larger than what is actually scrollable, leaving the last lines unreachable or causing the render-side clamp to silently reset cursor on next interaction. Fix: mirror the same `visible -= N` adjustment for tabStats inside clampCursor (or factor a shared helper that both clamp and render call).

**Location:** `cli/internal/tui/dashboard.go` line 556

---

sergiotejon:
@Muriano @vbuenog @ivanmunozruiz — Aprovecho esta PR para plantear algo que vengo notando: cambios pequeños y localizados como este (solo afecta al CLI, cero impacto en el daemon) fuerzan un bump de versión del monorepo completo y un ciclo de release que reconstruye daemon + CLI + Flutter app + Docker image.

¿Tendría sentido separar las aplicaciones cliente (CLI, Flutter app) en su propio repo con versionado y pipeline de release independientes? Así:

- **Cambios solo-CLI** (como este) no generan una release del daemon ni rebuilds de la imagen Docker
- **Cambios solo-daemon** (como el #385 de `requested_reviewers`) no rebuildan el CLI ni la app
- Cada componente tiene su propio semver, changelog y cadencia de release

No es urgente — para cambios tan simples como este no merece la pena mover nada. Pero a medida que el proyecto crece, el acoplamiento de releases empieza a pesar. ¿Qué opináis?

PR #391, opened 2026-05-05
feat: implement #390 — bug: archived repos not purged from monitored list

### LINKED ISSUE #390: bug: archived repos not purged from monitored list
## Problem

Archived repos remain in the monitored list after being archived on GitHub. The discovery filter (`archived:false`) only applies to **new** discoveries — repos already in the SQLite store are never re-checked.

**Example:** `freepik-company/ai-ilb` is archived but still appears in `heimdallm-cli repos` with 3 PRs, wasting API calls and poll cycles.

## Expected behavior

Heimdallm should periodically check whether monitored repos are still active and remove archived (or deleted) ones from the working set.

## Suggested approach

In the discovery/poll cycle, after merging static + topic-discovered repos, verify each repo's `archived` status via the GitHub API and drop archived ones from the active set. Optionally log a warning when a previously monitored repo is detected as archived.

sergiotejon:
Auto-generated by Heimdallm in response to #390.

Closes #390

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -552,7 +552,7 @@ func main() {
 			}
 
 			// Publish initial repos immediately
-			sendDiscoveryRepos(ctx, discoverySvc, limiter, repoPublisher, tier1ConfigFn)
+			sendDiscoveryRepos(ctx, discoverySvc, limiter, repoPublisher, tier1ConfigFn, ghClient)
 
 			ticker := time.NewTicker(discoveryInterval)
 			defer ticker.Stop()
@@ -561,7 +561,7 @@ func main() {
 				case <-ctx.Done():
 					return
 				case <-ticker.C:
-					sendDiscoveryRepos(ctx, discoverySvc, limiter, repoPublisher, tier1ConfigFn)
+					sendDiscoveryRepos(ctx, discoverySvc, limiter, repoPublisher, tier1ConfigFn, ghClient)
 				}
 			}
 		}()
@@ -1708,6 +1708,7 @@ func sendDiscoveryRepos(
 	limiter *scheduler.RateLimiter,
 	pub scheduler.Tier1Publisher,
 	configFn func() scheduler.Tier1Config,
+	archiveChecker discovery.ArchivedChecker,
 ) {
 	cfg := configFn()
 	var discovered []string
@@ -1725,6 +1726,15 @@ func sendDiscoveryRepos(
 	}
 
 	repos := discovery.MergeRepos(cfg.StaticRepos, discovered, cfg.NonMonitored)
+
+	if archiveChecker != nil {
+		active, archived := discovery.FilterArchived(repos, archiveChecker, discovered)
+		for _, r := range archived {
+			slog.Warn("tier1: dropping archived/deleted repo from active set", "repo", r)
+		}
+		repos = active
+	}
+
 	slog.Info("tier1: discovery complete", "repos", len(repos))
 	if err := pub.PublishRepos(ctx, repos); err != nil {
 		slog.Error("tier1: publish repos failed", "err", err)
--- a/daemon/cmd/heimdallm/main_discovery_tier1_test.go
+++ b/daemon/cmd/heimdallm/main_discovery_tier1_test.go
@@ -67,6 +67,7 @@ func TestSendDiscoveryReposRefreshesTopicBeforePublish(t *testing.T) {
 				DiscoveryOrgs:  []string{"org"},
 			}
 		},
+		nil,
 	)
 
 	if disc.refreshCalls != 1 {
@@ -101,6 +102,7 @@ func TestSendDiscoveryReposSkipsRefreshWhenTopicDisabled(t *testing.T) {
 				StaticRepos: []string{"org/static"},
 			}
 		},
+		nil,
 	)
 
 	if disc.refreshCalls != 0 {
@@ -131,6 +133,7 @@ func TestSendDiscoveryReposPublishesCachedReposWhenRefreshFails(t *testing.T) {
 				DiscoveryOrgs:  []string{"org"},
 			}
 		},
+		nil,
 	)
 
 	if disc.refreshCalls != 1 {
--- a/daemon/internal/discovery/discovery.go
+++ b/daemon/internal/discovery/discovery.go
@@ -172,6 +172,42 @@ func MergeRepos(static, discovered, nonMonitored []string) []string {
 	return out
 }
 
+// ArchivedChecker verifies whether a single repository is archived or deleted.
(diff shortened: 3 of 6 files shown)


Muriano:
## ⚠️ MEDIUM Issue

IsRepoArchived does not accept a context.Context. The discovery loop passes ctx into sendDiscoveryRepos but the per-repo archive check cannot be cancelled or deadlined; on shutdown or slow GitHub responses these calls will keep running until the underlying HTTP client's own timeout. Recommended fix: change the ArchivedChecker interface and Client method to IsRepoArchived(ctx context.Context, repo string) and thread ctx through c.do.

**Location:** `daemon/internal/github/client.go` line 451

---

Muriano:
## ⚠️ MEDIUM Issue

FilterArchived performs one synchronous GitHub API call per non-discovered repo on every discovery tick, with no caching or TTL. For deployments with many static repos and a short discoveryInterval this multiplies API consumption and tail latency of sendDiscoveryRepos linearly with the static repo set, and can hit secondary rate limits. Recommended fix: cache archive status with a TTL (e.g. a few hours) keyed by repo, or batch via GraphQL, so a steady-state tick performs ~0 API calls.

**Location:** `daemon/cmd/heimdallm/main.go` line 1735

---

Muriano:
## ⚠️ MEDIUM Issue

Treating any HTTP 404 as 'archived=true' will permanently drop repos that return 404 for reasons unrelated to deletion — most notably a token that lost access to a private repo, an org SSO re-auth requirement, or a temporary GitHub edge cache miss. Combined with the fact that 404s are not subject to the fail-open path used for other errors, a transient access issue silently purges the repo from the active set. Recommended fix: only treat 404 as archived after a confirmation (e.g. second attempt or counter of consecutive 404s), or distinguish 'deleted' from 'access denied' via /orgs/{org} reachability before purging.

**Location:** `daemon/internal/github/client.go` line 461

---

Muriano (approved):

Solid implementation with fail-open semantics, dedicated checker interface, and good unit test coverage. Main concerns are performance (N API calls per discovery tick for static repos with no caching) and lack of context propagation in IsRepoArchived which prevents cancellation and timeouts.

### Suggestions

- Add a metric/counter for archived-or-deleted repos dropped per tick to make this observable in production.
- Consider including the org in the warn log ('repo', r is fine but adding 'reason', 'archived|404' would help debugging).
- Add an integration-level test in main_discovery_tier1_test.go that exercises sendDiscoveryRepos with a non-nil archiveChecker to lock in the wiring.
- FilterArchived could pre-size the active/archived slices with make([]string, 0, len(repos)) to avoid repeated growth.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Adds GitHub archive detection so archived/deleted repos are filtered from the monitored set on each discovery cycle. Implementation is clean: well-scoped interface, fail-open on transient errors, skipCheck optimization for topic-discovered repos (which already filter archived:false), and good test coverage on both the discovery filter and the GitHub client method.

### Issues

🟡 **daemon/internal/github/client.go:462** — 404 is treated as 'archived=true', but a 404 can also occur for transferred repos or repos the token has lost access to (org membership/permission changes). In those cases the repo gets silently purged from the monitored set with only a 'dropping archived/deleted' warn log, which can be misleading during incident triage. Consider distinguishing 404 from archived in the warn log (e.g. include the reason) so operators can tell why a repo disappeared.
🟡 **daemon/cmd/heimdallm/main.go:1735** — FilterArchived issues one GET /repos/{owner}/{repo} per non-discovered repo on every discovery tick. For deployments with many static repos and a short discoveryInterval, this can consume a noticeable share of the GitHub REST quota. Not blocking, but worth noting in the PR description / docs, or considering a lower-frequency check (e.g. every Nth tick) in a follow-up.

### Suggestions

- Consider logging the underlying reason (archived vs 404) when dropping a repo, so operators can disambiguate archive vs deleted/transferred/permission-loss.
- If static-repo lists grow large, consider running the archive check less frequently than the main discovery interval to conserve REST quota.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid implementation with good fail-open semantics, clear separation of concerns, and thorough tests. However, three legitimate concerns raised by the existing AI review remain unaddressed: missing context.Context propagation, no caching/TTL leading to N API calls per tick, and 404 being conflated with 'archived' which can purge repos on transient access issues.

### Issues

⚠️ **daemon/internal/github/client.go:451** — IsRepoArchived does not accept context.Context. The caller passes ctx into sendDiscoveryRepos but per-repo archive checks cannot be cancelled or deadlined on shutdown. Change signature to IsRepoArchived(ctx context.Context, repo string) and thread through c.do; also update the ArchivedChecker interface and FilterArchived accordingly.
⚠️ **daemon/cmd/heimdallm/main.go:1735** — FilterArchived issues one synchronous GitHub API call per non-discovered repo on every discovery tick with no caching. With many static repos and a short discoveryInterval this scales linearly, increases tail latency of sendDiscoveryRepos, and risks secondary rate limits. Add a TTL cache (hours) keyed by repo, or batch via GraphQL, so steady-state ticks perform ~0 API calls.
⚠️ **daemon/internal/github/client.go:461** — Treating any 404 as archived=true will permanently drop repos for non-deletion reasons: token losing access to a private repo, org SSO re-auth required, or transient GitHub edge cache miss. Unlike other errors, 404 bypasses the fail-open path so a transient access issue silently purges the repo. Require N consecutive 404s before purging, or distinguish 'deleted' from 'access denied' (e.g. probe /orgs/{org}) before treating as archived.
🟡 **daemon/internal/discovery/discovery.go:184** — FilterArchived performs sequential per-repo lookups. For larger static repo sets, even with caching this could be parallelized (bounded worker pool) to reduce tick latency. Consider after caching is added.
🟡 **daemon/internal/github/client.go:461** — The body is read via io.ReadAll before checking the status code, but on 404 the body is discarded. Minor: the LimitReader read is wasted on the 404 path. Reorder status check before reading the body for 404.

### Suggestions

- Thread context.Context through ArchivedChecker.IsRepoArchived and the underlying do() call to support cancellation and per-call deadlines.
- Add a TTL-based cache for archive status (e.g. 6-24h) so steady-state discovery ticks make ~0 API calls; archived state changes infrequently.
- Disambiguate 404 (deleted) from 403/SSO/access-revoked: either probe the parent /orgs/{org} or require N consecutive 404s before treating a repo as archived, to avoid silent purges on access regressions.
- Consider also pulling 'disabled' from the repo response — disabled repos behave similarly to archived (no pushes/PRs) and may also be worth dropping.
- Tests are good; add a test for sendDiscoveryRepos integration with a non-nil ArchivedChecker so the wiring change in main.go is covered end-to-end.

---

PR #400, opened 2026-05-07
feat: auto-clone repo context for agents

### LINKED ISSUE #394: feat: auto-clone repo to temp directory for agent context
## Context

Several pipeline stages (refinement, development) need access to the full repository to explore code, run git blame, read architecture, etc. Currently this requires the user to map a \`local_dir\` per repo. This proposal adds automatic cloning as a fallback.

## Behavior

1. If the user has configured \`local_dir\` for a repo → use that path (existing behavior)
2. If no \`local_dir\` is set → clone the repo to a default directory:
   - User-configurable base path via config: \`clone_dir\` (per-repo, per-org, global)
   - Default: a \`heimdallm/\` directory inside the OS temp directory (\`os.TempDir()/heimdallm/<org>/<repo>\`)
3. Clone is shallow (\`--depth=1\`) for speed unless the stage needs git history (refinement/triage need blame → full clone or \`--unshallow\` on demand)
4. Reuse existing clones if present — \`git pull\` to update rather than re-clone

## Agent Context Passing

Always pass the repo directory (whether mapped or auto-cloned) to the AI agent via the preferred mechanism:

| CLI | Flag |
|-----|------|
| Claude | \`--directory <path>\` |
| Gemini | \`--cwd <path>\` |
| Codex | \`--cwd <path>\` |
| OpenCode | TBD |

## Cleanup

- Temp clones should respect \`retention.max_days\` — delete clones for repos no longer monitored after the retention period
- Add a \`make clean-clones\` / API endpoint to manually purge all temp clones

## Config

New fields (per-repo > per-org > global hierarchy):
- \`clone_dir\` — base path for auto-cloned repos (default: OS temp dir)

## Related

- Required by refinement stage (#393) and improved triage (#392)
- Replaces the need for manual \`local_dir\` setup in many cases

ivanmunozruiz:
## Summary

- Add a repo-context manager that resolves `local_dir`, reuses read-only `local_dir_base` context, and auto-clones missing repos into managed shallow checkouts under `clone_dir` or `os.TempDir()/heimdallm/<org>/<repo>`.
- Protect managed checkouts with per-repo `Acquire`/`Release` locks and a `.heimdallm-managed` ownership marker before any `fetch/reset/clean` or cleanup.
- Wire repo context into PR review, issue triage, and auto-implement execution paths while keeping `GET /config` display-only resolution pure.
- Add authenticated manual clone cleanup via `DELETE /config/clones/{repo}` and document the behavior.

## Behavior

- PR review and issue triage degrade to diff-only / no local context if clone preparation fails.
- `auto_implement` fails visibly if a writable checkout cannot be prepared.
- `local_dir` remains operator-owned and can be used for write flows; `local_dir_base` remains read-only shared context.

Closes #394.

## Validation

- `make test-docker`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ endif
         release-local package-macos install-service verify-linux run-linux \
         install-linux uninstall-linux \
         setup up up-build up-daemon up-build-daemon down logs logs-daemon \
-        ps restart clean _check-docker _check-env _check-linux _post-up-hints
+        ps restart clean clean-clones _check-docker _check-env _check-linux _post-up-hints
 
 # ── Build ─────────────────────────────────────────────────────────────────────
 
@@ -104,6 +104,14 @@ dev-stop:
 	   kill "$$UI_PID" 2>/dev/null && echo "↓  UI parada (PID $$UI_PID)" || true; \
 	   rm -f "$$UI_PID_FILE"; \
 	 fi
+
+clean-clones:
+	@: "$${HEIMDALLM_API_TOKEN:?set HEIMDALLM_API_TOKEN to the daemon API token}"
+	@HEIMDALLM_SERVER_URL="$${HEIMDALLM_SERVER_URL:-http://localhost:7842}"; \
+	  curl -fsS -X DELETE \
+	    -H "X-Heimdallm-Token: $$HEIMDALLM_API_TOKEN" \
+	    "$$HEIMDALLM_SERVER_URL/config/clones"; \
+	  echo
 
 # ── Local release (macOS only: sign + notarize + DMG + GitHub release) ───────
 #
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -280,16 +281,47 @@ func main() {
 	srv := server.New(s, broker, p, apiToken)
 	srv.SetNATSConn(eventBus.Conn())
 	srv.SetConfigPath(cfgPath)
-	srv.SetCleanCloneFn(func(repo string) error {
-		cfgMu.Lock()
-		aiCfg := cfg.AIForRepo(repo)
-		cfgMu.Unlock()
-		return repoCtx.Purge(context.Background(), repo, aiCfg.CloneDir)
-	})
 	shutdownReq := make(chan struct{}, 1)
 
 	// discoverySvc holds the discovered repo cache.
 	discoverySvc := discovery.NewService(ghClient)
+
+	srv.SetCleanCloneFn(func(ctx context.Context, repo string) error {
+		cfgMu.Lock()
+		aiCfg := cfg.AIForRepo(repo)
+		cfgMu.Unlock()
+		return repoCtx.Purge(ctx, repo, aiCfg.CloneDir)
+	})
+	srv.SetCleanClonesFn(func(ctx context.Context) (int, error) {
+		cfgMu.Lock()
+		cfgSnap := cfg
+		cfgMu.Unlock()
+		return purgeAllManagedClones(ctx, repoCtx, cfgSnap)
+	})
+
+	runCloneRetention := func(reason string) {
+		cfgMu.Lock()
+		cfgSnap := cfg
+		var discovered []string
+		if cfg.GitHub.DiscoveryTopic != "" {
+			discovered = discoverySvc.Discovered()
+		}
+		cfgMu.Unlock()
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+		defer cancel()
+		removed, err := purgeStaleManagedClones(ctx, repoCtx, cfgSnap, discovered)
+		if err != nil {
+			slog.Warn("clone retention purge failed", "reason", reason, "err", err)
+			return
+		}
+		if removed > 0 {
+			slog.Info("clone retention purge removed managed clones", "reason", reason, "removed", removed)
+		}
+	}
+	runCloneRetention("startup")
+	clonePurge := scheduler.New(24*time.Hour, func() { runCloneRetention("periodic") })
+	clonePurge.Start()
+	defer clonePurge.Stop()
 
 	// loginMu guards cachedLogin against concurrent reads/writes from the
 	// poll cycle and HTTP goroutines.
@@ -3266,6 +3298,80 @@ func issueTrackingOverrideMap(ov *config.IssueTrackingOverride) map[string]any {
 	}
 	if ov.Assignees != nil {
 		out["assignees"] = ov.Assignees
+	}
+	return out
+}
+
+func purgeAllManagedClones(ctx context.Context, manager *repoctx.Manager, cfg *config.Config) (int, error) {
+	if manager == nil {
+		return 0, fmt.Errorf("repoctx: nil manager")
+	}
+	var total int
+	var errs []error
+	for _, cloneDir := range managedCloneDirs(cfg) {
+		report, err := manager.PurgeAll(ctx, cloneDir)
+		total += report.Removed
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return total, errors.Join(errs...)
+}
+
+func purgeStaleManagedClones(ctx context.Context, manager *repoctx.Manager, cfg *config.Config, discovered []string) (int, error) {
+	if manager == nil {
+		return 0, fmt.Errorf("repoctx: nil manager")
+	}
+	if cfg == nil || cfg.Retention.MaxDays <= 0 {
+		return 0, nil
+	}
+	monitored := monitoredRepoSet(cfg, discovered)
+	var total int
+	var errs []error
+	for _, cloneDir := range managedCloneDirs(cfg) {
+		report, err := manager.PurgeStale(ctx, cloneDir, monitored, cfg.Retention.MaxDays)
+		total += report.Removed
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return total, errors.Join(errs...)
+}
+
+func managedCloneDirs(cfg *config.Config) []string {
+	seen := map[string]struct{}{}
+	add := func(path string) {
+		seen[strings.TrimSpace(path)] = struct{}{}
+	}
+	add("")
+	if cfg != nil {
+		add(cfg.AI.CloneDir)
+		for _, org := range cfg.AI.Orgs {
+			add(org.CloneDir)
+		}
+		for _, repo := range cfg.AI.Repos {
+			add(repo.CloneDir)
+		}
+	}
+	out := make([]string, 0, len(seen))
+	for path := range seen {
+		out = append(out, path)
+	}
+	sort.Strings(out)
+	return out
+}
+
+func monitoredRepoSet(cfg *config.Config, discovered []string) map[string]struct{} {
+	out := map[string]struct{}{}
+	if cfg == nil {
+		return out
+	}
+	repos := discovery.MergeRepos(cfg.GitHub.Repositories, discovered, cfg.GitHub.NonMonitored)
+	for _, repo := range repos {
+		repo = strings.TrimSpace(repo)
+		if repo != "" {
+			out[repo] = struct{}{}
+		}
 	}
 	return out
 }
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/heimdallm/daemon/internal/keychain"
 	"github.com/heimdallm/daemon/internal/notify"
 	"github.com/heimdallm/daemon/internal/pipeline"
+	"github.com/heimdallm/daemon/internal/repoctx"
 	"github.com/heimdallm/daemon/internal/scheduler"
 	"github.com/heimdallm/daemon/internal/server"
 	"github.com/heimdallm/daemon/internal/sse"
@@ -210,6 +211,7 @@ func main() {
 	notifier := notify.New()
 	ghClient := gh.NewClient(token)
 	exec := executor.New()
+	repoCtx := repoctx.NewManager()
 
 	// Load or create the per-daemon API token.  All mutating HTTP endpoints
 	// require this token in X-Heimdallm-Token (security issue #3).
@@ -271,14 +273,20 @@ func main() {
 	}
 	issueFetcher := issuepipeline.NewFetcher(ghClient, ghClient, s, issuePipe)
 	issueFetcher.SetBotLogin(resolvedBotLogin) // break re-triage loop (#362)
-	srv := server.New(s, broker, p, apiToken)
-	srv.SetNATSConn(eventBus.Conn())
-	srv.SetConfigPath(cfgPath)
-	shutdownReq := make(chan struct{}, 1)
-
 	// cfgMu protects cfg and the pipeline so reload is safe from any goroutine.
 	var cfgMu sync.Mutex
 	var reloadMu sync.Mutex // serialises config reloads to prevent duplicate pipelines
+
+	srv := server.New(s, broker, p, apiToken)
+	srv.SetNATSConn(eventBus.Conn())
+	srv.SetConfigPath(cfgPath)
+	srv.SetCleanCloneFn(func(repo string) error {
+		cfgMu.Lock()
+		aiCfg := cfg.AIForRepo(repo)
+		cfgMu.Unlock()
+		return repoCtx.Purge(context.Background(), repo, aiCfg.CloneDir)
+	})
+	shutdownReq := make(chan struct{}, 1)
 
 	// discoverySvc holds the discovered repo cache.
 	discoverySvc := discovery.NewService(ghClient)
@@ -481,6 +489,7 @@ func main() {
 		pipeline:             p,
 		issuePipe:            issuePipe,
 		fetcher:              issueFetcher,
+		repoCtx:              repoCtx,
 		store:                s,
 		broker:               broker,
 		cfgMu:                &cfgMu,
@@ -636,7 +645,14 @@ func main() {
 		aiCfg := c.AIForRepo(pr.Repo)
 		localDirBase := c.GitHub.LocalDirBase
 		cfgMu.Unlock()
-		aiCfg.LocalDir = config.ResolveLocalDir(aiCfg.LocalDir, pr.Repo, localDirBase)
+		repoHandle, err := acquireRepoContext(ctx, repoCtx, pr.Repo, &aiCfg, localDirBase, token, repoctx.ModeRead)
+		if err != nil {
+			logRepoContextFallback("review-worker", pr.Repo, err)
+			aiCfg.LocalDir = ""
+		}
+		if repoHandle != nil {
+			defer repoHandle.Release()
+		}
 
 		rev := runReview(pr, aiCfg)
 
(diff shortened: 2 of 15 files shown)


Muriano:
## ⚠️ MEDIUM Issue

In tier2Adapter.ProcessRepo, when issue.Mode == IssueModeDevelop and acquireRepoContext fails, the code logs/publishes an SSE error and sets aiCfg.LocalDir = "" but does NOT skip the issue. Execution falls through to building RunOptions with RequireWorkDirForDevelop: true and empty WorkDir, so issuePipe.Run will fail again with 'requires local repo context'. This produces duplicate error reporting and wasted setup. Impact: noisy logs/SSE, conflicting failure semantics across workers (the implement-worker in main() returns early on the same condition — see lines ~895-905). Fix: after publishing the SSE error in the develop branch, `continue` to the next issue (or skip building opts/calling Run) to mirror implement-worker behavior.

**Location:** `daemon/cmd/heimdallm/main.go` line 2438

---

Muriano:
## ⚠️ MEDIUM Issue

ensureManagedClone uses --depth=1 for both ModeRead and ModeWrite. For ModeWrite (auto_implement), the pipeline branches, commits, and pushes; many git operations (merge-base against default branch, rebase, log/diff against history, push of branches whose base is older than depth=1) silently misbehave or fail on shallow clones. EnsureFullHistory exists but is never invoked in this PR. Impact: auto_implement may produce broken PRs or fail at push time on managed clones. Fix: call manager.EnsureFullHistory for ModeWrite handles before returning (or unshallow inline when Mode == ModeWrite), and add a test covering the unshallow path on auto_implement.

**Location:** `daemon/internal/repoctx/manager.go` line 215

---

Muriano:
## ⚠️ MEDIUM Issue

acquireRepoContext lazily creates a fresh Manager when `manager == nil` (`if manager == nil { manager = repoctx.NewManager() }`). This silently hides a wiring bug: if the global repoCtx ever becomes nil (e.g., a future refactor), each call gets its own Manager with its own lock map, defeating per-repo serialization and risking concurrent clones/resets of the same target dir on disk. Impact: race window where two goroutines fetch+reset the same managed clone simultaneously, corrupting working tree mid-execution. Fix: treat nil manager as a programmer error — return an error or panic — instead of constructing a throwaway Manager.

**Location:** `daemon/cmd/heimdallm/main.go` line 3055

---

Muriano (approved):

Solid feature adding managed auto-clone with proper locking, marker validation, and token handling via askpass. Two correctness issues stand out: (1) the tier2 develop error path publishes an SSE error but does not skip the issue, so the pipeline runs anyway and fails again with RequireWorkDirForDevelop, producing duplicate errors and wasted work; (2) ModeWrite checkouts are shallow (--depth=1) and EnsureFullHistory is never called, which can break operations that need history (e.g., merge-base, rebase) during auto_implement.

### Suggestions

- buildAskPassEnv writes the token only to env (good), but the helper script lives in a temp dir per call. Consider reusing a single per-process helper script to reduce mkdir/chmod churn on hot paths.
- validateMarker rejects markers whose Repo field doesn't match — good — but writeMarker uses 0o644. Consider 0o600 for consistency with the askpass dir's 0o700, since the marker reveals repo identity on shared hosts.
- ProcessRepo captures `releaseRepoContext` as `repoHandle.Release` only on success; on error it stays nil. Consider always setting it (with a no-op) to simplify the pipeline contract.
- TestPipeline_AutoImplementProceedsWhenStateCheckFails diff includes whitespace-only realignment of struct field padding — pulls reviewer attention away from substantive changes; revert if not needed.
- execGit.Run wraps stderr but discards stdout. For `git fetch --unshallow` and similar, surfacing stdout in the error message can speed up triage.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid feature: introduces a managed auto-clone subsystem with per-repo locking, marker-file ownership, and ModeRead/ModeWrite policy. Code is well-structured and thoughtfully tested. Main concerns are a potential lock leak in tier2Adapter.ProcessRepo when the handle release is delegated to the pipeline, and a no-op nil-receiver self-reassignment pattern in Manager methods.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2438** — In tier2Adapter.ProcessRepo, on success the handle release is assigned to opts.ReleaseRepoContext rather than deferred. If any code between Acquire and issuePipe.Run panics, returns early, or if Run is never invoked due to a future refactor, the per-repo lock and managed clone reference leak. Other call sites use `defer repoHandle.Release()` directly. Either also `defer repoHandle.Release()` here (pipeline release becomes a second no-op via sync.Once) or guarantee Run runs unconditionally after acquisition.
🟡 **daemon/internal/repoctx/manager.go:132** — Manager.Acquire and Manager.Purge use `if m == nil { m = NewManager() }`. Reassigning a method receiver does not affect the caller's pointer, and the ephemeral Manager has its own lock map, so concurrent calls with nil receivers would not actually serialize. The helper acquireRepoContext already substitutes a fresh Manager when nil; consider either removing the self-heal here (return an error) or documenting it as a defensive no-op so callers don't rely on it.
🟡 **daemon/cmd/heimdallm/main.go:3055** — acquireRepoContext mutates aiCfg.LocalDir through the pointer parameter to the managed clone path, but that path is only valid until Handle.Release runs (Purge can delete it). The current call sites don't read aiCfg.LocalDir after Release, but this contract is implicit. Consider either a comment on acquireRepoContext explaining the lifetime coupling or returning the path explicitly so the lifetime is visible at call sites.
🟡 **daemon/internal/repoctx/manager.go:380** — buildAskPassEnv exposes the token via HEIMDALLM_GIT_TOKEN in the child process env so any subprocess git spawns inherits it. This is reasonable for git's expected helpers, but unsetting the helper script and env entries after each git invocation rather than reusing one cleanup at the end of clone/fetch would minimize the window. Minor; current code already removes the script dir on cleanup.
🟡 **daemon/internal/executor/executor.go:322** — ValidateWorkDir uses `strings.HasPrefix(abs, tempAbs+"/")` with a hardcoded forward slash. Fine for Linux/macOS daemon, but if this code ever runs on Windows the separator would differ. Use string(filepath.Separator) or filepath.Rel-based containment for portability.
🟡 **daemon/internal/repoctx/manager.go:245** — updateManagedClone runs `git remote set-url` without the askpass env, then `git fetch` with it. set-url itself doesn't need credentials so this is functionally fine, but if a future change adds a credential-using step here, the asymmetry is easy to miss. Consider passing env consistently or commenting why set-url is intentionally credential-less.

### Suggestions

- In tier2Adapter.ProcessRepo, mirror the implement-worker pattern: `defer repoHandle.Release()` immediately after Acquire, and rely on sync.Once in Handle.Release to make the pipeline's deferred release a safe no-op. This eliminates the leak window.
- Add a unit test that exercises Acquire→Release ref counting under concurrent goroutines for the same repo, to lock in the per-repo serialization invariant.
- Consider documenting in the Manager package doc that managed clones are shallow by default and EnsureFullHistory must be invoked explicitly when full history is required (e.g., for blame, log-based prompts).

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Well-structured auto-clone feature with solid security fundamentals (askpass for tokens, marker-file safety, path-traversal checks). However, there are a few concurrency and correctness issues worth addressing before merge.

### Issues

⚠️ **daemon/internal/repoctx/manager.go:139** — Silent nil-manager fallback in Acquire (`if m == nil { m = NewManager() }`) creates a throwaway Manager with an isolated lock map, completely defeating per-repo locking. The same pattern appears in Purge. If the manager is nil, that's a caller bug — return an error instead of silently masking it.
⚠️ **daemon/cmd/heimdallm/main.go:3058** — Same issue in `acquireRepoContext`: `if manager == nil { manager = repoctx.NewManager() }` creates a one-off manager with no shared state. Two concurrent callers passing nil would bypass each other's locks. This should be `return nil, fmt.Errorf("repoctx: nil manager")` instead.
⚠️ **daemon/cmd/heimdallm/main.go:2438** — In `ProcessRepo`, when `acquireRepoContext` fails for develop-mode issues, the code publishes an SSE error event but does NOT skip the issue (no `continue`). Execution falls through, builds the full RunOptions with `RequireWorkDirForDevelop: true`, then the pipeline fails again with a duplicate error. Add a `continue` after the SSE publish block to avoid redundant work and double error reporting.
🟡 **daemon/internal/repoctx/manager.go:265** — The per-repo lock is exclusive (buffered channel of size 1), meaning ModeRead callers (review/triage) serialize against each other for the same repo. Under load, concurrent reviews of a popular repo will queue. Consider a reader-writer lock pattern, or at least document that this is intentional for v1.
🟡 **daemon/internal/repoctx/manager.go:230** — `updateManagedClone` calls `remote set-url origin <url>` unconditionally on every Acquire, even when the URL hasn't changed. This is a wasted subprocess call on the hot path. Consider checking the current remote URL first, or only setting it during initial clone.
🟡 **daemon/internal/repoctx/manager.go:228** — `buildAskPassEnv` is called inside `updateManagedClone`, which runs on every Acquire for an existing clone. The askpass helper creates a temp dir, writes a script, and cleans it up — for every single review/triage operation. This is fine for correctness but adds FS overhead per operation.
🔴 **daemon/internal/repoctx/manager.go:350** — `splitRepo` calls `config.ValidateOrgSlug(parts[0])` but this function doesn't exist in the current codebase — only the unexported `githubOrgPattern` regex. Either this function is added elsewhere in the PR (not visible in the truncated diff) or this won't compile.
🟡 **daemon/internal/issues/pipeline.go:288** — `ReleaseRepoContext` is deferred at the top of `Run`, meaning the repo lock is held for the entire pipeline execution including the full AI CLI invocation. For auto_implement this is necessary, but for review_only the lock could be released much earlier (after reading the repo). This compounds the serialization issue in the manager.
🟡 **daemon/internal/executor/executor.go:319** — The new `os.TempDir()` allowlist uses `resolvedAbs` which calls `EvalSymlinks`. If the temp dir doesn't exist yet (unlikely but possible in containers), `EvalSymlinks` returns an error and the check is silently skipped. The existing `/tmp` check above covers Linux, so the failure is non-critical, but worth a comment.

### Suggestions

- Replace the nil-manager guards in `Manager.Acquire`, `Manager.Purge`, and `acquireRepoContext` with explicit errors. Silent fallback to a throwaway Manager defeats the per-repo locking and hides caller bugs.
- Add a `continue` (or restructure the error handling) in `ProcessRepo`'s develop-mode branch after publishing the SSE error, to avoid falling through to the pipeline which will report a second error for the same failure.
- Consider releasing the repo context handle earlier in the review_only flow path, since the AI CLI only needs the files read at invocation time — holding the exclusive lock through the full Claude call blocks other reviews of the same repo.
- Add an integration test that verifies two concurrent Acquire calls for the same repo with ModeRead actually serialize (to document the current behavior), so future contributors understand the trade-off.
- The `updateManagedClone` always does `reset --hard` + `clean -fd` even for ModeRead. For read-only callers where the clone is already clean, a `git diff --quiet HEAD FETCH_HEAD` check could skip the expensive reset.

---

ivanmunozruiz (commented):
## :white_check_mark: LGTM

Implementación coincide con el plan acordado. Locks per-repo, marker `.heimdallm-managed`, validación de path/repo/org slug, separación `Mode==ModeWrite` rechaza `local_dir_base`, token via `GIT_ASKPASS` sin leak en argv, fallback degradado en read-paths y fallo visible en auto_implement. Tests cubren los escenarios sensibles (lock serialization, partial-clone cleanup, refuse-unmanaged, purge-only-managed, token no-leak, auth en endpoint).

CI: 4/4 verde.

### Observaciones no bloqueantes

1. **Per-issue fetch en `ProcessRepo`** (`main.go:2438-2461`). 10 issues del mismo repo → 10 `fetch+reset+clean`. Una caché "ya actualizado en este poll cycle" lo evitaría. Optimización para v2.

2. **Error response del endpoint leakea paths absolutos**. `Purge` returns `repoctx: clone target \"/var/folders/.../heimdallm/org/repo\" exists but is not managed`. El handler reenvía `err.Error()` literal. Info-leak menor (con auth token), pero conviene reescribir al borde HTTP.

3. **`cmd.Env = nil` en `set-url`/`reset`/`clean`**. Hereda `os.Environ()` (funciona). Por consistencia con las llamadas con env askpass, mejor `cmd.Env = os.Environ()` siempre. Cosmético.

4. **`gitTimeout = 3 * time.Minute` hardcoded**. Monorepos grandes pueden superarlo incluso shallow. Hacerlo configurable vía `[ai]` o env quedaría limpio. No urgente.

5. **Lock exclusivo en `ModeRead`**. Dos reviews concurrentes del mismo repo se serializan aunque ninguna escriba. Si se vuelve hot path, `RWMutex` lo arregla sin cambiar el contrato `Handle`.

6. **No hay LRU/retention sweep automático**. La issue #394 menciona "respect retention.max_days". Este PR sólo añade el `DELETE /config/clones/{repo}` manual. Si se interpreta como MVP cierra la issue; si no, sería issue de seguimiento.

7. **Falta test (nice-to-have)**: caso `splitRepo` falla con slug inválido — verificar que ningún lock queda residual. La lógica actual valida antes de `acquireRepoLock`, así que conceptualmente está bien.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review: all three prior medium findings have been addressed — nil manager now returns an error, develop-mode acquire failures now skip the issue via the new (RunOptions, bool) OptionsFn contract, and Acquire calls EnsureFullHistory for ModeWrite. The new release/ok plumbing in tier2 ProcessRepo is correct (releaseOnReturn flag transfers ownership to the pipeline via ReleaseRepoContext, which defers cleanup in pipeline.Run). No new confirmed issues.

### Suggestions

- Consider a test for the ModeWrite unshallow path in repoctx.Manager.Acquire to lock in the EnsureFullHistory wiring.
- ProcessRepo's optsFor closure has become long with intertwined defer/ownership-transfer logic; extracting the acquire+ownership-handoff into a helper would reduce the chance of a future refactor breaking the releaseOnReturn invariant.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of #400. The author addressed the prior MEDIUM and most LOW findings: tier2Adapter now skips develop-mode failures via the new (opts, ok) signature; nil Manager now returns an error instead of self-healing; Acquire invokes EnsureFullHistory for ModeWrite; ValidateWorkDir uses filepath.Separator via pathWithin; aiCfg.LocalDir lifetime is now documented. The new release pattern (deferred releaseOnReturn flipped to false on success, with pipeline.Run owning final Release) is reasonable; sync.Once on Handle.Release prevents double-release. Only minor residual concerns remain.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:280** — SetCleanCloneFn callback calls repoCtx.Purge with context.Background() — no timeout/cancel. Purge takes the per-repo lock and can block indefinitely behind an in-flight pipeline Run, tying up the HTTP handler. Consider plumbing the request context (or wrapping with a timeout) so a stuck Purge doesn't pin server goroutines.
🟡 **daemon/internal/repoctx/manager.go:184** — If EnsureFullHistory fails for a ModeWrite acquire after ensureManagedClone has created/updated the directory, Acquire returns (nil, err) and releases the lock, but the partially-prepared clone (still shallow) is left on disk. Next call will reuse it via the update path, which is fine, but a transient unshallow failure (e.g., token scope) will repeatedly fail until manually purged. Consider either retrying without --unshallow on update or surfacing a hint to call clean_clone.
🟡 **daemon/cmd/heimdallm/main.go:2456** — tier2Adapter.ProcessRepo's optsFor sets releaseRepoContext = repoHandle.Release on success but only flips releaseOnReturn = false at the very end. Between the assignment and the final flip there's option-building code that could panic; in that case the deferred releaseOnReturn=true path still fires, but the pipeline-side ReleaseRepoContext has also been planted in opts. sync.Once protects correctness, but the dual-ownership window is subtle. A short comment near `releaseOnReturn = false` explaining why double-release is safe (sync.Once) would help future readers.

### Suggestions

- Consider adding an integration test that exercises the ModeWrite path end-to-end and asserts the clone is unshallowed (no .git/shallow afterwards) — this is the highest-risk behavior change of the PR.
- Document in repoctx that managed clone directories survive failed Acquire attempts and are intended to be self-healed by subsequent updates or manually purged via clean_clone.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

The three MEDIUM findings from the previous review are properly addressed: nil manager now returns errors instead of creating throwaway instances, and ProcessRepo correctly skips develop-mode issues on context failure via the new (opts, ok) return signature. The ownership-transfer pattern for repo handles is correct but introduces two different lifetime management styles across workers. The diff is truncated so the previous HIGH finding about splitRepo/ValidateOrgSlug cannot be verified.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:936** — implement-worker manages handle lifetime via `defer repoHandle.Release()` in the worker closure, while ProcessRepo transfers ownership via `ReleaseRepoContext` in RunOptions (released by pipeline.Run). Both are correct, but two different lifetime patterns for the same resource in the same codebase invites future mistakes — a maintainer could remove the implement-worker's defer thinking ReleaseRepoContext handles it, or vice versa. Consider unifying on one pattern.
⚠️ **daemon/internal/repoctx/manager.go:350** — Previous HIGH finding: splitRepo reportedly calls config.ValidateOrgSlug which does not exist in the codebase (only the unexported githubOrgPattern regex exists). The diff is truncated so this cannot be verified as fixed. If this function still references ValidateOrgSlug, the package won't compile.
🟡 **daemon/internal/repoctx/manager.go:108** — Per-repo lock is exclusive (chan struct{} size 1), so concurrent ModeRead operations (reviews, triage) on the same repo serialize behind each other. Under load with a popular monorepo, this queues all review workers. Documented as intentional for v1, but worth a follow-up issue for reader-writer locking.

### Suggestions

- Verify that splitRepo compiles — the previous review flagged a reference to config.ValidateOrgSlug which doesn't exist in the codebase. The diff is truncated so this couldn't be confirmed as resolved.
- Unify handle lifetime management: either always transfer ownership via ReleaseRepoContext in RunOptions (letting pipeline.Run release it), or always use defer in the worker closure. Mixing both patterns works today but is fragile.
- Consider filing a follow-up issue for reader-writer per-repo locking to avoid serializing concurrent read-only operations on popular repos.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Previous findings addressed: nil manager now returns error, ModeWrite calls EnsureFullHistory to upgrade shallow clones, and ProcessRepo's OptionsFn now returns (opts, ok) with the fetcher honoring ok=false to skip develop-mode issues when context prep fails. New code (CLI flag detection, /config/clones DELETE endpoint, retention purge) is properly gated and tested. No new blocking issues.

### Suggestions

- ReleaseRepoContext is handed off via RunOptions; if the pipeline panics or returns early before invoking it, the per-repo lock and managed clone reservation could leak. Consider guarding the call with a defer in the pipeline's Run entry point and/or a sync.Once in the release closure.
- cliHelp caches the help string per resolved CLI path forever; if a user upgrades their CLI mid-process to gain --add-dir support, the daemon won't pick it up until restart. Acceptable for now, but worth a comment.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of PR #400. Most prior findings have been addressed: SetCleanCloneFn now accepts and passes through the request context; nil-manager case in acquireRepoContext now returns an error (with a dedicated test); develop-mode fall-through in tier2Adapter.ProcessRepo now returns (RunOptions{}, false) so the caller skips the issue. New executor work (CLI workdir flag detection + temp-dir allowance in ValidateWorkDir) looks reasonable, but the codex flag placement and a minor cache miss path are worth a closer look.

### Issues

🟡 **daemon/internal/executor/executor.go:540** — detectWorkDirFlags returns ['--cd', workDir] for codex, but buildArgs appends workDirFlags AFTER the 'exec' subcommand (args becomes 'exec --session-id ... --model ... --cd <dir>'). codex's --cd is a global flag that typically must precede the subcommand. Detecting it via top-level 'codex --help' but emitting it after 'exec' may cause codex to error or ignore the flag. Either insert workDirFlags before the subcommand for codex, or check 'codex exec --help' for the supported flag.
🟡 **daemon/internal/executor/executor.go:555** — cliHelp does not cache negative results: if `<cli> --help` times out or errors, every subsequent ExecuteRaw call with a WorkDir pays the full 2s timeout again. For a misconfigured/missing CLI under load this serializes a 2s delay per invocation. Consider caching a sentinel for failures (with maybe a TTL) so the negative path is bounded.
🟡 **daemon/cmd/heimdallm/main.go:2495** — Prior re-review note still applies: in tier2Adapter.ProcessRepo's optsFor, releaseRepoContext is planted into opts before releaseOnReturn is flipped to false at the very end. If any of the intermediate option-construction steps panicked, both the deferred handle.Release and pipeline-side ReleaseRepoContext would target the same handle. sync.Once protects correctness but a one-line comment near 'releaseOnReturn = false' explaining the dual-ownership-via-sync.Once invariant would help future readers.

### Suggestions

- For codex, verify --cd placement against the actual CLI by integration-testing with a real (or more faithful) fake codex; the current fake-CLI test only checks that the flag appears somewhere in argv, not that it precedes the 'exec' subcommand.
- Consider caching cliHelp failures (e.g., store an empty string or a 'not supported' marker) to bound repeated 2s --help probes when the CLI is unavailable or slow.

---

PR #401, opened 2026-05-07
feat: add lightweight issue triage heuristics

### LINKED ISSUE #392: feat: redefine triage as lightweight assessment with assignee heuristics
## Context

The current triage step classifies issues by label and dispatches them. This proposal redefines triage as a **lighter, faster evaluation** focused on:

1. **What's happening** — quick understanding of the issue
2. **Where the root cause likely is** — affected area of the codebase
3. **Priority** — severity / urgency assessment
4. **Who should own it** — assignee selection based on heuristics

## Assignee Heuristics

The triage agent must select who should handle the issue using the following criteria, in order:

1. **Bug introduced by a specific person** → assign to that person. Use `git blame` / commit history on the affected area to identify the author of the regression.
2. **New feature request** → assign to the person with the most changes in the presumably affected codebase area. Use `git shortlog` / file-level commit frequency to identify the domain expert.
3. **Neither can be determined** → assign to the person configured as the triage owner (new config field).

## Requirements

- Triage should be fast and cheap (smaller model or shorter prompt is fine)
- The output should be: a summary comment on the issue, priority label, and assignee
- New config field: `triage_owner` (per-repo, per-org, global) — the fallback assignee
- Must work with the repo cloned locally or via the auto-clone mechanism (see related issue)

## Related

- Part of the triage → refinement → development pipeline (see state machine issue)

ivanmunozruiz:
## Summary

- Redefine issue triage output with affected area, affected paths, priority label, owner rationale, confidence, and evidence.
- Resolve issue ownership from AI output plus daemon-side validation against git history when a repo workdir is available.
- Fall back to configured `triage_owner` when confidence/evidence is insufficient, history is shallow/unavailable, or no safe assignee can be selected.
- Apply priority labels and issue assignees best-effort without failing the triage run.
- Store the richer triage payload in the existing `issue_reviews.triage` JSON field for forward-compatible UI/API use.

Closes #392

## Validation

- `make test-docker`

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -1467,6 +1467,13 @@ func main() {
 
 	// Wire the issue-review trigger callback: re-run issue pipeline on a stored issue.
 	srv.SetTriggerIssueReviewFn(func(issueID int64) error {
+		// The HTTP handler queues this work in a goroutine and returns 202, so
+		// r.Context() would be cancelled as soon as the response is written.
+		// Use an explicit operation timeout instead of an unbounded background
+		// context so repo-context git operations remain cancellable.
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
+		defer cancel()
+
 		publishIssueErr := func(msg string) {
 			broker.Publish(sse.Event{
 				Type: sse.EventIssueReviewError,
@@ -1489,14 +1496,14 @@ func main() {
 		localDirBase := cfg.GitHub.LocalDirBase
 		globalTimeout := cfg.AI.ExecutionTimeout
 		cfgMu.Unlock()
-		repoHandle, err := acquireRepoContext(context.Background(), repoCtx, iss.Repo, &aiCfg, localDirBase, token, repoctx.ModeRead)
+		repoHandle, err := acquireRepoContext(ctx, repoCtx, iss.Repo, &aiCfg, localDirBase, token, repoctx.ModeRead)
 		if err != nil {
 			logRepoContextFallback("trigger issue review", iss.Repo, err)
 			aiCfg.LocalDir = ""
 		}
 		if repoHandle != nil {
 			defer repoHandle.Release()
-			ensureRepoContextFullHistory(context.Background(), repoCtx, repoHandle, token, "trigger issue review", iss.Repo)
+			ensureRepoContextFullHistory(ctx, repoCtx, repoHandle, token, "trigger issue review", iss.Repo)
 		}
 
 		// Reconstruct github.Issue from store data for the pipeline
@@ -1560,7 +1567,7 @@ func main() {
 		slog.Info("trigger issue review: running pipeline",
 			"store_issue_id", issueID, "repo", iss.Repo, "number", iss.Number)
 
-		_, err = issuePipe.Run(context.Background(), ghIssue, opts)
+		_, err = issuePipe.Run(ctx, ghIssue, opts)
 		if err != nil {
 			broker.Publish(sse.Event{Type: sse.EventIssueReviewError, Data: sseData(map[string]any{
 				"issue_id": issueID, "repo": iss.Repo, "error": err.Error(),
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -838,6 +838,7 @@ func main() {
 		}
 		if repoHandle != nil {
 			defer repoHandle.Release()
+			ensureRepoContextFullHistory(ctx, repoCtx, repoHandle, token, "triage-worker", msg.Repo)
 		}
 
 		extraFlags := agentCfg.ExtraFlags
@@ -870,6 +871,7 @@ func main() {
 			},
 			IssuePromptOverride:     issuePrompt,
 			IssueInstructions:       issueInstructions,
+			TriageOwner:             aiCfg.TriageOwner,
 			ImplementPromptOverride: implPrompt,
 			ImplementInstructions:   implInstructions,
 			PRReviewers:             aiCfg.PRReviewers,
@@ -970,6 +972,7 @@ func main() {
 			},
 			IssuePromptOverride:      issuePrompt,
 			IssueInstructions:        issueInstructions,
+			TriageOwner:              aiCfg.TriageOwner,
 			ImplementPromptOverride:  implPrompt,
 			ImplementInstructions:    implInstructions,
 			PRReviewers:              aiCfg.PRReviewers,
@@ -1493,6 +1496,7 @@ func main() {
 		}
 		if repoHandle != nil {
 			defer repoHandle.Release()
+			ensureRepoContextFullHistory(context.Background(), repoCtx, repoHandle, token, "trigger issue review", iss.Repo)
 		}
 
 		// Reconstruct github.Issue from store data for the pipeline
@@ -1543,6 +1547,7 @@ func main() {
 			},
 			IssuePromptOverride:     issuePrompt,
 			IssueInstructions:       issueInstructions,
+			TriageOwner:             aiCfg.TriageOwner,
 			ImplementPromptOverride: implPrompt,
 			ImplementInstructions:   implInstructions,
 			PRReviewers:             aiCfg.PRReviewers,
@@ -2502,6 +2507,9 @@ func (a *tier2Adapter) ProcessRepo(ctx context.Context, repo string) (int, error
 			}
 			aiCfg.LocalDir = ""
 		} else if repoHandle != nil {
+			if issue.Mode == config.IssueModeReviewOnly {
+				ensureRepoContextFullHistory(ctx, a.repoCtx, repoHandle, a.ghToken, "issue poll", issue.Repo)
+			}
 			releaseRepoContext = repoHandle.Release
 		}
 
@@ -2535,6 +2543,7 @@ func (a *tier2Adapter) ProcessRepo(ctx context.Context, repo string) (int, error
 			},
 			IssuePromptOverride:      issuePrompt,
 			IssueInstructions:        issueInstructions,
+			TriageOwner:              aiCfg.TriageOwner,
 			ImplementPromptOverride:  implPrompt,
 			ImplementInstructions:    implInstructions,
 			PRReviewers:              aiCfg.PRReviewers,
@@ -3124,6 +3133,16 @@ func acquireRepoContext(
 	// path.
 	aiCfg.LocalDir = h.Path()
 	return h, nil
+}
+
+func ensureRepoContextFullHistory(ctx context.Context, manager *repoctx.Manager, h *repoctx.Handle, token, scope, repo string) {
+	if manager == nil || h == nil {
+		return
+	}
+	if err := manager.EnsureFullHistory(ctx, h, token); err != nil {
+		slog.Warn(scope+": full git history unavailable; triage owner verification may fall back",
+			"repo", repo, "err", err)
+	}
 }
 
 func logRepoContextFallback(scope, repo string, err error) {
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -8,6 +8,7 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -27,6 +28,8 @@ const maxDiffBodyBytes = 10 * 1024 * 1024 // 10 MB for diffs
 // maxErrBodyLen limits the number of bytes included in error messages to avoid
 // leaking sensitive GitHub diagnostic information (e.g. token details).
 const maxErrBodyLen = 200
+
+var labelColorRE = regexp.MustCompile(`^[0-9A-Fa-f]{6}$`)
 
 // safeTruncate shortens s to at most max bytes, snapping on a rune boundary
 // so we never emit a split multi-byte UTF-8 character in an error message.
@@ -863,15 +866,20 @@ func (c *Client) FetchLabels(repo string) ([]string, error) {
 	return names, nil
 }
 
-// CreateLabel creates a repository label. A 422 is tolerated because GitHub
-// returns it when a racing process created the label after FetchLabels.
+// CreateLabel creates a repository label. A 422 is tolerated only when GitHub
+// reports an already-existing label, which can happen if another process wins
+// the race after FetchLabels.
 func (c *Client) CreateLabel(repo, name, color, description string) error {
 	if repo == "" || name == "" {
 		return nil
 	}
+	color = strings.TrimPrefix(color, "#")
+	if !labelColorRE.MatchString(color) {
+		return fmt.Errorf("github: invalid label color %q", color)
+	}
 	payload, err := json.Marshal(map[string]string{
 		"name":        name,
-		"color":       strings.TrimPrefix(color, "#"),
+		"color":       color,
 		"description": description,
 	})
 	if err != nil {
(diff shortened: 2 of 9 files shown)


Muriano:
## 🟡 LOW Issue

ensureRepoContextFullHistory is called with context.Background() in the trigger issue review path, losing cancellation/timeout propagation from the surrounding handler. Impact: a hung git fetch cannot be cancelled by the parent request lifecycle. Fix: thread the parent ctx (or a derived ctx with timeout) instead of context.Background().

**Location:** `daemon/cmd/heimdallm/main.go` line 1496

---

Muriano:
## ⚠️ MEDIUM Issue

contributorsForPaths builds 'git log --format=... --all -- <paths>' but does not pass '--' as a separator before user-derived affected_paths after potential refspec args. While '--all' precedes '--', any affected path beginning with '-' (after sanitization that only blocks leading '/') could still be interpreted as a flag by git. Impact: malformed/malicious LLM-suggested paths starting with '-' could alter git log behavior. Fix: in sanitizeAffectedPaths, also reject paths starting with '-' (or pass each path with explicit ':(literal)' pathspec magic).

**Location:** `daemon/internal/issues/triage_assignment.go`

---

Muriano (approved):

PR adds lightweight issue triage heuristics with priority labels, ownership verification via git history, and a triage_owner fallback. Logic is sound and well-tested with backward compatibility preserved. Minor concerns around git command argument parsing (pathspec separator) and goroutine context usage.

### Suggestions

- Consider adding a unit test exercising contributorsForPaths with shallow-repo fallback and noreply email parsing to lock in the regex behavior.
- Document the priorityLabelColorBySeverity color scheme choice (or move to config) so non-default themes don't surprise users.
- ensureIssueLabel ignores the case where FetchLabels returns the label with different casing — it returns 'existing' (good) but consider logging when a case-mismatch is detected for observability.

---

Muriano:
## 🟡 LOW Issue

enrichTriageResult is called before applyIssueTriageMetadata, but normalizeSeverity is also invoked inside parseIssueResult — severity is normalized twice. Not a bug, but indicates the normalization location is unclear. Low priority cleanup.

**Location:** `daemon/internal/issues/pipeline.go` line 521

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured addition of lightweight triage heuristics with good test coverage and graceful degradation. The git-history contributor resolution is a solid approach. A few issues around unbounded git log output, context propagation, and double normalization warrant attention.

### Issues

⚠️ **daemon/internal/issues/triage_assignment.go:139** — contributorsForPaths runs `git log --format=... --all -- <paths>` with no `--max-count` limit. On repos with long histories this can produce unbounded output, consuming significant memory and time (captureGit buffers stdout into a bytes.Buffer). Add a reasonable limit like `-n 500` or `--max-count=500` to cap the work while still producing meaningful contributor rankings.
🟡 **daemon/internal/issues/triage_assignment.go:139** — Using `--all` searches every branch's history. This means commits from abandoned feature branches or forks are counted equally, potentially skewing contributor rankings with irrelevant authorship. Consider using just the default branch (no `--all`) or `--first-parent` for more accurate ownership signal.
🟡 **daemon/internal/issues/triage_assignment.go:155** — contributorLoginsFromEmail falls back to the email local-part as a GitHub login guess. For corporate emails (e.g. `jsmith@company.com`), `jsmith` may not correspond to the actual GitHub login. This is inherently heuristic, but worth documenting or at minimum deprioritizing non-noreply-derived logins vs. noreply-derived ones (which are authoritative).
⚠️ **daemon/cmd/heimdallm/main.go:1499** — The `ensureRepoContextFullHistory` call in the trigger-issue-review handler uses `context.Background()` while the other three call sites use the ambient `ctx`. This means the git fetch cannot be cancelled if the parent operation is aborted. Should use the available context from the enclosing scope for consistency and cancellability.
🟡 **daemon/internal/issues/pipeline.go:931** — parseIssueResult now normalizes severity (lines 931-937 in the diff), and then enrichTriageResult normalizes severity again. The double normalization is harmless but adds maintenance burden — if the normalization logic changes, both sites must be updated. Consider keeping normalization only in enrichTriageResult and removing the added lines from parseIssueResult, or vice versa.
🟡 **daemon/internal/issues/pipeline.go:975** — IssueLabelCatalog is checked via type assertion (`gh.(IssueLabelCatalog)`) in ensureIssueLabel. This works but is a subtle contract: the real GitHub client implements it, but any issueGitHub implementation that doesn't will silently skip label creation with no log message. Consider logging when the type assertion fails so operators know why labels aren't being auto-created.
🟡 **daemon/internal/issues/triage_assignment.go:49** — enrichTriageResult unconditionally overwrites TentativeAssignee with SuggestedAssignee (`r.Triage.TentativeAssignee = r.Triage.SuggestedAssignee`). If the LLM independently sets TentativeAssignee to a different value, it's silently dropped. Either remove TentativeAssignee from the Triage JSON schema (so the LLM doesn't try to set it) or don't overwrite it.
🟡 **daemon/internal/github/client.go:884** — CreateLabel reads the response body but only uses it in the error path. The success path (StatusCreated) discards the body, which is fine. However, the body read happens before the 422 early-return, meaning the body IS read but the variable is unused on that path — minor but clean.

### Suggestions

- Add `--max-count=500` (or similar) to the `git log` call in contributorsForPaths to bound memory usage and execution time on large repositories.
- Use the ambient `ctx` instead of `context.Background()` in the trigger-issue-review handler for consistent cancellation behavior.
- Consider weighting noreply-derived GitHub logins higher than heuristic email-local-part logins, since noreply addresses are authoritative while local-part guesses may not match actual GitHub usernames.
- Consolidate severity normalization into a single location — either parseIssueResult or enrichTriageResult — to avoid duplicate logic.
- The Triage struct grew from 3 to 12 fields. Consider whether diagnostic-only fields (AssignmentSource, AssignmentDiagnostic) should be a separate nested struct to keep the user-facing triage JSON cleaner.

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Adds lightweight issue triage heuristics with priority labels, affected-path detection, and git-history-based assignee verification. Logic is mostly sound and well tested, but a few correctness/security concerns stand out: trusting git author emails as GitHub logins is impersonation-prone, context.Background() drops cancellation in the trigger path, and severity normalization is duplicated.

### Issues

🔴 **daemon/internal/issues/triage_assignment.go:0** — contributorLoginsFromEmail derives a 'GitHub login' from the local-part of any commit author email (e.g. 'victim@example.com' -> 'victim'). Git author email is attacker-controlled, so a malicious commit (or a contributor with a coincidental local-part) can impersonate any GitHub user and become the auto-assigned owner via SetAssignees. Restrict trusted derivation to the users.noreply.github.com pattern (or another verified mapping) and treat plain-email logins as unverified, or require the suggested_assignee path only.
🟡 **daemon/cmd/heimdallm/main.go:1496** — ensureRepoContextFullHistory uses context.Background() in the trigger-issue-review path, dropping the parent request's cancellation/timeout. A hung git fetch cannot be aborted. Thread the surrounding ctx (or a derived ctx with an explicit timeout) instead.
🟡 **daemon/internal/issues/triage_assignment.go:0** — sanitizeAffectedPaths only rejects paths starting with '/'. While the git log invocation does include a '--' separator (mitigating flag-injection), defense-in-depth would also reject paths starting with '-' or use pathspec magic (':(literal)<path>') so future call-sites that forget '--' don't reintroduce the issue.
🟡 **daemon/internal/issues/pipeline.go:521** — Severity is normalized inside parseIssueResult and again in enrichTriageResult. Not a bug but indicates ownership of normalization is unclear; consolidate in one place to avoid drift.
🟡 **daemon/internal/issues/triage_assignment.go:0** — git log invocation uses '--all' across every ref including refs/pull/* and dependabot branches when present, which can let unrelated authors dominate the contributor ranking for a path. Consider scoping to the default branch (or a configurable ref) for ownership signal.
🟡 **daemon/internal/github/client.go:0** — CreateLabel passes color verbatim after stripping a leading '#' but does not validate it as a 6-hex string; an LLM-supplied invalid color will produce a 422 that is silently swallowed (because 422 is treated as the 'race' case), masking a real misuse. Distinguish 'already exists' from 'invalid payload' by inspecting the response body or pre-validating color.

### Suggestions

- Tighten contributor-to-login mapping: only trust GitHub noreply emails, or cross-check candidate logins against repo collaborators / suggested_assignee before auto-assigning.
- Replace context.Background() in the trigger-issue-review handler with the request-scoped ctx (or context.WithTimeout) so git operations honour cancellation.
- Consolidate severity normalization in a single layer (either parseIssueResult or enrichTriageResult) and have the other call-site rely on it.
- Reject affected_paths starting with '-' in sanitizeAffectedPaths and/or use ':(literal)' pathspec magic for defense-in-depth against flag injection.
- In CreateLabel, distinguish 422-already-exists from 422-invalid-payload by checking the response body, so genuinely malformed labels surface as errors instead of silent no-ops.
- Consider restricting git-history scanning to the default branch rather than '--all' to avoid PR/fork refs skewing ownership.

---

ivanmunozruiz (commented):
## :white_check_mark: LGTM — cumple #392 y los 5 puntos del plan acordado

Verificación contra el plan que acordamos:

| Punto | Implementación | OK |
|---|---|---|
| Post-hoc verification del `suggested_assignee` vs git log | `resolveTriageAssignee` corre `git log --format=%ae%x00%an --all -- <paths>` y reemplaza por top contributor si el sugerido no aparece | ✅ |
| Unshallow explícito para triage | `ensureRepoContextFullHistory` cableado en 3 paths (worker, trigger, tier2 ReviewOnly) | ✅ |
| JSON retro-compatible | Bloque `triage` opcional + test `BackCompatOldIssueResultWithoutTriageBlock` | ✅ |
| `assignee_confidence` gateando assignment | low/vacío → fallback a triage_owner; `tentative_assignee` preservado para el comentario | ✅ |
| Storage JSON blob | Reutiliza `issue_reviews.triage`; sin migración | ✅ |

Cumplimiento de la issue: summary/severity/category/affected_area/paths/priority_label, sanitización (paths, slug GitHub, noreply email parsing), label auto-creada con colores por severity, best-effort label+assignee.

### Observaciones no bloqueantes

1. **Heurística simplificada vs spec**. La issue separa bug→`git blame` / feature→`git shortlog`; el PR usa `git log --format=%ae` para ambos. Para bugs no identifica al autor del regression-line específico — confía en que la IA lo haya hecho internamente con blame y verifica sólo que aparece en log. Defendible y cubre el espíritu, pero distinto a la letra.

2. **`git log` sin `--no-merges`**. En repos con muchas PRs el committer del merge infla su cuenta. Fix trivial: añadir `--no-merges` a los args en `contributorsForPaths`.

3. **Email canonicalization edge case**. Emails no-noreply tipo `alice.smith@company.com` → login descartado correctamente (regex rechaza el `.`); `alice@company.com` → mapea a login `alice` que puede no ser el GitHub login real. Aceptable para v1 porque GitHub moderno genera noreply para la mayoría de commits.

4. **`AssignmentDiagnostic`** sólo en JSON, no en el comentario público. Considera añadir una línea opcional al markdown si quieres exponerlo a operadores sin abrir la DB.

### Buen trabajo

- `githubNoreplyRE` maneja `N+login@users.noreply.github.com` → crítico para GitHub web/PR commits
- `contributorEvidence` muestra top-3 con cuenta de commits → comentario auditable
- `AssignmentSource` etiquetado (`suggested_assignee_verified`/`history_top_contributor`/`triage_owner`/`none`) → trazabilidad para UI futura
- Colores de severity siguen estándar industrial
- `normalizePriorityLabel` acepta tanto `"high"` como `"priority: high"` y canonicaliza

Cuando los 2 checks pendientes cierren verdes y los CODEOWNERS aprueben, mergeable.

ivanmunozruiz:
Gracias por las revisiones. He aplicado los cambios accionables en `e48c657`.

Resumen de fixes:

- `contributorLoginsFromEmail` ya no deriva GitHub logins desde emails normales tipo `victim@example.com`; sólo se consideran fiables los emails `users.noreply.github.com`.
- `git log` queda acotado con `--max-count=500`, se eliminó `--all` y se añadió `--no-merges` para evitar historial irrelevante y salidas no acotadas.
- `sanitizeAffectedPaths` rechaza paths que empiecen por `-`, además de paths absolutos/traversal/`.git`.
- El trigger manual de issue review ya no usa un `context.Background()` sin límite: ahora crea un contexto con timeout explícito de 30 minutos para que `acquireRepoContext`, `EnsureFullHistory` y el pipeline sean cancelables.
- La normalización de severity queda centralizada en `enrichTriageResult`; `parseIssueResult` sólo mantiene los fallbacks de compatibilidad.
- `CreateLabel` valida colores hex de 6 caracteres y sólo tolera `422` si GitHub devuelve un caso `already_exists`; otros `422` ahora salen como error.
- `TentativeAssignee` se conserva si venía en el JSON y sólo cae a `suggested_assignee` cuando está vacío.
- Añadidos tests para emails no-noreply, paths con `-`, preservación de tentative assignee y casos de `CreateLabel`.

Validación:

- `make test-docker` ✅

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review: all three previous findings (context propagation in trigger issue review, path sanitization for leading '-', severity normalization centralization) appear addressed per the diff and author's notes. No new blocking issues found in the visible changes.

### Suggestions

- Consider passing '--' as an explicit separator before pathspecs in git log invocations for defense-in-depth, even though sanitizeAffectedPaths now rejects leading '-'.
- The 30-minute timeout on the trigger-issue-review path is hardcoded; consider deriving from cfg.AI.ExecutionTimeout for consistency with other paths.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review confirms the author has addressed all previously raised findings. The trigger-issue-review path now uses an explicit 30-minute timeout context (main.go:1467-1473), CreateLabel validates hex colors via labelColorRE and only swallows 422 when the body matches already_exists (client.go:870-907), and the pipeline plumbs TriageOwner and applies labels/assignees best-effort with non-fatal failures. Tests cover the fallback assignment, missing-label creation, metadata-failure resilience, legacy result back-compat, invalid-color rejection, and the 422 already-exists vs invalid distinction. The full triage_assignment.go was truncated in the diff so the email-spoofing/sanitize-paths/--all fixes could not be visually verified, but the new test names and the back-compat test (which asserts no assignment without confidence) are consistent with the author's claim. No new high-severity issues introduced.

### Issues

🟡 **daemon/internal/issues/pipeline.go:1095** — BuildMarkdownComment renders AssigneeReason and AssigneeEvidence entries verbatim with fmt.Sprintf("%s"). LLM-supplied content with backticks, code fences, or stray newlines can break the surrounding bullet list or visually inject content into the GitHub comment. Consider escaping/sanitizing or stripping newlines before formatting. Cosmetic-only; GitHub still sanitizes scripts.
🟡 **daemon/internal/issues/pipeline.go:1005** — ensureIssueLabel falls back to applying the raw LLM-suggested label even if FetchLabels and CreateLabel both fail. AddLabels then logs a non-fatal warning, but a malformed label name produced by the LLM is attempted on every retriage. Consider only attempting AddLabels when the label is known to exist (skip on create failure that is not 'already exists').
🟡 **daemon/internal/issues/pipeline.go:982** — applyIssueTriageMetadata calls SetAssignees with the resolved AssignedAssignee unconditionally — including the configured TriageOwner fallback. If TriageOwner is misconfigured (typo or off-org user), every triaged issue silently fails assignment with only a warn-level log. Consider validating TriageOwner once at config load or surfacing the failure in the triage comment.

### Suggestions

- Since triage_assignment.go is the bulk of the new logic and was truncated in the supplied diff, please ensure CI/lint covers it; the visible tests look reasonable but full review of normalizeGitHubLogin/sanitizeAffectedPaths/contributorsForPaths could not be done in this re-review.
- Consider adding a startup-time sanity check that the configured TriageOwner is a valid GitHub login (regex) and optionally a collaborator on representative repos, to catch misconfigurations before they show up as silent assignment failures.
- Strip or escape newlines/backticks from AssigneeReason and items in AssigneeEvidence before embedding them in the Markdown comment to keep the bullet list well-formed.

---

Issue #223, opened 2026-04-22
bug: issue classified as IT (review_only) despite having develop labels — should be DEV


Muriano:
## Problem

`freepik-company/on-call#63` has a develop label but was classified as `review_only` instead of `auto_implement`. The daemon repeatedly tries to auto-implement it but the classification logic already ran as IT, so it enters an infinite retry loop (non-fast-forward push because the branch already exists from a previous attempt).

## Expected

Issues with develop labels should be classified as DEV (`auto_implement`) regardless of other labels. The develop label should take precedence over review-only classification.

## Observed

The daemon log shows the issue stuck in a loop:
```
level=INFO msg=notify title="Issue Auto-Implement Started" message="freepik-company/on-call #63"
level=ERROR msg="issues fetcher: pipeline run failed" ... err="non-fast-forward push"
```
Repeating every ~90 seconds indefinitely.

## Additional context

There's no backoff or max-retry logic for failed auto-implement pushes, which turns a classification bug into a resource-consuming infinite loop.

ivanmunozruiz:
Context after #395: this issue is now superseded by the staged issue state machine.

#395 intentionally changed classification precedence to the linear stage order `skip > blocked > review_only/triage > refinement > develop > default_action`, so an earlier-stage label now wins over a develop label. That avoids silently skipping triage/refinement during promotion.

Leaving this issue open for maintainers to decide, but if it stays open it should likely be reframed as a follow-up around retry/backoff or stale auto-implement branches, not as "develop labels should always beat review_only".

Muriano:
## Cerrada — superseded por el pipeline de stages y los fixes de retry

Los dos problemas originales están resueltos:

### 1. Clasificación: develop label vs review_only

La función `Classify()` (`config.go:276-302`) ahora implementa el orden de precedencia del state machine (#395):

```
skip > blocked > review_only (triage) > refinement > develop > default_action
```

Esto es **intencional**: en el nuevo pipeline, una issue debe pasar por triage antes de llegar a development. Si tiene ambos labels, el stage más temprano gana, y la issue avanza secuencialmente via promote (manual o automático). El bug original ("develop label debería ganar a review_only") ya no aplica porque el flujo es lineal por diseño.

### 2. Infinite retry loop en non-fast-forward push

Resuelto directamente en referencia a este issue (#223):
- `MaxAutoImplementFailures = 3` (`fetcher.go:33`) — cap de reintentos
[...]

PR #404, opened 2026-05-07
feat(issues): add issue stage promotion state machine

### LINKED ISSUE #395: feat: issue state machine — triage → refinement → development with promote
## Context

Currently issues can carry multiple classification labels simultaneously. This proposal enforces a **linear state machine** where an issue is in exactly one stage at a time.

## State Flow

\`\`\`
triage → refinement → development
\`\`\`

An issue **cannot** be in multiple stages simultaneously, even if it has labels for more than one. The pipeline must enforce mutual exclusivity.

## Label Management

Each stage transition (promote):
1. **Removes** the label(s) of the current stage
2. **Adds** the label(s) of the next stage
3. Posts an audit comment documenting the transition (who/what triggered it, timestamp)

Example label scheme (configurable):
- Triage: \`triage\`
- Refinement: \`refinement\` / \`refine\`
- Development: \`develop\` / \`auto_implement\` (existing)

### Manual Promote
- **API endpoint**: \`POST /issues/{id}/promote\` (already exists for review_only → develop, extend it)
- **GUI/TUI**: promote button on the issue detail view
- **GitHub label**: user manually swaps labels on GitHub → daemon detects on next poll and runs the new stage

### Automatic Promote (configurable per scope)
- After triage completes successfully → auto-promote to refinement
- After refinement completes successfully → auto-promote to development
- Config fields (per-repo > per-org > global):
  - \`auto_promote_triage\` — bool, default \`true\` (triage is fast, usually safe to auto-advance)
  - \`auto_promote_refinement\` — bool, default \`false\` (refinement output should be reviewed by a human before committing to development)

## Pipeline Integration

Each stage runs in its own control loop:
- **Triage loop**: picks up new issues, runs lightweight assessment, assigns priority + owner
- **Refinement loop**: picks up triaged issues, runs deep investigation, produces action plan
- **Development loop**: picks up refined issues, runs auto-implement (existing behavior)

An issue entering a loop is processed by that loop and then either:
- Auto-promoted to the next stage (if configured), or
- Left in current stage waiting for manual promote

## Guard Rails

- If an issue arrives with labels for multiple stages, the **earliest** stage wins (triage > refinement > development)
- If a stage fails (agent error, timeout), the issue stays in its current stage with an error comment — it is NOT promoted
- Blocked issues (existing \`blocked\` label) pause at whatever stage they're in until unblocked

## Related

- Triage definition: #392
- Refinement definition: #393
- Auto-clone for code access: #394

ivanmunozruiz:
## Summary

Closes #395.

Adds the issue stage state machine for `triage -> refinement -> development` while keeping the legacy `review_only` action name for stored triage rows and existing APIs.

## What changed

- Adds a stage transition helper that moves GitHub labels add-first/remove-after, creates target labels best-effort, posts audit comments, deduplicates recent promotion audits, and emits `issue_promoted` SSE events.
- Updates issue classification precedence to `skip > blocked > review_only > refinement > develop > default_action` so temporary multi-label states prefer the earliest stage.
- Changes `POST /issues/{id}/promote` to apply label transitions only; the next poll executes the new stage through the normal worker path.
- Adds auto-promotion after successful triage/refinement with `auto_promote_triage` defaulting on and `auto_promote_refinement` defaulting off.
- Detects manual GitHub label swaps on the next poll, posts an audit comment, and dispatches the newly visible stage.
- Adds CLI/TUI promotion support and updates Flutter issue screens/SSE handling for stage promotion.
- Updates README and configuration guide for the staged issue pipeline.

## Validation

- `make test-docker`
- CLI `go test ./...` inside Docker
- `flutter test`
- `flutter analyze`
- `make build-web`

## Notes

Local unrelated files were left out of the commit: `flutter_app/pubspec.lock` and `flutter_app/flutter_01.png`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 > **Note:** Versions 0.1.0–0.1.3 were originally published under [`theburrowhub/heimdallr-docker`](https://github.com/theburrowhub/heimdallr-docker) (now archived). The project was unified into this repository and renamed to Heimdallm in v0.1.1.
 
+## Unreleased
+
+### Upgrade Notes
+
+* **issues:** issue classification now follows the state-machine order `skip > blocked > review_only > refinement > develop > default_action`. If a repo intentionally double-labels issues with triage/review and develop labels, the earlier stage now wins so Heimdallm does not skip triage or refinement during promotion.
+* **issues:** `auto_promote_triage` defaults on only when `refinement_labels` is configured. Repos without a refinement target keep their previous review-only behavior; set `auto_promote_triage = false` explicitly to keep refinement manual even when refinement labels exist.
+
 ## [0.6.12](https://github.com/theburrowhub/heimdallm/compare/v0.6.11...v0.6.12) (2026-05-01)
 
 
--- a/README.md
+++ b/README.md
@@ -349,6 +349,8 @@ tagged with both a blocked and a develop label stays blocked until
 promotion. Stage promotion updates labels only; the next poll executes the newly visible stage. The feature is **opt-in**: leave `HEIMDALLM_ISSUE_BLOCKED_LABELS`
 empty and nothing about the existing pipeline changes.
 
+**Upgrade note for staged issues:** earlier-stage labels now win over later-stage labels. If you previously used overlapping triage/develop labels to force direct development, split them into dedicated labels or remove the triage label before applying the develop label. `auto_promote_triage` defaults on only when `refinement_labels` is configured; set it to `false` to keep refinement promotion manual.
+
 ### Automated install (for agents / scripts)
 
 See [LLM-HOW-TO-INSTALL.md](LLM-HOW-TO-INSTALL.md) for a step-by-step guide suitable for Claude Code, shell scripts, or any automation tool.
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Heimdallm runs in the background and does three things, in parallel, at your con
 Watches the PRs where you're requested as a reviewer, runs an AI code review, and submits it to GitHub as your account. No copy-pasting, no manual prompting.
 
 ### 2. Issue triage & auto-implement
-Fetches issues from monitored repos, classifies them by label (`review_only` vs `auto_implement` vs `skip` vs `blocked`), and for the develop-track ones optionally **creates a branch, commits the change, and opens a PR** against your default branch — fully autonomous on the issues you mark for it. Issues can declare dependencies on other issues/PRs; Heimdallm holds them in a `blocked` state until the prerequisites close, then promotes them automatically.
+Fetches issues from monitored repos, classifies them by label (`review_only` triage, `refinement`, `develop` / `auto_implement`, `skip`, `blocked`), and can move issues through triage -> refinement -> development. Develop-track issues optionally **create a branch, commit the change, and open a PR** against your default branch. Issues can declare dependencies on other issues/PRs; Heimdallm holds them in a `blocked` state until the prerequisites close, then promotes them automatically.
 
 ### 3. Self-monitoring UI
 A Flutter Web dashboard (`:3000`) with Dashboard, PR list, Issue list, prompt/agent editor, live config editor, and a live log stream. Opens alongside the daemon in Docker mode.
@@ -22,7 +22,7 @@ A Flutter Web dashboard (`:3000`) with Dashboard, PR list, Issue list, prompt/ag
 ### Headline features
 
 - **Automatic reviews** — polls `review-requested:@me` on GitHub and submits reviews as your account
-- **Issue pipeline** — label-driven triage + optional auto-implement with branch/commit/PR cycle
+- **Issue pipeline** — label-driven triage, refinement planning, and optional auto-implement with branch/commit/PR cycle
 - **Issue dependencies** — mark downstream work with a `blocked` label; declare deps via a `## Depends on` body section *or* GitHub's native sub-issues; Heimdallm auto-promotes when all blockers close
 - **Configurable prompts** — general review, security audit, performance, architecture, or your own with `{diff}` `{title}` `{author}` `{comments}` placeholders, managed from the web UI at `/agents`
 - **Two feedback modes** — *single* (one consolidated review) or *multi* (one GitHub comment per issue + summary), globally and per repo
@@ -335,7 +335,7 @@ label:
    label(s), add `HEIMDALLM_ISSUE_PROMOTE_TO_LABEL`, and leave an audit
    comment listing each dep and its state at check time.
 6. The same poll cycle's fetch pass then classifies the promoted issue
-   normally and dispatches it to triage or auto-implement.
+   normally and dispatches it to triage, refinement, or auto-implement.
 
 Issues with a blocked label but **no declared deps in either source**
 stay blocked — the daemon won't guess when to unblock them. Remove the
@@ -344,9 +344,9 @@ a given issue, Heimdallm skips that issue for the current cycle rather
 than risk promoting on incomplete information.
 
 Classification precedence is
-`skip > blocked > review_only > develop > default_action`, so an issue
+`skip > blocked > review_only > refinement > develop > default_action`, so an issue
 tagged with both a blocked and a develop label stays blocked until
-promotion. The feature is **opt-in**: leave `HEIMDALLM_ISSUE_BLOCKED_LABELS`
+promotion. Stage promotion updates labels only; the next poll executes the newly visible stage. The feature is **opt-in**: leave `HEIMDALLM_ISSUE_BLOCKED_LABELS`
 empty and nothing about the existing pipeline changes.
 
 ### Automated install (for agents / scripts)
--- a/cli/internal/api/client.go
+++ b/cli/internal/api/client.go
@@ -264,6 +264,12 @@ func (c *Client) TriggerIssueReview(id int64) error {
 	return err
 }
 
+// PromoteIssue moves an issue to its next configured stage by changing labels.
+func (c *Client) PromoteIssue(id int64) error {
+	_, err := c.do("POST", fmt.Sprintf("/issues/%d/promote", id))
+	return err
+}
+
 // Shutdown asks the daemon to stop gracefully.
 func (c *Client) Shutdown() error {
 	_, err := c.do("POST", "/shutdown")
--- a/cli/internal/cli/follow.go
+++ b/cli/internal/cli/follow.go
@@ -184,6 +184,11 @@ func formatEventData(data string) string {
 			parts = append(parts, from+" → "+to)
 		}
 	}
+	if from, ok := m["from_stage"].(string); ok {
+		if to, ok := m["to_stage"].(string); ok {
+			parts = append(parts, from+" → "+to)
+		}
+	}
 
 	if reason, ok := m["reason"].(string); ok && reason != "" {
 		parts = append(parts, styleMuted.Render(reason))
--- a/cli/internal/cli/promote_issue.go
+++ b/cli/internal/cli/promote_issue.go
@@ -1,0 +1,59 @@
+package cli
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/spf13/cobra"
+)
+
+func newPromoteIssueCmd() *cobra.Command {
+	var repo string
+
+	cmd := &cobra.Command{
+		Use:   "promote-issue <number>",
+		Short: "Promote an issue to its next configured stage",
+		Long: "Move the issue identified by its GitHub number to the next stage by updating GitHub labels.\n" +
+			"Use --repo to specify the repository (required when the daemon monitors multiple repos).",
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := clientFromContext(cmd.Context())
+			number, err := strconv.Atoi(args[0])
+			if err != nil {
+				return fmt.Errorf("invalid issue number: %w", err)
+			}
+
+			issues, err := c.ListIssues()
+			if err != nil {
+				return fmt.Errorf("fetching issues: %w", err)
+			}
+
+			var matched []int64
+			for _, iss := range issues {
+				if iss.Number == number && (repo == "" || iss.Repo == repo) {
+					matched = append(matched, iss.ID)
+				}
+			}
+
+			switch len(matched) {
+			case 0:
+				if repo != "" {
+					return fmt.Errorf("no issue #%d found in %s", number, repo)
+				}
+				return fmt.Errorf("no issue #%d found (try --repo org/repo)", number)
+			case 1:
+				if err := c.PromoteIssue(matched[0]); err != nil {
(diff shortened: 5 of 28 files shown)


Muriano:
## ⚠️ MEDIUM Issue

Classification precedence is reversed: review_only > refinement > develop replaces the prior develop > refinement > review_only ordering that issue #223 explicitly introduced to avoid an infinite retry loop when an issue carried both DEV and IT labels. The PR comment claims the state machine makes this safe, but existing user configs with overlapping labels (e.g. legacy issues tagged 'bug' + 'question') will now silently downgrade to review_only and, with auto-promote-triage defaulting to true, will cycle review_only → refinement on each poll instead of going straight to develop. Impact: behavioral regression for any deployment that depended on #223 semantics. Fix: either keep #223 ordering for non-state-machine repos, gate the new ordering behind an explicit opt-in, or document this as a breaking change in the changelog and call it out in the README migration notes.

**Location:** `daemon/internal/config/config.go` line 240

---

Muriano:
## ⚠️ MEDIUM Issue

autoPromoteStageEnabled defaults AutoPromoteTriage to true when unset. Existing users on review_only-only workflows (no refinement_labels, or refinement_labels they don't want to use) will, after upgrade, see triage runs auto-promote — moving labels and adding audit comments without any opt-in. NextStage will return ErrStageTargetLabelMissing when refinement_labels is empty and the warn log mitigates this, but for users who DO have refinement_labels configured for manual use, the automatic transition is an unannounced behavior change. Recommend defaulting AutoPromoteTriage to false (matching AutoPromoteRefinement) so promotion is always opt-in, or documenting the new default prominently in CHANGELOG/README upgrade notes.

**Location:** `daemon/cmd/heimdallm/main.go` line 3430

---

Muriano:
## 🟡 LOW Issue

canPromoteIssue gates the [p] hotkey on issue.LatestReview.ActionTaken being 'review_only' or 'refinement'. ActionTaken reflects the last completed pipeline run, not the current GitHub stage, so an issue that was just promoted (labels updated, no new review yet) or one whose labels were edited externally can be in a different stage than what the TUI shows. The server-side promote handler does validate via Classify+NextStage, so this is a UX nit rather than a correctness bug, but pressing [p] on a stale row will hit ErrPromoteConflict and surface as 'Promotion failed' to the user. Consider using the current label-derived stage when available, or refreshing before showing the prompt.

**Location:** `cli/internal/tui/dashboard.go` line 1222

---

Muriano (approved):

Substantial PR adding an issue stage state machine (triage → refinement → develop) with manual and auto-promotion paths. Architecture is reasonable: promotion only mutates labels and the next poll executes the new stage, with a stale-job guard before each worker runs the pipeline. Tests for classification and recorder are updated. Two behavioral changes deserve attention: classification precedence has been reversed (review_only now beats develop), which intentionally walks back the issue #223 fix, and auto-promote from triage defaults to true.

### Suggestions

- The promote callback uses context.Background() with a 30s timeout; consider deriving from the daemon's root context so a shutdown cancels in-flight promotions cleanly.
- issueStageStillCurrent re-classifies against the in-memory issue snapshot captured at dispatch time, not a fresh GitHub fetch. If labels change between dispatch and worker start without the issue being re-fetched, the stale check passes. Worth a comment clarifying this is best-effort.
- TestRecorder_IssueStagePromoted only covers the from_stage/to_stage path; add a case where both legacy from_label and new from_stage are present to lock in the precedence rule in recordIssuePromoted.
- The README change merges 'develop / auto_implement' into one bullet but the body still uses both terms inconsistently — a small editorial pass would help.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Refactor introduces a state-machine-driven issue stage promotion pipeline with manual + auto-promotion paths and stale-job guards. The implementation is well-structured (worker pre-flight stage checks, single TransitionIssueStage helper, audit comments) but the classification precedence reversal and auto-promote-on-by-default are behavioral regressions that contradict the explicit fix from #223 and lack a migration story.

### Issues

⚠️ **daemon/internal/config/config.go:240** — Classification precedence reversed from develop > refinement > review_only to review_only > refinement > develop. Issue #223 introduced the prior order specifically to break an infinite retry loop when an issue carried both a DEV and IT label. With AutoPromoteTriage defaulting to true, existing deployments where issues are double-labelled will now silently downgrade to review_only and cycle review_only → refinement → develop on successive polls instead of going straight to develop. Either gate behind an opt-in, preserve #223 ordering when refinement_labels is empty, or document as a breaking change with migration notes.
⚠️ **daemon/cmd/heimdallm/main.go:3430** — autoPromoteStageEnabled returns true when AutoPromoteTriage is nil. Existing review_only-only users will, after upgrade, see triage runs auto-promote — adding labels and audit comments without opt-in. ErrStageTargetLabelMissing on missing refinement_labels softens this but only logs a warn; users who configured refinement_labels for manual use will get unannounced auto-transitions. Default to false (matching AutoPromoteRefinement) for parity, or call this out prominently in CHANGELOG/README upgrade notes.
🟡 **cli/internal/tui/dashboard.go:1222** — canPromoteIssue gates the [p] hotkey on issue.LatestReview.ActionTaken, which reflects the last completed pipeline run rather than the current GitHub stage. After a promotion (labels updated, new review not yet recorded) or external label edits, pressing [p] on a stale row hits server-side ErrPromoteConflict and surfaces as 'Promotion failed'. UX nit (server validation is correct) but consider refreshing or deriving the stage from current labels.
🟡 **daemon/internal/activity/recorder.go:274** — When a stage-promotion event arrives with from_stage/to_stage set, the recorded details map still includes empty from_label/to_label keys alongside the stage keys. Downstream consumers (CLI follow output, UI activity feed) that branch on label presence may render empty arrows. Consider omitting label keys when stage is the source of truth.
🟡 **daemon/cmd/heimdallm/main.go:887** — issueStageStillCurrent runs after the worker has already fetched the GitHub issue, but the ghIssue used for classification is the same one passed into the worker — there's a TOCTOU window between the fetcher's classification and the worker's run. The check helps for queue-backed staleness but does not re-fetch labels from GitHub, so a label edit between fetcher dispatch and worker pickup is detected only if the queue payload's labels are stale. Acceptable for v1 but worth noting.

### Suggestions

- Add a CHANGELOG entry and README migration section calling out the precedence reversal and auto-promote-triage default; both are user-visible behavior changes.
- Default AutoPromoteTriage to false for parity with AutoPromoteRefinement, and require explicit opt-in to surface the new state machine.
- In canPromoteIssue, fall back to current label-derived classification when available, or trigger a refresh before showing the prompt to avoid stale [p] presses.
- Consider re-fetching the GitHub issue inside issueStageStillCurrent (or accept its limitation in a comment) so the guard catches label edits that happened while the job sat in the queue.
- In the activity recorder, drop empty from_label/to_label keys when from_stage/to_stage are set so consumers can rely on a single source of truth per event.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Well-structured state machine for issue stage promotion with good test coverage and audit trails. However, the default auto-promote-triage being true is a silent behavioral change for existing deployments, the classification precedence reversal is breaking, and the async promote endpoint masks errors from callers.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:3397** — autoPromoteStageEnabled defaults to true for triage when AutoPromoteTriage is nil. Every existing deployment without this new config field will silently start auto-promoting issues from triage to refinement on upgrade. This should default to false for backward compatibility — opt-in, not opt-out.
⚠️ **daemon/internal/config/config.go:240** — Classification precedence reversed from develop > refinement > review_only to review_only > refinement > develop. This is a breaking behavioral change: existing users who dual-label issues (e.g., review_only + develop) relied on the develop-wins semantics from issue #223 to prevent misclassification. The PR should document migration guidance or, at minimum, add a release note warning. Consider whether the #223 infinite-retry-loop scenario is still prevented under the new ordering.
⚠️ **cli/internal/tui/dashboard.go:335** — promoteIssueMsg error handler sets d.connected = false. A promotion conflict (e.g., already at final stage) is a normal business error, not a connectivity failure. This will incorrectly show the dashboard as disconnected and may trigger reconnection logic. Only set d.connected = false for actual connection/transport errors.
⚠️ **daemon/cmd/heimdallm/main.go:1738** — The promote callback now performs classification and NextStage resolution which can return ErrPromoteConflict, but the HTTP handler (handlePromoteIssue) fires the callback in a goroutine and has already returned 202. The caller (CLI or TUI) never learns that promotion was rejected — it always sees 'promote queued'. Consider doing the cheap classify+next-stage check synchronously and returning 409 for conflicts, then running only the label transition async.
🟡 **daemon/internal/server/handlers.go:165** — Doc comment on SetTriggerPromoteFn still says 'The callback must run the auto_implement pipeline for the given issue regardless of its current classification (review_only → develop promotion)'. The PR changes the callback to do label transitions only, not run the pipeline. This comment is now misleading.
🟡 **cli/internal/tui/dashboard.go:1219** — canPromoteIssue checks issue.LatestReview.ActionTaken to determine promotability, but the issue's labels may have changed on GitHub since the last review. This could offer the promote action when labels no longer match, or hide it when they do. Consider also (or instead) checking the current label-derived classification.
🟡 **cli/internal/tui/dashboard.go:450** — The 'p'/'P' key handler with canPromoteIssue guard is duplicated identically in handleKey (list view) and handleDetailKey (detail view). Extract to a shared method to prevent the two copies from drifting.

### Suggestions

- Flip autoPromoteStageEnabled default for triage from true to false — existing deployments should not gain new autonomous behavior on upgrade without explicit opt-in.
- Make the classify+NextStage check in the promote callback synchronous so the HTTP handler can return 409 for conflicts instead of always returning 202.
- Add a CHANGELOG/migration note about the classification precedence reversal (develop no longer beats review_only) so operators can audit their label configurations before upgrading.
- Update the SetTriggerPromoteFn doc comment in handlers.go to reflect the new label-transition-only semantics.
- In promoteIssueMsg error handling, remove d.connected = false — promotion failure is not a connection error and shouldn't trigger disconnected UI state.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: most prior findings addressed. Classification precedence reversal is now documented as a breaking change in CHANGELOG/README. AutoPromoteTriage now uses a smart default (on only when refinement_labels is configured), which is a reasonable middle ground and is called out in upgrade notes. Recorder no longer emits empty legacy label keys when stage data is present. TOCTOU in stage-current check is now documented as best-effort. The TUI canPromoteIssue UX nit is unchanged.

### Issues

🟡 **cli/internal/tui/dashboard.go:1228** — canPromoteIssue still gates the [p] hotkey on issue.LatestReview.ActionTaken (last completed pipeline run) rather than current GitHub stage. After a promotion completes label changes but before the next pipeline run records a review, or after external label edits, the TUI will offer/deny [p] inconsistently with server validation, surfacing as 'Promotion failed' via ErrPromoteConflict. Unchanged from prior review.
🟡 **daemon/cmd/heimdallm/main.go:3430** — AutoPromoteTriage smart default still flips behavior for users who configured refinement_labels intending manual control — after upgrade, label transitions and audit comments occur without explicit opt-in. CHANGELOG/README now document this, but consider whether matching AutoPromoteRefinement's default-false would be safer; documentation-only mitigation is acceptable if the team prefers the staged default.

### Suggestions

- Derive canPromoteIssue from the issue's current label set (using IssueTrackingConfig.Classify) rather than LatestReview.ActionTaken, or refresh the issue row immediately after a successful promotion request so the TUI matches server-side stage truth.
- Consider an integration test that simulates a queued worker job whose labels change between fetcher dispatch and worker pickup to lock in issueStageStillCurrent's stale-job guard against future regressions.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review: previous medium findings have been addressed. The precedence reversal is now documented in CHANGELOG and README upgrade notes, and auto_promote_triage now defaults on only when refinement_labels is configured (with explicit test coverage in main_test.go). The low-severity TUI staleness concern remains unchanged but is non-blocking.

### Suggestions

- cli/internal/tui/dashboard.go:1228 — canPromoteIssue still gates [p] on LatestReview.ActionTaken, which reflects the last completed pipeline run rather than the current GitHub stage. Pressing [p] on a stale row will surface 'Promotion failed' via ErrPromoteConflict. Consider deriving the stage from current labels or refreshing before enabling the prompt. (Carried over from prior review; non-blocking.)
- daemon/cmd/heimdallm/main.go:3430 — Smart-default for AutoPromoteTriage is now gated on refinement_labels, which removes the surprise for review_only-only deployments. For users who configured refinement_labels for manual use, the new auto-on default is still a behavior change; the CHANGELOG/README note covers this, but consider an explicit log line on first auto-promotion so operators see the transition in their daemon logs.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

The major issues from the first review are addressed: auto_promote_triage now has a smart default gated on refinement_labels being configured, the classification precedence change is documented in the CHANGELOG and README, and the d.connected bug is fixed. Three prior findings remain unchanged, and the handler doc comments are now stale in two places.

### Issues

⚠️ **daemon/internal/server/handlers.go:1032** — Still unaddressed from prior review: handlePromoteIssue fires triggerPromoteFn in a goroutine and unconditionally returns 202. The callback now does a cheap Classify+NextStage check that can detect ErrPromoteConflict synchronously, but the result is never surfaced to the caller. The CLI/TUI will always show 'Promotion requested' even when the issue is already at its final stage. Consider running the validation (store lookup + Classify + NextStage) synchronously and returning 409 for conflicts, then running only TransitionIssueStage async.
🟡 **daemon/internal/server/handlers.go:164** — Still unaddressed from prior review: SetTriggerPromoteFn doc comment says 'The callback must run the auto_implement pipeline for the given issue regardless of its current classification (review_only → develop promotion)'. The callback now performs stage-label transitions only, not pipeline runs, and supports multiple stage transitions, not just review_only→develop.
🟡 **daemon/internal/server/handlers.go:1012** — handlePromoteIssue doc comment says 'promotes a review_only-classified issue to auto_implement, triggering the full develop pipeline immediately'. Both claims are stale: promotion now transitions between arbitrary stages via label changes, and the next poll executes the stage rather than the handler running the pipeline directly.
🟡 **cli/internal/tui/dashboard.go:519** — promoteSelectedIssue and the promoteIssueMsg handler reuse d.shutdownMessage for promotion status ('Promoting issue #N...', 'Promotion requested', 'Promotion failed: ...'). This text persists in the header bar with no mechanism to clear it after a promotion completes. On success d.fetchData is triggered but the data-fetch completion handler does not reset shutdownMessage, so the stale status stays until the user triggers the shutdown dialog. Consider a separate statusMessage field that auto-clears on the next refresh, or clear shutdownMessage in the data-fetch response handler.

### Suggestions

- Split handlePromoteIssue into a synchronous validation phase (store lookup, Classify, NextStage) that can return 409/ErrPromoteConflict, and an async label-transition phase that runs in the goroutine. This lets the CLI/TUI distinguish 'conflict' from 'queued' without waiting for SSE events.
- Update both doc comments in handlers.go (lines 164-166 and 1012-1015) to reflect the new label-transition semantics: the callback changes stage labels and posts an audit comment; the next poll cycle executes the newly-visible stage.

---

Issue #410, opened 2026-05-08
test: staged pipeline from triage to refinement


ivanmunozruiz:
This is a fresh smoke test for the Heimdallm staged issue pipeline.\n\nExpected flow:\n1. Heimdallm picks up this issue from the `heimdallm_triage` label.\n2. It posts a triage comment and priority label.\n3. With `refinement_labels = ["heimdallm_refine"]` configured, it promotes the issue from `heimdallm_triage` to `heimdallm_refine`.\n4. On the next poll, it runs refinement and posts a deep implementation plan.\n5. If `auto_promote_refinement = true`, it may then promote to `heimdallm_develop`; otherwise it should remain in refinement until manual promote.\n\nThis issue is intentionally low-risk and does not request product code changes. It can be closed after the staged flow is verified.

ivanmunozruiz:
## 🟡 Heimdallm triage — LOW

This issue is a smoke test for the Heimdallm staged issue pipeline, validating the flow from triage label pickup, through refinement promotion, to optional auto-promotion to develop. It explicitly requests no product code changes and exists only to verify end-to-end staged behavior with `refinement_labels` and `auto_promote_refinement` configured. It can be closed once the staged flow is observed working as described.

### Classification

- **Category:** other
- **Suggested severity:** low
- **Priority label:** priority: low
- **Likely affected area:** heimdallm staged issue pipeline (triage → refinement → develop)
- **Tentative owner:** @ivanmunozruiz
- **Suggested assignee:** @ivanmunozruiz
- **Owner rationale:** Issue author and recent committer of the staged pipeline features (refinement stage, promotion state machine, refinement labels config) per recent commits 53d0b98, 2a124de, 4305ea0.
- **Owner confidence:** high
- **Owner evidence:**
  - 53d0b98 feat(app): expose refinement labels config (#409) by ivanmunozruiz
  - 2a124de feat(issues): add issue stage promotion state machine (#404) by ivanmunozruiz
  - 4305ea0 feat: add issue refinement stage (#402) by ivanmunozruiz

### Next steps

- Observe Heimdallm's behavior across two poll cycles: confirm triage comment + priority label, then promotion from heimdallm_triage to heimdallm_refine, then a refinement plan comment.
- Verify auto_promote_refinement config: if true, confirm subsequent promotion to heimdallm_develop; if false, confirm it stays in refinement pending manual promote.
- Close the issue once the staged flow is verified end-to-end; no code changes required.
[...]

PR #412, opened 2026-05-08
fix(issues): dedupe stage promotion audits

### LINKED ISSUE #411: Bug: stage promotion repeats after refinement label transition
## Problem

While testing the staged issue flow on #410, Heimdallm correctly moved the issue from triage to refinement and posted a refinement plan, but subsequent polls kept recording or promoting triage -> refinement again.

Observed symptoms:
- Multiple activity entries for Promoted: triage -> refinement on the same issue.
- Multiple GitHub audit comments for the same stage transition.
- Duplicate refinement comments can be queued while the first refinement run is still in flight.
- The issue remains on heimdallm_refine unless auto_promote_refinement=true is explicitly configured, but the app does not make that repo/global behavior obvious enough during setup/testing.

## Expected

- A completed/audited stage transition should be idempotent across polls.
- Manual GitHub label-change audit should not re-run when an earlier auto/manual promotion comment already records the same from/to stage after the latest stored review.
- After refinement completes, promotion to develop should happen only when auto_promote_refinement=true and a develop label is configured.
- The UI should make the auto-promote refinement setting easy to verify/configure.

## Reproduction

1. Configure issue tracking labels for a repo:
   - review-only: heimdallm_triage
   - refinement: heimdallm_refine
   - develop: heimdallm_develop
2. Create an issue with heimdallm_triage.
3. Wait for triage and auto-promotion to refinement.
4. Observe repeated triage -> refinement promotion activity/comments on following polls.

## Notes

Test issue: #410.

ivanmunozruiz:
## Summary

- Prevent repeated manual-stage audit/promote handling when a stage promotion comment already records the same transition after the latest stored review.
- Only skip the transition when labels already reflect the target stage; if labels are stale, retry label normalization without posting a duplicate audit comment.
- Add regression coverage for the repeated `triage -> refinement` audit loop seen on #410 and related edge cases.

## Root cause

After an issue was auto-promoted from triage to refinement, the next poll could still see the latest stored review as `review_only` until the refinement row existed. The fetcher interpreted the current refinement label as a fresh manual GitHub stage change on every poll and re-ran the transition audit path.

## Review follow-up

- Removed the Flutter auto-promote UI changes from this PR so the fix stays scoped to #411.
- Centralized stage-promotion audit matching.
- Added tests for already-applied labels, stale labels after an audit, different transitions, old audit comments, and the `CreatedAt` fallback when `CommentedAt` is zero.

## Validation

- `make test-docker GO_TEST_ARGS="-run TestFetcher_ManualStageLabelChange ./internal/issues/..."`
- `make test-docker`

Fixes #411.

### DIFF
--- a/daemon/internal/issues/fetcher.go
+++ b/daemon/internal/issues/fetcher.go
@@ -354,10 +354,15 @@ func (f *Fetcher) auditManualStageChange(ctx context.Context, issue *github.Issu
 	if ref.IsZero() {
 		ref = latest.CreatedAt
 	}
-	if hasStagePromotionCommentSince(comments, from, to, ref) {
+	alreadyAudited := hasStagePromotionCommentSince(comments, from, to, ref)
+	if alreadyAudited && stageTransitionApplied(issue, cfg, to) {
 		slog.Debug("issues fetcher: stage transition already audited since latest review",
 			"repo", issue.Repo, "number", issue.Number, "from", from, "to", to)
 		return nil
+	}
+	if alreadyAudited {
+		slog.Debug("issues fetcher: stage transition audit exists but labels need normalization",
+			"repo", issue.Repo, "number", issue.Number, "from", from, "to", to)
 	}
 
 	return TransitionIssueStage(ctx, f.stageClient, StageTransition{
@@ -369,6 +374,7 @@ func (f *Fetcher) auditManualStageChange(ctx context.Context, issue *github.Issu
 		Trigger:        StagePromotionManualGitHub,
 		Time:           time.Now().UTC(),
 		RecentComments: comments,
+		SuppressAudit:  alreadyAudited,
 		Broker:         f.stageBroker,
 	})
 }
--- a/daemon/internal/issues/fetcher.go
+++ b/daemon/internal/issues/fetcher.go
@@ -350,6 +350,15 @@ func (f *Fetcher) auditManualStageChange(ctx context.Context, issue *github.Issu
 			comments = got
 		}
 	}
+	ref := latest.CommentedAt
+	if ref.IsZero() {
+		ref = latest.CreatedAt
+	}
+	if hasStagePromotionCommentSince(comments, from, to, ref) {
+		slog.Debug("issues fetcher: stage transition already audited since latest review",
+			"repo", issue.Repo, "number", issue.Number, "from", from, "to", to)
+		return nil
+	}
 
 	return TransitionIssueStage(ctx, f.stageClient, StageTransition{
 		Issue:          issue,
--- a/daemon/internal/issues/fetcher.go
+++ b/daemon/internal/issues/fetcher.go
@@ -335,6 +335,9 @@ func (f *Fetcher) auditManualStageChange(ctx context.Context, issue *github.Issu
 		}
 		return err
 	}
+	if latest == nil {
+		return nil
+	}
 	from, ok := StageFromAction(latest.ActionTaken)
 	if !ok || from == to {
 		return nil
--- a/daemon/internal/issues/fetcher_test.go
+++ b/daemon/internal/issues/fetcher_test.go
@@ -163,6 +163,23 @@ func fixture(number int, updated time.Time) *github.Issue {
 	}
 }
 
+func issueLabels(labels ...string) []github.Label {
+	out := make([]github.Label, 0, len(labels))
+	for _, label := range labels {
+		out = append(out, github.Label{Name: label})
+	}
+	return out
+}
+
+func stagePipelineCfg() config.IssueTrackingConfig {
+	return config.IssueTrackingConfig{
+		Enabled:          true,
+		ReviewOnlyLabels: []string{"triage"},
+		RefinementLabels: []string{"refine"},
+		DevelopLabels:    []string{"develop"},
+	}
+}
+
 func enabledCfg() config.IssueTrackingConfig {
 	return config.IssueTrackingConfig{Enabled: true}
 }
@@ -731,6 +748,7 @@ func TestFetcher_ManualStageLabelChangeSkipsAlreadyAuditedTransition(t *testing.
 		Title:     "Needs plan",
 		UpdatedAt: now,
 		Mode:      config.IssueModeRefinement,
+		Labels:    issueLabels("refine"),
 	}
 	latestReviewAt := now.Add(-5 * time.Minute)
 	dedup := &fakeDedup{byGithubID: map[int64]dedupEntry{
@@ -763,12 +781,7 @@ func TestFetcher_ManualStageLabelChangeSkipsAlreadyAuditedTransition(t *testing.
 	f.SetPublisher(pub)
 	f.SetStageTransitioner(stage, nil)
 
-	processed, err := f.ProcessRepo(context.Background(), "org/repo", config.IssueTrackingConfig{
-		Enabled:          true,
-		ReviewOnlyLabels: []string{"triage"},
-		RefinementLabels: []string{"refine"},
-		DevelopLabels:    []string{"develop"},
-	}, "alice", noOpts)
+	processed, err := f.ProcessRepo(context.Background(), "org/repo", stagePipelineCfg(), "alice", noOpts)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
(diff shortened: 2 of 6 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

PR adds dedup check for stage promotion audit comments to prevent duplicate audits, plus Flutter UI for auto-promote toggles. The dedup logic is correct and well-tested. Minor concern: the dedup check happens after fetching comments but the early-return skips the actual stage transition entirely, which may be intentional but worth confirming.

### Suggestions

- Consider adding a test case where ref falls back to latest.CreatedAt (CommentedAt is zero) to cover that branch
- The Flutter `globalValue` in OverrideDropdown for auto-promote triage falls back to `effectiveRefinementLabels.isNotEmpty` which mixes a boolean default with label-presence heuristic — document this implicit default or make it explicit
- hasStagePromotionCommentSince could share the heading/from/to matching logic with hasRecentStagePromotionComment to reduce duplication

---

vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Backend dedup fix is correct and well-tested. However, the PR conflates an unrelated feature (auto-promote toggles in Flutter UI) with the bug fix, which complicates review and bisecting.

### Issues

⚠️ **PR scope:0** — PR title says 'dedupe stage promotion audits' but the diff also introduces new global/org/repo auto-promote configuration fields and UI in Flutter. These should be split into separate PRs — the fix should be reviewable and revertable independently from the new feature.
🟡 **flutter_app/lib/features/config/config_screen.dart:462** — Default value for the triage switch is `_issueTracking.refinementLabels.isNotEmpty`, which conflates 'has labels configured' with 'auto-promote enabled'. Once the user toggles, the value becomes explicit, but the initial implicit default may surprise users who configured refinement labels without intending auto-promotion.
🔴 **flutter_app/lib/features/repositories/repo_detail_screen.dart:460** — The diff adds `autoPromoteTriage`/`autoPromoteRefinement` on `RepoConfig` and `globalAutoPromoteTriage`/`globalAutoPromoteRefinement` on the app config, but model/serialization changes are not shown in this diff. Verify these fields are wired through copyWith, JSON (de)serialization, and the backend config schema; otherwise the UI silently drops state.
⚠️ **daemon/internal/issues/fetcher.go:359** — Early return on detected duplicate skips the entire `TransitionIssueStage` call. If a previous run posted the audit comment but failed to apply the labels (partial failure), this dedup will permanently prevent a retry. Consider whether the dedup should only suppress the comment, or document the assumption that comment presence implies a completed transition.
🟡 **daemon/internal/issues/stage_transition.go:284** — `hasStagePromotionCommentSince` matches by substring on '- From:' / '- To:' lines, which is fragile if the audit comment template ever reorders or renames fields. Consider centralizing the format with a parser, or reusing the same builder used to render the comment to derive the match strings.
🟡 **daemon/internal/issues/fetcher_test.go:723** — Test only covers the positive dedup case. Missing coverage for: (a) comment present but for a different from→to transition (should still audit), (b) comment present but older than `ref` (should still audit), (c) `latest.CommentedAt` zero with non-zero `CreatedAt` fallback path.

### Suggestions

- Split the Flutter auto-promote feature into a separate PR so the dedup fix can ship and be reverted independently.
- Add tests for the negative dedup paths (different transition, older comment, fallback to CreatedAt) to lock in the boundary conditions of `hasStagePromotionCommentSince`.
- Extract the audit-comment marker generation/matching into a shared helper so the producer and `hasStagePromotionCommentSince` cannot drift apart.
- Confirm the new auto-promote fields are persisted end-to-end (model copyWith, JSON, daemon config struct, fetcher behavior) — the diff alone does not demonstrate this.
- Document or test the assumption that an existing audit comment implies the label transition succeeded; otherwise consider checking actual label state before skipping.

---

ivanmunozruiz:
Addressed the requested changes in `5f823eb`:

- Removed the Flutter auto-promote UI changes from this PR so #412 stays focused on the backend bug from #411.
- Tightened manual stage promotion dedup: an existing audit comment only skips the transition when the issue labels already reflect the target stage. If labels are stale, Heimdallm retries `AddLabels`/`RemoveLabels` while suppressing a duplicate audit comment.
- Centralized stage-promotion comment matching in one helper used by both dedup paths.
- Added regression coverage for:
  - already-audited + labels already applied,
  - already-audited + stale labels needing normalization,
  - different transition audit not suppressing the current transition,
  - old audit comment not suppressing a new transition,
  - `CreatedAt` fallback when `CommentedAt` is zero.

Validation: `make test-docker` passes.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #412 after addressing previous feedback. The PR now correctly dedupes stage promotion audits by checking both audit comment existence and current label state, with a centralized matcher helper and comprehensive regression tests covering happy/sad paths (already-audited+applied, stale labels, different transitions, old audits, CreatedAt fallback). Logic is sound, tests are deterministic, and the suppression flag flows cleanly through StageTransition. No new issues found.

### Suggestions

- Consider extracting the `ref := latest.CommentedAt; if ref.IsZero() { ref = latest.CreatedAt }` pattern into a small helper if it gets reused elsewhere.
- The debug log message 'audit exists but labels need normalization' is informative; consider promoting to Info level since it represents a recoverable inconsistency worth surfacing in production logs.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: the author addressed all prior findings. PR scope is now narrowed to the backend fix; dedup no longer permanently blocks label retries when labels are stale (uses SuppressAudit to skip the comment but still call TransitionIssueStage); comment matching is centralized in stagePromotionCommentMatches; and regression tests cover different-transition, older-than-ref, label-normalization, and CreatedAt-fallback cases. Remaining nits are minor.

### Issues

🟡 **daemon/internal/issues/stage_transition.go:158** — When SuppressAudit is true, `commented` stays false and is then passed to publishStagePromotionEvent. If downstream consumers of the broker payload use `commented` to decide whether an audit exists, this will under-report (an audit comment does exist, it just wasn't posted by this call). Consider passing `commented || tr.SuppressAudit` or renaming the field to reflect 'audit posted in this call' vs 'audit present'.
🟡 **daemon/internal/issues/stage_transition.go:293** — hasRecentStagePromotionComment now delegates to stagePromotionCommentMatches with from="", which silently disables the From check. That preserves prior behavior, but the empty-string sentinel is easy to misuse. A small comment on stagePromotionCommentMatches noting that empty from/to means 'don't filter on that field' would prevent future foot-guns.
⚠️ **daemon/internal/issues/fetcher.go:353** — `ref := latest.CommentedAt; if ref.IsZero() { ref = latest.CreatedAt }` assumes `latest` is non-nil at this point. Worth a quick check that earlier code in auditManualStageChange guarantees `latest != nil`; if it can be nil this will panic. (Not visible in the diff — flag for verification.)
⚠️ **daemon/internal/issues/stage_transition.go:248** — stageTransitionApplied requires the target label present AND no other stage labels present. If a repo configures multiple labels per stage (e.g. RefinementLabels=["refine","in-progress"]), `stageTargetLabel` returns only one — issues with the alternate label set will be treated as not-applied and trigger label rewrites on every poll. Confirm stageTargetLabel/stageLabelsExcept handle multi-label stages correctly.

### Suggestions

- Add a doc comment to stagePromotionCommentMatches clarifying that empty from/to disables that filter.
- Verify `latest` non-nil precondition at fetcher.go:353 (or guard it) since dereferencing a nil review pointer would panic.
- Confirm stageTransitionApplied behaves correctly when a stage has multiple configured labels; if not, dedup may oscillate.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured dedup logic for stage promotion audits with comprehensive test coverage. The separation into early-return (labels already correct) vs. suppress-comment-only (stale labels) paths is clean. Five targeted regression tests cover the key edge cases. A few minor observations below.

### Issues

🟡 **daemon/internal/issues/stage_transition.go:305** — hasStagePromotionCommentSince does not filter by comment author (e.g. bot login). A human who pastes the exact audit-comment format would be treated as a valid audit. The existing hasRecentStagePromotionComment has the same gap, so this is pre-existing, but the new since-based check amplifies the window in which a false positive could suppress a real audit.
🟡 **daemon/internal/issues/stage_transition.go:243** — stageTransitionApplied references stageTargetLabel and stageLabelsExcept which are not included in this diff. Unable to verify their correctness or confirm they handle all stage configurations (e.g. empty label lists, multi-label stages). Please confirm these helpers exist and have test coverage.
🟡 **daemon/internal/issues/fetcher.go:351** — If both latest.CommentedAt and latest.CreatedAt are zero (e.g. a store row with default timestamps), ref is zero, hasStagePromotionCommentSince returns false, and the code falls through to the normal audit path. This is correct behavior, but worth a brief comment since the zero-time semantics are load-bearing for correctness.
🟡 **daemon/internal/issues/stage_transition.go:315** — stagePromotionCommentMatches relies on substring matching against formatted lines (e.g. '- From: `triage`'). If the audit comment template ever changes its markdown format, this matcher silently breaks. Consider extracting the format strings into shared constants used by both StagePromotionAuditComment and stagePromotionCommentMatches to keep them in sync.

### Suggestions

- Consider adding an author filter (bot login) to hasStagePromotionCommentSince to prevent false-positive dedup from non-bot comments that happen to match the heading format.
- Extract the '- From: `%s`' and '- To: `%s`' format patterns into package-level constants shared between StagePromotionAuditComment and stagePromotionCommentMatches so template changes don't silently break matching.
- The test helper stagePipelineCfg() hard-codes single-label stages. Consider adding a test variant with multi-label stage configs to verify stageTransitionApplied handles them correctly.

---

ivanmunozruiz:
Small follow-up pushed in `208164b` for the latest review nits:

- `audit_comment` in the promotion event now reports true when `SuppressAudit` means an audit already existed and was intentionally not reposted.
- Added a unit test for that `SuppressAudit` event payload behavior.
- Added a defensive `latest == nil` guard in `auditManualStageChange`.
- Documented that empty `from` / `to` in `stagePromotionCommentMatches` intentionally disables that filter.

Validation:
- `make test-docker GO_TEST_ARGS="-run 'TestFetcher_ManualStageLabelChange|TestTransitionIssueStage' ./internal/issues/..."`
- `make test-docker`

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: author addressed three of four prior findings (nil-check on latest, commented||SuppressAudit propagation, doc comment on empty-stage sentinel). Multi-label stage concern remains unverified in the diff. New test coverage is solid and exercises the dedup paths well.

### Issues

⚠️ **daemon/internal/issues/stage_transition.go:248** — stageTransitionApplied still relies on stageTargetLabel returning a single label. If a stage is configured with multiple labels (e.g. RefinementLabels=["refine","in-progress"]), an issue carrying only the non-primary label will be treated as not-applied, triggering label rewrites every poll and defeating the dedup. None of the new tests cover the multi-label case — please confirm stageTargetLabel/stageLabelsExcept handle this, or add a test pinning the behavior.

### Suggestions

- Add a regression test for stageTransitionApplied with a stage configured for multiple labels to lock down behavior and prevent silent dedup bypass.
- Consider renaming SuppressAudit to something like AuditAlreadyExists since that is what it now semantically signals to publishStagePromotionEvent (the OR with tr.SuppressAudit when setting commented makes the field's meaning broader than 'suppress').

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Follow-up addresses prior review nits cleanly: nil guard on latest review, audit_comment now reflects suppressed-but-present audit, doc comment on stagePromotionCommentMatches, and a focused unit test for the SuppressAudit event payload. Logic is consistent with the dedup intent and tests cover the relevant branches.

### Suggestions

- Consider extracting the `commented || tr.SuppressAudit` expression into a named local (e.g. `auditPresent`) to make the publish call self-documenting.
- The `stageTransitionApplied` check uses `LabelNames()` which may include extra labels; consider asserting that no other stage's labels are present is already handled, but documenting the invariant in a short comment would help future readers.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

The author addressed the key feedback from the previous review: added the `latest == nil` guard, documented the empty from/to semantics in `stagePromotionCommentMatches`, fixed the `audit_comment` event payload semantics, and added thorough test coverage. The remaining prior findings (no author filtering on comment matching, format string coupling) are pre-existing patterns and low-risk. No new issues introduced in this revision. The dedup logic is sound and well-tested across five distinct scenarios.

### Suggestions

- Consider extracting the `- From: ` and `- To: ` format patterns into package-level constants shared between StagePromotionAuditComment and stagePromotionCommentMatches to prevent silent drift if the template changes (carried forward from prior review, still low priority).

---

Issue #414, opened 2026-05-08
docs: clarify staged issue pipeline smoke-test flow


ivanmunozruiz:
## Goal

Add a short documentation note that explains how to smoke-test the staged issue pipeline end to end.

## Context

Heimdallm now supports a staged issue flow:

1. `heimdallm_triage` runs issue triage.
2. `heimdallm_refine` runs refinement / implementation planning.
3. `heimdallm_develop` runs auto-implementation and opens a PR.

This issue intentionally starts with only `heimdallm_triage` so Heimdallm can exercise the configured promotions.

## Requested Change

Add or update a small docs section that describes the recommended smoke-test labels and expected transitions:

```toml
review_only_labels = ["heimdallm_triage"]
refinement_labels = ["heimdallm_refine"]
develop_labels = ["heimdallm_develop"]
auto_promote_triage = true
auto_promote_refinement = true
```

The note should make clear that if `auto_promote_refinement=false`, the issue should stop in refinement until manually promoted.

## Acceptance Criteria

- Documentation mentions the three labels above.
- Documentation explains triage -> refinement -> develop -> PR at a high level.
- Documentation states that refinement only auto-promotes to develop when `auto_promote_refinement=true`.
- Keep the change small and limited to docs.

## Test Purpose

This issue is also a live smoke test for Heimdallm staged issue processing. If the configured flow is working, Heimdallm should triage, refine, promote, implement a docs-only PR, and link that PR here.

ivanmunozruiz:
## 🟡 Heimdallm triage — LOW

Request to add a small docs section in the configuration guide explaining how to smoke-test the staged issue pipeline (heimdallm_triage → heimdallm_refine → heimdallm_develop → PR), the recommended toml snippet using those labels, and the expectation that refinement only auto-promotes to develop when auto_promote_refinement=true. Acceptance criteria are clear and bounded to docs only. The issue itself is also intended as a live end-to-end smoke test of the staged flow.

### Classification

- **Category:** docs
- **Suggested severity:** low
- **Priority label:** priority: low
- **Likely affected area:** docs/configuration-guide.md (issue tracking / stage promotion section)
- **Affected paths:** docs/configuration-guide.md, README.md
- **Assigned owner:** @vbuenog
- **Suggested assignee:** @ivanmunozruiz
[...]

PR #419, opened 2026-05-08
fix(issues): scope triage breaker to triage runs

### LINKED ISSUE #418: Bug: triage circuit breaker blocks promoted refinement/develop stages
## Problem

A promoted issue can reach the refinement worker but fail before refinement starts because the issue triage circuit breaker runs for every issue mode.

Observed on #414 after rebuilding with the stage-grace fix:

```text
refinement-worker: processing repo=theburrowhub/heimdallm number=414
issues pipeline: CIRCUIT BREAKER TRIPPED — skipping triage reason="per-issue cap reached: 3 triages in last 24h (cap 3)"
refinement-worker: pipeline run failed err="issues pipeline: circuit breaker tripped..."
```

## Expected

The circuit breaker should cap only triage/re-triage loops. It should not block refinement or develop after a stage promotion.

## Fix direction

- Run the issue circuit breaker only when the effective mode is triage/review_only.
- Count only `action_taken='review_only'` rows for issue triage breaker budgets, so refinement/develop rows do not consume triage quota.
- Add regression tests for refinement and develop with a tripped triage breaker.

## Acceptance Criteria

- An issue with `heimdallm_refine` still runs refinement even if the triage breaker would be tripped.
- An issue with `heimdallm_develop` still runs develop even if the triage breaker would be tripped.
- The breaker still protects triage/re-triage loops.
- Docker Go tests pass.

ivanmunozruiz:
## Summary

- Run the issue circuit breaker only for effective triage/review_only runs, so promoted refinement/develop stages are not blocked by prior triage activity.
- Count only `action_taken='review_only'` rows in issue triage breaker budgets; refinement/develop rows no longer consume triage quota.
- Add regression coverage for refinement/develop with a tripped triage breaker and for store-level action filtering.

## Root Cause

In the #414 smoke test, the daemon correctly classified the issue as refinement and queued the `refinement-worker`, but `Pipeline.Run` checked the issue circuit breaker before dispatching on mode. The breaker was originally designed to cap runaway triage/re-triage loops, but with staged issue processing it also blocked refinement/develop once a promoted issue had enough prior issue_reviews rows.

## Validation

- `make test-docker GO_TEST_ARGS="-run 'TestPipeline_CircuitBreakerDoesNotGate|TestCountIssueReviewsForIssue_IgnoresNonTriageActions|TestCountIssueTriagesForRepo_IgnoresNonTriageActions|TestCheckIssueCircuitBreaker|TestPipeline_RunRefinementHappyPath|TestPipeline_AutoImplementHappyPath' ./internal/issues/... ./internal/store/..."`
- `make test-docker`

Fixes #418.

### DIFF
--- a/daemon/internal/issues/pipeline.go
+++ b/daemon/internal/issues/pipeline.go
@@ -420,12 +420,12 @@ func (p *Pipeline) Run(ctx context.Context, issue *github.Issue, opts RunOptions
 		return nil, fmt.Errorf("issues pipeline: refusing an ignore-classified issue (fetcher should have filtered it out)")
 	}
 
-	// Upsert + initial SSE events are common to both flows so we do them
+	// Upsert + initial SSE events are common to all flows so we do them
 	// here. issue_detected fires before the flow forks, issue_review_started
 	// fires after so the UI can show the correct "triaging" vs "implementing"
 	// copy — the runner sets the exact flavour it wants.
 	//
-	// Upsert runs BEFORE the circuit breaker so the breaker's per-issue
+	// For triage runs, upsert runs BEFORE the circuit breaker so its per-issue
 	// count (which keys on the internal store ID via issue_reviews.issue_id)
 	// sees the correct row. The upsert is idempotent — on a breaker-trip
 	// the issue row stays but no issue_reviews row is written for this
@@ -439,12 +439,14 @@ func (p *Pipeline) Run(ctx context.Context, issue *github.Issue, opts RunOptions
 		return nil, fmt.Errorf("issues pipeline: upsert issue: %w", err)
 	}
 
-	// Circuit breaker: hard cap on triage count per issue / per repo.
+	// Circuit breaker: hard cap on triage count per issue / per repo. It must
+	// not gate refinement or develop, otherwise a correctly promoted issue can
+	// get stuck after earlier triage/refinement activity.
 	// Runs AFTER the in-flight claim and upsert so it only fires when
 	// both dedup layers missed; returns *CircuitBreakerError so the
 	// caller (fetcher) can distinguish it from a genuine pipeline
 	// failure. See theburrowhub/heimdallm#292.
-	if p.breaker != nil {
+	if p.breaker != nil && effective == config.IssueModeReviewOnly {
 		tripped, reason, err := p.store.CheckIssueCircuitBreaker(issueID, issue.Repo, *p.breaker)
 		if err != nil {
 			slog.Warn("issues pipeline: circuit breaker check failed, proceeding",
--- a/daemon/internal/issues/pipeline_test.go
+++ b/daemon/internal/issues/pipeline_test.go
@@ -633,6 +633,29 @@ func TestPipeline_RunRefinementHappyPath(t *testing.T) {
 	}
 }
 
+func TestPipeline_CircuitBreakerDoesNotGateRefinement(t *testing.T) {
+	s := &fakeStore{
+		breakerTripped: true,
+		breakerReason:  "triage cap reached",
+	}
+	gh := &fakeGH{}
+	exec := &fakeExec{detectCLI: "claude", rawOutput: []byte(validRefinementResult)}
+	p := issues.New(s, gh, exec, nil, &fakeBroker{}, nil)
+	p.SetCircuitBreakerLimits(&store.IssueCircuitBreakerLimits{PerIssue24h: 3, PerRepoHr: 10})
+
+	rev, err := p.Run(context.Background(), newIssue(config.IssueModeRefinement), issues.RunOptions{
+		Primary:                     "claude",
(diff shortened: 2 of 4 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Correctly scopes the triage circuit breaker to review_only runs so refinement/develop are not blocked. SQL filter, pipeline gate, and tests all align. Minor concern: action_taken value is hardcoded as the literal 'review_only' in SQL rather than referencing config.IssueModeReviewOnly, creating a drift risk if the constant ever changes.

### Suggestions

- Replace the hardcoded 'review_only' literal in the two SQL queries with a parameter bound from string(config.IssueModeReviewOnly) to keep the breaker filter in sync if the mode constant is ever renamed.
- Consider adding a store-level test that asserts CheckIssueCircuitBreaker does not trip when only refinement/develop rows exist at-or-above the cap, complementing the existing pipeline-level tests.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Clean, focused fix that scopes the issue circuit breaker to triage runs only, preventing legitimate refinement/develop runs from being blocked by historical triage activity. Defense-in-depth: gated both at the pipeline call site and at the SQL count level. Good test coverage for both layers.

### Issues

🟡 **daemon/internal/store/issue_circuitbreaker.go:17** — The literal 'review_only' is hardcoded in two SQL queries but corresponds to config.IssueModeReviewOnly elsewhere in the codebase. If that constant's underlying string ever changes, these queries will silently undercount and the breaker will stop tripping without any test failure pointing here. Consider plumbing the constant through (e.g. via a package-level var or query parameter) so the coupling is explicit.
🟡 **daemon/internal/issues/pipeline.go:447** — The pipeline-level guard `effective == config.IssueModeReviewOnly` makes the SQL-level filter redundant for the current call path. That redundancy is fine as defense-in-depth, but it's worth confirming there is no other caller of CountIssueReviewsForIssue/CountIssueTriagesForRepo that expected the old 'count all action types' semantics (e.g. a metrics or admin endpoint) — the rename of the error string to 'count issue triages' suggests intent, but a quick grep for callers would confirm no behavioral regression elsewhere.
🟡 **daemon/internal/issues/pipeline.go:423** — Updated comment says 'For triage runs, upsert runs BEFORE the circuit breaker' but upsert still runs before the breaker for ALL flows (it's outside the new `effective == IssueModeReviewOnly` guard). The original comment's ordering rationale applies whenever the breaker runs, which is now only for triage — so the wording is technically correct but could mislead a reader into thinking upsert ordering is conditional. Minor.

### Suggestions

- Consider extracting the 'review_only' literal in the two SQL queries to a named constant alongside config.IssueModeReviewOnly (or pass it as a query parameter) so a future rename of the mode value can't silently break the breaker.
- Add a regression test that asserts the breaker DOES still trip on a 3rd review_only run when prior refinement/develop rows exist — i.e. that the action_taken filter doesn't accidentally exclude the rows it's supposed to count when mixed with non-triage rows.
- Tighten the pipeline.go comment on line 423 to reflect that the breaker only runs for triage now, so the upsert-before-breaker ordering only matters in that flow.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-tested fix that correctly scopes the circuit breaker to triage-only runs. The two-layer approach (pipeline gate + SQL filter) provides good defense-in-depth. The downgrade path (develop without workdir falls back to review_only) correctly still hits the breaker. No correctness issues found.

### Issues

🟡 **daemon/internal/store/issue_circuitbreaker.go:12** — The SQL literal `'review_only'` is a silent coupling to `config.IssueModeReviewOnly` (defined as `"review_only"`). If that constant ever changes, these queries will silently stop matching. Consider adding a brief code comment (e.g., `// must match config.IssueModeReviewOnly`) or extracting the string via a package-level const to keep them in sync.
🟡 **daemon/internal/store/issue_circuitbreaker.go:34** — Same hardcoded `'review_only'` literal as above in `CountIssueTriagesForRepo`. If one is updated the other must be too — easy to miss.
🟡 **daemon/internal/issues/pipeline_test.go:646** — The test for refinement bypassing the breaker sets `breakerTripped: true` on the fake store, which is good, but doesn't verify that `CheckIssueCircuitBreaker` was never called. An explicit assertion (e.g., a call counter on the fake) would distinguish 'breaker skipped at pipeline level' from 'breaker called but didn't trip due to the SQL filter', strengthening confidence in the two-layer design.

### Suggestions

- Consider defining a package-level constant like `actionTriageOnly = string(config.IssueModeReviewOnly)` in the store package and referencing it in both SQL queries to eliminate the silent string coupling.
- A small follow-up to add an index on `(issue_id, created_at, action_taken)` could help if the issue_reviews table grows large, since both queries now filter on all three columns.

---

PR #425, opened 2026-05-08
fix(issues): dedupe stale stage promotion audits

### LINKED ISSUE #424: fix(issues): do not re-emit stale stage promotion audits
## Problem

During the staged issue pipeline smoke test on #414, the issue correctly moved from refinement to development, but the activity log then kept recording repeated `Promoted: triage -> development` entries on each poll.

The issue already had a valid stage-promotion audit for the target stage (`To: development`) and the labels were already normalized (`heimdallm_develop` present, previous stage labels removed). The manual-label-change auditor was still matching the exact stale `from` stage from the latest stored review, so it treated the already-audited target as a new transition.

## Expected behavior

Once a promotion to the current target stage is already audited after the latest issue review and the stage labels are applied, the fetcher should not mutate labels, post duplicate audit comments, or emit duplicate activity events.

## Acceptance criteria

- Existing exact transition dedup still works.
- Existing target-stage audit dedups stale `from` transitions when labels are already applied.
- If labels are stale, the daemon may still normalize labels but must suppress duplicate audit comments.
- The activity log no longer shows repeated stale `triage -> development` promote rows for the same already-applied target stage.

ivanmunozruiz:
## What changed

- Dedupes manual stage-promotion audits by target stage as well as exact `from -> to` pair.
- Keeps label-normalization retry behavior when labels are stale, while suppressing duplicate audit comments.
- Adds regression coverage for the observed #414 case where `refinement -> development` was already audited but a stale local `from=triage` caused repeated `triage -> development` activity rows.

## Root cause

The manual GitHub label-change auditor only treated an audit as existing when the stored latest review's inferred `from` stage exactly matched the audit comment. If the latest stored review lagged behind the actual stage path, the issue could already have a valid `To: development` audit and normalized labels, but the fetcher still emitted another promotion event using the stale `from` stage.

## Validation

- `make test-docker GO_TEST_ARGS="-run TestFetcher_ManualStageLabelChange ./internal/issues"`
- `make test-docker`

Closes #424

### DIFF
--- a/daemon/internal/issues/fetcher.go
+++ b/daemon/internal/issues/fetcher.go
@@ -367,12 +367,13 @@ func (f *Fetcher) auditManualStageChange(ctx context.Context, issue *github.Issu
 		ref = latest.CreatedAt
 	}
 	alreadyAudited := hasStagePromotionCommentSince(comments, from, to, ref)
-	if alreadyAudited && stageTransitionApplied(issue, cfg, to) {
+	targetAudited := alreadyAudited || hasStagePromotionTargetCommentSince(comments, to, ref)
+	if targetAudited && stageTransitionApplied(issue, cfg, to) {
 		slog.Debug("issues fetcher: stage transition already audited since latest review",
 			"repo", issue.Repo, "number", issue.Number, "from", from, "to", to)
 		return nil
 	}
-	if alreadyAudited {
+	if targetAudited {
 		slog.Debug("issues fetcher: stage transition audit exists but labels need normalization",
 			"repo", issue.Repo, "number", issue.Number, "from", from, "to", to)
 	}
@@ -386,7 +387,7 @@ func (f *Fetcher) auditManualStageChange(ctx context.Context, issue *github.Issu
 		Trigger:        StagePromotionManualGitHub,
 		Time:           time.Now().UTC(),
 		RecentComments: comments,
-		SuppressAudit:  alreadyAudited,
+		SuppressAudit:  targetAudited,
 		Broker:         f.stageBroker,
 	})
 }
--- a/daemon/internal/issues/fetcher_test.go
+++ b/daemon/internal/issues/fetcher_test.go
@@ -1010,6 +1010,60 @@ func TestFetcher_ManualStageLabelChangeRetriesLabelsWhenAuditExistsButLabelsAreS
 	}
 }
 
+func TestFetcher_ManualStageLabelChangeSkipsStaleFromWhenTargetAlreadyAudited(t *testing.T) {
+	now := time.Now()
+	issue := &github.Issue{
+		ID:        int64(1008),
+		Number:    8,
+		Repo:      "org/repo",
+		Title:     "Ready to build",
+		UpdatedAt: now,
+		Mode:      config.IssueModeDevelop,
+		Labels:    issueLabels("develop"),
+	}
+	latestReviewAt := now.Add(-5 * time.Minute)
+	dedup := &fakeDedup{byGithubID: map[int64]dedupEntry{
+		issue.ID: {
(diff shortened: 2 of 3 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Targeted dedup fix that treats any prior audit comment matching the target stage as sufficient to suppress duplicate stale-from stage promotion audits. The change is minimal, well-scoped, and accompanied by a focused unit test covering the new behavior.

### Suggestions

- Consider adding a sad-path test where the target-stage audit predates the latest review (since.Before check) to lock in that stale audits do not suppress new transitions.
- hasStagePromotionTargetCommentSince duplicates the iteration shape of hasStagePromotionCommentSince; a small shared helper parameterized by from could reduce drift if future filters are added.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-scoped fix that broadens stage-promotion dedup to match on the target stage alone, preventing duplicate audit comments when an earlier promotion to the same target was already recorded under a different 'from' stage. Logic, naming, and test coverage are solid.

### Issues

🟡 **daemon/internal/issues/fetcher.go:370** — The broadened dedup (matching any promotion to `to` regardless of `from`) could theoretically suppress a legitimate audit if two semantically distinct transitions happen to share the same target stage within the same review window. This is likely acceptable given the 'since latest review' scoping, but worth confirming there's no scenario where the `from` stage carries audit-significant information that should produce a distinct comment even when the target matches.
🟡 **daemon/internal/issues/stage_transition.go:327** — The new `hasStagePromotionTargetCommentSince` duplicates the iteration/filtering skeleton of `hasStagePromotionCommentSince` (zero-time guard, loop, CreatedAt check). Consider whether a single helper with an optional `from` parameter (the wildcard behavior already exists in `stagePromotionCommentMatches`) would reduce surface area, though this is a minor maintainability nit.

### Suggestions

- Add a brief inline comment on the `targetAudited` line explaining *why* matching on target-only is safe (e.g., 'a stale from stage should not re-trigger an audit when the target was already recorded'), since the dedup broadening is the core behavioral change and the reasoning isn't obvious from the code alone.
- Consider a negative test case where the existing audit comment targets a *different* stage than `to`, confirming that the new path does NOT suppress the audit in that scenario.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Targeted fix that broadens stage-promotion audit dedup to also recognize comments matching the target stage with any from-stage. Logic, naming, and test coverage are all reasonable.

### Issues

🟡 **daemon/internal/issues/stage_transition.go:319** — hasStagePromotionTargetCommentSince duplicates almost all of hasStagePromotionCommentSince; consider parameterizing the existing helper (or having it call this one with a non-empty from) to avoid two near-identical loops drifting apart.
🟡 **daemon/internal/issues/fetcher.go:370** — alreadyAudited is still computed even though targetAudited is a strict superset (alreadyAudited implies targetAudited). The first call is now redundant and could be dropped, or short-circuited by calling the broader check first.
🟡 **daemon/internal/issues/stage_transition.go:319** — Target-only matching means any prior promotion ending at `to` (e.g. from a different from-stage on an older transition) suppresses a fresh audit. That is the intended dedup behavior here, but it widens the suppression window — worth a brief comment on the helper documenting that intent so future readers don't tighten it back up.

### Suggestions

- Consolidate hasStagePromotionCommentSince and hasStagePromotionTargetCommentSince into one helper that accepts an optional from stage, reusing stagePromotionCommentMatches' empty-string convention.
- Add a short doc comment on hasStagePromotionTargetCommentSince explaining why matching on target alone is correct (mirroring the rationale already on stagePromotionCommentMatches).
- Consider a second test where a target-only audit exists AND labels are still stale, to lock in that the normalization-debug branch is also hit via the new path.

---

PR #429, opened 2026-05-08
fix(issues): scope staged pipeline by assignee

ivanmunozruiz:
## Summary
- default issue-tracking assignee scope to the authenticated GitHub login instead of treating empty config as a shared queue
- skip stale triage/refinement/development jobs when the issue has been reassigned outside this daemon's scope
- allow issues assigned to this daemon with a refinement label to enter refinement directly after another Heimdallm completed triage
- update docs/config copy and tests for assignee-scoped staged issue processing

## Tests
- `make test-docker GO_TEST_ARGS="-run 'Test(IssueTrackingWithDefaultAssignee|IssueTrackingMatchesAssignees|FetchIssues_AssigneeScopedStageCanStartAtRefinement|Fetcher_DefaultsAssigneeScopeToAuthenticatedUser|Fetcher_SkipsWhenAssigneeScopeUnavailable)' ./internal/config ./internal/github ./internal/issues"`

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -334,7 +334,7 @@ func main() {
 	// loginMu guards cachedLogin against concurrent reads/writes from the
 	// poll cycle and HTTP goroutines.
 	var loginMu sync.Mutex
-	var cachedLogin string
+	var cachedLogin = resolvedBotLogin
 
 	buildRunOpts := func(pr *gh.PullRequest, aiCfg config.RepoAI) pipeline.RunOptions {
 		cli := aiCfg.Primary
@@ -840,6 +840,14 @@ func main() {
 		localDirBase := c.GitHub.LocalDirBase
 		globalTimeout := c.AI.ExecutionTimeout
 		cfgMu.Unlock()
+		loginMu.Lock()
+		authUser := cachedLogin
+		loginMu.Unlock()
+		var ok bool
+		repoIT, ok = issueTrackingWithAssigneeScope("triage-worker", msg.Repo, repoIT, authUser)
+		if !ok {
+			return
+		}
 		if !issueStageStillCurrent("triage-worker", ghIssue, repoIT, config.IssueModeReviewOnly) {
 			return
 		}
@@ -944,6 +952,14 @@ func main() {
 		localDirBase := c.GitHub.LocalDirBase
 		globalTimeout := c.AI.ExecutionTimeout
 		cfgMu.Unlock()
+		loginMu.Lock()
+		authUser := cachedLogin
+		loginMu.Unlock()
+		var ok bool
+		repoIT, ok = issueTrackingWithAssigneeScope("refinement-worker", msg.Repo, repoIT, authUser)
+		if !ok {
+			return
+		}
 		if !issueStageStillCurrent("refinement-worker", ghIssue, repoIT, config.IssueModeRefinement) {
 			return
 		}
@@ -1013,6 +1029,14 @@ func main() {
 		localDirBase := c.GitHub.LocalDirBase
 		globalTimeout := c.AI.ExecutionTimeout
 		cfgMu.Unlock()
+		loginMu.Lock()
+		authUser := cachedLogin
+		loginMu.Unlock()
+		var ok bool
+		repoIT, ok = issueTrackingWithAssigneeScope("implement-worker", msg.Repo, repoIT, authUser)
+		if !ok {
+			return
+		}
 		if !issueStageStillCurrent("implement-worker", ghIssue, repoIT, config.IssueModeDevelop) {
 			return
 		}
@@ -1240,6 +1264,10 @@ func main() {
 		nonMonList := append([]string(nil), c.GitHub.NonMonitored...)
 		localDirBaseList := append([]string(nil), c.GitHub.LocalDirBase...)
 		cfgMu.Unlock()
+		loginMu.Lock()
+		authUser := cachedLogin
+		loginMu.Unlock()
+		issueTracking := c.GitHub.IssueTracking.WithDefaultAssignee(authUser)
 		orgOverrides := make(map[string]map[string]any)
 		for org, ai := range c.AI.Orgs {
 			orgOverrides[org] = orgAIOverrideMap(ai)
@@ -1320,7 +1348,7 @@ func main() {
 			"ai_fallback":                 c.AI.Fallback,
 			"review_mode":                 c.AI.ReviewMode,
 			"retention_days":              c.Retention.MaxDays,
-			"issue_tracking":              c.GitHub.IssueTracking,
+			"issue_tracking":              issueTracking,
 			"repo_overrides":              repoOverrides,
 			"org_overrides":               orgOverrides,
 			"agent_configs":               agentConfigs,
@@ -2258,6 +2286,41 @@ type tier2Adapter struct {
 
 const breakerTripDedupTTL = 24 * time.Hour
 
+func (a *tier2Adapter) cachedAuthenticatedUser() string {
+	if a == nil || a.login == nil {
+		return ""
+	}
+	if a.loginMu == nil {
+		return *a.login
+	}
+	a.loginMu.Lock()
+	defer a.loginMu.Unlock()
+	return *a.login
+}
+
+func (a *tier2Adapter) resolveAuthenticatedUser() string {
+	if authUser := a.cachedAuthenticatedUser(); authUser != "" {
+		return authUser
+	}
+	if a == nil || a.ghClient == nil {
+		return ""
+	}
+	u, err := a.ghClient.AuthenticatedUser()
+	if err != nil {
+		return ""
+	}
+	if a.login != nil {
+		if a.loginMu == nil {
+			*a.login = u
+		} else {
+			a.loginMu.Lock()
+			*a.login = u
+			a.loginMu.Unlock()
+		}
+	}
+	return u
+}
+
 type breakerTripKey struct {
 	Repo    string
 	Number  int
@@ -2661,17 +2724,14 @@ func (a *tier2Adapter) ProcessRepo(ctx context.Context, repo string) (int, error
 		return 0, nil
 	}
 
-	// Resolve authenticated user
-	a.loginMu.Lock()
-	authUser := *a.login
-	a.loginMu.Unlock()
-	if authUser == "" {
-		if u, err := a.ghClient.AuthenticatedUser(); err == nil {
-			authUser = u
-			a.loginMu.Lock()
-			*a.login = u
-			a.loginMu.Unlock()
-		}
+	// Resolve authenticated user before applying issue tracking defaults: an
+	// empty assignee list means "this daemon's user", not a shared queue.
+	authUser := a.resolveAuthenticatedUser()
+
+	var ok bool
+	repoIT, ok = issueTrackingWithAssigneeScope("tier2 issue processing", repo, repoIT, authUser)
+	if !ok {
+		return 0, nil
 	}
 
 	optsFor := func(issue *gh.Issue) (issuepipeline.RunOptions, bool) {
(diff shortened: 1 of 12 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Scopes the staged issue pipeline by assignee, defaulting to the authenticated GitHub login when no explicit assignees are configured. Changes are well-tested, documented, and add a fail-safe skip when no scope can be resolved. This is a deliberate behavior change (empty assignees no longer means 'anyone') but is documented in the config example and configuration guide.

### Suggestions

- issueTrackingWithAssigneeScope logs at Warn level on every invocation when scope is missing; since worker handlers call this on every dispatched job, this can become noisy in logs if auth resolution is temporarily unavailable. Consider rate-limiting the warning or downgrading repeat warnings to Debug.
- In daemon/cmd/heimdallm/main.go around line 1267, after computing issueTracking via WithDefaultAssignee(authUser), there is no fail-safe skip if authUser is empty — only the workers and tier2 paths skip. The status snapshot will then advertise an unscoped issue_tracking. Consider using issueTrackingWithAssigneeScope here too for consistency, or document why the snapshot path differs.
- tier2Adapter.PromoteReady duplicates the auth-user resolution block from ProcessRepo (lock/unlock + AuthenticatedUser fallback). Extracting a small helper (e.g., a.resolveAuthUser()) would reduce duplication and keep the locking pattern in one place.
- The unrelated whitespace/struct alignment changes in daemon/internal/github/fetch_issues_test.go (issueFixture realignment, trailing-comment indentation in TestFetchIssues_FilterModeInclusive) add diff noise unrelated to the PR scope; consider isolating cosmetic formatting in a separate commit.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid change that scopes issue processing to the daemon's authenticated user by default, preventing multi-instance conflicts. Core logic is correct, concurrency patterns are sound, and test coverage is thorough. A few minor issues around normalization consistency and a stale-comment edge case.

### Issues

🟡 **daemon/internal/config/config.go:233** — WithDefaultAssignee normalizes with TrimLeft(login, "@") but does not lowercase, while MatchesAssignees lowercases during comparison. This works correctly at runtime, but the stored Assignees value (e.g. "Alice") differs in casing from what MatchesAssignees compares against. If any future code path does a direct string equality check on Assignees instead of going through MatchesAssignees, it will silently fail. Consider normalizing to lowercase in WithDefaultAssignee as well, or extract a shared normalizeLogin helper.
🟡 **daemon/internal/config/config.go:233** — strings.TrimLeft(login, "@") treats "@" as a character set, not a prefix — it strips all leading '@' characters. While GitHub logins cannot start with '@', strings.TrimPrefix(login, "@") would be semantically clearer and is used elsewhere for this purpose.
🟡 **daemon/cmd/heimdallm/main.go:3452** — The new MatchesAssignees check in issueStageStillCurrent logs "issue assigned outside this daemon scope" but MatchesAssignees also returns false for *unassigned* issues (empty assignee list). The log message is misleading in that case — an unassigned issue isn't "assigned outside" the scope. Consider distinguishing the two cases for clearer operational debugging.
🟡 **daemon/cmd/heimdallm/main.go:1267** — c.GitHub.IssueTracking is accessed after cfgMu.Unlock(), unlike NonMonitored and LocalDirBase which are defensively copied before unlock. Since c is a value copy and WithDefaultAssignee returns a new struct, this is safe in practice, but the inconsistency with the surrounding defensive-copy pattern could mislead future maintainers. Consider moving the WithDefaultAssignee call above the Unlock or adding a brief comment explaining why no defensive copy is needed.
🟡 **daemon/cmd/heimdallm/main.go:2821** — In tier2Adapter.PromoteReady, the login-resolution fallback (AuthenticatedUser API call) runs outside any lock. Two concurrent PromoteReady calls could both see authUser=="" and both call AuthenticatedUser(). This is benign (idempotent + same result) but wastes an API call. Consider a sync.Once or check-lock-check pattern if this hot path matters.

### Suggestions

- Extract a normalizeLogin(s string) string helper (TrimSpace + TrimPrefix @ + ToLower) to DRY up the normalization in WithDefaultAssignee and MatchesAssignees and guarantee consistent casing.
- In issueStageStillCurrent, split the log message: 'issue unassigned, skipping' vs 'issue assigned to X, outside daemon scope' for better operational clarity.
- Consider adding a test for WithDefaultAssignee when the login is empty string or whitespace-only to document the no-op behavior explicitly.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid scoping change that closes a real correctness gap: workers and promoters previously ran on issues that didn't belong to this daemon's user. The `WithDefaultAssignee` / `MatchesAssignees` API is well-factored, the duplicate `hasAnyAssignee` is removed cleanly, and tests cover the new paths. A few minor concerns around silent skipping when the auth user is unresolved and one possibly-noisy log path.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:337** — `cachedLogin = resolvedBotLogin` makes the worker handlers' early-return safe only if `resolvedBotLogin` is reliably populated at startup. If it can be empty (e.g., bot login resolution deferred to the first poll), every triage/refinement/implement message that arrives before the first successful `AuthenticatedUser()` call will silently return via `issueTrackingWithAssigneeScope` with only a warn log — no requeue, no retry. Worth confirming `resolvedBotLogin` is guaranteed non-empty here, or having the workers fall back to `AuthenticatedUser()` like `tier2Adapter.ProcessRepo` does.
🟡 **daemon/cmd/heimdallm/main.go:2832** — In `PromoteReady`, if `*a.login` is still empty after the `AuthenticatedUser()` fallback (transient API error), every repo will hit the warn log inside `issueTrackingWithAssigneeScope` and be skipped. That's correct behavior, but emits one warn per repo per promotion cycle — could be noisy on multi-repo deployments during a GitHub outage. Consider logging once at the call site when `authUser == ""` and short-circuiting the loop.
🟡 **daemon/internal/config/config.go:254** — `MatchesAssignees` lower-cases/trims `c.Assignees` on every call. Hot paths (`issueStageStillCurrent`, per-issue promote loop, per-issue filter in `fetch_issues.go`) now repeat this normalization for every issue. Not a correctness issue, but if assignee lists are long or issue volume is high, normalizing once when the config is loaded would be cheaper. Low priority.
🟡 **daemon/cmd/heimdallm/main.go:1268** — The status payload now reports `issueTracking` with the auth-user-defaulted assignees rather than the raw config. Operators reading `/status` will see a synthetic `assignees: [<bot-login>]` even when their config has no `assignees` set. That's arguably the desired effective view, but is a user-visible change worth calling out in the PR description / CHANGELOG so existing dashboards aren't surprised.

### Suggestions

- Add a test (or comment) confirming `resolvedBotLogin` is always populated before the worker subscriptions start, since the new worker early-return depends on it.
- Consider hoisting the `AuthenticatedUser()` fallback into a small helper shared by `ProcessRepo`, `PromoteReady`, and the worker handlers — the pattern is now duplicated four times.
- Document in `IssueTrackingConfig` doc comment that `MatchesAssignees` does its own normalization, so callers don't need to pre-trim the input slice (e.g., `issue.AssigneeLogins()`).
- The unrelated whitespace-only change on `fetch_issues_test.go:316` (extra spaces in a comment) is harmless but inflates the diff — easy to drop.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: PR scopes the staged issue pipeline by authenticated assignee, defaulting empty assignee config to the bot login. Previously raised concerns appear unchanged in this diff. No new regressions spotted; main remaining concerns are around startup ordering of resolvedBotLogin and minor noise/perf details.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:337** — `cachedLogin = resolvedBotLogin` still relies on resolvedBotLogin being non-empty at this point. If resolution is deferred or fails at startup, the new worker handlers (triage/refinement/implement) will silently drop messages via issueTrackingWithAssigneeScope's warn-only path with no requeue or retry, unlike tier2Adapter.ProcessRepo which falls back to AuthenticatedUser(). Confirm resolvedBotLogin is guaranteed populated here, or have the workers fall back to a live lookup.
🟡 **daemon/cmd/heimdallm/main.go:2849** — PromoteReady now resolves authUser once outside the loop and lazily inside, but on a transient AuthenticatedUser() error each repo iteration will retry and emit one warn per repo via issueTrackingWithAssigneeScope. During a GitHub outage on multi-repo deployments this is noisy. Consider short-circuiting the whole promote cycle once authUser resolution fails, with a single warn at the call site.
🟡 **daemon/internal/config/config.go:254** — MatchesAssignees lower-cases/trims c.Assignees on every call. Now invoked from hot paths (issueStageStillCurrent, per-issue promote loop, per-issue filter in fetch_issues.go). Normalizing once at config-load time would be cheaper for large assignee lists / high issue volume. Not a correctness issue.
🟡 **daemon/cmd/heimdallm/main.go:1268** — The /status payload now reports issueTracking with the auth-user-defaulted assignees rather than the raw config, so operators will see a synthetic `assignees: [<bot-login>]` they did not configure. Worth calling out in the PR description / CHANGELOG so dashboards / runbooks aren't surprised.

### Suggestions

- Add a startup-time invariant or test: if issue_tracking.enabled and no explicit assignees, resolvedBotLogin must be non-empty before workers start consuming messages — otherwise the first batch is silently dropped.
- Consider precomputing a normalized assignee set on IssueTrackingConfig (e.g., during applyDefaults) to avoid repeating ToLower/TrimSpace per issue in MatchesAssignees.
- Note in the PR description that /status now reports the effective (defaulted) assignees, since this is a visible behavior change for operators.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Re-review of PR #429. Most previous LOW findings remain unaddressed but are minor. One new medium-severity concern: the PromoteReady method calls resolveAuthenticatedUser() (which may make a GitHub API call) while holding cfgMu, blocking all config readers for the duration of a network round-trip.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2855** — In PromoteReady, resolveAuthenticatedUser() is called inside the cfgMu critical section (locked at ~line 2851). resolveAuthenticatedUser may call ghClient.AuthenticatedUser(), a network API call. Holding cfgMu during a network round-trip blocks all other goroutines that need config access. Move the lazy resolution before cfgMu.Lock() or release cfgMu before resolving and re-acquire after.
🟡 **daemon/internal/config/config.go:233** — Previous finding still unaddressed: TrimLeft(login, "@") is a character-set strip, not a prefix strip. strings.TrimPrefix(login, "@") is semantically correct and consistent with typical usage.
🟡 **daemon/cmd/heimdallm/main.go:3468** — Previous finding still unaddressed: MatchesAssignees returns false for both 'assigned to someone else' and 'unassigned' cases, but the log message only says 'issue assigned outside this daemon scope'. An unassigned issue isn't assigned outside scope — it's unassigned. Consider distinguishing for operational clarity.
🟡 **daemon/cmd/heimdallm/main.go:2296** — In resolveAuthenticatedUser, the 'if a == nil' check on the second guard is unreachable dead code — cachedAuthenticatedUser() already handles a==nil and returns empty string, which causes the first guard to fall through. The second nil check can never trigger since we already dereferenced a.cachedAuthenticatedUser() successfully. Minor dead code.

### Suggestions

- In PromoteReady, resolve authUser lazily but outside the cfgMu lock: collect repos needing resolution in a first pass, release cfgMu, resolve the user, re-acquire cfgMu, then proceed with the grouping loop.
- Extract a normalizeLogin helper (TrimPrefix + ToLower) shared by WithDefaultAssignee and MatchesAssignees to ensure consistent normalization.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Scopes the staged issue pipeline (triage/refinement/implement, tier2 process+promote, fetcher, promoter) by assignee, defaulting the empty-assignees case to the authenticated bot login so daemons no longer treat shared queues as 'anyone'. Initialization of cachedLogin from resolvedBotLogin closes the startup race that previously left the scope empty. MatchesAssignees/WithDefaultAssignee are correctly implemented (case-insensitive, @ trimming, value-receiver immutability) with adequate unit coverage; semantics in fetch_issues.issueMatchesFilters are preserved. The stale-job guard now also rejects out-of-scope assignees, and the early-return in issueTrackingWithAssigneeScope prevents accidental wide-open processing. No confirmed bugs in the diff.

### Suggestions

- tier2Adapter.cachedAuthenticatedUser dereferences *a.login without a lock when a.loginMu is nil; in production loginMu is always set, but consider documenting that invariant or removing the nil-mutex branch to avoid a latent data race if a future caller wires login without the mutex.
- Status export now serializes the post-default issue_tracking (with the bot login injected into Assignees). Worth a brief note in the UI/status consumers, since operators will see an assignee they didn't configure.
- Minor: the test file fetch_issues_test.go reformatting only changes alignment whitespace on the issueFixture struct and one inline comment — drop those unrelated cosmetic edits to keep the diff focused.
- Consider an explicit opt-in (e.g. assignees = ["*"] or a shared_queue=true flag) for operators who legitimately want a shared queue, since the new default makes that configuration impossible to express today.

---

PR #432, opened 2026-05-08
fix(issues): enforce scoped auto-implementation changes

### LINKED ISSUE #430: fix(issues): constrain staged issue ownership and real implementation PRs
## Problem

The staged issue workflow must only be processed by the Heimdallm instance that owns the issue, and auto-implementation must only open PRs with real implementation changes.

Recent runs exposed three related failures:

- An issue can end up with multiple assignees, so more than one Heimdallm instance can consider it in scope.
- Auto-implementation opened a PR that only committed `.heimdallm-managed`, which is operational metadata, not product code.
- The generated PR was not explicitly assigned to the current Heimdallm operator.

## Desired behavior

- Only process staged issues that have one of the configured Heimdallm stage labels and exactly one assignee matching this daemon scope.
- Triage may choose or hand off an owner, but after the handoff the issue must have exactly one assignee.
- Refinement and development must skip issues with zero assignees, multiple assignees, or an assignee outside the configured scope.
- Auto-implementation must ignore Heimdallm operational metadata files when detecting/staging changes.
- If the agent produces no real code changes, Heimdallm must not create a PR and must leave an explicit log/activity/comment explaining that no implementation changes were produced.
- Auto-created implementation PRs must be assigned to the current operator, unless an explicit `pr_assignee` override is configured.

## Acceptance criteria

- A staged issue assigned to exactly `@ivanmunozruiz` and carrying the configured stage label is eligible for processing by this daemon.
- A staged issue assigned to someone else, unassigned, or assigned to multiple users is skipped with visible diagnostics.
- A triage handoff from `@alice` to `@bob` results in assignees exactly `[bob]`, not `[alice, bob]`.
- `.heimdallm-managed` never makes `HasChanges` true and is never committed to implementation PRs.
- Marker-only or no-op agent output does not open a PR.
- A real implementation PR is assigned to the current user by default.
- Tests cover assignee scope, single-assignee handoff, marker exclusion, no-op implementation behavior, and PR assignment.

## Notes

This supersedes the narrower stale-worktree test issue #422 and the invalid auto-generated PR #431.

ivanmunozruiz:
## Summary
- Require scoped staged issues to have exactly one matching assignee before triage/refinement/development processing.
- Replace issue assignees on triage handoff, exclude .heimdallm-managed from auto-implement change detection/commits, and assign generated PRs to the current operator by default.
- Strengthen auto-implement so refinement plans are passed as prioritized implementation context while still allowing repository investigation when refinement omits files/symbols.
- Record execute/push failures as auto_implement_failed and avoid marker-only/no-op PRs.

## Tests
- make test-docker

Closes #430

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -897,7 +897,7 @@ func main() {
 			ImplementPromptOverride: implPrompt,
 			ImplementInstructions:   implInstructions,
 			PRReviewers:             aiCfg.PRReviewers,
-			PRAssignee:              aiCfg.PRAssignee,
+			PRAssignee:              defaultAutoImplementPRAssignee(aiCfg.PRAssignee, authUser),
 			PRLabels:                aiCfg.PRLabels,
 			PRDraft:                 aiCfg.PRDraft != nil && *aiCfg.PRDraft,
 			GeneratePRDescription:   aiCfg.GeneratePRDescription != nil && *aiCfg.GeneratePRDescription,
@@ -1092,7 +1092,7 @@ func main() {
 			ImplementPromptOverride:  implPrompt,
 			ImplementInstructions:    implInstructions,
 			PRReviewers:              aiCfg.PRReviewers,
-			PRAssignee:               aiCfg.PRAssignee,
+			PRAssignee:               defaultAutoImplementPRAssignee(aiCfg.PRAssignee, authUser),
 			PRLabels:                 aiCfg.PRLabels,
 			PRDraft:                  aiCfg.PRDraft != nil && *aiCfg.PRDraft,
 			GeneratePRDescription:    aiCfg.GeneratePRDescription != nil && *aiCfg.GeneratePRDescription,
@@ -2824,7 +2824,7 @@ func (a *tier2Adapter) ProcessRepo(ctx context.Context, repo string) (int, error
 			ImplementPromptOverride:     implPrompt,
 			ImplementInstructions:       implInstructions,
 			PRReviewers:                 aiCfg.PRReviewers,
-			PRAssignee:                  aiCfg.PRAssignee,
+			PRAssignee:                  defaultAutoImplementPRAssignee(aiCfg.PRAssignee, authUser),
 			PRLabels:                    aiCfg.PRLabels,
 			PRDraft:                     aiCfg.PRDraft != nil && *aiCfg.PRDraft,
 			GeneratePRDescription:       aiCfg.GeneratePRDescription != nil && *aiCfg.GeneratePRDescription,
@@ -3495,6 +3495,13 @@ func issueTrackingWithAssigneeScope(scope, repo string, it config.IssueTrackingC
 		return it, false
 	}
 	return it, true
+}
+
+func defaultAutoImplementPRAssignee(configured, authUser string) string {
+	if assignee := strings.TrimSpace(configured); assignee != "" {
+		return assignee
+	}
+	return strings.TrimSpace(strings.TrimLeft(authUser, "@"))
 }
 
 func autoPromoteAfterStage(
--- a/daemon/cmd/heimdallm/main_test.go
+++ b/daemon/cmd/heimdallm/main_test.go
@@ -45,6 +45,15 @@ func TestAcquireRepoContextNilManagerIsError(t *testing.T) {
 	}
 }
 
+func TestDefaultAutoImplementPRAssignee(t *testing.T) {
+	if got := defaultAutoImplementPRAssignee("", "@ivanmunozruiz"); got != "ivanmunozruiz" {
+		t.Fatalf("default assignee = %q, want ivanmunozruiz", got)
+	}
+	if got := defaultAutoImplementPRAssignee("configured", "ivanmunozruiz"); got != "configured" {
+		t.Fatalf("configured assignee = %q, want configured", got)
+	}
+}
+
 func TestAutoPromoteTriageSmartDefaultRequiresRefinementLabel(t *testing.T) {
 	aiCfg := config.RepoAI{}
 	it := config.IssueTrackingConfig{}
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -239,12 +239,14 @@ func (c IssueTrackingConfig) WithDefaultAssignee(login string) IssueTrackingConf
 
 // MatchesAssignees reports whether the current assignee filter permits an
 // issue assigned to the provided GitHub logins. An inactive filter permits all;
-// an active filter never matches an unassigned issue.
+// an active filter only matches issues with exactly one assignee in scope.
+// That single-owner invariant prevents two Heimdallm instances from processing
+// the same staged issue when GitHub temporarily shows multiple assignees.
 func (c IssueTrackingConfig) MatchesAssignees(assignees []string) bool {
 	if len(c.Assignees) == 0 {
 		return true
 	}
-	if len(assignees) == 0 {
+	if len(assignees) != 1 {
 		return false
 	}
 	want := make(map[string]struct{}, len(c.Assignees))
@@ -257,13 +259,9 @@ func (c IssueTrackingConfig) MatchesAssignees(assignees []string) bool {
 	if len(want) == 0 {
 		return false
 	}
-	for _, a := range assignees {
-		a = strings.ToLower(strings.TrimSpace(strings.TrimLeft(a, "@")))
-		if _, ok := want[a]; ok {
-			return true
-		}
-	}
-	return false
+	a := strings.ToLower(strings.TrimSpace(strings.TrimLeft(assignees[0], "@")))
+	_, ok := want[a]
+	return ok
 }
 
 // Classify returns the processing mode for an issue given its labels.
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -316,6 +316,13 @@ func labelSetIntersects(set map[string]struct{}, list []string) bool {
 // JSON tags must match the snake_case keys written by the PUT /config handler
 // (handlers.go:normalizeAgentConfigsForPut) so ApplyStore can unmarshal a
 // stored row back into this struct symmetrically.
+//
+// Boolean fields deliberately do NOT use `omitempty`. With omitempty a future
+// caller that marshals this struct directly (instead of the current map[string]any
+// path in the handler) would silently drop a `false` value, and ApplyStore's
+// "unmarshal into the existing struct" merge would then preserve a TOML `true`
+// the operator was trying to override. Keeping the zero value in the JSON
+// guarantees the override semantic regardless of how the JSON is produced.
 type CLIAgentConfig struct {
 	Model        string `toml:"model" json:"model,omitempty"`                 // e.g. "claude-opus-4-6"
 	MaxTurns     int    `toml:"max_turns" json:"max_turns,omitempty"`         // claude: --max-turns (0 = not set)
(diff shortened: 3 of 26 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

Targeted fix that hardens auto-implement scope: PR assignee now defaults to authUser/single issue assignee, assignee filter becomes a hard single-owner gate, SetAssignees switches from additive POST to replacing PATCH, and a managed-clone marker is excluded from auto_implement diffs. The intent is sound and well-tested, but several semantic changes are breaking and a couple of small correctness/robustness concerns deserve attention.

### Issues

⚠️ **daemon/internal/config/config.go:248** — MatchesAssignees now requires exactly one assignee. This is a breaking semantic change: an issue assigned to {alice, bob} no longer matches a filter for {alice}. The comment justifies this for the staged auto_implement path (avoid double-processing), but MatchesAssignees is also used by the generic fetch/filter pipeline, so triage/refinement of multi-assignee issues will silently stop working too. Consider scoping the single-owner invariant to the auto_implement decision instead of the shared filter, or document the operator-visible behavior change in release notes.
⚠️ **daemon/internal/github/repos.go:253** — SetAssignees changed from POST /issues/{n}/assignees (additive) to PATCH /issues/{n} (replace). The function name always implied 'set', but the previous implementation appended. Any caller that relied on additive behavior (e.g. adding the daemon to an already-assigned human) will now wipe existing assignees. Worth grepping all call sites and confirming they expect replace semantics; otherwise add an explicit PR note for operators.
🟡 **daemon/internal/issues/prompt.go:327** — The refinement guidance block is appended whenever ctx.TriageContext is non-empty, but TriageContext can be a triage summary rather than a refinement plan. The phrase 'If a previous refinement plan is present above' becomes misleading and may push the agent to invent a 'contract' that does not exist. Gate this block on actually having a refinement plan, not just any triage context.
🟡 **daemon/internal/issues/gitops.go:111** — The pathspec `:(exclude).heimdallm-managed` only excludes the file at repository root. If the marker ever lands in a subdirectory (or a future variant uses a path), it will be staged/committed. Using `:(exclude,top)**/.heimdallm-managed` or the equivalent would be more robust and matches the documented intent ('repoctx-managed clones').
🟡 **daemon/internal/worker/implement_test.go:65** — time.Sleep(100*time.Millisecond) before publishing is a startup race; on a loaded CI runner this can flake even with the 5s outer deadline. Prefer signalling readiness from the worker (or polling NATS subscription state) rather than sleeping.
🟡 **daemon/internal/issues/pipeline.go:797** — metadataOpts := opts is a shallow copy; PRReviewers/PRLabels slices are still shared with the caller. Not a bug today (applyPRMetadata appears read-only), but the pattern is fragile — a future mutation in applyPRMetadata would silently leak back into opts. Either deep-copy the slices or pass the resolved assignee as a separate argument.
🟡 **daemon/cmd/heimdallm/main.go:3500** — defaultAutoImplementPRAssignee only strips a single leading '@' via TrimLeft, which would also strip multiple '@@@' if present. More importantly, authUser is used unvalidated — if it ever contains whitespace or a display name (e.g. 'Ivan Munoz'), the resulting assignee call will fail at GitHub. A small TrimSpace + validation log on the unexpected-format path would harden this.

### Suggestions

- Add a regression test for MatchesAssignees with zero assignees to lock in the new behavior alongside the multi-assignee case.
- Document the SetAssignees POST→PATCH change and the single-owner filter invariant in CHANGELOG/release notes — both are operator-visible.
- Consider extracting the single-owner gate into a clearly named helper (e.g. IsSingleOwnerInScope) used only by the auto_implement decision, so the generic IssueTrackingConfig.MatchesAssignees can remain permissive for triage/refinement.
- For the refinement prompt addition, assert in tests that the new contract language is NOT emitted when only a triage summary (no refinement) is present.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: most prior medium/low findings remain unaddressed in the visible diff. defaultAutoImplementPRAssignee now wraps with TrimSpace (improvement), but the breaking semantic change to MatchesAssignees, the POST→PATCH replace semantics of SetAssignees, the unconditional refinement-contract prompt block, and the non-recursive marker-exclude pathspec are all still present.

### Issues

⚠️ **daemon/internal/config/config.go:248** — MatchesAssignees still rejects multi-assignee issues globally (`len(assignees) != 1`). The doc comment frames this as an auto_implement single-owner invariant, but the function is shared with triage/refinement filtering — issues assigned to {alice,bob} now silently stop matching {alice} for non-staged paths too. Scope the invariant to the auto_implement decision (e.g. an explicit MatchesSingleOwner method) or call it out in release notes.
⚠️ **daemon/internal/github/repos.go:253** — SetAssignees still uses PATCH /issues/{n} (replace) instead of POST /issues/{n}/assignees (additive). The renamed doc comment ('replaces assignees') confirms the intent, but no audit of call sites is shown — any caller that previously relied on additive behavior will silently wipe other assignees. Document the change for operators or grep all call sites and confirm replace semantics is what each one wants.
🟡 **daemon/internal/issues/prompt.go:327** — Refinement-contract block is still appended whenever `ctx.TriageContext != ""`, but TriageContext can be a plain triage summary with no refinement plan. The phrase 'If a previous refinement plan is present above, treat it as the implementation contract' may push the agent to fabricate a contract. Gate this block on a real refinement-plan signal, not any non-empty triage context.
🟡 **daemon/internal/issues/gitops.go:111** — Pathspec `:(exclude).heimdallm-managed` only excludes the marker at repo root. Use `:(exclude,top)**/.heimdallm-managed` (or equivalent) so the exclusion is robust if the marker ever appears in a subdirectory.
🟡 **daemon/internal/issues/pipeline.go:797** — `metadataOpts := opts` is a shallow copy; PRReviewers/PRLabels slices remain shared with the caller. Not a live bug, but a future mutation in applyPRMetadata would silently leak back into opts. Pass the resolved assignee as a separate parameter, or deep-copy the slices.
🟡 **daemon/cmd/heimdallm/main.go:3500** — defaultAutoImplementPRAssignee now wraps with TrimSpace (good), but `strings.TrimLeft(authUser, "@")` still strips any number of leading '@' chars and there is no validation/log when authUser is malformed (whitespace inside, display name, etc.). Consider TrimPrefix for a single '@' and a debug log on unexpected formats so a misconfigured token doesn't silently produce a broken assignee call.

### Suggestions

- Either narrow MatchesAssignees back to ANY-match and introduce a dedicated single-owner check used only by the auto_implement gate, or add release notes calling out that multi-assignee issues are no longer matched by assignee filters.
- Add a brief PR description note: 'SetAssignees now replaces assignees (PATCH) instead of appending (POST). Verified all current call sites expect replace semantics.' so operators are not surprised.
- Detect the presence of a refinement plan explicitly (e.g. a flag on PromptContext) and gate the new contract block on that, not on any TriageContext.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of PR #432: most previously flagged issues remain unaddressed in the current diff. The MEDIUM findings on MatchesAssignees scope leakage and SetAssignees PATCH semantics are still present. LOW findings on the pathspec exclude pattern and authUser sanitization are also unchanged.

### Issues

⚠️ **daemon/internal/config/config.go:248** — MatchesAssignees still uses `len(assignees) != 1`, applying the single-owner invariant globally. The function is shared with triage/refinement filtering, not just auto_implement, so multi-assigned issues silently stop matching for all paths. Scope the invariant (e.g. a separate MatchesSingleOwner) or document the behavior change.
⚠️ **daemon/internal/github/repos.go:253** — SetAssignees still switched to PATCH /issues/{n} (replace) from POST /issues/{n}/assignees (additive). No audit of call sites visible in the diff — any caller relying on the previous additive semantics will now wipe pre-existing assignees. Verify each call site or call out the semantic change.
🟡 **daemon/internal/issues/gitops.go:111** — Pathspec `:(exclude)".heimdallm-managed"` only excludes the marker at repo root. Use `:(exclude,top)**/.heimdallm-managed` (or equivalent) so the exclusion is robust if the marker ever appears in a subdirectory.
🟡 **daemon/cmd/heimdallm/main.go:3500** — defaultAutoImplementPRAssignee uses `strings.TrimLeft(authUser, "@")`, which strips any number of leading '@' chars. Consider TrimPrefix and a debug log on unexpected formats so a misconfigured token doesn't silently produce a broken assignee call.

### Suggestions

- Add a release note / migration warning for the SetAssignees replace semantics so operators reconfigure callers that depended on additive behavior.
- Split MatchesAssignees into MatchesAssigneeFilter (any-of, original semantics) and MatchesSoleOwner (single-owner invariant) and call the right one at each site to avoid leaking the auto_implement invariant into triage/refinement paths.
- Make the managed-clone marker exclusion path-agnostic via `:(exclude,top)**/.heimdallm-managed` and add a unit test covering a marker placed in a subdirectory.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of #432: prior MEDIUM/LOW findings remain unaddressed in the latest diff (MatchesAssignees scope, SetAssignees PATCH semantics, gitops pathspec scope, TrimLeft sloppiness). New ApplyStore agent_configs branch and MaxAutoImplementNoChanges cap look reasonable; one new concern with omitempty on boolean fields breaking explicit-override semantics.

### Issues

⚠️ **daemon/internal/config/config.go:248** — MatchesAssignees still enforces len(assignees) != 1 globally. The doc comment now acknowledges the single-owner invariant, but the function is still shared by triage/refinement/auto_implement filtering — multi-assignee issues silently drop out of all of them. Either split into MatchesSingleOwner used only by the auto_implement path, or confirm in the PR description that the behavior change is intentional for every caller. Unchanged from prior review.
⚠️ **daemon/internal/github/repos.go:253** — SetAssignees still uses PATCH /repos/{owner}/{repo}/issues/{n} (replace semantics) instead of the prior POST /assignees (additive). The diff still shows no audit of call sites. Any caller that previously relied on additive behavior now wipes pre-existing assignees on the issue. Verify every call site (or document the semantic change in the PR body).
⚠️ **daemon/internal/config/config.go:322** — New JSON tags use `omitempty` on boolean fields (Bare, DangerouslySkipPerms, NoSessionPersistence) and ints. If the PUT handler also uses omitempty when writing to the store, an operator who explicitly sets bare=false in the UI cannot override a TOML bare=true — the field is omitted from JSON and the existing struct's true survives the merge in store.go. Confirm the handler serializes booleans unconditionally, or use pointer fields for tri-state semantics.
🟡 **daemon/internal/issues/gitops.go:111** — Pathspec `:(exclude)<file>` is still anchored at repo root only. If the marker is ever written into a subdirectory (or a future repoctx change moves it), HasChanges/CommitAll silently regress. Use `:(exclude,top)**/.heimdallm-managed` or anchor explicitly. Unchanged from prior review.
🟡 **daemon/cmd/heimdallm/main.go:3503** — defaultAutoImplementPRAssignee still uses strings.TrimLeft(authUser, "@") which strips any number of leading '@' chars. Prefer strings.TrimPrefix and a debug log when the input shape is unexpected, so misconfigured tokens surface rather than producing a malformed assignee. Unchanged from prior review.

### Suggestions

- Add a unit test for CLIAgentConfig JSON round-trip that explicitly sets a boolean to false in the store payload and asserts it overrides a TOML true — that will catch the omitempty issue if it bites.
- Audit SetAssignees callers in this PR's branch (grep for SetAssignees) and either preserve additive semantics or note the migration in the commit message.
- Consider renaming MatchesAssignees to MatchesSingleOwnerAssignee or adding a separate predicate for the auto_implement path to avoid surprising triage/refinement consumers.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: the JSON-tag omitempty issue on booleans is fixed (tags now omit `omitempty` on bool fields, with a comment and a regression test covering false-overrides-true). The other four findings are unchanged in the code — three of them now have explanatory comments/docs but the underlying behavior is the same.

### Issues

⚠️ **daemon/internal/config/config.go:249** — MatchesAssignees still rejects multi-assignee issues globally (len(assignees) != 1). The doc comment now justifies the single-owner invariant for the auto_implement race, but this method is still the sole filter used by triage/refinement paths via issueMatchesFilters; multi-assignee issues silently drop from those flows too. Either split into a separate MatchesSingleOwner used only on the auto_implement path, or call out in the PR body that triage/refinement intentionally now ignore multi-assignee issues.
⚠️ **daemon/internal/github/repos.go:253** — SetAssignees still switched from POST /assignees (additive) to PATCH /issues/{n} (replace). The doc comment was updated from 'sets' to 'replaces', but there's no evidence of a call-site audit — any caller that relied on additive semantics will now wipe pre-existing assignees. Please confirm every caller (grep for SetAssignees) and either pass the union of existing+new, or document the semantic change in the PR description.
🟡 **daemon/internal/issues/gitops.go:111** — Pathspec `:(exclude)`+managedCloneMarkerFile is still anchored at the repo root. The new test only places the marker at the root, so a future repoctx change that nests the marker (e.g. .heimdallm/managed) would silently regress HasChanges/CommitAll. Use `:(exclude,top)**/.heimdallm-managed` or otherwise match by basename to make the exclusion location-independent.
🟡 **daemon/cmd/heimdallm/main.go:3503** — defaultAutoImplementPRAssignee still uses strings.TrimLeft(authUser, "@") which greedily strips any number of leading '@' characters, so a malformed token like "@@bot" silently produces "bot". Prefer strings.TrimPrefix(authUser, "@") so unexpected shapes surface rather than being normalized away.

### Suggestions

- Add a call-site audit (or a code comment listing them) for SetAssignees so reviewers can verify the POST→PATCH switch was intentional everywhere.
- If the multi-assignee filter change is intentional for triage/refinement, document it in the PR body and in MatchesAssignees' godoc beyond the single-owner-race rationale.
- Consider a small test that places .heimdallm-managed inside a subdirectory to lock in the intended pathspec semantics.

---

PR #435, opened 2026-05-08
fix(github): paginate FetchCollaborators via Link header

vbuenog:
## Summary

Some GitHub users (e.g. `vbuenog`, `dieloren`) were missing from the **Issue Triage assignee combobox** in the Flutter UI on every repository whose collaborator list exceeds 100 entries. This PR fixes the underlying daemon bug so the combobox shows the complete list of collaborators GitHub returns for the repo.

## Root cause

The combobox is populated by merging two sources: collaborators returned by `GET /api/repos/{repo}/collaborators` (served by the daemon) and logins already saved in user config (`knownGitHubUsers`). The daemon's `Client.FetchCollaborators` (`daemon/internal/github/client.go`) called `GET /repos/{owner}/{repo}/collaborators?per_page=100` exactly once and returned whatever fit on page 1 — there was **no pagination**.

Verified with the `gh` CLI against `freepik-company/ai-platform-iqs-v3`:

| Check | Result |
|---|---|
| Total collaborators returned with full pagination | **253** |
| Collaborators returned by the daemon today (page 1 only) | **100** |
| `vbuenog` / `dieloren` on page 1 | ❌ neither |
| `vbuenog` / `dieloren` past page 1 | ✅ both present |
| Both are confirmed `freepik-company` org members (`HTTP 204` from `/orgs/freepik-company/members/{login}`) | ✅ |

So 153 of 253 collaborators were silently dropped on this repo, including these two users.

## Change

- **`daemon/internal/github/client.go`** — `FetchCollaborators` now follows GitHub's `Link: rel=\"next\"` header until exhausted, accumulating logins from every page. A new unexported helper `parseNextLink(string) string` extracts the `next` URL from the `Link` header.
- **Pagination strategy** — uses GitHub's documented `Link: rel=\"next\"` mechanism (not page-count math), which is robust to filtered responses where a page may be shorter than `per_page` even though more data follows.
- **Safety bound** — capped at 100 pages (~10,000 collaborators). If the cap is reached, the function returns a loud error mentioning \"pagination\" rather than silently truncating (which is precisely the failure mode this PR fixes). Malformed `Link` headers also fall through to a clean stop instead of looping forever.
- **`c.do` contract preserved** — the absolute `next` URL from GitHub is parsed with `net/url` and `RequestURI()` (path + query) is passed to `c.do`, so `c.baseURL` remains the source of truth for both real GitHub and `httptest` servers.

## Out of scope

- No caching of the collaborators list (today's behavior fetches on demand; unchanged).
- No switch to `/orgs/{org}/members` as the data source (a possible alternative discussed during design — kept the existing repo-collaborators semantics).
- No audit of other GitHub list endpoints in the daemon. Other endpoints likely have the same bug, but the change is intentionally focused on the one observed.
- No UI changes in `flutter_app` — the combobox already merges and dedupes the list it receives.

## Tests

New black-box tests in `daemon/internal/github/collaborators_test.go` (matching the existing `package github_test` convention; uses `httptest.NewServer`):

- `TestFetchCollaborators_MultiPage` — server returns 3 pages with `Link: rel=\"next\"` on pages 1 & 2; asserts all logins from all pages are concatenated in order.
- `TestFetchCollaborators_SinglePage` — no `Link` header → exactly one HTTP call, single-page logins returned.
- `TestFetchCollaborators_RunawayCap` — server always returns `Link: rel=\"next\"` pointing back to itself → returns an error mentioning \"pagination\" and terminates well under the test timeout.
- `TestFetchCollaborators_MidPaginationError` — page 1 returns 200 with `next`, page 2 returns 500 → returns an error and a `nil` slice (no silent partial success).
- `TestFetchCollaborators_LinkHeaderVariants` — exercises `parseNextLink` indirectly through `FetchCollaborators` for: `only_last_no_next`, `empty_header`, `malformed_no_brackets`, `malformed_no_rel`, and a `next_and_last` sanity case where both `rel` values are present and `next` is followed correctly.

## Test plan

- [ ] CI green (`go test ./daemon/...` and `go vet ./daemon/...`).
- [ ] After merge, open the Issue Triage section in the Flutter UI for a Freepik repo with >100 collaborators and confirm `vbuenog`, `dieloren`, and other previously-missing org members now appear in the assignee combobox.

## Docs

- Spec: `docs/superpowers/specs/2026-05-08-collaborators-pagination-design.md`
- Plan: `docs/superpowers/plans/2026-05-08-collaborators-pagination.md`

### DIFF
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -943,29 +943,71 @@ func (c *Client) RemoveIssueLabel(repo string, number int, label string) error {
 	return nil
 }
 
-// FetchCollaborators returns the login names of repository collaborators.
+// parseNextLink extracts the URL whose rel parameter is "next" from a GitHub
+// Link header. Returns "" when no such URL exists. The header format is:
+//
+//	<https://api.github.com/...&page=2>; rel="next", <...>; rel="last"
+//
+// We do not pull in a parser dependency; a small string scan is sufficient.
+func parseNextLink(header string) string {
+	for _, part := range strings.Split(header, ",") {
+		segs := strings.Split(strings.TrimSpace(part), ";")
+		if len(segs) < 2 {
+			continue
+		}
+		urlPart := strings.TrimSpace(segs[0])
+		if !strings.HasPrefix(urlPart, "<") || !strings.HasSuffix(urlPart, ">") {
+			continue
+		}
+		linkURL := urlPart[1 : len(urlPart)-1]
+		for _, s := range segs[1:] {
+			s = strings.TrimSpace(s)
+			if s == `rel="next"` || s == "rel=next" {
+				return linkURL
+			}
+		}
+	}
+	return ""
+}
+
+// FetchCollaborators returns the login names of repository collaborators,
+// following GitHub's Link: rel="next" header to walk every page.
 func (c *Client) FetchCollaborators(repo string) ([]string, error) {
 	if repo == "" {
 		return nil, nil
 	}
-	resp, err := c.do("GET", fmt.Sprintf("/repos/%s/collaborators?per_page=100", repo), "application/vnd.github+json")
-	if err != nil {
-		return nil, fmt.Errorf("github: fetch collaborators: %w", err)
+	const maxPages = 100
+	path := fmt.Sprintf("/repos/%s/collaborators?per_page=100", repo)
+	var logins []string
+	for page := 0; page < maxPages; page++ {
+		resp, err := c.do("GET", path, "application/vnd.github+json")
+		if err != nil {
+			return nil, fmt.Errorf("github: fetch collaborators: %w", err)
+		}
+		body, _ := io.ReadAll(io.LimitReader(resp.Body, maxBodyBytes))
+		linkHeader := resp.Header.Get("Link")
+		resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("github: fetch collaborators %s: status %d", repo, resp.StatusCode)
+		}
+		var raw []struct {
+			Login string `json:"login"`
+		}
+		if err := json.Unmarshal(body, &raw); err != nil {
+			return nil, fmt.Errorf("github: decode collaborators: %w", err)
+		}
+		for _, u := range raw {
+			logins = append(logins, u.Login)
+		}
+		next := parseNextLink(linkHeader)
+		if next == "" {
+			return logins, nil
+		}
+		nextURL, err := url.Parse(next)
+		if err != nil {
+			return nil, fmt.Errorf("github: parse next link %q: %w", next, err)
+		}
+		path = nextURL.RequestURI()
 	}
-	body, _ := io.ReadAll(io.LimitReader(resp.Body, maxBodyBytes))
-	resp.Body.Close()
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("github: fetch collaborators %s: status %d", repo, resp.StatusCode)
-	}
-	var raw []struct {
-		Login string `json:"login"`
-	}
-	if err := json.Unmarshal(body, &raw); err != nil {
-		return nil, fmt.Errorf("github: decode collaborators: %w", err)
-	}
-	logins := make([]string, len(raw))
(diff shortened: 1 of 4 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

PR correctly fixes pagination bug in FetchCollaborators by following GitHub's Link rel="next" header. Implementation is clean, uses stdlib only, includes a 100-page safety cap, and has thorough test coverage (multi-page, single-page, runaway cap, mid-pagination errors, header variants). Code follows existing patterns in the file.

### Suggestions

- parseNextLink only matches exact strings `rel="next"` or `rel=next`; RFC 8288 allows additional whitespace and parameters (e.g. `rel="next" title="..."`). GitHub's current format is matched, but a slightly more permissive check (e.g. trimming and splitting on `=`) would be more robust to future format variations.
- Consider logging a warning when pagination cap is hit so operators can detect the condition in production rather than only seeing the returned error.
- The 100-page cap (10,000 collaborators) is reasonable but undocumented as a configurable limit; if very large enterprise repos exist, consider making it configurable via Client options.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid, well-tested pagination implementation for FetchCollaborators. The approach is correct — follows Link rel="next" headers, caps at 100 pages, and handles error/edge cases. A few minor issues worth addressing.

### Issues

🟡 **daemon/internal/github/client.go:972** — SSRF risk: parseNextLink blindly trusts the URL from the Link header, and the result is passed to c.do() which will make an HTTP request to baseURL + path. A malicious or compromised GitHub response could include a Link header pointing to an internal service (e.g., http://169.254.169.254/...). Since url.Parse().RequestURI() only extracts the path+query, and c.do() prepends c.baseURL, this is mitigated in practice — but only if baseURL always points to GitHub. Consider validating that the parsed next URL's host matches the expected GitHub API host, or at minimum document why this is safe.
🟡 **daemon/internal/github/client.go:968** — Body is not closed on non-200 error paths. When resp.StatusCode != http.StatusOK on page > 0, the function returns an error but resp.Body was already read and closed (line 963-964), so this is actually fine. However, if c.do() itself returns an error, the resp may or may not be nil depending on the http client — the current code handles this the same way as the rest of the file, so this is consistent.
🟡 **daemon/internal/github/client.go:955** — Inconsistency with existing pagination patterns. Other paginated methods in this file (fetchReposForOrg, GetPRTimelineEventsForReviewer) use page-number query parameters, not Link header parsing. While Link-header pagination is arguably more correct for the collaborators endpoint, this creates two different pagination strategies in the same file. Not a bug, but worth a comment or eventual unification.
⚠️ **docs/superpowers/plans/2026-05-08-collaborators-pagination.md:1** — A 657-line implementation plan document is included in the PR. This appears to be an agentic workflow artifact that should not be committed to the repo — it duplicates information already in the code and PR description, and will become stale immediately.
🟡 **daemon/internal/github/collaborators_test.go:109** — The RunawayCap test makes 100 real HTTP round-trips to the httptest server. While fast in practice, the maxPages constant (100) is hardcoded in the production code and the test implicitly depends on that value. If maxPages changes, this test silently becomes wrong. Consider either exporting the cap as a package-level constant or adding a comment noting the coupling.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Adds Link-header pagination to FetchCollaborators with a 100-page safety cap, plus comprehensive black-box tests covering multi-page, single-page, runaway, mid-pagination error, and Link header parsing variants. Implementation is solid and the test coverage is thorough.

### Suggestions

- daemon/internal/github/client.go:970 — minor: the initial fetch error `"github: fetch collaborators: %w"` doesn't include the repo name unlike the status/pagination errors below it; trivial consistency nit.
- daemon/internal/github/client.go:968 — minor: loop counter named `page` is slightly confusing given GitHub's own `page=N` query param; something like `i` or `pageCount` would read clearer. Optional.
- daemon/internal/github/client.go:946 — parseNextLink does strict equality on `rel="next"` / `rel=next`; this matches GitHub's actual output but won't tolerate extra whitespace inside the quotes or unusual quoting. Acceptable for GitHub's stable format.

---

Issue #439, opened 2026-05-11
chore(flutter): bump in-constraint deps + flutter_lints 3 → 6


ivanmunozruiz:
## Context

`flutter run` was reporting *25 packages have newer versions incompatible with dependency constraints* on every launch. Splitting these into tiers based on risk:

- **Tier 1 — in-constraint patch/minor bumps** (no `pubspec.yaml` edit needed): build/codegen tooling (`build_runner`, `json_serializable`, `built_value`, `source_gen`, etc.) plus a few runtime transitives.
- **Tier 2 — `flutter_lints` 3 → 6** (dev-dependency only, no runtime impact, just new lint info-level rules to clean up).
- **Tier 3 — runtime majors** deliberately left out: `file_picker` 8 → 11, `go_router` 14 → 17, `flutter_riverpod` 2 → 3. Each deserves its own PR + smoke test to keep regressions bisectable.

## What this issue tracks

Tiers 1 + 2 only, bundled as one PR. Goal: shrink the `flutter pub outdated` list, refresh dev tooling, and pick up the new Dart lints, all without touching runtime behavior.

## Acceptance

- [ ] `flutter analyze` reports zero issues
- [ ] `flutter test` passes (167/167)
- [ ] Manual smoke test on macOS covers Repositories / Agents / PRs tabs
- [ ] CI green

PR #446, opened 2026-05-11
chore(flutter): migrate to flutter_riverpod 3

### LINKED ISSUE #445: chore(flutter): migrate to flutter_riverpod 3
## Context

Last Tier 3 dependency bump in the cleanup series. Follow-up to #443 / #441 / #439.

Riverpod 3 is a substantial API overhaul that touches ~80 sites across 26 files. Unlike `go_router` and `file_picker` (which were quick), this one is a real migration. Cataloged via empirical analysis of the codebase before starting:

- `AsyncValue.valueOrNull` removed (use `.value`, which returns `T?` in v3)
- `StateProvider` removed from main library (moved to `legacy.dart`)
- `StateNotifier` deprecated in favour of `Notifier`
- `AutoDisposeAsyncNotifier` fused into `AsyncNotifier`
- `Provider<void>` side-effect pattern no longer cleanly compatible with v3 lifecycle
- `ref.invalidate` no longer triggers rebuild without an active listener
- `StreamProvider` pauses when not actively listened
- Automatic retry of failed providers (timers can outlive tests)

## Acceptance

- [ ] `flutter analyze` reports zero issues
- [ ] `flutter test` passes
- [ ] Manual smoke test on macOS: dashboard, PR list, activity tab live mode, repositories, agents config save, stats filters
- [ ] CI green

ivanmunozruiz:
Closes #445.

## Summary

Final Tier 3 bump (closes the deps-cleanup series started with #439). flutter_riverpod 2.6.1 → 3.3.1: ~80 sites across 26 files in the core migration commit, plus a follow-up that compensates for v3's stricter provider lifecycle in tab navigation.

## Migration breakdown (commit `14bff17`)

| Area | Sites | Change |
|---|---|---|
| `AsyncValue.valueOrNull` → `.value` | 27 | v3 removed `valueOrNull`; `.value` returns `T?` now |
| `StateProvider<T>` → `NotifierProvider<LocalStateNotifier<T>, T>` | 11 | v3 moves `StateProvider` to legacy. New helper `LocalStateNotifier<T>` (under `core/state/`) keeps `.set(T)` / `.update((T)=>T)` ergonomics so consumers stay close to the old shape |
| `extends StateNotifier<T>` → `extends Notifier<T>` | 1 | `ActivityQueryNotifier`: constructor body moves into `build()` override |
| `Provider<void>` side-effect → `Notifier<void>` | 1 | `activityLiveRefreshProvider` re-implemented as `ActivityLiveRefreshNotifier`; integrates cleanly with v3 lifecycle when watched |
| `AutoDisposeAsyncNotifier` → `AsyncNotifier` | 1 | v3 fuses them; `.autoDispose` modifier stays on provider declaration |
| `.notifier.state = v` → `.notifier.set(v)` | 12 | Consumer-side; `update((s)=>...)` callers unchanged thanks to helper |

## Tab keep-alive fix (commit `39eb9d1`)

Smoke testing surfaced an intermittent "click tab twice to open" feel after the migration. Root cause: Riverpod 3 pauses providers without active listeners, and Flutter's `TabBarView` disposes off-screen children — so switching away from a tab released the only `ref.watch` keeping its providers awake. Re-entering rendered an empty frame while the chain reactivated.

Added a `KeepAliveTab` wrapper (`AutomaticKeepAliveClientMixin`) and applied it to the three `TabBarView`s with persistent provider state:
- Dashboard tabs (Activity / Activity Log / Repositories / Prompts / Agents / Stats)
- Agents prompt-category tabs (PR Review / Issue Triage / Development)
- Server tabs (Status / Events / Logs)

The modal-internal TabBarView in the agents prompt editor is left untouched — that one is recreated per dialog open.

## v3 behaviour changes that needed test compensation

- **Providers without active listeners get paused** in v3. Production paths use `ref.watch` and are fine; tests have to attach explicit listeners (`c.listen`).
- **`ref.invalidate` no longer forces a rebuild** without an active listener. Same compensation in the affected test.
- **New auto-retry timer** can outlive a test if a provider's build throws. The two `saveAndStartDaemon` tests now pass `retry: (_, _) => null` to disable retry for that container.

## Test plan

- [x] `flutter analyze` — zero issues
- [x] `flutter test` — 167/167
- [x] Manual smoke test on macOS: dashboard, PR list, activity live mode, repositories, agents config save, stats filters, server tab
- [x] Tab switching is now consistently single-click after the keep-alive fix
- [ ] CI green

### DIFF
--- a/flutter_app/lib/core/state/local_state_notifier.dart
+++ b/flutter_app/lib/core/state/local_state_notifier.dart
@@ -1,0 +1,15 @@
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+/// Drop-in replacement for Riverpod 2's `StateProvider`. Exposes the same
+/// `.set(T)` / `.update((T)=>T)` ergonomics so existing call sites keep their
+/// shape; the only consumer-side change is `notifier.state = v` → `.set(v)`.
+class LocalStateNotifier<T> extends Notifier<T> {
+  LocalStateNotifier(this._initial);
+  final T _initial;
+
+  @override
+  T build() => _initial;
+
+  void set(T value) => state = value;
+  void update(T Function(T) updater) => state = updater(state);
+}
--- a/flutter_app/lib/features/activity/activity_providers.dart
+++ b/flutter_app/lib/features/activity/activity_providers.dart
@@ -121,6 +121,10 @@ const _activityLogEventTypes = {
 /// Implemented as a Notifier (not `Provider<void>`) so it integrates cleanly with
 /// Riverpod 3's lifecycle: when the screen watches it, the notifier stays
 /// alive, and the nested `ref.listen(sseStreamProvider, ...)` stays subscribed.
+/// The provider is declared without `.autoDispose`, so it relies on v3's
+/// default keep-alive semantics — adding `.autoDispose` here would re-introduce
+/// the "paused while between watchers" behaviour this notifier was created to
+/// avoid.
 class ActivityLiveRefreshNotifier extends Notifier<void> {
   @override
   void build() {
--- a/flutter_app/lib/features/activity/activity_providers.dart
+++ b/flutter_app/lib/features/activity/activity_providers.dart
@@ -3,13 +3,15 @@ import 'dart:async';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 
 import '../../core/models/activity.dart';
+import '../../core/state/local_state_notifier.dart';
 import '../dashboard/dashboard_providers.dart'
     show apiClientProvider, sseStreamProvider;
 
 /// Notifier managing the current query (date, range, filter sets).
 /// Widgets read & mutate this; the entries provider watches it.
-class ActivityQueryNotifier extends StateNotifier<ActivityQuery> {
-  ActivityQueryNotifier() : super(_today());
+class ActivityQueryNotifier extends Notifier<ActivityQuery> {
+  @override
+  ActivityQuery build() => _today();
 
   static ActivityQuery _today() {
     final now = DateTime.now();
@@ -88,11 +90,14 @@ class ActivityQueryNotifier extends StateNotifier<ActivityQuery> {
 
 /// Current query state.
 final activityQueryProvider =
-    StateNotifierProvider<ActivityQueryNotifier, ActivityQuery>(
-      (ref) => ActivityQueryNotifier(),
+    NotifierProvider<ActivityQueryNotifier, ActivityQuery>(
+      ActivityQueryNotifier.new,
     );
 
-final activityLiveUpdatesProvider = StateProvider<bool>((ref) => false);
+final activityLiveUpdatesProvider =
+    NotifierProvider<LocalStateNotifier<bool>, bool>(
+      () => LocalStateNotifier<bool>(false),
+    );
 
 const _activityLiveRefreshDelay = Duration(milliseconds: 750);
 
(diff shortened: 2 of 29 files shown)


Muriano:
## ⚠️ MEDIUM Issue

ReviewTriggerNotifier changed from AutoDisposeAsyncNotifier<void> to AsyncNotifier<void>. This is a lifecycle change, not just a rename: the notifier (and any in-flight trigger state) will no longer be disposed when the PR detail screen unmounts, so state can leak across navigations. In Riverpod 3 auto-dispose is controlled at the provider level (e.g. NotifierProvider.autoDispose) — if the previous auto-dispose behavior was intentional, the corresponding provider declaration needs `.autoDispose` to preserve it. Recommend either using NotifierProvider.autoDispose for the matching provider or documenting why retaining the notifier across disposals is now desired. Severity: medium.

**Location:** `flutter_app/lib/features/pr_detail/pr_detail_providers.dart` line 10

---

Muriano:
## ⚠️ MEDIUM Issue

Inside PrListRefreshNotifier.build(), `Future.microtask(() => state++)` is scheduled from a ref.listen callback. If the notifier is disposed between the microtask being scheduled and running (e.g. provider rebuild after SSE reconnect), accessing `state` on a disposed Notifier throws a StateError. The previous code had the same shape via `ref.read(...notifier).update(...)`, which is also unsafe but at least re-reads the current notifier. Recommend guarding with `if (!ref.mounted) return;` (Riverpod 3 exposes Ref.mounted) before mutating state. Severity: medium.

**Location:** `flutter_app/lib/features/dashboard/dashboard_providers.dart` line 78

---

Muriano (approved):

Clean mechanical migration from flutter_riverpod 2 → 3. StateNotifier/StateProvider replaced with Notifier-based equivalents via a LocalStateNotifier helper, .valueOrNull → .value, and KeepAliveTab wrappers added to preserve tab state. One behavior change (AutoDisposeAsyncNotifier → AsyncNotifier on ReviewTriggerNotifier) and one fragile microtask pattern worth verifying.

### Suggestions

- LocalStateNotifier.update shadows no inherited method today, but Notifier may gain similar helpers in future Riverpod versions; consider renaming to `mutate` or marking as @override-safe to avoid silent collisions.
- PrListRefreshNotifier defines both `bump()` and `update(int Function(int))`, but only `state++` is used inside build(); confirm external callers actually need both methods or drop the unused one (YAGNI).
- The doc comment on activityLiveRefreshProvider says the notifier 'stays alive' while watched — worth adding a note that this relies on Riverpod 3's default keep-alive semantics (it is NOT auto-dispose unless declared so), to prevent future regressions if someone adds `.autoDispose`.
- Consider a brief CHANGELOG / migration note documenting the `.valueOrNull` → `.value` sweep and the StateProvider → LocalStateNotifier pattern, so contributors don't reintroduce v2 idioms.

---

ivanmunozruiz:
@Muriano — follow-up applied in `6415e22`:

**Issue #2 (microtask race)** — Valid. Added `if (!ref.mounted) return;` guard inside the `Future.microtask` body.

**Suggestion (unused `bump()`)** — Valid. Dropped. `update((s)=>s+1)` from the shared `LocalStateNotifier` shape covers external callers, and `state++` is used directly inside `build()`.

**Suggestion (activityLiveRefresh docstring)** — Applied. Documented the intentional absence of `.autoDispose`.

**Issue #1 (autoDispose) — refuting**: the change from `extends AutoDisposeAsyncNotifier<void>` to `extends AsyncNotifier<void>` does not remove autoDispose behaviour. In Riverpod 3, autoDispose is declared at the **provider** level, not the class. The provider declaration on `pr_detail_providers.dart:28` is `AsyncNotifierProvider.autoDispose<ReviewTriggerNotifier, void>(ReviewTriggerNotifier.new)` — `.autoDispose` is preserved. The class-side change is the v3 idiom (the `AutoDispose*` class hierarchy was fused into the base classes — see the v3 changelog). No state leak across navigations.

**Suggestion (rename `LocalStateNotifier.update` to avoid future collision)** — Not applied. No collision with `Notifier`'s public API today; if a future Riverpod version introduces an `update` method, dealing with it then is cheaper than pre-emptive renames here. YAGNI.

CI green; 167/167 tests passing.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of Riverpod 3 migration follow-up. Both previously flagged issues have been resolved: (1) the autoDispose concern is correctly refuted by the author — in Riverpod 3 the AutoDispose* class hierarchy was fused into the base classes, and `reviewTriggerProvider` is declared as `AsyncNotifierProvider.autoDispose<ReviewTriggerNotifier, void>(...)`, so autoDispose semantics are preserved at the provider level; (2) `PrListRefreshNotifier.build()` now guards the microtask body with `if (!ref.mounted) return;` before mutating state, eliminating the dispose race. No new issues found in the latest changes.

---

PR #448, opened 2026-05-11
fix(notifications): migrate from local_notifier to flutter_local_notifications

### LINKED ISSUE #438: bug: macOS system notifications stopped working
## Problem

System notifications on macOS are no longer being delivered. The daemon triggers notification events (review started, review completed, errors, etc.) but no macOS notification banners appear.

## Expected Behavior

Heimdallm should show native macOS notifications for key events:
- Review started / completed
- Issue triage completed
- Errors and failures
- Stage promotions

## Steps to Reproduce

1. Run Heimdallm on macOS with notifications previously working
2. Trigger a review or wait for the poll cycle to process an item
3. No notification banner appears

## Investigation Hints

- Check if the notification entitlement is present in the app bundle
- Verify `UNUserNotificationCenter` authorization status
- Check if the LaunchAgent-spawned daemon has permission to post notifications (daemons vs user-agents have different notification scopes)
- macOS Sequoia / Sonoma may have changed notification permission behavior for unsigned apps
- Verify the `xattr -cr` step isn't stripping notification entitlements

ivanmunozruiz:
Closes #438.

## Problem

The `local_notifier` package's macOS backend is built on `NSUserNotification` / `NSUserNotificationCenter`, deprecated by Apple in macOS 10.14 (2018). On recent macOS releases the system increasingly drops notifications posted through this path with no error, no banner, and no entry in Notification Center — exactly the "stopped working" symptom @Muriano reported. Permissions don't help because the API path itself is the deprecated one.

## Fix

Swap `local_notifier ^0.1.6` → `flutter_local_notifications ^21.0.0`, whose macOS backend uses **`UNUserNotificationCenter`** (the modern User Notifications framework). On first run the plugin calls `requestAuthorization` and macOS shows the standard permission prompt; subsequent notifications go through the normal Notification Center pipeline.

The new dependency also covers Linux (libnotify via dbus), so the team member on Linux keeps working without a code branch.

## API shape

The `PlatformServices.showNotification` / `setupNotifier` abstract surface is unchanged — callers in `main.dart` aren't touched. Internal differences absorbed inside `DesktopPlatformServices`:

| Before (`local_notifier`) | After (`flutter_local_notifications`) |
|---|---|
| `LocalNotification(title:body:)` → `n.onClick = ...; n.show()` | `_notifier.show(id:title:body:notificationDetails:payload:)` |
| Per-instance `onClick` callback | Global `onDidReceiveNotificationResponse`, dispatched via a `Map<int, VoidCallback>` keyed by the notification id (passed as `payload`) |
| `localNotifier.setup(appName:)` | `_notifier.initialize(settings: InitializationSettings(macOS:..., linux:...), onDidReceiveNotificationResponse: ...)` |

## Test plan

- [x] `flutter analyze` — zero issues
- [x] `flutter test` — 167/167
- [ ] Manual smoke test on macOS: trigger a review, expect a system notification banner and an entry in Notification Center; click → app focuses + navigates to PR detail
- [ ] @Muriano confirms notifications now appear on his machine
- [ ] CI green

### DIFF
--- a/flutter_app/devtools_options.yaml
+++ b/flutter_app/devtools_options.yaml
@@ -1,0 +1,3 @@
+description: This file stores settings for Dart & Flutter DevTools.
+documentation: https://docs.flutter.dev/tools/devtools/extensions#configure-extension-enablement-states
+extensions:
--- a/flutter_app/lib/core/platform/platform_services_desktop.dart
+++ b/flutter_app/lib/core/platform/platform_services_desktop.dart
@@ -144,6 +144,12 @@ class DesktopPlatformServices with WindowListener implements PlatformServices {
   // Notification plumbing. flutter_local_notifications wires onClick through
   // a single global callback (per payload) instead of per-instance handlers,
   // so we keep a small id → callback map here and use the id as payload.
+  //
+  // The map is bounded by `_maxNotifierHandlers` with FIFO eviction (Dart's
+  // default Map preserves insertion order). Without this, notifications that
+  // are dismissed or expire without being clicked would leak their closures
+  // indefinitely in a long-running tray app.
+  static const int _maxNotifierHandlers = 128;
   final FlutterLocalNotificationsPlugin _notifier =
       FlutterLocalNotificationsPlugin();
   final Map<int, VoidCallback> _notifierHandlers = {};
@@ -182,8 +188,18 @@ class DesktopPlatformServices with WindowListener implements PlatformServices {
     required String body,
     VoidCallback? onClick,
   }) {
-    final id = _nextNotifierId++;
-    if (onClick != null) _notifierHandlers[id] = onClick;
+    // Bitmask keeps the id in 32-bit positive range; some platform notifier
+    // backends use 32-bit ids internally, so we avoid relying on Dart's
+    // 64-bit ints surviving the boundary.
+    final id = _nextNotifierId = (_nextNotifierId + 1) & 0x7FFFFFFF;
+    if (onClick != null) {
+      _notifierHandlers[id] = onClick;
+      if (_notifierHandlers.length > _maxNotifierHandlers) {
+        // Drop the oldest handler. Map preserves insertion order, so the
+        // first key is the eldest entry.
+        _notifierHandlers.remove(_notifierHandlers.keys.first);
+      }
+    }
     _notifier.show(
       id: id,
       title: title,
--- a/flutter_app/lib/core/platform/platform_services_desktop.dart
+++ b/flutter_app/lib/core/platform/platform_services_desktop.dart
@@ -2,7 +2,7 @@ import 'dart:io';
 import 'dart:ui' show VoidCallback;
 import 'package:flutter/foundation.dart' show debugPrint;
 import 'package:flutter/painting.dart' show Size;
-import 'package:local_notifier/local_notifier.dart';
+import 'package:flutter_local_notifications/flutter_local_notifications.dart';
 import 'package:tray_manager/tray_manager.dart';
 import 'package:window_manager/window_manager.dart';
 import '../api/api_client.dart';
@@ -17,9 +17,9 @@ import 'platform_services.dart';
 
 /// Desktop implementation of [PlatformServices].
 ///
-/// Wraps dart:io, tray_manager, window_manager, local_notifier, and the
-/// existing FirstRunSetup / DaemonLifecycle helpers so that shared code
-/// never has to import them directly.
+/// Wraps dart:io, tray_manager, window_manager, flutter_local_notifications,
+/// and the existing FirstRunSetup / DaemonLifecycle helpers so that shared
+/// code never has to import them directly.
 class DesktopPlatformServices with WindowListener implements PlatformServices {
   DesktopPlatformServices({
     int apiPort = 7842,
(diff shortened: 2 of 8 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean migration from local_notifier to flutter_local_notifications. The callback-routing pattern via payload IDs is the standard approach for this plugin, but the handler map has an unbounded growth problem when notifications are dismissed without being clicked.

### Issues

⚠️ **flutter_app/lib/core/platform/platform_services_desktop.dart:170** — Memory leak: `_notifierHandlers` entries are only removed in `onDidReceiveNotificationResponse` (i.e., when the user clicks). Notifications that expire or are dismissed by the user will leave their VoidCallback (and anything it captures) in the map forever. In a long-running desktop app this is unbounded growth.
🟡 **flutter_app/lib/core/platform/platform_services_desktop.dart:155** — No `WindowsInitializationSettings` provided. The lock file pulls in `flutter_local_notifications_windows` as a transitive dependency, and the old `local_notifier` package supported Windows. If the app ships on Windows, notifications will silently fail or use undocumented defaults.
🟡 **flutter_app/lib/core/platform/platform_services_desktop.dart:148** — `_nextNotifierId` increments without bound. While 64-bit Dart ints won't overflow in practice, some platform notification APIs use 32-bit IDs internally. After 2^31 notifications the platform-side ID could wrap and collide, producing undefined behavior. A simple `_nextNotifierId = (_nextNotifierId + 1) & 0x7FFFFFFF` keeps it in safe range.

### Suggestions

- Cap `_notifierHandlers` size or add eviction: e.g., when inserting a new handler, evict entries with IDs more than N behind `_nextNotifierId`. This bounds the map without needing a timer.
- Add `windows: WindowsInitializationSettings()` to `InitializationSettings` for parity with the old local_notifier support, or document that Windows is intentionally unsupported.
- Consider wrapping `_nextNotifierId` with a bitmask (e.g., `& 0x7FFFFFFF`) to stay within 32-bit positive range for platform safety.

---

Muriano:
## ⚠️ MEDIUM Issue

Memory leak in `_notifierHandlers`: an entry is inserted on every `showNotification` call with a non-null `onClick`, but it is only removed inside `onDidReceiveNotificationResponse`. Notifications that the user dismisses, ignores, or that expire from Notification Center never fire that callback, so the closure (and anything it captures) is retained for the lifetime of the process. In a long-running tray app this grows unboundedly. Fix: track a TTL/cap, remove the handler after a timeout, or clear stale entries when `_nextNotifierId` wraps a threshold.

**Location:** `flutter_app/lib/core/platform/platform_services_desktop.dart` line 186

---

Muriano (approved):

Clean migration from local_notifier to flutter_local_notifications. The plumbing is reasonable, but the id→callback map leaks entries whenever a notification is dismissed/expires without being clicked, and Windows initialization is missing even though the Windows plugin is pulled in transitively.

### Suggestions

- `flutter_local_notifications_windows` is now a transitive dependency but `InitializationSettings` only configures macOS and Linux. If the desktop build ever targets Windows, `initialize()` will leave it unconfigured — consider adding `WindowsInitializationSettings` or documenting that Windows is intentionally unsupported here.
- `_nextNotifierId` increments forever; consider resetting or using a bounded id space (the plugin uses 32-bit ints on Android — not relevant here, but worth a comment).
- `_notifier.show(...)` returns a `Future` that is now discarded; `showNotification` could be made `async` to surface plugin errors instead of swallowing them.
- Consider exposing a `dispose()` that clears `_notifierHandlers` and cancels pending notifications when the desktop services shut down.

---

Muriano:
## ⚠️ MEDIUM Issue

Memory leak in _notifierHandlers still present: entries are inserted on every showNotification call with a non-null onClick but only removed inside onDidReceiveNotificationResponse. Notifications the user dismisses, ignores, or that expire never fire the callback, so closures are retained for the process lifetime. In a long-running tray app this grows unboundedly. Fix: cap the map size with LRU eviction, add a TTL that clears stale entries, or periodically prune when _nextNotifierId crosses a threshold.

**Location:** `flutter_app/lib/core/platform/platform_services_desktop.dart` line 186

---

Muriano (approved):

Migration from local_notifier to flutter_local_notifications is clean and well-documented. The previously flagged memory leak in _notifierHandlers remains unaddressed in this diff — entries are still only removed on user click, not on dismiss/expire.

### Suggestions

- Consider resetting _nextNotifierId or using a bounded id space — current int increments are fine in practice but document the assumption.
- _notifier and _notifierHandlers could be made private final at construction; currently fine but worth noting if DesktopPlatformServices is ever instantiated multiple times in tests.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Author addressed the previous memory leak finding by adding a bounded FIFO eviction (`_maxNotifierHandlers = 128`) on the `_notifierHandlers` map. The migration from local_notifier to flutter_local_notifications is otherwise clean: plugin registrants, Podfile.lock, and pubspec are consistently updated; macOS permissions are requested explicitly; payload is used to route clicks to the correct callback. No new blocking issues.

### Suggestions

- FIFO eviction drops the *oldest* pending handler, which may not match user intent (the oldest notification is also the one most likely to still be visible to the user). LRU-by-access isn't meaningful here since entries are only accessed on click; consider instead bounding by recent activity or adding a TTL that prunes when `_nextNotifierId` advances past a threshold.
- `_nextNotifierId` is masked to 31 bits and can theoretically wrap to 0 after ~2B notifications, colliding with a still-pending handler. Practically unreachable in a tray app, but a brief comment noting the assumption would help.
- Consider exposing a `dispose()` path that clears `_notifierHandlers` if `DesktopPlatformServices` is ever recreated, to avoid retaining captured closures across hot-restart in development.

---

ivanmunozruiz:
Follow-up applied in `7a0b036`:

**Memory leak in `_notifierHandlers`** — Valid. Capped the map at 128 entries with FIFO eviction (Dart's default Map preserves insertion order, so removing `keys.first` drops the eldest). Notifications that are dismissed or expire without being clicked no longer leak their closures.

**`_nextNotifierId` overflow** — Applied. Masked with `& 0x7FFFFFFF` to stay in 32-bit positive range, since some platform backends use 32-bit ids internally.

**Suggestions not applied, with rationale:**

- **`WindowsInitializationSettings`** — Heimdallm doesn't target Windows. `flutter_launcher_icons` in `pubspec.yaml` configures macOS and Linux only; there's no Windows runner bundle in the build, and CI/release pipelines don't produce Windows artifacts. The Windows-side plugin pod that comes in transitively is dead weight, not a runtime concern. Will document if a Windows target is ever added.

- **`show()` Future discarded** — Intentional. Converting `showNotification` to `async` would change the abstract `PlatformServices.showNotification` signature (currently `void`), forcing every caller in `main.dart`/etc. to `await` for a fire-and-forget operation. The plugin handles its own internal errors; we already log via the Dart VM if something throws. Net cost of the change > benefit.

- **`dispose()` method** — Out of scope here. `DesktopPlatformServices` is a process-wide singleton in this tray app; the only "disposal" event is app exit, at which point the OS reclaims everything. Adding a no-op dispose() ceremony would suggest a lifecycle that doesn't exist.

CI green (Flutter run is still in progress from the previous push, will rebuild on this commit); 167/167 tests passing locally.

PR #457, opened 2026-05-11
fix(issues): scope auto-promote to daemon assignee + record real PR author (#456)

### LINKED ISSUE #456: Heimdallm: corregir lógica de autopromote y asignación de autor en PRs
## Contexto

Detectados dos bugs en el comportamiento de Heimdallm al crear PRs automáticas.

### 1. Lógica de autopromote incorrecta
Heimdallm hace autopromote siempre que el assignee triageado sea cualquier usuario. El comportamiento esperado es:
- Si el assignee triageado **es el mismo usuario** que lanza la acción → autopromote.
- Si el assignee triageado **es otro usuario** → ofrecer la opción de "pasarle la pelota" (asignarle la PR a ese usuario), sin autopromote automático.

### 2. Asignación incorrecta del autor de la PR
Heimdallm asigna como autor de la PR a un usuario distinto del que realmente la creó. Debe asignar correctamente al autor real de la PR.

---
_Originalmente reportado en freepik-company/ai-platform-workspace#304 (Slack #ai-platform-heimdallm-feedback, creado por Hermod). Copiado aquí porque GitHub no permite transferir issues entre orgs._

ivanmunozruiz:
## Summary

Fixes the two bugs reported in #456:

1. **Auto-promote bypassed assignee scope.** When triage reassigned an issue to a different user, the daemon kept advancing the stage labels. Root cause: `autoPromoteAfterStage` operated on the pre-pipeline `ghIssue`, so the post-pipeline assignee change (`applyIssueTriageMetadata` → `SetAssignees`) was invisible.
2. **Auto-created PRs were credited to the issue reporter.** `CreatedPR` discarded `user.login` from the GitHub response and the pipeline wrote `Author: issue.User.Login`.

## Changes

- `autoPromoteAfterStage` now refetches the issue, then gates the transition on `it.MatchesAssignees(fresh.AssigneeLogins())`. Out-of-scope → skip with an INFO log; the issue's new assignee owns the next stage (existing "ball passed" handoff). Refetch failure falls back to the cached copy so a transient API error never fail-opens.
- `CreatedPR` carries `Author` (parsed from `user.login`). The auto-implement pipeline uses it; falls back to a new `RunOptions.AuthUser` (the daemon's cached login) when the API response omits the field.
- `autoPromoteAfterStage` narrowed to an `autoPromoteGitHub` interface — purely a testability seam; `*gh.Client` still satisfies it.

## Tests (TDD)

- `TestCreatePR_CapturesAuthorFromResponse` / `TestCreatePR_AuthorEmptyWhenMissing` — wire-level parsing.
- `TestAutoImplement_StoresCreatedPRAuthorFromGitHubResponse` / `TestAutoImplement_FallsBackToAuthUserWhenCreatedPRAuthorEmpty` — pipeline records the right author.
- `TestAutoPromoteSkippedWhenAssigneeOutOfScope` — no label mutation when refetch shows a different assignee.
- `TestAutoPromoteProceedsWhenAssigneeInScope` — happy path still promotes.
- `TestAutoPromoteRefetchFailureFallsBackToCachedIssue` — transient refetch error → continues with the cached issue.

## Test plan
- [x] `make test-docker` — full suite green locally.
- [ ] Manual: triage an issue assigned to a different operator, confirm no stage advance and no spurious PR.
- [ ] Manual: trigger auto-implement, verify Activity view shows the bot login as PR author.

Closes #456

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -3536,15 +3536,15 @@ func autoPromoteAfterStage(
 	// Refetch the issue so the scope check sees post-pipeline assignees:
 	// triage / refinement may have reassigned it to another user ("passed
 	// the ball"), and the worker's cached copy still reflects the old
-	// owner. A refetch failure falls back to the cached copy — preferable
-	// to fail-open with no assignee data when the issue has, e.g., a
-	// transient GitHub outage.
-	fresh := issue
-	if updated, err := client.GetIssue(issue.Repo, issue.Number); err != nil {
-		slog.Warn(scope+": auto-promote refetch failed, using cached issue for scope check",
+	// owner. Refetch failure is fail-closed: falling back to the cached
+	// copy would reintroduce the exact race this gate is closing (a stale
+	// in-scope snapshot during a transient outage could auto-promote past
+	// a handoff). Skip the promotion; the next pipeline event will retry.
+	fresh, err := client.GetIssue(issue.Repo, issue.Number)
+	if err != nil {
+		slog.Warn(scope+": auto-promote skipped — issue refetch failed; will retry on next pipeline event",
 			"repo", issue.Repo, "number", issue.Number, "err", err)
-	} else if updated != nil {
-		fresh = updated
+		return
 	}
 
 	if !it.MatchesAssignees(fresh.AssigneeLogins()) {
@@ -3559,18 +3559,18 @@ func autoPromoteAfterStage(
 	if err != nil {
 		if errors.Is(err, issuepipeline.ErrStageTargetLabelMissing) {
 			slog.Warn(scope+": auto-promote target label missing; leaving issue in current stage",
-				"repo", fresh.Repo, "number", fresh.Number, "from", from, "err", err)
+				"repo", issue.Repo, "number", issue.Number, "from", from, "err", err)
 			return
 		}
 		slog.Warn(scope+": auto-promote skipped",
-			"repo", fresh.Repo, "number", fresh.Number, "from", from, "err", err)
+			"repo", issue.Repo, "number", issue.Number, "from", from, "err", err)
 		return
 	}
 
-	comments, err := client.FetchIssueCommentsOnly(fresh.Repo, fresh.Number)
+	comments, err := client.FetchIssueCommentsOnly(issue.Repo, issue.Number)
 	if err != nil {
 		slog.Warn(scope+": auto-promote comment fetch failed, continuing without audit dedup context",
-			"repo", fresh.Repo, "number", fresh.Number, "err", err)
+			"repo", issue.Repo, "number", issue.Number, "err", err)
 	}
 	if err := issuepipeline.TransitionIssueStage(ctx, client, issuepipeline.StageTransition{
 		Issue:          fresh,
@@ -3584,7 +3584,7 @@ func autoPromoteAfterStage(
 		Broker:         broker,
 	}); err != nil {
 		slog.Warn(scope+": auto-promote failed",
-			"repo", fresh.Repo, "number", fresh.Number, "from", from, "to", to, "err", err)
+			"repo", issue.Repo, "number", issue.Number, "from", from, "to", to, "err", err)
 	}
 }
 
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -901,6 +901,7 @@ func main() {
 			PRLabels:                aiCfg.PRLabels,
 			PRDraft:                 aiCfg.PRDraft != nil && *aiCfg.PRDraft,
 			GeneratePRDescription:   aiCfg.GeneratePRDescription != nil && *aiCfg.GeneratePRDescription,
+			AuthUser:                authUser,
 		}
 
 		rev, err := issuePipe.Run(ctx, ghIssue, opts)
@@ -1096,6 +1097,7 @@ func main() {
 			PRLabels:                 aiCfg.PRLabels,
 			PRDraft:                  aiCfg.PRDraft != nil && *aiCfg.PRDraft,
 			GeneratePRDescription:    aiCfg.GeneratePRDescription != nil && *aiCfg.GeneratePRDescription,
+			AuthUser:                 authUser,
 			RequireWorkDirForDevelop: true,
 		}
 
@@ -2828,6 +2830,7 @@ func (a *tier2Adapter) ProcessRepo(ctx context.Context, repo string) (int, error
 			PRLabels:                    aiCfg.PRLabels,
 			PRDraft:                     aiCfg.PRDraft != nil && *aiCfg.PRDraft,
 			GeneratePRDescription:       aiCfg.GeneratePRDescription != nil && *aiCfg.GeneratePRDescription,
+			AuthUser:                    authUser,
 			RequireWorkDirForDevelop:    requireWorkDir,
 			RequireWorkDirForRefinement: requireRefinementWorkDir,
 			ReleaseRepoContext:          releaseRepoContext,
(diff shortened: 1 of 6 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured fix addressing two distinct bugs: (1) scoping auto-promote to the daemon's assignee after a pipeline stage, and (2) recording the real PR author from the GitHub API response instead of the issue reporter. Good test coverage, clean interface extraction, and robust refetch-with-fallback pattern. A few minor concerns around edge cases and test fragility.

### Issues

⚠️ **daemon/cmd/heimdallm/main_test.go:153** — TestAutoPromoteProceedsWhenAssigneeInScope passes nil for the broker parameter, yet the transition is expected to proceed end-to-end (label mutations observed). If TransitionIssueStage ever publishes an event via the broker, this test will panic instead of failing gracefully. Consider passing a no-op broker stub.
🟡 **daemon/cmd/heimdallm/main.go:3546** — When GetIssue fails AND the cached issue has no assignees (empty AssigneeLogins()), the behavior depends entirely on MatchesAssignees([]string{}). If that returns true (vacuous match), auto-promote proceeds unscoped — the fail-open intent is correct for transient errors but could silently promote unowned issues. This edge case has no test coverage.
🟡 **daemon/cmd/heimdallm/main_test.go:91** — autoPromoteFakeGH.RemoveLabels appends individual label strings (via '...') into a flat []string, while AddLabels appends the whole []string slice into [][]string. This inconsistency is not a bug today (tests only check len()), but makes future assertions harder to write correctly and could mask bugs where RemoveLabels is called multiple times.
🟡 **daemon/internal/issues/pipeline.go:791** — The TrimLeft(opts.AuthUser, "@") call treats '@' as a set of runes to strip from the left, not a prefix. This is fine for a single '@' but would incorrectly strip leading '@' characters from usernames like '@@user'. TrimPrefix would be more precise, though this is unlikely to matter in practice since GitHub logins cannot start with '@'.

### Suggestions

- Add a no-op broker stub (e.g., &fakeBroker{}) in TestAutoPromoteProceedsWhenAssigneeInScope and TestAutoPromoteRefetchFailureFallsBackToCachedIssue to prevent future panics if TransitionIssueStage gains publish calls.
- Consider adding a test case for auto-promote when the cached issue has zero assignees and the refetch fails, to document the expected behavior of MatchesAssignees on an empty list.
- Use strings.TrimPrefix(opts.AuthUser, "@") instead of strings.TrimLeft(opts.AuthUser, "@") for correctness — TrimLeft strips a character set, not a prefix.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Targeted two-part fix: scopes auto-promote to the daemon's assignee via a post-pipeline issue refetch, and records the real PR author from the GitHub response with an AuthUser fallback. Changes are well-tested (unit tests for in-scope, out-of-scope, refetch failure, author capture, author fallback), use a narrow interface seam for testability, and preserve safe fallback behavior. No confirmed bugs found.

### Suggestions

- The refetch-failure path falls back to the cached issue; if the cached issue itself has no assignees (e.g., a freshly unassigned issue), MatchesAssignees behavior on empty input should be confirmed to fail-closed rather than fail-open — worth a brief comment or explicit test.
- Consider extracting the trim/normalize logic (`strings.TrimSpace(strings.TrimLeft(x, "@"))`) into a small helper since it now appears in both defaultAutoImplementPRAssignee and the new prAuthor fallback.
- The autoPromoteGitHub interface doc comment is helpful; consider also noting that callers must pass a non-nil client since GetIssue is invoked unconditionally.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Well-structured fix for two related #456 bugs: scoping auto-promote to the daemon's assignee via post-pipeline refetch, and persisting the actual PR author returned by GitHub with AuthUser fallback. Good test coverage including the refetch-failure fallback path. Interface-based testability seam is clean and locally scoped.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:3540** — Refetch-failure fallback to the cached issue is documented as 'fail-open' but can silently auto-promote on a stale assignee snapshot during transient GitHub outages — exactly the race this PR is trying to close. Consider skipping promotion (fail-closed) on refetch error, or at least gating the fallback on whether the cached issue has any assignees at all. The test only covers the in-scope cached path.
🟡 **daemon/cmd/heimdallm/main.go:3543** — `if updated != nil` after a nil error is defensive against a contract `GetIssue` likely doesn't violate. If GetIssue never returns (nil, nil), this branch is dead; if it can, the fallback semantics should be documented in GetIssue rather than guarded at call sites.
🟡 **daemon/cmd/heimdallm/main.go:3556** — Log-field renames from `issue.Repo/Number` to `fresh.Repo/Number` are cosmetic — repo/number cannot change across a refetch — and add visual churn to the diff without behavior change.
🟡 **daemon/cmd/heimdallm/main_test.go:175** — Missing test for the out-of-scope cached + refetch-failure combination (i.e., GetIssue errors and the cached copy is also out of scope). Would lock in expected behavior for the worst-case of the fail-open policy.
🟡 **daemon/internal/issues/pipeline.go:791** — `prAuthor` falls back to AuthUser but there is no final fallback if both are empty — the PR row would persist Author="". Worth a quick check whether the store/Activity view tolerates an empty author, or whether a last-resort fallback (e.g., issue.User.Login with a warn log) is preferable to silently storing blank.

### Suggestions

- Consider fail-closed on refetch error: if the whole point of the refetch is to avoid promoting after a handoff, falling back to the cached copy reintroduces the exact race during transient outages. A skip-with-warn is safer and the next pipeline tick will retry.
- If `*gh.Client.GetIssue` can legitimately return (nil, nil), document it on the method; otherwise drop the `updated != nil` guard.
- Add a regression test for the out-of-scope cached + refetch-failure case to pin down the fail-open behavior explicitly.
- In `runAutoImplement`, log a warn when both `createdPR.Author` and `opts.AuthUser` are empty so the stored-blank case is observable rather than silent.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

PR correctly addresses two issues from #456: scopes auto-promote to the daemon's assignee via a fresh refetch (fail-closed on error) and records the real PR author from the GitHub API response with a fallback to AuthUser. Code is well-tested with unit tests covering happy path, out-of-scope handoff, refetch failure, and author fallback. The new autoPromoteGitHub interface is appropriately narrow and serves as a testability seam. No confirmed bugs found.

### Suggestions

- Consider distinguishing transient errors (network/5xx) from permanent ones (404 issue deleted) on GetIssue refetch — fail-closed is correct for transient failures but a 404 will retry forever on each pipeline event.
- The autoPromoteGitHub interface comment mentions FetchIssueCommentsOnly is used for audit dedup but the diff does not show the call site; verify the interface members are all actually exercised by autoPromoteAfterStage and its TransitionIssueStage path.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: the medium-severity fail-open concern is fully addressed — refetch failure is now fail-closed with a clear rationale comment and a dedicated regression test (TestAutoPromoteSkippedWhenRefetchFailsEvenIfCachedOutOfScope). The defensive `if updated != nil` and the cosmetic log-field renames from the prior round are gone. The new autoPromoteGitHub interface is a tight, well-scoped testability seam. Remaining nits are minor.

### Issues

🟡 **daemon/internal/issues/pipeline.go:801** — When both createdPR.Author and opts.AuthUser are empty, the code logs a warning but still proceeds to upsert prRow with Author="". The comment acknowledges the NOT NULL constraint will surface it, but relying on a DB error for a foreseeable case mixes layers — consider either short-circuiting the store insert or using a last-resort fallback (e.g., issue.User.Login) so the warn-and-fail path is explicit rather than dependent on schema enforcement.
🟡 **daemon/cmd/heimdallm/main.go:3517** — The autoPromoteGitHub interface embeds StageTransitionClient and adds GetIssue + FetchIssueCommentsOnly, but FetchIssueCommentsOnly is not visibly called in autoPromoteAfterStage in the diff — it's presumably invoked transitively via TransitionIssueStage's audit-dedup path. The doc comment claims it but a reader of just this file can't verify; a one-line pointer to where it's consumed would help future maintainers (or drop it if it's only there to satisfy the *gh.Client surface).

### Suggestions

- Consider adding a final fallback for the auto-PR author column (e.g., 'unknown' sentinel or the issue reporter with an explicit log) if relying on the NOT NULL constraint is not the intended UX — otherwise document that the upsert is expected to fail loudly in that path.
- Optionally tighten the autoPromoteGitHub interface comment to name the indirect FetchIssueCommentsOnly caller (TransitionIssueStage) so the surface is self-documenting.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Re-review: previous medium finding (nil broker) is still present but less impactful than originally assessed — broker is not dereferenced in the auto-promote path. The RemoveLabels inconsistency was fixed. Both fixes (assignee scope gating and real PR author recording) are well-structured, well-tested, and address genuine bugs. No new issues found.

### Issues

🟡 **daemon/cmd/heimdallm/main_test.go:153** — TestAutoPromoteProceedsWhenAssigneeInScope still passes nil for the broker parameter. While broker is not dereferenced in the current auto-promote path (TransitionIssueStage does not take a broker), adding a broker.Publish call in autoPromoteAfterStage in the future would cause a nil-pointer panic in this test instead of a clean failure. A no-op stub (e.g. &fakeBroker{}) would be more defensive.

### Suggestions

- Consider adding a simple no-op Publisher stub for auto-promote tests so future additions of SSE events in that path don't silently turn into panics.
- The NOT NULL comment on the empty-prAuthor warning (pipeline.go:798) implies an INSERT failure, but in SQLite an empty string satisfies NOT NULL — if the intent is to reject empty authors, a CHECK constraint or explicit guard would be needed.

---

PR #459, opened 2026-05-11
fix(issues): restore handoff flow + single-flight triage claim per issue (#458)

### LINKED ISSUE #458: bug: handoff flow regression + duplicate triage on bot's own comment
Detectado al transferir un issue entre orgs y aplicar la label de triage en `freepik-company/ai-platform-iqs-v3#511`. Dos bugs encadenados, los meto juntos porque el reproductor es el mismo y el fix toca un fichero común.

### Síntoma

Tras triagear y asignar a otro operador, el daemon **no transiciona la label** a `refinement`. El issue se queda con label `heimdallm-triage` para siempre. El daemon del assignee real (si está corriendo) re-triagea en vez de refinar.

### Evidencia (logs de iqs-v3#511)

```
14:58:40.231 issues pipeline: triage complete #511 severity=medium posted=true
14:58:40.606 triage-worker: auto-promote skipped — issue assignee out of scope (handoff to another operator)
              issue_assignees=[asanmartinfreepik] scope_assignees=[ivanmunozruiz]
```

### Causa raíz

#457 metió un gate de assignee scope en `autoPromoteAfterStage` (`daemon/cmd/heimdallm/main.go`). Pensaba que sin él el daemon ejecutaría todas las etapas hasta crear PR en nombre del usuario equivocado.

Lo que no vi: **el worker entry ya tiene ese gate** (`issueTrackingWithAssigneeScope` + `issueStageStillCurrent` en `main.go:847,959,1034`, añadido por #429). Se ve en spam constante de los logs:

```
14:59:35.753 github: issue skipped by assignee scope repo=ai-platform-iqs-v3 number=511 assignees=[asanmartinfreepik] allowed_assignees=[ivanmunozruiz]
```

Resultado: **doble gate**. La label nunca pasa de triage, y el handoff queda muerto.

### Modelo correcto

- Daemon A triagea, reasigna a B, **mueve la label a refinement**.
- Worker entry de refinement en A: scope falla → skip (ya funciona).
- Worker entry de refinement en B: scope match → corre refinement.

La frase \"sin autopromote automático\" de #456 se refería a \"no ejecutes la siguiente etapa\", no a \"no muevas la label\". El worker entry scope es el gate apropiado; mi gate en auto-promote es redundante y nocivo.

### Fix

Revertir el `MatchesAssignees` skip de `autoPromoteAfterStage`. El refetch (también introducido por #457) se vuelve innecesario al no haber decisión que tomar — fuera también. Mantener únicamente:
- El parseo de `Author` en `CreatedPR` (bug 2 de #456, sin relación, sigue siendo correcto).
- El campo `RunOptions.AuthUser` y su uso como fallback.

---

### Síntoma

Dos comentarios de triage independientes sobre el mismo issue en ~22 segundos. El segundo se identifica a sí mismo como \"Re-triage\". Ambos consumen una run del LLM.

### Evidencia (logs + comentarios de iqs-v3#511)

```
14:57:36.028 triage-worker: processing #511        [run #1 arranca]
14:58:36.139 triage-worker: processing #511        [run #2 arranca]
14:58:40.231 pipeline: triage complete #511        [run #1 termina, PostComment]
14:59:00.265 pipeline: triage complete #511        [run #2 termina, PostComment]
```

Comments en GitHub:
- `12:58:36Z` — \"Heimdallm triage — MEDIUM ... Assigned owner: @asanmartinfreepik\"
- `12:58:58Z` — \"Heimdallm triage — MEDIUM ... Tentative owner: @sergiotejon\" (Re-triage)

### Causa raíz

`Pipeline.Run` en `daemon/internal/issues/pipeline.go:373` claves el in-flight con:

```go
claimKey = issue.UpdatedAt.UTC().Format(time.RFC3339)
ok, err := p.store.ClaimIssueTriageInFlight(claimIssueID, claimKey)
```

El claim es `(github_issue_id, updated_at)`. Cuando `PostComment` del run #1 bumpea el `updated_at` del issue al final del run, el tier2 poller (cada 60s) refresca la lista de issues, ve un `updated_at` distinto → toma un claim distinto → arranca run #2 en paralelo o casi-paralelo.

Misma forma que #362 (cerrado: \"claim key invalidated by bot own comment\"), pero éste era para una ruta distinta. La heurística de \"updated_at en el claim key\" permite re-triage legítimo cuando un humano comenta — pero también permite re-triage por el propio comentario del bot.

### Fix propuesto

Marker de \"last successful triage at\" por issue, separado del in-flight claim:
- Tabla `issue_last_triage_at(github_issue_id PK, posted_at TIMESTAMP)`.
- Al final de un run exitoso de triage, UPSERT con `now()`.
- Antes de claimar, consultar — si hay marker con `posted_at > now - cooldown` (sugerencia: 2 min, mismo orden que `stagePromotionWindow`), skip con log INFO.

Esto preserva la semántica de \"re-triage on new human activity\" después del cooldown, y bloquea la race del propio comentario del bot dentro del cooldown.

---

## Repro

1. Tener Heimdallm corriendo con scope `[userA]` sobre repo X.
2. Transferir issue desde otro repo a X con `gh issue transfer`.
3. Añadir label `heimdallm-triage` (o equivalente configurada como triage).
4. Esperar 2 min.

Observado: dos comentarios de triage y la label sigue en `heimdallm-triage`.

## Scope del PR

Una sola PR fixea ambos:
- Revertir el gate de auto-promote scope + refetch (Bug 1).
- Añadir cooldown marker para triage (Bug 2).
- Tests TDD para cada uno.

Closes #456 follow-up.

ivanmunozruiz:
## Summary

Fixes both bugs reported in #458, found while testing transfer + label on `freepik-company/ai-platform-iqs-v3#511`:

1. **Handoff broken** — #457 over-blocked by adding an assignee-scope skip in `autoPromoteAfterStage`. The next stage's worker entry already gates by scope (`issueTrackingWithAssigneeScope` + `issueStageStillCurrent`, added in #429). The double-gate left issues stuck at the current stage label; the new assignee's daemon never picked them up.
2. **Duplicate triage** — in-flight claim was keyed on `(issue_id, updated_at)`. The bot's own `PostComment` at end-of-triage bumps the issue's `updated_at` on GitHub. The next tier2 poll (within ~140ms in iqs-v3#511) saw a different `updated_at` → got a different claim → started a second triage run.

## Changes

- `autoPromoteAfterStage`: revert the assignee-scope skip + refetch. Label transition is unconditional when auto-promote is enabled; handoff happens at next-stage worker entry. The `autoPromoteGitHub` interface introduced by #457 also goes away (nothing left to mock here).
- `ClaimIssueTriageInFlight`: SQL changed from `INSERT OR IGNORE INTO ... (issue_id, updated_at)` to `INSERT ... WHERE NOT EXISTS (SELECT 1 ... WHERE issue_id=?)`. Atomic single-flight per `issue_id`, regardless of `updated_at`. The `updated_at` column stays as diagnostic context (which snapshot started the run). After release, any future updated_at can re-claim normally.

The `AuthUser` + `CreatedPR.Author` parts of #457 stay — independent, correct.

## Tests

- `TestIssueInflight_ClaimAndRelease` rewritten: a second claim with a *different* updated_at on an active issue now returns false; releasing the active claim lets any future updated_at re-claim. Independent issues remain independent.
- Removed the `autoPromoteFakeGH` fake and four `TestAutoPromote*` tests from `main_test.go` — they pinned the redundant gate. Nothing left to assert at this layer (transition behavior is owned by `TransitionIssueStage`, which has its own test suite).

## Test plan
- [x] `make test-docker` green.
- [ ] Manual: re-test transfer + label flow on a fresh issue; verify single triage comment + label moves to refinement.

Closes #458


vbuenog (approved):
## 🤖 Heimdallm AI Review

Reverts the over-broad #457 scope gate that was blocking handoffs and tightens issue triage single-flight to be keyed by issue_id only (so the bot's own comment, which bumps updated_at, cannot bypass the claim). Logic changes are coherent with the cited regression, the SQL is race-safe via atomic INSERT...SELECT WHERE NOT EXISTS, and test coverage is updated to reflect new semantics. Net effect is a targeted, well-justified fix.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:3509** — Reverting the autoPromoteGitHub seam to take *gh.Client directly removes unit-test coverage of autoPromoteAfterStage entirely (the four tests it enabled were deleted with no replacement). The new behavior is simpler but now has zero direct tests; correctness rests on integration paths and on the next worker's scope guard. Consider keeping the seam (even minimally) so the smart-default + label-transition paths remain unit-testable.
🟡 **daemon/internal/store/issue_inflight.go:30** — The previous INSERT OR IGNORE relied on a UNIQUE/PK constraint on (issue_id, updated_at). With single-flight now keyed on issue_id alone, that composite constraint is overconstrained relative to intent — harmless today, but if a future change broadens claims (e.g., multiple concurrent worker classes) the schema will silently block them. Worth a follow-up to align the schema (UNIQUE on issue_id) and drop the WHERE-NOT-EXISTS belt-and-braces, or document why both layers stay.
⚠️ **daemon/internal/store/issue_inflight.go:30** — ReleaseIssueTriageInFlight is presumably still keyed by (issue_id, updated_at). With claims now keyed by issue_id only, a release using a different updated_at than what was stored would leave the row in place and permanently block re-claims on that issue. The test only releases with the exact claimed updated_at — please confirm Release was updated to key off issue_id (or that callers are guaranteed to pass the claim's updated_at).
🟡 **daemon/cmd/heimdallm/main.go:3524** — Removing the post-pipeline refetch means TransitionIssueStage now uses the worker's possibly-stale `issue` (assignees, labels) for the label flip. For the label transition itself this is fine (assignees aren't read), but if downstream audit-comment dedup or label resolution relies on the current label set, a stale snapshot could re-add a label that triage already removed. Confirm TransitionIssueStage tolerates stale label state on `Issue`.

### Suggestions

- Add at least one regression test exercising autoPromoteAfterStage end-to-end without the scope gate — even a thin one against a fake client — to pin the #458 fix so a future refactor cannot silently re-introduce a scope check at the wrong layer.
- Verify ReleaseIssueTriageInFlight's WHERE clause matches the new single-flight key; if it still uses (issue_id, updated_at), realign it with the claim so a snapshot-bump between claim and release cannot strand the row.
- Consider a schema migration to drop updated_at from the uniqueness constraint (keep the column for diagnostics as the comment notes) so the contention model is expressed in one place instead of two.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

The PR fixes a real regression (#458) by removing the double-gated handoff check and tightening the triage claim to single-flight-per-issue. The claim query change is correct, but several surrounding artifacts — schema, release function, fakeStore, and pipeline.go comments — were not updated to match the new per-issue-only semantics, creating an inconsistency that risks leaked claims and misleading tests.

### Issues

⚠️ **daemon/internal/store/issue_inflight.go:42** — ReleaseIssueTriageInFlight still deletes by the composite key (issue_id, updated_at) but the claim is now single-flight per issue_id alone. If any caller ever passes a mismatched updated_at (e.g., a future code path or a subtle race), the DELETE matches zero rows, the claim silently leaks, and the issue is locked until the 30-min stale sweep. Release should be `DELETE … WHERE issue_id = ?` for consistency with the new claim semantics.
⚠️ **daemon/internal/store/store.go:143** — The table schema still has `PRIMARY KEY (issue_id, updated_at)`. The new INSERT…SELECT…WHERE NOT EXISTS enforces per-issue uniqueness at query level, but the underlying constraint is wider. Any code path that does a raw INSERT (e.g., the test helper InsertStaleIssueTriageInFlight) can still create multiple rows per issue, violating the intended invariant. A migration to `PRIMARY KEY (issue_id)` (or at least a UNIQUE constraint on issue_id) would enforce the new single-flight invariant at the database level.
⚠️ **daemon/internal/issues/pipeline_test.go:91** — The fakeStore.ClaimIssueTriageInFlight still keys on `issueID|updatedAt` (composite), so it simulates the OLD behavior, not the new single-flight-per-issue semantics. Tests using this fake won't catch regressions where the same issue is claimed with a different updated_at — they'll allow the second claim when the real store would now block it.
⚠️ **daemon/internal/issues/pipeline.go:332** — Stale comment: 'a genuine activity bump (new updated_at) produces a new claim key that bypasses the held one' is no longer true with single-flight-per-issue. This means a circuit-breaker-held issue is now locked until the 30-min stale sweep rather than being bypassable by new activity — an undocumented behavioral change that could delay legitimate retries.
🟡 **daemon/internal/issues/pipeline.go:232** — Comment at the issueStore interface says 'persistent (github_issue_id, updated_at) key' — no longer accurate after this PR changes the claim to per-issue-only.
🟡 **daemon/cmd/heimdallm/main.go:3510** — autoPromoteAfterStage now takes concrete *gh.Client instead of the autoPromoteGitHub interface, eliminating the testability seam. The four tests covering the removed assignee-scope gate were deleted with no replacement tests for the new unconditional-promotion path.

### Suggestions

- Update ReleaseIssueTriageInFlight to DELETE by issue_id only (dropping the updated_at condition), and migrate the schema to PRIMARY KEY (issue_id) with updated_at as a plain column. This makes all three layers — schema, claim, release — consistently enforce single-flight-per-issue.
- Update the fakeStore in pipeline_test.go to key claims on issueID alone, matching the new real-store behavior, so integration-level tests actually exercise the new dedup semantics.
- Update the circuit-breaker hold comment in pipeline.go (~line 332) to reflect that a held claim now blocks ALL retries for the issue until the stale sweep, and consider whether that extended lock-out is the desired behavior or whether the release function should also be called in the breaker path now that new-updated_at bypass is gone.
- Add at least one test for the new unconditional auto-promote behavior (no assignee scope check) to prevent accidental reintroduction of the gate.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: most prior findings addressed. Release semantics now correctly key off issue_id only; a schema-level UNIQUE index on issue_id was added (good defense-in-depth). The deleted autoPromoteAfterStage tests still have no replacement. One new minor concern: the schema-migration comment is slightly misleading.

### Issues

🟡 **daemon/cmd/heimdallm/main_test.go:54** — Still no unit-test coverage for autoPromoteAfterStage after the seam was reverted to *gh.Client. The four previous tests were removed and not replaced; correctness of the smart-default + label transition path now rests entirely on integration paths and the next-worker scope guard. Consider re-introducing a minimal seam (or a thin test harness around TransitionIssueStage's deps) so the auto-promote happy/skip paths remain unit-testable.
🟡 **daemon/internal/store/store.go:234** — The comment claims CREATE UNIQUE INDEX IF NOT EXISTS 'silently no-ops if existing rows already violate uniqueness', but SQLite will actually error in that case (IF NOT EXISTS only suppresses the 'already exists' case, not constraint failures). Combined with discarding db.Exec's error, an upgrade path with pre-existing in-flight duplicates would silently fail to create the index, leaving only the WHERE NOT EXISTS query as the guard. Acceptable in practice (the SQL still enforces single-flight), but the comment should be corrected and ideally the Exec error logged.
🟡 **daemon/cmd/heimdallm/main.go:3551** — Still unverified: TransitionIssueStage now receives the worker's possibly-stale `issue` (label set may not reflect post-pipeline mutations by triage). If TransitionIssueStage reads `Issue.Labels` for resolution or audit-dedup decisions, a stale snapshot could re-add a label triage already removed. Worth confirming this function only reads identity (repo+number) and not the label slice.
🟡 **daemon/internal/store/store.go:225** — The original composite PRIMARY KEY (issue_id, updated_at) is now strictly weaker than the new UNIQUE INDEX on (issue_id). It's harmless but creates two overlapping uniqueness constraints; a future migration that broadens claims (e.g., per-worker-class) would have to drop both. Consider documenting that the PK is now vestigial, or scheduling a follow-up to drop it.

### Suggestions

- Add at least one unit test covering autoPromoteAfterStage's enabled/disabled and stage-resolution branches, even without the prior fake-GH seam — a smaller helper that exercises the early-return paths would close the regression-detection gap left by the deleted tests.
- Log the error from db.Exec on the CREATE UNIQUE INDEX statement and fix the comment to reflect SQLite's actual behavior on constraint violations during index creation.
- Confirm TransitionIssueStage's read surface on Issue (identity-only vs label set) and document the assumption next to the call site to make the 'stale issue is fine' decision durable.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Previous medium-severity findings have been addressed well. The ReleaseIssueTriageInFlight fix, fakeStore alignment, and comment updates are all correct. Two new issues emerge from the UNIQUE index approach: silently swallowed errors on index creation failure and a misleading comment about that failure mode. The auto-promote test gap from the prior review remains.

### Issues

⚠️ **daemon/internal/store/store.go:235** — db.Exec for CREATE UNIQUE INDEX discards the error return. The comment claims 'The CREATE silently no-ops if existing rows already violate uniqueness' — this is incorrect for SQLite. If existing duplicate issue_id rows exist (from pre-fix runs), CREATE UNIQUE INDEX will return an error (not silently no-op), and that error is swallowed. The index will NOT be created, leaving the DB-level invariant unenforced until ClearStaleIssueTriageInFlight happens to clean the duplicates and a daemon restart re-runs Open(). At minimum capture and log the error so operators know the guardrail is absent; ideally clean duplicates before creating the index.
🟡 **daemon/cmd/heimdallm/main.go:3510** — autoPromoteAfterStage now takes concrete *gh.Client and the four tests covering the assignee-scope gate were deleted with no replacement tests for the new unconditional-promotion path. Even a basic test asserting that auto-promote calls TransitionIssueStage without the refetch/scope-check gate would pin the new intended behavior and prevent accidental reintroduction of the old logic.

### Suggestions

- In store.go Open(), capture the error from the CREATE UNIQUE INDEX exec. If it fails (likely due to existing duplicate rows), log a warning and run a cleanup query (e.g., DELETE duplicates keeping the latest started_at per issue_id) before retrying the index creation. This makes the upgrade path self-healing rather than relying on a timed sweep.
- Add at least one test for the unconditional auto-promote path to pin the new behavior — e.g., verify TransitionIssueStage is called regardless of assignee mismatch between the cached issue and scope config.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Revert of #457's assignee scope gate in autoPromoteAfterStage to fix the handoff regression, plus tightening of the issue triage in-flight claim from (issue_id, updated_at) to single-flight per issue_id. Logic changes are sound and well-documented, but the new UNIQUE INDEX migration has a misleading comment and swallows errors, and the deleted scope tests leave the regression they were pinning untested.

### Issues

⚠️ **daemon/internal/store/store.go:236** — The comment claims `CREATE UNIQUE INDEX IF NOT EXISTS ... silently no-ops if existing rows already violate uniqueness`, but SQLite actually returns an error (`UNIQUE constraint failed`) and skips index creation in that case. The error is then silently discarded because `db.Exec(...)` ignores its return values. For a daemon upgrading from the pre-#458 schema with multiple (issue_id, updated_at) rows for the same issue still resident, the unique index will fail to install and the schema-level invariant the PR advertises (`lifts the invariant from a query convention to a DB guarantee`) will silently not exist — leaving only the `INSERT ... WHERE NOT EXISTS` query as the guard. Fix: either (a) document accurately that operators must clear `issue_triage_in_flight` before upgrading and check the error, or (b) make Open() best-effort delete duplicates first, then create the index, and surface a clear log/error if it still fails. At minimum, capture and log the err from db.Exec so silent failures are observable.
⚠️ **daemon/cmd/heimdallm/main_test.go:56** — All four `TestAutoPromote*` tests added in #457 were deleted, but no replacement test pins the new contract. The new behavior — that auto-promote intentionally does NOT gate on assignee scope and relies on the next stage's worker entry (`issueTrackingWithAssigneeScope` + `issueStageStillCurrent`) — is precisely the contract the PR is restoring, and the failure mode of #457 (label transition double-gated, issue stranded) has no regression pin. Add at least one test that proves auto-promote transitions the label even when the issue has been reassigned out of scope (the handoff case), so a future change re-introducing the gate is caught immediately.

### Suggestions

- Consider an explicit force-release path for retry-marker comments rather than waiting up to 30 min for ClearStaleIssueTriageInFlight — the PR comment itself flags this as a likely follow-up; filing the issue alongside the merge would keep the operator-retry gap visible.
- In `ReleaseIssueTriageInFlight`, the `updatedAt` parameter is now purely vestigial. If you don't expect to reintroduce it, either remove it (with the call-site update) or rename to `_updatedAtForLogging` to make the intent obvious at the signature, since the `_ = updatedAt` inside the body is easy to miss.
- The schema comment `The composite PK above is now narrower than this index but harmless` is inverted — the composite PK (issue_id, updated_at) is broader (allows multiple rows per issue_id) than the new UNIQUE index on (issue_id). Worth fixing for future readers.

---

PR #473, opened 2026-05-12
fix(issues): dispatch re-review by current GitHub labels, not stored action (#462)

### LINKED ISSUE #462: bug: auto-promote does not update stored classification — re-review falls back to triage
## Problem

When an issue is auto-promoted through the pipeline (e.g., triage → refinement → development), the stored classification/action in the database is not updated to reflect the new stage. The issue retains its original classification (e.g., `review_only` from triage).

If the user then clicks "Re-review" on that issue, the daemon re-processes it using the **stored classification**, not the current stage — so it runs triage again instead of the stage the issue is actually in.

## Steps to Reproduce

1. Issue enters pipeline, gets triaged (classification: `review_only`, action: `triage`)
2. Auto-promote fires → issue moves to refinement (labels updated on GitHub)
3. Refinement completes → auto-promote fires → issue moves to development (labels updated)
4. User clicks "Re-review" on the issue
5. **Expected**: daemon runs the development stage (auto-implement)
6. **Actual**: daemon runs triage again, because the stored action is still `review_only`

## Root Cause

The stage promotion logic (`stage_transition.go`) updates GitHub labels correctly but does not update the stored `action_taken` / classification in the issues table or the latest issue review record. The re-review trigger reads from the store and dispatches based on the stored action.

## Expected Behavior

When an issue is promoted to a new stage:
1. Update the stored classification to match the new stage
2. Re-review should dispatch to the **current** stage, not the original one
3. Alternatively, re-review should read the issue's current GitHub labels to determine the correct stage rather than relying on the stored value

## Related

- Issue state machine (#395)
- Stage promotion logic: `stage_transition.go`

ivanmunozruiz:
## Summary

Fixes #462: clicking "Re-review" on an issue that was auto-promoted to refinement or develop used to re-run triage, because the callback reconstructed the issue from SQLite and hardcoded `Mode = review_only`. With this PR the callback refetches the issue from GitHub and dispatches to the worker matching the issue's *current* labels — the same classification path the fetcher uses.

## Changes

- **New helper** `dispatchIssueRunByCurrentMode(ctx, pub, repoIT, issue)` in `main.go`: classifies via `IssueTrackingConfig.Classify` and publishes to the matching NATS subject (triage / refinement / implement). Blocked / Ignore return errors without publishing. Takes a narrow `issueRunPublisher` interface (3 methods) as a testability seam; `*bus.NATSIssuePublisher` satisfies it.
- **`SetTriggerIssueReviewFn` callback** shrinks from ~110 lines of in-process pipeline wiring to ~35 lines: get issue from store, refetch from GitHub, resolve `repoIT`, hand off to the helper. The triage / refinement / implement workers already do the end-to-end work (acquire repo context, build opts, run pipeline, single-flight claim #458, auto-promote #456) — no duplication needed at the trigger site.

## What stays

- Endpoint path `POST /issues/{id}/review` — backwards compat with the Flutter app.
- `SetTriggerIssueRefineFn` (force-refine semantic) untouched.

## Tests (TDD)

`TestDispatchIssueRunByCurrentMode_*` covers every Classify branch:
- `_Triage`, `_Refinement`, `_Develop` — label → expected subject.
- `_BlockedReturnsErrorWithoutPublishing`, `_SkipLabelReturnsErrorWithoutPublishing` — gates surface a reason.
- `_DefaultsToTriageWhenConfigured`, `_DefaultsToIgnoreReturnsError` — fallback parity with the fetcher.

A fake `issueRunPublisher` records `(subject, repo, number, githubID)` so the test asserts both routing and argument propagation.

## Test plan
- [x] `make test-docker` green.
- [ ] Manual: trigger Re-review on an issue with `heimdallm-develop` label, confirm auto-implement worker handles it (not triage).
- [ ] Manual: trigger Re-review on an issue with `blocked` label, confirm the GUI gets the error event.

Closes #462


Muriano (approved):
## 🤖 Heimdallm AI Review

Refactors the manual re-review trigger to dispatch by current GitHub labels via NATS instead of running the pipeline in-process with stored classification. Fix is well-scoped, addresses the root cause of #462 (stale ActionTaken after auto-promote), and ships with focused unit tests covering all classification branches via a clean publisher seam. No confirmed defects found.

### Suggestions

- The 1-minute context timeout covers a GitHub GetIssue + NATS publish, which is reasonable, but consider making it configurable alongside other AI/HTTP timeouts so ops can tune it without a code change.
- Error messages returned from dispatchIssueRunByCurrentMode are surfaced verbatim to the SSE stream via publishIssueErr; confirm they don't leak internal repo paths or tokens (current strings look safe — repo/number only).
- The Blocked/Ignore branches return errors that the HTTP handler likely maps to a 5xx via the trigger callback; a 4xx-style signal (or a typed error) might be more accurate since these are user-state conditions, not failures.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Refactor replaces stored-ActionTaken re-review with a fresh GitHub fetch + label-driven dispatch via NATS. The change is well-scoped, removes ~80 lines of duplicated pipeline wiring, adds a clean seam (issueRunPublisher) and a thorough table of unit tests covering all classification branches, scope gating, nil issue, and blocked/skip cases. Logic mirrors the fetcher's classification path, error messages are operator-friendly, and the assignee scope gate prevents the silent-drop UX bug. No issues found.

### Suggestions

- Consider emitting an SSE event on successful dispatch (mirroring publishIssueErr on failure) so the GUI can transition the spinner state deterministically rather than relying on subsequent worker events.
- The 1-minute context timeout is tight but reasonable for a GetIssue + NATS publish; if GitHub rate-limit retries kick in this could trip — worth a metric/log if it ever fires in practice.

---

PR #476, opened 2026-05-12
feat(repoctx): per-execution git worktrees for concurrent pipelines (#461)

### LINKED ISSUE #461: feat: use git worktrees for concurrent pipeline executions
## Problem

When multiple pipeline stages run concurrently on the same repo (e.g., two auto-implement issues, two PR reviews, or a refinement + a development task), they share the same local directory. This causes:

- Git conflicts between branches being worked on simultaneously
- Dirty working tree errors when one execution's uncommitted changes interfere with another
- Non-fast-forward push failures when branches collide
- Race conditions on checkout/commit/push sequences
- Even read-only operations (PR reviews, refinements) conflict if concurrent: one checkout changes the branch for the other

## Proposal

Use **git worktrees** to isolate **every** pipeline execution in its own working directory, so concurrent operations on the same repo never interfere with each other.

Worktrees are always created **inside the repo directory** (managed by git under \`.git/worktrees/\`), following the same convention used by Claude Code and other AI agents. This applies to both auto-cloned repos and user-mapped \`local_dir\` repos — the worktree is a lightweight, ephemeral checkout that git manages natively.

### How It Works

1. The repo has one **main clone** (the existing \`local_dir\` or auto-cloned directory) — used as the shared object store
2. When **any** pipeline stage needs to work on a repo, it creates a **worktree** inside the repo:
   \`\`\`bash
   cd /path/to/org/repo
   git worktree add .worktrees/issue-42-develop -b heimdallm/issue-42
   # or for read-only:
   git worktree add .worktrees/pr-review-123 --detach HEAD
   \`\`\`
3. The agent receives the **worktree path** as its working directory — never the main working tree
4. After execution completes (success or failure), the worktree is cleaned up:
   \`\`\`bash
   git worktree remove .worktrees/issue-42-develop
   \`\`\`

### Directory Structure

\`\`\`
/path/to/org/repo/                   <- main clone (shared object store)
  .git/
    worktrees/                       <- git's internal worktree tracking
  .worktrees/                        <- Heimdallm's worktree checkout directories
    pr-review-123/                   <- worktree for PR review #123
    pr-review-456/                   <- worktree for PR review #456 (concurrent, same repo)
    issue-42-refinement/             <- worktree for issue #42 refinement
    issue-57-develop/                <- worktree for issue #57 auto-implement
\`\`\`

### Lifecycle

Every execution gets its own worktree, regardless of whether it writes or only reads:

| Stage | Worktree | Branch |
|-------|----------|--------|
| Triage | Yes (detached HEAD at main) | — |
| Refinement | Yes (detached HEAD at main or PR branch) | — |
| Development | Yes (dedicated branch) | \`heimdallm/issue-<number>\` |
| PR Review | Yes (detached HEAD at PR head SHA) | — |

This ensures two PR reviews on the same repo, or a PR review + a refinement, or any other combination can run simultaneously without interfering.

### Benefits

- **True concurrency** — any combination of pipeline stages can run in parallel on the same repo
- **No git conflicts** — each execution has its own working directory and branch/checkout
- **Clean failures** — if an execution fails, its worktree is removed without affecting others
- **Efficient disk usage** — worktrees share the \`.git\` object store, only the working files are duplicated
- **Convention-aligned** — same pattern used by Claude Code and other AI agents, all managed by git natively inside the repo
- **Deterministic state** — each agent starts with a known-clean working tree, no leftover artifacts from previous runs

## Config

\`\`\`toml
[github]
max_worktrees_per_repo = 5           # prevent unbounded disk usage; queue if at capacity
\`\`\`

The \`.worktrees/\` directory should be added to \`.gitignore\` by the auto-clone mechanism. For user-mapped repos, Heimdallm should check/warn if it's not gitignored.

## Cleanup

- Worktrees are removed after each execution (success or failure)
- On daemon startup, prune stale worktrees from previous runs: \`git worktree prune\`
- Respect \`retention.max_days\` for any worktrees that somehow survive
- Before creating a worktree, check \`max_worktrees_per_repo\` — if at capacity, queue the execution instead of failing

## Related

- Auto-clone mechanism (#394) — worktrees are created inside the cloned repo
- Issue state machine (#395) — multiple stages can run concurrently on different issues in the same repo
- Retry cap (#223 fix) — failed worktrees should be cleaned up, not retried indefinitely

ivanmunozruiz:
## Summary

Closes #461. Replaces the single-clone-per-repo lock with per-execution
git worktrees so multiple pipeline stages (PR review, triage,
refinement, develop) can run against the same repository in parallel.

- New `WorktreeToken` / `WorktreeBaseRef` / `Branch` / `Inspect` fields
  on `repoctx.Request`. Every Acquire that supplies a token now runs
  inside its own `<clone>/.worktrees/<token>/` directory.
- Two-tier concurrency model: short-lived critical-section mutex
  guards `git worktree add`/`remove`; long-lived N-slot semaphore
  caps concurrent worktrees per repo (`ai.max_worktrees_per_repo`,
  default 5). The cap is enforced via a buffered channel with FIFO
  queuing and `ctx.Done()` cancellation.
- `.worktrees/` is appended to the managed clone's `.gitignore` on
  bootstrap (idempotent) and explicitly excluded from `git clean`.
- `PruneStaleWorktrees` walks each clone at daemon startup and
  removes any worktree directory the in-memory manager does not own
  — covers crash recovery for daemons that exited mid-execution.
- All 7 `acquireRepoContext` callsites pass a stage-specific token
  (`pr-review-<n>`, `triage-<n>`, `develop-<n>`, `refinement-<n>`,
  `pr-tier2-<n>`).

Full design + step-by-step TDD log lives at
`docs/superpowers/plans/2026-05-12-issue-461-worktrees-concurrent-pipelines.md`.

## Test plan

- [x] `make test-docker` passes (repoctx + cmd/heimdallm + config).
- [x] New repoctx tests: token sanitisation, worktree add/remove,
  cap blocking, ctx cancel during queue, Inspect short-circuit,
  base ref + branch arg builder, gitignore bootstrap idempotency,
  prune of orphan worktrees.
- [x] New config tests for `ai.max_worktrees_per_repo` default + override.
- [ ] Manual smoke: trigger a PR review and an issue triage on the
  same repo simultaneously, confirm both run and produce distinct
  `<clone>/.worktrees/...` paths in logs.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -219,6 +219,21 @@ func main() {
 	exec := executor.New()
 	repoCtx := repoctx.NewManager()
 
+	// Sweep worktrees left behind by a previous daemon process. At
+	// startup the manager has no active worktrees, so every directory
+	// under `<clone>/.worktrees/` is by definition stale and safe to
+	// remove. Mirrors the in-flight DB sweeps above. (#461)
+	{
+		ctx := context.Background()
+		for _, cloneDir := range managedCloneDirs(cfg) {
+			if n, err := repoCtx.PruneStaleWorktreesUnder(ctx, cloneDir); err != nil {
+				slog.Warn("startup: prune stale worktrees", "dir", cloneDir, "err", err)
+			} else if n > 0 {
+				slog.Info("startup: pruned stale worktrees", "dir", cloneDir, "count", n)
+			}
+		}
+	}
+
 	// Load or create the per-daemon API token.  All mutating HTTP endpoints
 	// require this token in X-Heimdallm-Token (security issue #3).
 	apiToken, err := loadOrCreateAPIToken(dataDir())
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -703,7 +703,7 @@ func main() {
 		aiCfg := c.AIForRepo(pr.Repo)
 		localDirBase := c.GitHub.LocalDirBase
 		cfgMu.Unlock()
-		repoHandle, err := acquireRepoContext(ctx, repoCtx, pr.Repo, &aiCfg, localDirBase, token, repoctx.ModeRead)
+		repoHandle, err := acquireRepoContext(ctx, repoCtx, pr.Repo, &aiCfg, localDirBase, token, repoctx.ModeRead, wtTokenFor("pr-review", pr.Number), "", "")
 		if err != nil {
 			logRepoContextFallback("review-worker", pr.Repo, err)
 			aiCfg.LocalDir = ""
@@ -870,7 +870,7 @@ func main() {
 		}
 		ghIssue.Mode = config.IssueModeReviewOnly
 
-		repoHandle, err := acquireRepoContext(ctx, repoCtx, msg.Repo, &aiCfg, localDirBase, token, repoctx.ModeRead)
+		repoHandle, err := acquireRepoContext(ctx, repoCtx, msg.Repo, &aiCfg, localDirBase, token, repoctx.ModeRead, wtTokenFor("triage", msg.Number), "", "")
 		if err != nil {
 			logRepoContextFallback("triage-worker", msg.Repo, err)
 			aiCfg.LocalDir = ""
@@ -983,7 +983,7 @@ func main() {
 		}
 		ghIssue.Mode = config.IssueModeRefinement
 
-		opts, releaseRepoContext, err := buildRefinementRunOptions(ctx, s, repoCtx, msg.Repo, token, aiCfg, agentCfg, localDirBase, globalTimeout, false, "refinement-worker")
+		opts, releaseRepoContext, err := buildRefinementRunOptions(ctx, s, repoCtx, msg.Repo, msg.Number, token, aiCfg, agentCfg, localDirBase, globalTimeout, false, "refinement-worker")
 		if err != nil {
 			slog.Error("refinement-worker: prepare repo context failed",
 				"repo", msg.Repo, "number", msg.Number, "err", err)
@@ -1060,7 +1060,7 @@ func main() {
 		}
 		ghIssue.Mode = config.IssueModeDevelop
 
-		repoHandle, err := acquireRepoContext(ctx, repoCtx, msg.Repo, &aiCfg, localDirBase, token, repoctx.ModeWrite)
+		repoHandle, err := acquireRepoContext(ctx, repoCtx, msg.Repo, &aiCfg, localDirBase, token, repoctx.ModeWrite, wtTokenFor("develop", msg.Number), "", "")
 		if err != nil {
 			slog.Error("implement-worker: prepare repo context failed",
 				"repo", msg.Repo, "number", msg.Number, "err", err)
@@ -1515,7 +1515,7 @@ func main() {
 		aiCfg := cfg.AIForRepo(pr.Repo)
 		localDirBase := cfg.GitHub.LocalDirBase
 		cfgMu.Unlock()
-		repoHandle, err := acquireRepoContext(context.Background(), repoCtx, pr.Repo, &aiCfg, localDirBase, token, repoctx.ModeRead)
+		repoHandle, err := acquireRepoContext(context.Background(), repoCtx, pr.Repo, &aiCfg, localDirBase, token, repoctx.ModeRead, wtTokenFor("pr-review", pr.Number), "", "")
 		if err != nil {
 			logRepoContextFallback("trigger review", pr.Repo, err)
 			aiCfg.LocalDir = ""
@@ -1705,7 +1705,7 @@ func main() {
 		localDirBase := c.GitHub.LocalDirBase
 		globalTimeout := c.AI.ExecutionTimeout
 		cfgMu.Unlock()
-		opts, releaseRepoContext, err := buildRefinementRunOptions(ctx, s, repoCtx, iss.Repo, token, aiCfg, agentCfg, localDirBase, globalTimeout, force, "trigger issue refinement")
+		opts, releaseRepoContext, err := buildRefinementRunOptions(ctx, s, repoCtx, iss.Repo, iss.Number, token, aiCfg, agentCfg, localDirBase, globalTimeout, force, "trigger issue refinement")
 		if err != nil {
 			publishIssueErr(fmt.Sprintf("Failed to prepare repo context: %v", err))
 			return fmt.Errorf("trigger issue refinement: prepare repo context: %w", err)
@@ -2605,7 +2605,7 @@ func (a *tier2Adapter) ProcessPR(ctx context.Context, pr scheduler.Tier2PR) erro
 	aiCfg := c.AIForRepo(pr.Repo)
 	localDirBase := c.GitHub.LocalDirBase
 	a.cfgMu.Unlock()
-	repoHandle, err := acquireRepoContext(ctx, a.repoCtx, pr.Repo, &aiCfg, localDirBase, a.ghToken, repoctx.ModeRead)
+	repoHandle, err := acquireRepoContext(ctx, a.repoCtx, pr.Repo, &aiCfg, localDirBase, a.ghToken, repoctx.ModeRead, wtTokenFor("pr-tier2", pr.Number), "", "")
 	if err != nil {
 		logRepoContextFallback("tier2 PR", pr.Repo, err)
 		aiCfg.LocalDir = ""
@@ -2728,7 +2728,16 @@ func (a *tier2Adapter) ProcessRepo(ctx context.Context, repo string) (int, error
 		}
 		var releaseRepoContext func()
 		releaseOnReturn := true
-		repoHandle, err := acquireRepoContext(ctx, a.repoCtx, issue.Repo, &aiCfg, localDirBase, a.ghToken, mode)
+		wtPrefix := "stage"
+		switch issue.Mode {
+		case config.IssueModeDevelop:
+			wtPrefix = "develop"
+		case config.IssueModeRefinement:
+			wtPrefix = "refinement"
+		case config.IssueModeReviewOnly:
+			wtPrefix = "triage"
+		}
+		repoHandle, err := acquireRepoContext(ctx, a.repoCtx, issue.Repo, &aiCfg, localDirBase, a.ghToken, mode, wtTokenFor(wtPrefix, issue.Number), "", "")
 		defer func() {
 			if releaseOnReturn && repoHandle != nil {
 				repoHandle.Release()
@@ -3359,6 +3368,7 @@ func buildRefinementRunOptions(
 	s *store.Store,
 	manager *repoctx.Manager,
 	repo string,
+	issueNumber int,
 	token string,
 	aiCfg config.RepoAI,
 	agentCfg config.CLIAgentConfig,
@@ -3367,7 +3377,7 @@ func buildRefinementRunOptions(
 	force bool,
 	scope string,
 ) (issuepipeline.RunOptions, func(), error) {
-	repoHandle, err := acquireRepoContext(ctx, manager, repo, &aiCfg, localDirBase, token, repoctx.ModeRead)
+	repoHandle, err := acquireRepoContext(ctx, manager, repo, &aiCfg, localDirBase, token, repoctx.ModeRead, wtTokenFor("refinement", issueNumber), "", "")
 	if err != nil {
 		return issuepipeline.RunOptions{}, nil, err
 	}
@@ -3638,6 +3648,9 @@ func acquireRepoContext(
 	localDirBase []string,
 	token string,
 	mode repoctx.Mode,
+	wtToken string,
+	wtBaseRef string,
+	branch string,
 ) (*repoctx.Handle, error) {
 	if manager == nil {
 		return nil, fmt.Errorf("repoctx: nil manager")
@@ -3649,6 +3662,9 @@ func acquireRepoContext(
 		CloneDir:           aiCfg.CloneDir,
 		Token:              token,
 		Mode:               mode,
+		WorktreeToken:      wtToken,
+		WorktreeBaseRef:    wtBaseRef,
+		Branch:             branch,
 	})
 	if err != nil {
 		return nil, err
@@ -3658,6 +3674,14 @@ func acquireRepoContext(
 	// path.
 	aiCfg.LocalDir = h.Path()
 	return h, nil
+}
+
+// wtTokenFor produces a sanitisation-safe worktree token for a
+// pipeline stage. The prefix names the stage (`pr-review`, `triage`,
+// `develop`, `refinement`, `pr-tier2`) so operators can correlate
+// `<clone>/.worktrees/<token>/` with the running execution.
+func wtTokenFor(prefix string, n int) string {
+	return fmt.Sprintf("%s-%d", prefix, n)
 }
 
 func ensureRepoContextFullHistory(ctx context.Context, manager *repoctx.Manager, h *repoctx.Handle, token, scope, repo string) {
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -217,7 +217,9 @@ func main() {
 	notifier := notify.New()
 	ghClient := gh.NewClient(token)
 	exec := executor.New()
-	repoCtx := repoctx.NewManager()
+	repoCtx := repoctx.NewManagerWithOptions(repoctx.ManagerOptions{
+		MaxWorktreesPerRepo: cfg.AI.MaxWorktreesPerRepo,
+	})
 
 	// Sweep worktrees left behind by a previous daemon process. At
 	// startup the manager has no active worktrees, so every directory
--- a/daemon/cmd/heimdallm/main_test.go
+++ b/daemon/cmd/heimdallm/main_test.go
@@ -39,7 +39,7 @@ func newMemStore(t *testing.T) *store.Store {
 
 func TestAcquireRepoContextNilManagerIsError(t *testing.T) {
 	aiCfg := config.RepoAI{}
-	_, err := acquireRepoContext(context.Background(), nil, "org/repo", &aiCfg, nil, "secret", repoctx.ModeRead)
+	_, err := acquireRepoContext(context.Background(), nil, "org/repo", &aiCfg, nil, "secret", repoctx.ModeRead, "", "", "")
 	if err == nil || !strings.Contains(err.Error(), "nil manager") {
 		t.Fatalf("err = %v, want nil manager error", err)
 	}
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -375,6 +375,13 @@ type AIConfig struct {
 	AutoPromoteTriage     *bool  `toml:"auto_promote_triage,omitempty"`
 	AutoPromoteRefinement *bool  `toml:"auto_promote_refinement,omitempty"`
 
+	// MaxWorktreesPerRepo caps how many per-execution git worktrees the
+	// daemon will hold concurrently for a single repository (#461). A
+	// fresh value of 0 inherits the daemon default (5). Set higher if
+	// the repo has many independent stages running in parallel; lower
+	// if disk pressure dominates.
+	MaxWorktreesPerRepo int `toml:"max_worktrees_per_repo"`
+
 	// GeneratePRDescription enables LLM-generated PR titles and descriptions
 	// for auto_implement PRs. When true, after the implementation commit,
 	// a second LLM call generates a rich PR description from the diff.
@@ -852,6 +859,9 @@ func (c *Config) applyDefaults() {
 	}
 	if c.AI.RefinementTimeout == "" {
 		c.AI.RefinementTimeout = "30m"
+	}
+	if c.AI.MaxWorktreesPerRepo == 0 {
+		c.AI.MaxWorktreesPerRepo = 5
 	}
 	if c.ActivityLog.Enabled == nil {
 		v := true
--- a/daemon/internal/config/config_test.go
+++ b/daemon/internal/config/config_test.go
@@ -59,6 +59,23 @@ func TestApplyDefaults_PreservesExisting(t *testing.T) {
 	}
 }
 
+func TestApplyDefaults_MaxWorktreesPerRepo(t *testing.T) {
+	cfg := &Config{}
+	cfg.applyDefaults()
+	if cfg.AI.MaxWorktreesPerRepo != 5 {
+		t.Errorf("MaxWorktreesPerRepo = %d, want 5", cfg.AI.MaxWorktreesPerRepo)
+	}
+}
+
+func TestApplyDefaults_MaxWorktreesPerRepo_PreservesExisting(t *testing.T) {
+	cfg := &Config{}
+	cfg.AI.MaxWorktreesPerRepo = 12
+	cfg.applyDefaults()
+	if cfg.AI.MaxWorktreesPerRepo != 12 {
+		t.Errorf("MaxWorktreesPerRepo overwritten: %d", cfg.AI.MaxWorktreesPerRepo)
+	}
+}
+
 // ── applyEnvOverrides ────────────────────────────────────────────────────────
 
 func TestApplyDefaults_MaxConcurrentWorkers(t *testing.T) {
--- a/daemon/internal/repoctx/manager.go
+++ b/daemon/internal/repoctx/manager.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -201,6 +202,8 @@ func (m *Manager) Acquire(ctx context.Context, req Request) (*Handle, error) {
 	}()
 
 	if local := resolveLocal(req); local != "" {
+		// User-mapped paths are returned as-is for now; worktrees only
+		// run inside managed clones until the bootstrap step lands.
 		return &Handle{path: local, managed: false, release: release}, nil
 	}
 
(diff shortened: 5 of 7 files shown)


Muriano (changes requested):
## 🤖 Heimdallm AI Review

Solid two-tier locking design with semaphore-bounded concurrent worktrees, defensive token sanitisation, startup pruning, and good test coverage. Two notable concerns: (1) potential worktree-path collision when the same (stage, issue/PR number) is acquired concurrently from different entry points, and (2) interaction between ensureWorktreeGitignore and `git reset --hard` when the upstream repo already tracks .gitignore.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:706** — wtTokenFor() builds tokens as `<stage>-<number>`, but the same stage/number pair is used from multiple entry points that can fire concurrently (e.g. the review-worker at line 706 and the manual `trigger review` path at line 1518 both use `wtTokenFor("pr-review", pr.Number)`; tier2 ProcessRepo also reuses `develop`/`refinement`/`triage` prefixes that overlap with the per-worker goroutines). Two concurrent Acquire() calls with the same token resolve to the same `<clone>/.worktrees/<token>/` path; the second `git worktree add` will fail because the directory already exists, causing the second pipeline to error out (or, worse, both operate on the same dir if the second call somehow proceeds). Impact: spurious failures or data corruption under realistic schedules. Fix: include a uniqueness component (uuid, monotonic counter, or attempt id) in the token, or document and enforce that the per-(stage,number) lock is held for the entire pipeline lifetime.
⚠️ **daemon/internal/repoctx/manager.go:565** — ensureWorktreeGitignore writes/edits `<clone>/.gitignore`, but updateManagedClone runs `git reset --hard FETCH_HEAD` on every refresh. If upstream tracks its own .gitignore (common), the reset overwrites the daemon's edits; ensureWorktreeGitignore then rewrites the file, leaving a persistent dirty working tree relative to FETCH_HEAD. Any downstream code/AI tool that reads `git status` will see uncommitted changes. Impact: misleading git state for AI runs and noisy diffs. Fix: rely on `$GIT_DIR/info/exclude` (untracked-only, never collides with upstream) instead of `.gitignore`, or add the entry only when no upstream .gitignore exists.
🟡 **daemon/internal/repoctx/manager.go:370** — In acquireWorktree the deferred cleanup checks `err != nil` via closure capture, but several subsequent assignments use `:=` (e.g. `unlock, err := ...`, `cloneRoot, err := ...`). These are reassignments only because `err` is already in scope — but if the function is later edited to introduce a new outer name, the defer would silently stop firing. Impact: latent maintainability hazard for cap/lock leak on error. Fix: use named return values `(h *Handle, err error)` so the defer's reliance on the outer `err` is explicit and refactor-safe.
🟡 **daemon/internal/repoctx/manager.go:880** — PruneStaleWorktrees reads `m.active` under lock but only checks a single absolute path. Worktrees registered by `markActive` use `filepath.Join(cloneRoot, ".worktrees", token)`. If cloneRoot ever resolves via a symlink or non-canonical path on one side and the other side reads the directory through a different canonicalisation, the `live` lookup misses and an active worktree could be pruned. Impact: low-probability but destructive (concurrent `worktree remove --force` on a live execution). Fix: canonicalise paths (filepath.EvalSymlinks or filepath.Clean+Abs) on both insert and lookup.

### Suggestions

- Emit a metric/counter for cap waits and prune counts so operators can tune MaxWorktreesPerRepo from real data rather than guessing.
- Consider logging the chosen worktree path at INFO when acquiring, so operators can correlate `<clone>/.worktrees/<token>/` with the execution without grepping multiple log lines.
- Add a test that exercises two concurrent Acquire() calls with the *same* WorktreeToken to lock in the intended behaviour (collision error vs. queueing) — the current tests only cover distinct tokens.
- buildWorktreeAddArgs places `--detach` and `<ref>` together; document (or test) that `git worktree add <path> --detach <ref>` is the supported form across the git versions you target — some older gits prefer `--detach` after the ref.
- Startup sweep iterates `managedCloneDirs(cfg)` synchronously; for many configured clone roots this serialises behind git. Consider parallelising or bounding with a small errgroup.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Per-execution worktree support is a sensible concurrency win and the two-tier locking model (binary crit lock + counting cap semaphore) is well thought out. Validation, gitignore bootstrap, and startup pruning are all reasonable. A couple of cleanup edge cases in the release path and token collision semantics deserve attention before merge.

### Issues

⚠️ **daemon/internal/repoctx/manager.go:369** — In acquireWorktree's release closure, if acquireRepoLock(bgCtx) fails (e.g. lock held longer than gitTimeout), the code logs and skips both `worktree remove` AND `m.unmarkActive(wtPath)`. That leaves a stale on-disk worktree AND a permanent entry in m.active, which causes PruneStaleWorktrees to subsequently skip the orphan (it treats live entries as in-use). Move unmarkActive outside the lock branch so the in-memory tracker always clears, and consider best-effort os.RemoveAll on the path when the lock can't be obtained.
⚠️ **daemon/cmd/heimdallm/main.go:3683** — wtTokenFor returns `<prefix>-<n>` which is deterministic per (stage, number). The Request doc says concurrent operations on the same key may collide — and `git worktree add` will fail with `already exists` on the second concurrent caller. If two events for the same PR/issue are ever processed in parallel (retries before a release completes, scheduler hiccup, manual trigger overlapping a worker), the second acquireRepoContext will hard-fail rather than degrade. Either detect ENOTEMPTY/`already exists` and recover, or include a uniqueness suffix (e.g. nanoseconds / counter) when single-flighting isn't guaranteed by the caller.
🟡 **daemon/internal/repoctx/manager.go:638** — ensureWorktreeGitignore only recognises a small set of exact spellings (`.worktrees/`, `.worktrees`, `/.worktrees/`, `/.worktrees`). A pre-existing repo .gitignore containing e.g. `**/.worktrees/` or a glob pattern would not be detected and the function would append a duplicate line. Minor, but committing such a duplicate into the user's repo would be a visible artefact. Consider a more tolerant check or scope the write strictly to managed clones (it already is, but the failure mode is worth documenting).
🟡 **daemon/internal/repoctx/manager.go:554** — ensureWorktreeGitignore is invoked on every Acquire that hits ensureManagedClone (both the existing-marker branch and the fresh-clone branch), not just at bootstrap. Repeated ReadFile/WriteFile on every pipeline run is harmless but unnecessary; gating on a per-clone in-memory flag, or only running it once after marker write, would avoid the IO.
🟡 **daemon/internal/repoctx/manager.go:215** — Inspect mode is documented and implemented (acquireInspect) but no caller in this diff sets req.Inspect=true. If the /config/clones endpoint wiring lands in a follow-up that's fine, but worth confirming it's not dead code in the meantime.

### Suggestions

- In the release closure, always call m.unmarkActive(wtPath) (e.g. via defer) so the in-memory active set can't diverge from on-disk state when the bg lock acquisition fails.
- Make `git worktree add` resilient to a pre-existing path: detect `already exists` and either reuse the worktree (if it points at the right ref) or rm+retry. Pairs naturally with the deterministic token scheme.
- Consider a fast-path in PruneStaleWorktrees / startup sweep that also reconciles git's `worktree list` against the filesystem, so a directory git already knows about isn't double-removed (you already fall back to rm on git failure, which is good; the inverse — git registry entry without a dir — is what `git worktree prune` catches, so that's covered).
- Add a test for the release-path lock-failure scenario (simulate acquireRepoLock returning ctx.Err()) to lock in the unmarkActive-always behaviour once fixed.

---

ivanmunozruiz:
Addressed all four issues in commits 46df803..d33e745:

- **HIGH (token collision)**: manager now appends a monotonic seq to the path, so `pr-review-1234` from two concurrent entry points becomes `pr-review-1234.1` and `pr-review-1234.2`. Prefix stays operator-greppable. New test `TestAcquireSameTokenProducesDistinctWorktrees` pins the contract.
- **WARN (gitignore × reset --hard)**: switched to `.git/info/exclude`. Never tracked, never reverted by reset.
- **INFO (closure-captured err)**: `acquireWorktree` now uses named returns (`h *Handle, err error`) so the deferred rollback's reliance on the outer `err` is explicit and refactor-safe.
- **INFO (symlink canonicalisation)**: `markActive` / `unmarkActive` / `PruneStaleWorktrees` membership lookup runs through `EvalSymlinks` (with `Abs`/`Clean` fallbacks) so a clone reached via different path forms still resolves to the same key.

The suggestions list (metric counters, info-level acquire log, parallel startup sweep, git-version compat note on `worktree add --detach <ref>`) is good but out of scope for this PR — opening as follow-ups.

Muriano (approved):
## 🤖 Heimdallm AI Review

All four prior findings have been addressed: token collisions resolved via monotonic seq suffix, gitignore replaced with .git/info/exclude (never reverted by reset --hard), named returns make defer rollback explicit, and active-path lookup canonicalises via EvalSymlinks. New code is well-structured with a clear two-tier lock model (binary crit-section + counting semaphore), proper rollback on partial failure, startup sweep for stale worktrees, and reasonable test coverage including the collision-contract test. No new blocking issues.

### Suggestions

- Emit metrics (worktree_acquire_total, worktree_active_gauge, prune_count) so operators can alert on cap saturation and leak rates.
- Run the startup sweep across managedCloneDirs in parallel — a single slow repo currently blocks the rest of daemon startup.
- Document the minimum git version required for `worktree add --detach <ref>` (git ≥ 2.16) in the config docs near MaxWorktreesPerRepo.
- Add an info-level slog entry on successful Acquire including repo, token, seq, and path so on-call can correlate `<clone>/.worktrees/<token>.<seq>/` back to a pipeline run without DB lookup.
- Consider capping the wtSeq path component length or wrapping it — over a very long-lived daemon process this is unbounded (uint64 so practically fine, but cosmetic in logs).

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of changes since last review. The token-collision fix (monotonic seq), named returns, EvalSymlinks canonicalisation, and the switch to .git/info/exclude all look correct. However, one previously-flagged MEDIUM issue is NOT addressed: in acquireWorktree's release closure, m.unmarkActive(wtPath) is still nested inside the `else` branch of the lock-acquisition check, so a bgCtx lock-acquire failure leaves a permanent m.active entry that causes PruneStaleWorktrees to treat the orphan as live forever. A doc-vs-behaviour inconsistency was introduced by the seq fix.

### Issues

⚠️ **daemon/internal/repoctx/manager.go:384** — Release closure regression unaddressed: m.unmarkActive(wtPath) is still inside the `else` branch of `if err := m.acquireRepoLock(bgCtx, ...); err != nil`. If the lock cannot be acquired within gitTimeout, the on-disk worktree stays AND the in-memory active entry is never cleared — making the leak permanent (PruneStaleWorktrees will treat the canonical path as live and skip it). Move unmarkActive unconditionally (e.g. before the `if` or after the `if/else`), and consider a best-effort os.RemoveAll(wtPath) on lock failure so the disk side is also reclaimed.
🟡 **daemon/internal/repoctx/manager.go:65** — Docstring on Request.WorktreeToken still claims callers should derive a deterministic value so 'retries land on the same path'. With the new monotonic seq suffix this is no longer true — each Acquire produces a fresh `<token>.<seq>` directory. Update the doc to reflect that the token is now only an operator-readable correlation prefix, not a deterministic identity.
🟡 **daemon/internal/repoctx/manager.go:570** — ensureWorktreeExclude is still called on every Acquire that traverses ensureManagedClone (both the existing-marker and fresh-clone branches). Each call reads and potentially rewrites .git/info/exclude. Harmless but unnecessary on every pipeline run — gate behind a per-clone sync.Once or a one-shot at bootstrap.
🟡 **daemon/internal/repoctx/manager.go:240** — Inspect mode (acquireInspect) is implemented but the diff has no caller setting req.Inspect=true. If the /config/clones inspection wiring is landing in a follow-up PR, fine; worth confirming it isn't unused code in main.

### Suggestions

- Add a regression test that simulates the release closure failing to re-acquire the repo lock (e.g. inject a runner/lock stub) and asserts that the worktree path is no longer in m.active afterwards, so PruneStaleWorktreesUnder can reclaim it.
- Consider exposing PruneStaleWorktrees on a periodic ticker (not just startup) to defensively reclaim any future leaks if the unmarkActive bug recurs.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review confirms the four prior findings are addressed: monotonic seq disambiguates colliding tokens, `.git/info/exclude` survives reset --hard, named returns make the rollback refactor-safe, and active-path lookups canonicalise via EvalSymlinks. The two-tier lock model (crit lock + cap semaphore) is sound, release cleanup is robust to lock-timeout (cap is always released, filesystem fallback removes the dir), and startup prune covers leftover worktrees. No new high/medium issues spotted in the updated diff.

### Suggestions

- Consider emitting metrics (worktree acquire/release/prune counters, cap-wait duration) so operators can size MaxWorktreesPerRepo from data rather than guesswork.
- Startup prune sweeps clone dirs serially; parallelising via errgroup would shorten daemon boot when many repos are configured.
- `buildWorktreeAddArgs` uses `worktree add --detach <ref>` which requires git ≥ 2.21 — a one-line note in the manager doc or a startup git-version check would prevent silent regressions on older hosts.
- `wtTokenFor` accepts any int including negatives; if `issueNumber` is ever 0 or negative the resulting token (`develop-0`) is technically valid but semantically meaningless — consider asserting `n > 0` at the callsite or in the helper.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Author's named-return refactor and the unconditional defers in release() resolve the previous MEDIUM lock-failure leak (m.unmarkActive + best-effort os.RemoveAll now run regardless of lock acquisition). Three prior LOW findings are still unaddressed in the diff.

### Issues

🟡 **daemon/internal/repoctx/manager.go:65** — Stale docstring on Request.WorktreeToken: still claims callers should derive a deterministic value 'so that retries land on the same path'. With the monotonic seq suffix appended in acquireWorktree (`<token>.<seq>`), retries no longer land on the same path — the token is purely an operator-correlation prefix now. Update the comment to match the new contract; otherwise callers may reason about idempotency that the manager does not provide.
🟡 **daemon/internal/repoctx/manager.go:594** — ensureWorktreeExclude is still invoked on every Acquire via both branches of ensureManagedClone (existing-marker reuse and fresh-clone). Each call re-reads .git/info/exclude and may rewrite it. Harmless but wasteful per-pipeline-run I/O; gate behind a sync.Once keyed on the clone path, or perform it once during the manager bootstrap.
🟡 **daemon/internal/repoctx/manager.go:248** — Inspect mode (acquireInspect) is implemented and exercised, but no caller in this diff sets req.Inspect=true. If the /config/clones inspection wiring is intended to land in a follow-up PR, leave a TODO comment pointing to it; otherwise this is dead code paying maintenance cost in tests and review surface.

### Suggestions

- Consider adding a short note in NewManagerWithOptions godoc that MaxWorktreesPerRepo<=0 disables the cap, so callers reading the config side (default=5) and the manager side together don't have to derive the behaviour.
- The startup sweep in main.go uses managedCloneDirs(cfg) and calls PruneStaleWorktreesUnder serially per dir; with many configured clone roots this can delay daemon start. Consider parallelising once the operational pattern stabilises (already noted as out-of-scope follow-up).

---

Muriano (approved):
## 🤖 Heimdallm AI Review

All four prior findings (token collision, .gitignore vs reset --hard, closure-captured err, symlink canonicalisation) are addressed correctly in the new diff. The monotonic seq, .git/info/exclude entry, named-returns in acquireWorktree, and canonicalWorktreePath all look sound. The two-tier lock model (cap semaphore held across run, crit lock released after worktree add) is consistent, deferred rollbacks correctly cascade on error, and Release falls back to filesystem cleanup if the lock can't be reacquired so the cap slot is never leaked. Startup sweep is gated on the manager's empty active set, which is by construction safe at boot.

### Suggestions

- Follow-ups already acknowledged by author: metric counters for cap waits/worktree lifecycle, parallel startup sweep across clone dirs, and a git-version compatibility note for `worktree add --detach <ref>` (requires git ≥ 2.9).
- Consider asserting in tests that `PruneStaleWorktreesUnder` is a no-op when invoked against a clone base with no `.worktrees/` directories, to guard against future regressions in the os.ErrNotExist branch.
- `wtTokenFor` in main.go could live next to the other repoctx helpers (or in the repoctx package) so the prefix taxonomy is centralised — minor.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured concurrency upgrade from single-lock to two-tier (crit-section + counting-semaphore) worktree model. The core lifecycle logic is sound, cleanup/rollback paths are thorough, and the monotonic seq addresses the token-collision concern. A few race and idempotency edges remain.

### Issues

⚠️ **daemon/internal/repoctx/manager.go:975** — PruneStaleWorktrees checks m.active under m.mu but does NOT hold the per-repo critical-section lock. acquireWorktree calls markActive() AFTER `git worktree add` succeeds (while still holding crit), leaving a window where a concurrent prune could see the new directory on disk, find it absent from m.active, and remove it. Safe at startup (no concurrent acquires), but the doc advertises use for 'periodic sweeps that catch leaks', which could race with in-flight acquireWorktree calls.
⚠️ **daemon/internal/repoctx/manager.go:370** — Handle.Release() returned from acquireWorktree has no sync.Once guard. A double-call would re-acquire the repo lock, attempt `git worktree remove --force` on an already-removed path (logged as warning), and drain the cap channel a second time — potentially unblocking a waiter that should still be blocked. Adding sync.Once would make the release idempotent.
🟡 **daemon/internal/repoctx/manager.go:260** — acquireInspect calls ensureManagedClone, which will trigger a full `git clone` over the network if the clone doesn't already exist. For a read-only /config/clones inspection endpoint, silently cloning a repo is surprising and could be slow or fail unexpectedly. Consider returning an error when the clone does not already exist in the inspect path.
🟡 **daemon/internal/repoctx/manager.go:1020** — validateWorktreeToken enforces a character-class regex but has no maximum length check. On platforms with 255-byte filename limits, a very long token combined with the seq suffix could exceed NAME_MAX and cause `git worktree add` to fail with a cryptic OS error. A cap (e.g., 128 chars) would produce a clearer error.
🟡 **daemon/cmd/heimdallm/main.go:228** — The startup prune loop iterates managedCloneDirs(cfg) sequentially. For deployments with many clone roots, this serialises O(repos × worktrees) filesystem and git operations on a cold start. Consider parallelising with an errgroup (author noted this as a follow-up, flagging for tracking).
🟡 **daemon/cmd/heimdallm/main.go:3648** — acquireRepoContext grew three new positional string params (wtToken, wtBaseRef, branch) that are empty strings at most call sites. A struct or option pattern would be more self-documenting and less error-prone than positional empty strings, reducing the risk of argument-order mistakes.

### Suggestions

- Wrap Handle.Release in sync.Once to guarantee idempotency — double-release currently drains the cap channel twice, which could violate the semaphore invariant.
- Have PruneStaleWorktrees acquire the per-repo crit lock (or at minimum document that it must not be called concurrently with acquireWorktree for the same repo) to close the mark-before-prune race window.
- Consider bundling the three new acquireRepoContext params (wtToken, wtBaseRef, branch) into a small AcquireOpts struct to avoid the growing positional-empty-string smell at every call site.
- Add a length cap to validateWorktreeToken (e.g., 128 chars) so callers get a clear validation error instead of a platform-dependent filesystem error on long tokens.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Well-structured concurrency model with two-tier locking (binary critical-section + counting semaphore) and thorough input validation. The worktree lifecycle is mostly sound, but there is a lock-leak path in acquireClone inherited from pre-refactor code, and the release closure's error branch leaves the active-set inconsistent. Tests cover the happy path and pruning well.

### Issues

🔴 **daemon/internal/repoctx/manager.go:287** — Lock leak in acquireClone: if EnsureFullHistory fails, the function returns an error without calling `unlock()` (the release func is only attached to the returned Handle). The repo lock is permanently held for this repo key. The parallel `acquireWorktree` path fixed this with deferred rollbacks, but `acquireClone` still has the pre-existing bug. Add a `defer` that calls `unlock()` on error, matching the pattern in `acquireWorktree`.
⚠️ **daemon/internal/repoctx/manager.go:370** — In the release closure of acquireWorktree: when acquireRepoLock fails, `unmarkActive(wtPath)` is skipped but `releaseCap()` is still called. This leaves a phantom entry in `m.active` that prevents PruneStaleWorktrees from ever cleaning the orphaned worktree directory. Either call `unmarkActive` unconditionally (the active-set is guarded by m.mu, not the repo lock), or attempt the os.RemoveAll fallback outside the lock branch.
⚠️ **daemon/cmd/heimdallm/main.go:3648** — acquireRepoContext now takes 10 positional parameters (3 new string args), and most call sites pass empty strings for wtBaseRef and branch. This is fragile — a caller swapping the order of the three trailing strings gets no compile-time error. Consider an options struct or at minimum passing the Request fields through a smaller struct.
🟡 **daemon/internal/repoctx/manager.go:253** — acquireInspect calls ensureManagedClone (which may clone from GitHub) but returns a Handle with a no-op release. If the clone was freshly created solely for inspection and the inspect caller doesn't trigger any pipeline, the clone persists with no further cleanup path until an explicit purge API call. This is technically correct but worth a doc comment noting the intentional trade-off.
🟡 **daemon/internal/repoctx/manager.go:659** — ensureWorktreeExclude checks for several variants of the .worktrees pattern (with/without leading slash, with/without trailing slash) but misses the case where a user or tool may have added a comment or whitespace-only variant. More importantly, if the exclude file uses CRLF line endings (unlikely on Linux but possible on shared clones), strings.Split on '\n' would leave trailing '\r', causing the match to miss and duplicate entries to accumulate.
🟡 **daemon/internal/repoctx/manager.go:345** — acquireWorktree calls EnsureFullHistory on the clone root for ModeWrite before creating the worktree. If two concurrent ModeWrite worktree acquisitions race, one holds the crit lock and does the full-history fetch while the other waits. This is correct but the second caller re-runs EnsureFullHistory redundantly. Minor performance concern only.

### Suggestions

- Fix the lock leak in `acquireClone`: add `defer func() { if err != nil { unlock() } }()` after acquiring the lock (using named returns as in `acquireWorktree`), so `EnsureFullHistory` failures don't permanently deadlock the repo.
- In the release closure's lock-failure branch, call `m.unmarkActive(wtPath)` unconditionally and attempt `os.RemoveAll(wtPath)` as a best-effort fallback — the active-set is protected by `m.mu`, not the repo lock, so this is safe and prevents phantom entries.
- Replace the three trailing string parameters of `acquireRepoContext` (`wtToken`, `wtBaseRef`, `branch`) with a small struct to prevent silent argument transposition bugs.
- Add a metrics counter (or at least an info-level log) when cap semaphore wait begins, so operators can distinguish 'slow because waiting for cap' from 'slow because git is slow'.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-designed concurrency upgrade from a single exclusive lock to a two-tier model (binary crit-section + counting semaphore) with per-execution git worktrees. The rollback/cleanup logic is careful, path traversal validation is solid, and the symlink canonicalisation addresses a real-world class of bugs. A few configuration-layer gaps and one subtle semantic mismatch remain.

### Issues

⚠️ **daemon/internal/config/config.go:863** — Zero-value semantics mismatch: ManagerOptions documents that MaxWorktreesPerRepo <= 0 disables capping, but applyDefaults() silently overrides 0 → 5. A user who writes `max_worktrees_per_repo = 0` in TOML to disable the cap gets 5 instead. Use a *int pointer (nil = use default) or document that a negative value is the disable signal.
⚠️ **daemon/internal/config/config.go:380** — No upper-bound validation on MaxWorktreesPerRepo. The value becomes a buffered channel capacity (`make(chan struct{}, n)`). An unreasonably large value (e.g. 999999) wastes memory. Add a ceiling check in Config.validate() (e.g. max 100).
⚠️ **daemon/internal/repoctx/manager.go:370** — When resolveLocal() returns a user-mapped directory, acquireWorktree holds a cap slot without creating a worktree. This means user-mapped repos count against MaxWorktreesPerRepo even though no disk worktree is allocated, potentially starving real worktree callers when many local-dir pipelines run concurrently on the same repo key.
🟡 **daemon/internal/repoctx/manager.go:404** — In the release closure, if `git worktree remove --force` fails, the on-disk worktree directory may persist until the next daemon restart (PruneStaleWorktrees). Between restarts there is no mechanism to reclaim the space. Consider a deferred os.RemoveAll fallback after the git failure (same pattern as the lock-timeout branch above it), rather than only logging a warning.
🟡 **daemon/internal/repoctx/manager.go:678** — ensureWorktreeExclude checks trimmed lines for exact matches but doesn't skip comment lines (prefixed with '#'). A line like `#.worktrees/` (no space) would match `trimmed == ".worktrees/"` after TrimSpace, silently treating a commented-out entry as present and skipping the real append.
🟡 **daemon/cmd/heimdallm/main.go:228** — Startup prune loop uses a bare `context.Background()` with no timeout. If a managed clone has a very large `.worktrees/` directory or a broken git state, the prune could hang indefinitely and block daemon startup. Consider deriving the context from a startup-timeout or the global shutdown signal.
🟡 **daemon/cmd/heimdallm/main.go:3662** — acquireRepoContext now takes three trailing string parameters (wtToken, wtBaseRef, branch). Every call site passes "" for the last two. This is a readability hazard—callers must remember positional meaning for three unlabelled empty strings. Consider an options struct or at minimum collapsing the unused params into a future PR.

### Suggestions

- Use a *int (pointer) for MaxWorktreesPerRepo in the TOML config so nil means 'use default (5)' and an explicit 0 means 'disable', aligning the config layer with the ManagerOptions semantics.
- Add an os.RemoveAll fallback after `git worktree remove --force` failure in the release closure (lines ~400-406), matching the existing fallback pattern used when the lock-acquire times out. This prevents silent disk leaks between restarts.
- Replace the three trailing string params on acquireRepoContext (wtToken, wtBaseRef, branch) with a small WorktreeHint struct to prevent positional confusion, especially as more call sites are migrated.
- Add a startup-timeout context (e.g. 60s) for the prune loop rather than using context.Background(), so a corrupted clone can't block the daemon from starting.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of #476. The author addressed the high/medium concerns (path collision via seq suffix, gitignore→info/exclude switch, named-returns for rollback safety, symlink canonicalisation of active-map keys). Three previously-noted LOW items remain unchanged in the current diff: a stale docstring on Request.WorktreeToken, redundant per-Acquire I/O in ensureWorktreeExclude, and the still-uncalled Inspect mode. None are blockers.

### Issues

🟡 **daemon/internal/repoctx/manager.go:65** — Stale docstring on Request.WorktreeToken still claims callers should derive a deterministic value 'so that retries land on the same path'. With the monotonic seq suffix appended in acquireWorktree (`<token>.<seq>`), retries no longer land on the same path — the token is purely an operator-correlation prefix now. Update the comment to match the contract introduced by wtSeq; otherwise callers may reason about an idempotency property the manager does not provide.
🟡 **daemon/internal/repoctx/manager.go:594** — ensureWorktreeExclude is still invoked on every Acquire via both branches of ensureManagedClone (existing-marker reuse and fresh-clone). Each call re-reads .git/info/exclude and may rewrite it. Harmless but wasteful per-pipeline-run I/O — consider gating behind a sync.Once keyed on the clone path or running once during manager bootstrap / first ensureManagedClone success.
🟡 **daemon/internal/repoctx/manager.go:248** — Inspect mode (acquireInspect) is implemented and tested but still has no caller in this diff setting req.Inspect=true. If the /config/clones inspection wiring is intended to land in a follow-up PR, add a TODO comment referencing the tracking issue; otherwise this is dead code paying maintenance cost in tests and review surface.

### Suggestions

- Follow-ups the author already flagged as out-of-scope are reasonable (metric counters for cap waits and prune counts, info-level acquire log already added, parallel startup sweep across clone roots, and a note about `git worktree add --detach <ref>` requiring git ≥ 2.7 — worth confirming against the deployment baseline).
- Consider extending TestAcquireSameTokenProducesDistinctWorktrees with a parallel-goroutine variant to assert wtSeq.Add ordering is safe under contention (the atomic guarantees it, but a regression test is cheap insurance).
- PruneStaleWorktreesUnder enumerates managedClones sequentially; on hosts with many configured clone roots a goroutine-per-root pool would shorten daemon startup, but only worth doing if startup latency is actually observed as a problem.

---

PR #479, opened 2026-05-12
fix(server): strip dangerously_skip_perms from PATCH bodies (#477)

### LINKED ISSUE #477: 🔴 SECURITY: PATCH /config bypasses M-5 gate — allows setting dangerously_skip_perms
## Vuln: Security control bypass via PATCH /config

**Severity:** HIGH  
**Confidence:** 0.95  
**Category:** `privilege_escalation`

### Description

`PUT /config` (handled by `handlePutConfig`) validates `agent_configs` through `normalizeAgentConfigsForPut`, which explicitly strips `dangerously_skip_perms: true` with the comment _"MUST NOT be exposed via HTTP API"_ (security gate M-5).

However, `PATCH /config` (handled by `handlePatchConfig` at `handlers.go:632`) accepts arbitrary JSON, deep-merges it into the live TOML config via `config.DeepMerge`, and **does NOT call `normalizeAgentConfigsForPut` or any equivalent field-level denylist**. The only post-merge validation is `config.ValidateMap`, which calls `cfg.Validate()` — but `Validate()` only checks `PermissionMode` and `ApprovalMode` for agent configs. `DangerouslySkipPerms` is never validated.

### Exploit Scenario

An attacker with a valid API token sends:

```http
PATCH /api/config HTTP/1.1
X-Heimdallm-Token: <valid-token>
Content-Type: application/json

{"ai": {"agents": {"claude": {"dangerously_skip_perms": true}}}}
```

The daemon deep-merges this into `config.toml`, persists it, and reloads. On the next AI invocation, `buildRunOpts` (main.go:379) reads `agentCfg.DangerouslySkipPerms = true` and passes `--dangerously-skip-permissions` to the claude CLI (executor.go:482-484), bypassing the AI permission sandbox entirely.

### Confirmed Exploit Path

1. `handlePatchConfig` decodes arbitrary JSON → `config.DeepMerge` merges without filtering
2. `ValidateMap` → `cfg.Validate()` does NOT check `DangerouslySkipPerms` — accepts silently
3. `AtomicWriteTOML` persists `dangerously_skip_perms = true` to disk
4. `reloadFn()` loads new config → `CLIAgentConfig.DangerouslySkipPerms = true`
5. `buildRunOpts` propagates to `ExecOptions.DangerouslySkipPerms`
6. `buildArgs` appends `--dangerously-skip-permissions` unconditionally

### Affected Files

- `daemon/internal/server/handlers.go` — `handlePatchConfig` (line 632), `normalizeAgentConfigsForPut` (line 1746)
- `daemon/internal/config/writer.go` — `DeepMerge` (line 29), `ValidateMap` (line 139)
- `daemon/internal/config/config.go` — `Validate()` missing check (line 1060), `CLIAgentConfig` struct (line 337)
- `daemon/cmd/heimdallm/main.go` — `buildRunOpts` (line 379)
- `daemon/internal/executor/executor.go` — `buildArgs` (line 482)

### Recommended Fix

In `handlePatchConfig`, after `config.DeepMerge` and before writing to TOML, apply the same normalization as PUT. Options:

1. Extract a shared helper `stripDangerousAgentFlags(m map[string]any)` called in both PUT and PATCH handlers
2. Add an explicit post-merge scrubbing step that traverses `merged["ai"]["agents"][*]` and deletes any `dangerously_skip_perms` key
3. Add `DangerouslySkipPerms` validation in `cfg.Validate()` that rejects `true` (defense-in-depth)

ivanmunozruiz:
## Summary

Closes #477. PUT /config validated `ai.agents.<cli>.dangerously_skip_perms` and 400'd if present (security gate M-5). PATCH /config took the deep-merge path through `patchTOML` with no equivalent denylist, so an attacker with a valid API token could persist the flag via:

```http
PATCH /config
{"ai":{"agents":{"claude":{"dangerously_skip_perms":true}}}}
```

The next AI invocation then ran with `--dangerously-skip-permissions`, bypassing the permission sandbox.

## Fix

`patchTOML` now invokes `stripDangerousAgentFlags(m)` between `mutateFn` and `ValidateMap`, so every PATCH endpoint (global, per-repo, per-org) inherits the gate without each handler having to remember. The scrubber walks `m["ai"]["agents"][*]` and deletes any `dangerously_skip_perms` key.

Direct edits to `config.toml` still flip the flag — the filesystem is the trusted boundary; HTTP is not. This matches the M-5 design.

## Test plan

- [x] `make test-docker` passes.
- [x] `TestHandlePatchConfig_StripsDangerouslySkipPerms` — single agent: flag stripped, legal fields land.
- [x] `TestHandlePatchConfig_StripsDangerouslySkipPermsAcrossAllAgents` — every agent (claude, gemini, codex, opencode).
- [x] Existing PUT tests still 400 on the same key.

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"13c5928f-827a-408a-a8e7-e9c5f58f4739","pid":70360,"procStart":"Mon May 11 11:55:26 2026","acquiredAt":1778573897562}
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,1 +1,0 @@
-{"sessionId":"13c5928f-827a-408a-a8e7-e9c5f58f4739","pid":70360,"procStart":"Mon May 11 11:55:26 2026","acquiredAt":1778573897562}
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ docker/config/
 daemon/heimdallm
 cli/heimdallm-cli
 .superpowers/
+.claude/scheduled_tasks.lock
--- a/daemon/internal/server/handlers.go
+++ b/daemon/internal/server/handlers.go
@@ -236,8 +236,13 @@ func (srv *Server) patchTOML(mutateFn func(m map[string]any) error) (map[string]
 	// merged map is validated and persisted. PUT rejects the key with
 	// a 400 via normalizeAgentConfigsForPut, but PATCH deep-merges
 	// arbitrary JSON, so the gate has to be enforced here regardless
-	// of which handler invoked patchTOML.
-	stripDangerousAgentFlags(m)
+	// of which handler invoked patchTOML. Audit-log non-zero strips so
+	// operators can detect bypass attempts even though they're now
+	// neutralized.
+	if stripped := stripDangerousAgentFlags(m); stripped > 0 {
+		slog.Warn("config: PATCH attempted to set dangerously_skip_perms; stripped (M-5 gate)",
+			"count", stripped)
+	}
 	if err := config.ValidateMap(m); err != nil {
 		return nil, err
 	}
@@ -1740,31 +1745,79 @@ func tailLines(f *os.File, n int) []string {
 	return lines
 }
 
+// dangerousAgentFlagKey names the security-gated flag that grants
+// `--dangerously-skip-permissions` to the AI CLI. Both
+// normalizeAgentConfigsForPut and stripDangerousAgentFlags reference
+// this constant so the M-5 denylist has a single source of truth.
+const dangerousAgentFlagKey = "dangerously_skip_perms"
+
 // stripDangerousAgentFlags removes the HTTP-forbidden
-// dangerously_skip_perms flag from every agent under ai.agents in a
-// merged config map. The flag is still settable via direct edits to
-// config.toml (security gate M-5: only filesystem-trusted inputs can
-// grant the AI permission-sandbox bypass), but never via the HTTP API.
+// dangerously_skip_perms flag from every agent map reachable in the
+// merged config: top-level ai.agents.<cli>, per-repo
+// ai.repos.<repo>.agents.<cli>, and per-org ai.orgs.<org>.agents.<cli>.
+// The flag is still settable via direct edits to config.toml
+// (security gate M-5: only filesystem-trusted inputs can grant the
+// permission-sandbox bypass), but never via the HTTP API.
+//
+// Returns the number of entries stripped so callers can audit-log
+// bypass attempts. Key comparison is case-insensitive because the
+// downstream koanf/mapstructure decoder is case-insensitive when
+// mapping into CLIAgentConfig.DangerouslySkipPerms — an exact-case
+// match would leave Dangerously_Skip_Perms / DANGEROUSLY_SKIP_PERMS
+// shaped payloads as a live bypass.
 //
 // Invoked from patchTOML so every PATCH endpoint (global, per-repo,
 // per-org) inherits the gate without each handler having to remember
 // to call it.
-func stripDangerousAgentFlags(m map[string]any) {
+func stripDangerousAgentFlags(m map[string]any) int {
+	if m == nil {
+		return 0
+	}
+	stripped := 0
+	scrub := func(agents map[string]any) {
+		for _, raw := range agents {
+			inner, ok := raw.(map[string]any)
+			if !ok {
+				continue
+			}
+			for k := range inner {
+				if strings.EqualFold(k, dangerousAgentFlagKey) {
+					delete(inner, k)
+					stripped++
+				}
+			}
+		}
+	}
 	ai, ok := m["ai"].(map[string]any)
 	if !ok {
-		return
+		return 0
 	}
-	agents, ok := ai["agents"].(map[string]any)
-	if !ok {
-		return
+	if agents, ok := ai["agents"].(map[string]any); ok {
+		scrub(agents)
 	}
-	for _, raw := range agents {
-		inner, ok := raw.(map[string]any)
-		if !ok {
-			continue
+	if repos, ok := ai["repos"].(map[string]any); ok {
+		for _, raw := range repos {
+			repo, ok := raw.(map[string]any)
+			if !ok {
+				continue
+			}
+			if agents, ok := repo["agents"].(map[string]any); ok {
+				scrub(agents)
+			}
 		}
-		delete(inner, "dangerously_skip_perms")
 	}
+	if orgs, ok := ai["orgs"].(map[string]any); ok {
+		for _, raw := range orgs {
+			org, ok := raw.(map[string]any)
+			if !ok {
+				continue
+			}
+			if agents, ok := org["agents"].(map[string]any); ok {
+				scrub(agents)
+			}
+		}
+	}
+	return stripped
 }
 
 // normalizeAgentConfigsForPut validates the agent_configs payload from PUT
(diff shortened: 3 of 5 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Targeted security fix that strips dangerously_skip_perms from PATCH bodies before validation/persist. Implementation is minimal, well-commented, and covered by two tests (single + all four agents). Defense-in-depth placement inside patchTOML ensures all PATCH callers inherit the gate.

### Suggestions

- Consider adding a negative test asserting the flag is also stripped when the agent map is nested under a per-repo/per-org path, if those PATCH routes ever serialize ai.agents under a different key — the current stripDangerousAgentFlags only scans top-level m["ai"]["agents"], so the comment claim 'every PATCH endpoint (global, per-repo, per-org) inherits the gate' should be verified against the actual schema for repo/org overrides.
- Minor: stripDangerousAgentFlags silently no-ops on type mismatches; an info/debug log when a non-map agent entry is encountered could aid future debugging, though it's not required.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Solid, focused security fix that strips dangerously_skip_perms from PATCH bodies in a centralized location (patchTOML). The placement after mutateFn/deep-merge but before ValidateMap+persist is correct. Two issues worth considering: a potential case-sensitivity bypass and an unrelated .gitignore change bundled in.

### Issues

⚠️ **daemon/internal/server/handlers.go:1762** — Case-sensitivity bypass risk: `delete(inner, "dangerously_skip_perms")` only matches the exact key. Go's JSON decoder into map[string]any preserves the original casing of JSON keys, so a payload like `"Dangerously_Skip_Perms": true` or `"DANGEROUSLY_SKIP_PERMS": true` would survive the strip. Whether this is exploitable depends on whether config.ValidateMap rejects unknown keys and whether downstream config readers do case-insensitive matching. If either is lenient, this is a real bypass. Consider a case-insensitive key scan (ranging over inner's keys and comparing via strings.EqualFold) instead of a single delete.
⚠️ **daemon/internal/server/handlers_test.go:1462** — No test coverage for case-variant bypass attempts (e.g. Dangerously_Skip_Perms, DANGEROUSLY_SKIP_PERMS). Adding a test case that sends alternate casings would verify the gate is robust regardless of downstream validation strictness.
🟡 **.gitignore:25** — Unrelated change (.claude/scheduled_tasks.lock) bundled into a security fix PR. Not harmful, but pollutes the security-relevant commit history and makes targeted reverts harder.
🟡 **daemon/internal/server/handlers_test.go:1525** — Tests only exercise the global /config PATCH endpoint. The patchTOML gate is shared, but per-repo and per-org PATCH endpoints may wrap the mutateFn differently or pre-process the body. At least one integration test hitting a per-repo PATCH path with the flag would strengthen confidence that M-5 holds across all entry points.

### Suggestions

- Replace the single `delete(inner, "dangerously_skip_perms")` with a case-insensitive scan: range over `inner` keys and delete any that match via `strings.EqualFold(k, "dangerously_skip_perms")`. This closes the case-variant bypass regardless of how strict ValidateMap is.
- Add a test case that sends mixed-case variants of the flag (e.g. `Dangerously_Skip_Perms`, `DANGEROUSLY_SKIP_PERMS`) and asserts they are also stripped.
- Split the .gitignore change into its own commit or PR to keep the security fix's diff minimal and auditable.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Targeted defense-in-depth fix that strips dangerously_skip_perms in the central patchTOML path so every PATCH endpoint (global/per-repo/per-org) inherits the M-5 gate. Implementation is minimal and well-placed, tests cover single-agent and all-agents cases, and the asymmetry with PUT (which 400s) is explained. Low risk.

### Issues

🟡 **daemon/internal/server/handlers.go:1762** — Silently stripping a security-relevant key diverges from PUT's 400-on-presence behavior. Returning 200 with the flag dropped means a buggy/malicious client gets no signal that its intent was rejected — consider at least logging at warn level when the key is stripped so operators can detect bypass attempts in audit logs.
🟡 **.gitignore:25** — Unrelated change (.claude/scheduled_tasks.lock) bundled into a security fix PR. Minor, but mixing unrelated hygiene with security commits muddies the audit trail for M-5.

### Suggestions

- Consider emitting a structured audit log entry when stripDangerousAgentFlags actually removes a key, so the security team can detect HTTP-side attempts to set dangerously_skip_perms even though they're now neutralized.
- Add a negative test asserting that a PATCH body without ai.agents (or with ai.agents.<name> as a non-map, e.g. null/string) doesn't panic — the type-assertion guards already handle this, but a regression test would lock the behavior in.
- Consider centralizing the forbidden-key list (currently 'dangerously_skip_perms' is referenced both here and in normalizeAgentConfigsForPut) as a single const/var to avoid drift if more flags get gated under M-5 later.
- Move the .gitignore line to a separate commit/PR to keep the security fix focused.

---

ivanmunozruiz:
Hardening tras los reviews (d0766e1):

- **Case-insensitive scrub** (sergiotejon ⚠️): `strings.EqualFold` en lugar de `delete` exacto. La asimetría era explotable porque mapstructure/koanf hace case-insensitive matching al deserializar a `CLIAgentConfig.DangerouslySkipPerms`. Aplicado también en `normalizeAgentConfigsForPut` para alinear PUT/PATCH.
- **Nested paths** (Muriano + sergiotejon): scrubber ahora recorre `ai.repos.<r>.agents.<cli>` y `ai.orgs.<o>.agents.<cli>` además del top-level. El schema actual no lee esos paths como CLIAgentConfig pero la defense-in-depth blinda contra drift futuro.
- **Audit log** (vbuenog): `stripDangerousAgentFlags` ahora retorna un count y `patchTOML` emite `slog.Warn` cuando count > 0.
- **DRY** (vbuenog): const `dangerousAgentFlagKey` compartido entre PUT y PATCH.

Tests añadidos:
- `TestHandlePatchConfig_StripsDangerouslySkipPermsCaseInsensitive` (3 spellings)
- `TestHandlePatchConfig_StripsDangerouslySkipPermsInRepoOverride`
- `TestHandlePatchConfig_StripsDangerouslySkipPermsInOrgOverride`
- `TestStripDangerousAgentFlags_HandlesUnexpectedShapes` (package-internal, robust contra nil/scalar)
- `TestStripDangerousAgentFlags_ReportsCount`

El `.gitignore` ya estaba en commit separado del fix de seguridad.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review tras hardening: el PR aborda correctamente el bypass M-5 en PATCH mediante un scrubber case-insensitive aplicado en patchTOML, cubre top-level y rutas anidadas (repos/orgs), comparte una constante con el validador PUT, emite audit log con count, y añade cobertura de tests amplia (case variants, repo/org overrides, shapes inesperadas). Las observaciones previas (case-insensitive, nested paths, DRY, audit log) están resueltas. No se detectan issues nuevos.

### Suggestions

- Considerar exponer una métrica contadora (no solo slog.Warn) para alertar sobre intentos repetidos de bypass M-5 en producción.
- El audit log no incluye request ID ni identidad del caller; añadir contexto del request facilitaría correlacionar intentos sospechosos.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Author addressed both prior findings: warn-level audit log now fires on non-zero strips, and the case-insensitive scrub closes a real koanf/mapstructure bypass vector. Scrubber correctly walks top-level, per-repo, and per-org agent maps, with strong test coverage (case variants, nested paths, panic-safety on malformed shapes). DRY constant shared between PUT and PATCH. The .gitignore nit was acknowledged as a separate commit and is benign. Implementation looks solid; no new issues.

### Suggestions

- Consider also scrubbing keys whose JSON value is not a bool (e.g. string "true") — koanf/mapstructure WeaklyTypedInput would coerce them. Current EqualFold delete catches the key regardless of value type, so this is already covered, but worth a comment or a dedicated test asserting `{"dangerously_skip_perms": "true"}` is stripped.
- Minor: the slog.Warn message could include a request-scoped identifier (remote addr or auth principal) so audit trails can attribute bypass attempts to a client, not just count them.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previous findings addressed. The case-insensitive scrub via strings.EqualFold closes the bypass risk, nested paths (repos/orgs) are now walked, audit logging is in place, and comprehensive tests cover each vector. The security gate is solid. Only a cosmetic nit remains.

### Issues

🟡 **daemon/internal/server/handlers_test.go:1555** — TestHandlePatchConfig_StripsDangerouslySkipPermsInRepoOverride and the org-override variant use silent early `return` when the post-PATCH structure is missing. Since the PATCH body contains only the stripped key (no legal sibling field), the entire sub-tree may be pruned during TOML serialization, causing the test to pass vacuously without actually proving the strip ran. Adding a legal sibling field (like permission_mode) and asserting it lands — as the global test already does — would make these tests truly affirmative rather than relying on absence-of-structure.
🟡 **.gitignore:25** — Unrelated .gitignore addition still bundled in the PR diff. Author acknowledged it is in a separate commit, so this is cosmetic only.

### Suggestions

- In the repo-override and org-override integration tests, send a legal sibling field (e.g. permission_mode) alongside dangerously_skip_perms and assert the sibling survives — this turns the test from 'structure absent so flag absent' into 'flag stripped but sibling preserved', matching the stronger pattern used in TestHandlePatchConfig_StripsDangerouslySkipPerms.

---

PR #485, opened 2026-05-12
fix(discovery): include [ai.repos.*] in Tier 2 merge (closes #281)

### LINKED ISSUE #281: bug: repos drop out of issue tracking when they have no active PRs
## Problem

Tier 1 (discovery) rebuilds the monitored repo list every 15 minutes by searching for PRs where the user is a reviewer. Tier 2 receives this list and iterates it for both PR review AND issue processing.

When a repo has no active PRs (all merged/closed, or the user was never a reviewer), Tier 1 doesn't discover it. The repo disappears from the Tier 2 list, and issue tracking (triage + auto-implement) stops working for that repo — even if:
- The repo is explicitly configured in \`[ai.repos."org/repo"]\` with issue tracking enabled
- The global issue tracking config is enabled and the repo was previously discovered

## Example

\`freepik-company/fc-vertexai-api\` has issue #141 with label \`heimdallm-triage\`, global IT is enabled with \`review_only_labels = ["heimdallm-triage"]\`, but the issue never enters the triage pipeline because the repo has no open PRs where the bot user is a reviewer.

## Expected Behavior

Once a repo is discovered (or manually configured), it should remain in the monitored list permanently for issue tracking purposes. The repo list for Tier 2 should be the **union** of:
1. Repos discovered by Tier 1 (PR-based discovery)
2. Repos explicitly configured in \`[ai.repos.*]\`
3. Repos in the \`repositories\` config list

A repo should only be removed from monitoring by explicit user action (remove from config or dismiss).

## Root Cause

\`scheduler/tier2.go\` receives repos exclusively from Tier 1's discovery channel. There is no merge with configured repos or previously-discovered repos persisted in the store.

Muriano:
## Summary
- Add a `configured` parameter to `MergeRepos` so repos with an explicit `[ai.repos.*]` entry are part of the union AND exempt from the `NonMonitored` blacklist
- Stop `upsertDiscoveredRepos` from sending an explicitly-configured repo to `NonMonitored` when `AutoEnablePRForDiscovery` is off
- Wire the configured list through `Tier1Config` so the NATS-published list matches the Tier 2 in-process merge

## Why
Without this, a repo wired up under `[ai.repos."org/repo"]` but missing from the `repositories` list (or absent from topic discovery because it has no active PRs) silently dropped out of Tier 2 issue polling. The companion bug — `upsertDiscoveredRepos` writing such a repo to `NonMonitored` — made the previous `MergeRepos` blacklist it, which is exactly the regression in #281's status check comment.

## Test plan
- [x] Updated existing `MergeRepos` tests for the new signature
- [x] New `TestMergeRepos_ConfiguredIncludedInUnion`, `TestMergeRepos_ConfiguredOverridesNonMonitored`, `TestMergeRepos_ConfiguredDeduplicatesWithStatic`, `TestMergeRepos_OnlyConfigured`
- [x] New `TestUpsertDiscoveredRepos_ExplicitAIConfigOverridesAutoEnableOff` verifies both halves of the fix (no demotion, lands in `Repositories`)
- [x] `go test ./...` passes for the whole daemon module
- [x] `go vet ./...` clean
- [x] `go build ./cmd/heimdallm` produces a working binary

Closes #281


vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid bug fix for #281 that correctly threads explicit [ai.repos.*] entries through the discovery merge so wired-up repos with no active PRs still receive issue polling. The MergeRepos signature change is mechanical and well-tested, the exemption semantics (configured wins over NonMonitored) are clearly documented, and the new tests cover the regression scenarios. A few minor concerns around symmetry with upsertDiscoveredRepos and a potential redundant call site.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:2243** — upsertDiscoveredRepos now appends an explicit-AI-config repo to c.GitHub.Repositories unconditionally — but if that repo is already present in NonMonitored from a prior tick, it will now appear in BOTH lists. MergeRepos handles this correctly via the exempt set, but other code reading c.GitHub.NonMonitored directly may behave inconsistently. Consider also removing it from NonMonitored when promoting.
⚠️ **daemon/cmd/heimdallm/main.go:615** — aiRepoKeys(cfg) is called once when building Tier1Config but the config can mutate at runtime (upsertDiscoveredRepos modifies cfg.AI? — actually only GitHub fields). Verify that AI.Repos is immutable post-startup, otherwise Tier1Config.ConfiguredRepos becomes stale. If config can be hot-reloaded, this should be a closure/getter rather than a snapshot.
🟡 **daemon/internal/discovery/discovery.go:138** — Doc comment says 'Order: static first, then configured, then discovered' which is correct, but the previous behavior callers relied on (static-then-discovered) is now subtly different — a repo present only in configured will now appear before discovered repos. Worth confirming no downstream consumer assumes the old positional ordering (e.g., for prioritization in rate-limited polling).
🟡 **daemon/cmd/heimdallm/main.go:2188** — The aiRepoKeys helper filters empty strings but does not trim whitespace, while other parts of the codebase (e.g., monitoredRepoSet at line 3978) do strings.TrimSpace on repo names. Minor inconsistency — a TOML key like ' org/repo ' would pass through unnormalized.

### Suggestions

- Consider whether upsertDiscoveredRepos should also strip the repo from NonMonitored when it has explicit [ai.repos.*] config, to keep config state consistent (not just the merged output).
- If config can be reloaded at runtime, change Tier1Config.ConfiguredRepos to a func()[]string or compute it inside the tier1 loop rather than capturing once at startup — symmetric with how StaticRepos/NonMonitored are read live elsewhere.
- Add a test covering the case where a repo appears in static, configured, AND nonMonitored simultaneously — to lock in that exempt-wins semantics persist across all three input sources.
- Consider normalizing repo keys (TrimSpace, lowercase?) consistently in aiRepoKeys to match monitoredRepoSet's handling.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Solid fix for #281 — configured repos correctly seed MergeRepos and override the NonMonitored blacklist. Tests are thorough. One compilation blocker: missing `sort` import.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:2192** — Missing `"sort"` import: `aiRepoKeys` calls `sort.Strings(out)` but the diff does not add `"sort"` to the import block, and it is not present on the base branch. This is a compilation error.
🟡 **daemon/cmd/heimdallm/main.go:2243** — The `hasExplicitAIConfig` check in `upsertDiscoveredRepos` is placed after the `discovery_orgs` org filter (line ~1932). A repo with explicit `[ai.repos.*]` config but whose org is outside `discovery_orgs` will be silently skipped before the new logic fires. This is partially mitigated because `MergeRepos` includes configured repos directly via the `configured` parameter, but the inconsistency means such a repo won't land in `Repositories` via upsert — it only appears in the merged output. If any downstream code inspects `cfg.GitHub.Repositories` directly rather than calling `MergeRepos`, the repo would be missing.

### Suggestions

- Add `"sort"` to the stdlib import block in `daemon/cmd/heimdallm/main.go` (between `"os/signal"` and `"strconv"`).
- Consider adding a standalone unit test for `aiRepoKeys` covering nil config, empty map, and map with empty-string keys to lock down the helper's behavior independently of callers.
- Consider moving the `hasExplicitAIConfig` check in `upsertDiscoveredRepos` above the org filter, or at least document why it's acceptable for the org filter to block explicitly-configured repos during the upsert phase.

---

Muriano:
## Review responses (commit `8e68b93`)

Rebased onto latest `main` (resolved conflicts in `main.go` and `discovery_test.go`, keeping both the new `MergeRepos(ConfiguredRepos)` signature and main's `FilterArchived` logic / tests). Addressing each finding:

### 🔴 Missing `sort` import — resolved
`"sort"` is now present in the import block (it landed on `main` and is also used by this branch). `go build ./...` and `go test ./internal/discovery/... ./cmd/...` both pass after the merge. No longer a compilation blocker.

### 🟡 Normalize repo keys in `aiRepoKeys` (TrimSpace) — fixed
Applied `strings.TrimSpace` in `aiRepoKeys` so a TOML key like ` org/repo ` is normalized identically to `monitoredRepoSet`. Commit `8e68b93`.

### 🟡 `upsertDiscoveredRepos` double-listing / promote-removal — out of scope → follow-up
Valid, but the #281 fix only requires configured repos to appear in the merged output (which `MergeRepos` exempt-set handling already guarantees). Tracked in #527.

### ⚠️ `hasExplicitAIConfig` placed after org filter — out of scope → follow-up
Mitigated today because `MergeRepos` includes configured repos directly. Tracked in #527.

### ⚠️ `Tier1Config.ConfiguredRepos` snapshot vs hot-reload — out of scope → follow-up
Config is not hot-reloaded today (the snapshot matches current behavior). If hot-reload is added this must become a getter; tracked in #527.

### 🟡 Ordering change (configured before discovered) — by design
The doc comment documents the intended order (static → configured → discovered). No downstream consumer relies on positional ordering for prioritization; polling tiers are membership-based, not order-based.

### Suggestion: extra test (static+configured+nonMonitored) — follow-up
Tracked in #527.

Re-requesting review.

PR #486, opened 2026-05-13
feat(tier2): parallel PR/issue polling + per-repo concurrency (#481)

### LINKED ISSUE #481: Tier 2 processTick is sequential — PR/issue polling block each other and repo discovery races with review start
## Problem

`runTier2.processTick()` (`main.go:2113-2187`) runs PR polling and issue polling **sequentially in a single goroutine**. This causes two related problems:

### 1. Cross-tier blocking within Tier 2

The tick processes PRs first, then issues repo-by-repo in a sequential `for range`:

```go
processTick := func() {
    // PR processing (all repos)
    prs := adapter.FetchPRsToReview()
    for _, pr := range prs { prPublisher.PublishPRReview(...) }

    // Issue processing (sequential, repo by repo)
    for _, repo := range currentRepos {
        adapter.ProcessRepo(ctx, repo)  // blocks until done
    }
}
```

If you have 20 repos and issue processing for repo #5 is slow (rate limit, large response), repos #6-#20 wait. Meanwhile, the next PR poll tick is delayed because the whole `processTick` hasn't returned.

This means a slow issue cycle in repoC delays PR detection in repoA — the tiers are not independent despite having separate NATS workers downstream.

### 2. Repo discovery races with review start (UX issue)

`FetchPRsToReview()` calls `upsertDiscoveredRepos()` which adds a new repo to config and emits `repo_discovered` via SSE. Immediately after, it publishes `SubjPRReview` to NATS. The `ReviewWorker` picks up the message and starts the review **in parallel**.

The Flutter UI can receive the `review_started` SSE event before (or simultaneously with) `repo_discovered`, causing the user to see a PR being reviewed for a repo that doesn't yet appear in the Repositories list. This creates a "something is broken" perception.

**Sequence today:**
```
FetchPRsToReview() → upsertDiscoveredRepos() → SSE: repo_discovered
                   → PublishPRReview()        → ReviewWorker picks up → SSE: review_started
                                                 ↑ can arrive before repo_discovered in the UI
```

## Possible directions (to be refined)

1. **Split PR and issue polling into independent goroutines** with their own tickers, so they don't block each other
2. **Parallelize per-repo issue processing** within each tick (bounded by a semaphore), replacing the sequential `for range`
3. **Guarantee SSE ordering** — ensure `repo_discovered` is delivered and acknowledged before publishing the PR review to NATS
4. **Per-repo concurrency** — give each repo its own polling goroutine so repoA's slow issue processing never delays repoB

## Impact

- Users see PRs for repos not yet visible in the Repositories dashboard
- Slow issue processing in one repo delays PR detection across all repos
- With many repos, the sequential loop can push effective poll intervals well beyond the configured value

ivanmunozruiz:
## Summary

Closes #481. Three independent improvements to the Tier 2 polling loop, grouped under one PR because they share state and tests.

### Before

```go
processTick := func() {
    fetchPRs()                                 // synchronous
    for _, repo := range currentRepos {        // sequential, repo by repo
        adapter.ProcessRepo(ctx, repo)         // can block for seconds
    }
}
```

A slow issue cycle on `repoC` delayed PR detection across every other repo. With 20 repos the effective poll interval crept well past the configured value.

### After

```go
// Two independent tickers — neither tier can block the other.
go func() {
    ticker := time.NewTicker(interval); defer ticker.Stop()
    if coldStart { prTick() }
    for { select { case <-ctx.Done(): return; case <-ticker.C: prTick() } }
}()
go func() {
    ticker := time.NewTicker(interval); defer ticker.Stop()
    if coldStart { issueTick() }
    for { select { case <-ctx.Done(): return; case <-ticker.C: issueTick() } }
}()

// inside issueTick → runIssueTier:
processReposInParallel(ctx, repos, cap, func(ctx, repo) (int, error) {
    return adapter.ProcessRepo(ctx, repo)
})
```

Three changes:

1. **PR and issue tiers run on independent tickers.** Each tier has its own goroutine with its own `time.NewTicker(interval)`. `time.Ticker` drops redundant ticks when the previous run is still in flight, so each tier serialises against itself without ever blocking the other. An issue cycle that exceeds `poll_interval` no longer delays PR detection.
2. **Per-repo issue processing fans out** through `processReposInParallel`, a bounded worker-pool helper. Cap comes from a new config knob `ai.tier2_repo_concurrency` (default 5) read live on every tick. Cancellation-safe: a labelled break + ctx-aware semaphore acquire stops dispatching the moment `ctx.Done()` fires.
3. **Newly-discovered repos defer their first review by one tick** so the Flutter UI receives `repo_discovered` before `review_started`. The filter lives inside `FetchPRsToReview`; the `scheduler.Tier2PRFetcher` interface signature stays unchanged. The next tick (after `upsertDiscoveredRepos` no longer reports the repo as "added") picks the PR up normally.

The GitHub API rate limiter (`scheduler.RateLimiter`) still throttles network usage. The new cap is wall-clock parallelism only.

Full design + step-by-step TDD log: `docs/superpowers/plans/2026-05-13-issue-481-tier2-parallel-polling.md`.

## Test plan

- [x] `make test-docker` passes (entire suite).
- [x] `processReposInParallel` unit tests: all-repos-processed, cap respected, runs concurrently (wall-clock check), continues after per-repo failure, ctx cancellation, **cancel stops scheduling (catches the labelled-break / ctx-aware semaphore bug)**, nil/empty no-op, non-positive cap falls back.
- [x] `TestTier2Adapter_FetchPRsToReview_DefersNewlyDiscoveredRepo`: first cycle publishes `repo_discovered` and withholds the PR; second cycle returns the PR normally.
- [x] `TestApplyDefaults_Tier2RepoConcurrency` + preserves-existing.
- [x] Existing tests updated where they implicitly relied on the immediate-review behaviour (`main_inflight_plumb_test.go`).

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -2052,6 +2052,12 @@ func bridgeDiscovery(ctx context.Context, conn *nats.Conn, out chan<- []string) 
 // A non-positive concurrency falls back to a sane default so a
 // misconfiguration cannot deadlock the daemon. Nil or empty repo
 // list is a no-op.
+//
+// Cancellation: both the per-repo scheduling loop and the semaphore
+// acquire observe ctx.Done so a daemon shutdown does not get stuck
+// waiting for the last free slot when workers are still draining.
+// Already-running workers receive ctx through workFn and are
+// responsible for their own short-circuit.
 func processReposInParallel(
 	ctx context.Context,
 	repos []string,
@@ -2071,16 +2077,17 @@ func processReposInParallel(
 	sem := make(chan struct{}, concurrency)
 	var total int64
 	var wg sync.WaitGroup
+schedule:
 	for _, repo := range repos {
-		// Honour cancellation before queuing more work — keeps shutdown
-		// snappy when the user pulls the plug mid-tick.
 		select {
 		case <-ctx.Done():
-			break
-		default:
+			// Stop dispatching new work; the previous `break` only
+			// exited the select, leaving the loop to keep queuing.
+			break schedule
+		case sem <- struct{}{}:
+			// Acquired a slot. Fall through to spawn the worker.
 		}
 		wg.Add(1)
-		sem <- struct{}{}
 		go func(repo string) {
 			defer wg.Done()
 			defer func() { <-sem }()
@@ -2142,8 +2149,11 @@ func runTier2(
 		return
 	}
 
-	ticker := time.NewTicker(interval)
-	defer ticker.Stop()
+	snapshotRepos := func() []string {
+		mu.Lock()
+		defer mu.Unlock()
+		return append([]string(nil), repos...)
+	}
 
 	// runPRTier publishes review requests for every reviewable PR.
 	// Independent of the issue tier so a slow GitHub Search call does
@@ -2219,41 +2229,64 @@ func runTier2(
 		})
 	}
 
-	processTick := func() {
-		mu.Lock()
-		currentRepos := append([]string(nil), repos...)
-		mu.Unlock()
-
+	// PR and issue tiers run on independent tickers so a slow issue
+	// cycle (taking longer than `interval`) cannot delay the next PR
+	// poll. Each tier serialises against itself: a Ticker drops
+	// extra ticks when its previous run is still in flight, so we
+	// never spawn two concurrent issue cycles. The PublishPending
+	// retry runs at the tail of each PR tick — the typical pending
+	// publishes come from PR fetch, and retries are idempotent.
+	prTick := func() {
+		currentRepos := snapshotRepos()
 		if len(currentRepos) == 0 {
 			return
 		}
-
-		// PR and issue tiers run in parallel so a slow issue cycle
-		// cannot delay PR detection (and vice versa). They still
-		// share the tick; the next tick waits for both to finish so
-		// we never let one tier accumulate unbounded backlog.
-		var wg sync.WaitGroup
-		wg.Add(2)
-		go func() { defer wg.Done(); runPRTier(currentRepos) }()
-		go func() { defer wg.Done(); runIssueTier(currentRepos) }()
-		wg.Wait()
-
-		// Retry pending publishes
+		runPRTier(currentRepos)
 		adapter.PublishPending()
 	}
-
-	if coldStart {
-		processTick()
-	}
-
-	for {
-		select {
-		case <-ctx.Done():
+	issueTick := func() {
+		currentRepos := snapshotRepos()
+		if len(currentRepos) == 0 {
 			return
-		case <-ticker.C:
-			processTick()
 		}
+		runIssueTier(currentRepos)
 	}
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		if coldStart {
+			prTick()
+		}
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				prTick()
+			}
+		}
+	}()
+	go func() {
+		defer wg.Done()
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		if coldStart {
+			issueTick()
+		}
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				issueTick()
+			}
+		}
+	}()
+	wg.Wait()
 }
 
 // upsertDiscoveredRepos adds PRs' repos to the monitored (or non-monitored)
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -2045,13 +2045,14 @@ func bridgeDiscovery(ctx context.Context, conn *nats.Conn, out chan<- []string) 
 
 // processReposInParallel runs workFn for every repo in repos with at
 // most `concurrency` calls in flight at once. Returns the sum of the
-// integer return values from the workers (intended for issue-count
-// aggregation). Errors are logged via slog and never abort siblings —
-// a single repo's failure must not freeze the whole tick. (#481)
+// integer return values from the workers. workFn errors are silently
+// counted as zero — the helper has no domain context to log them
+// usefully; callers wrap workFn to log per-repo failures in the
+// vocabulary that fits their tier. (#481)
 //
-// A non-positive concurrency falls back to a sane default so a
-// misconfiguration cannot deadlock the daemon. Nil or empty repo
-// list is a no-op.
+// A non-positive concurrency falls back to
+// config.DefaultTier2RepoConcurrency so a misconfiguration cannot
+// deadlock the daemon. Nil or empty repo list is a no-op.
 //
 // Cancellation: both the per-repo scheduling loop and the semaphore
 // acquire observe ctx.Done so a daemon shutdown does not get stuck
@@ -2068,7 +2069,7 @@ func processReposInParallel(
 		return 0
 	}
 	if concurrency <= 0 {
-		concurrency = 5
+		concurrency = config.DefaultTier2RepoConcurrency
 	}
 	if concurrency > len(repos) {
 		concurrency = len(repos)
@@ -2081,8 +2082,8 @@ schedule:
 	for _, repo := range repos {
 		select {
 		case <-ctx.Done():
-			// Stop dispatching new work; the previous `break` only
-			// exited the select, leaving the loop to keep queuing.
+			// Stop dispatching new work; a plain `break` here would
+			// only exit the select, leaving the loop to keep queuing.
 			break schedule
 		case sem <- struct{}{}:
 			// Acquired a slot. Fall through to spawn the worker.
@@ -2093,7 +2094,6 @@ schedule:
 			defer func() { <-sem }()
 			n, err := workFn(ctx, repo)
 			if err != nil {
-				slog.Error("tier2: per-repo work failed", "repo", repo, "err", err)
 				return
 			}
 			atomic.AddInt64(&total, int64(n))
@@ -2208,18 +2208,19 @@ func runTier2(
 		} else if n > 0 {
 			slog.Info("tier2: promoted issues", "count", n)
 		}
-		cap := 5
+		concurrency := config.DefaultTier2RepoConcurrency
 		if repoConcurrencyFn != nil {
 			if v := repoConcurrencyFn(); v > 0 {
-				cap = v
+				concurrency = v
 			}
 		}
-		issueCount = processReposInParallel(ctx, currentRepos, cap, func(ctx context.Context, repo string) (int, error) {
+		issueCount = processReposInParallel(ctx, currentRepos, concurrency, func(ctx context.Context, repo string) (int, error) {
 			if err := limiter.Acquire(ctx, scheduler.TierRepo); err != nil {
 				return 0, err
 			}
 			n, err := adapter.ProcessRepo(ctx, repo)
 			if err != nil {
+				slog.Error("tier2: issue processing", "repo", repo, "err", err)
 				return 0, err
 			}
 			if n > 0 {
@@ -2233,15 +2234,19 @@ func runTier2(
 	// cycle (taking longer than `interval`) cannot delay the next PR
 	// poll. Each tier serialises against itself: a Ticker drops
 	// extra ticks when its previous run is still in flight, so we
-	// never spawn two concurrent issue cycles. The PublishPending
-	// retry runs at the tail of each PR tick — the typical pending
-	// publishes come from PR fetch, and retries are idempotent.
+	// never spawn two concurrent issue cycles.
 	prTick := func() {
 		currentRepos := snapshotRepos()
 		if len(currentRepos) == 0 {
 			return
 		}
 		runPRTier(currentRepos)
+		// PublishPending lives in the PR tick on purpose: pending
+		// publishes are almost exclusively PR-review NATS messages
+		// from runPRTier, and retries are idempotent. If we ever
+		// route issue-side NATS publishes through the same queue,
+		// add a sibling call inside issueTick rather than removing
+		// this one.
 		adapter.PublishPending()
 	}
 	issueTick := func() {
@@ -2252,40 +2257,26 @@ func runTier2(
 		runIssueTier(currentRepos)
 	}
 
+	runTickerLoop := func(tick func()) {
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		if coldStart {
+			tick()
+		}
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				tick()
+			}
+		}
+	}
+
 	var wg sync.WaitGroup
 	wg.Add(2)
-	go func() {
-		defer wg.Done()
-		ticker := time.NewTicker(interval)
-		defer ticker.Stop()
-		if coldStart {
-			prTick()
-		}
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			case <-ticker.C:
-				prTick()
-			}
-		}
-	}()
-	go func() {
-		defer wg.Done()
-		ticker := time.NewTicker(interval)
-		defer ticker.Stop()
-		if coldStart {
-			issueTick()
-		}
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			case <-ticker.C:
-				issueTick()
-			}
-		}
-	}()
+	go func() { defer wg.Done(); runTickerLoop(prTick) }()
+	go func() { defer wg.Done(); runTickerLoop(issueTick) }()
 	wg.Wait()
 }
 
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -2220,7 +2220,14 @@ func runTier2(
 			}
 			n, err := adapter.ProcessRepo(ctx, repo)
 			if err != nil {
-				slog.Error("tier2: issue processing", "repo", repo, "err", err)
+				// Demote to debug during graceful shutdown — a
+				// cancelled ctx is expected behaviour, not a fault
+				// worth paging an operator.
+				if ctx.Err() != nil {
+					slog.Debug("tier2: issue processing cancelled", "repo", repo, "err", err)
+				} else {
+					slog.Error("tier2: issue processing", "repo", repo, "err", err)
+				}
 				return 0, err
 			}
 			if n > 0 {
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -17,6 +17,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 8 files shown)


ivanmunozruiz:
Addressed both findings (bcad5a9):

1. **PR/issue independence — tickers separados**: el shape anterior (PR+issue en paralelo dentro del mismo tick con `wg.Wait()`) preservaba el bloqueo si issue tardaba más que `poll_interval`. Ahora cada tier corre en su propia goroutine con su propio `time.NewTicker(interval)`. Los tickers descartan ticks si la corrida anterior sigue en vuelo, así cada tier se serializa contra sí mismo sin nunca bloquear al otro. PR detection ya no espera a issue.
2. **Bug de cancelación en el worker pool**: `break` dentro del `select` solo rompía el select, no el for, y `sem <- struct{}{}` no observaba ctx → podía quedar bloqueado o seguir encolando. Refactor a `select { case <-ctx.Done(): break schedule; case sem <- struct{}{}: }` con labeled break. Nuevo `TestProcessReposInParallel_CancelStopsScheduling` reproduce el bug (cap=2 con 20 repos, cancel tras 2 workers; el helper no debe procesar los 20).

Plan doc actualizado para reflejar la nueva arquitectura de tickers independientes.

`PublishPending` lo dejé colgando del PR tick (es idempotente y los pending típicos vienen de PR fetch).

Muriano (approved):
## 🤖 Heimdallm AI Review

Solid refactor: independent PR/issue tickers, bounded per-repo parallelism, and a one-tick deferral for newly-discovered repos to fix SSE ordering. Code is well-tested (concurrency cap, cancellation, no-op cases) and well-documented. Minor concerns around shadowing the builtin `cap`, an `addedThisTick` computed from a stale `added` reference, and a potential issue with `EmitPollingCompleted` firing on rate-limiter Acquire failure via defer (changing prior behavior). No blocking bugs found.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:2211** — Local variable `cap` shadows the builtin `cap()`. While harmless here (no slice/chan capacity calls follow in scope), it is a lint smell and a footgun for future edits in this function. Rename to `concurrency` or `repoCap` for clarity.
🟡 **daemon/cmd/heimdallm/main.go:2566** — `addedThisTick` is built from `added` before the diff shows where `added` comes from in this function — the truncated diff hides the call site of `upsertDiscoveredRepos`. If `added` is computed under the cfg lock but the map is consumed after Unlock, ensure there's no aliasing back into config slices. Recommend verifying `added` is a freshly-allocated slice (not a subslice of cfg.GitHub.Repositories) so the deferral set cannot mutate under concurrent config reload.
🟡 **daemon/cmd/heimdallm/main.go:2160** — `defer sse.EmitPollingCompleted(...)` now fires even when `limiter.Acquire` fails — previously the code emitted completion only on the early-return rate-limit path explicitly. The behavior is arguably better (always emits a completion to match started), but downstream consumers that previously never saw a `polling_completed` event for rate-limited ticks will now see one with count=0. Confirm SSE consumers (Flutter UI) tolerate this. Not a bug, but a contract change worth verifying.
🟡 **daemon/cmd/heimdallm/main_tier2_defer_test.go:125** — `testEventDataContains`/`stringContains` reimplement `strings.Contains`. Replace with `strings.Contains` from stdlib to avoid maintaining a hand-rolled substring search.

### Suggestions

- Extract the duplicated ticker goroutine boilerplate (lines ~2261-2289) into a small helper `runTicker(ctx, interval, coldStart, tick func())` — both goroutines are identical except for the tick function.
- Consider documenting in the config guide that setting `tier2_repo_concurrency` higher than the GitHub rate-limit budget will not increase throughput, only burst pressure on the limiter.
- The default fallback of 5 is hardcoded in two places (config defaults and `processReposInParallel`). Consider exposing a single `const DefaultTier2RepoConcurrency = 5` to keep them in sync.
- `TestProcessReposInParallel_RunsConcurrently` uses a 400ms wall-clock threshold — may be flaky on heavily loaded CI runners. Consider relaxing or marking with `t.Parallel()` discipline.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Well-structured refactor that addresses the three concerns from #481 (cross-tier blocking, per-repo serialisation, SSE ordering race). The cancellation fix in processReposInParallel is correct and well-tested. A few minor concerns around shadowing, hardcoded log prefixes in a general helper, and potential test flakiness on slow CI.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:2211** — `cap := 5` shadows the Go builtin `cap`. While not a bug, this is poor style and can cause subtle issues if someone later tries to use the builtin in this scope. Rename to `repoConcurrency` or similar.
🟡 **daemon/cmd/heimdallm/main.go:2095** — `processReposInParallel` is presented as a general helper but hardcodes `slog.Error("tier2: per-repo work failed", ...)` with a tier2-specific prefix. Either inline the helper into the tier2 caller, take a logger prefix as a parameter, or have workFn handle its own error logging and return only the count.
🟡 **daemon/cmd/heimdallm/main_tier2_parallel_test.go:78** — TestProcessReposInParallel_RunsConcurrently relies on wall-clock timing (400ms threshold for ~200ms expected). On loaded CI runners this can be flaky. Consider using a barrier/channel-based assertion of concurrency rather than time.Sleep + elapsed.
🟡 **daemon/cmd/heimdallm/main_tier2_defer_test.go:124** — `testEventDataContains` and `stringContains` reimplement `strings.Contains`. Just import `strings` and call `strings.Contains(ev.Data, repo)`. The `data == needle` short-circuit is redundant with Contains.
⚠️ **daemon/cmd/heimdallm/main.go:2178** — Both PR and issue tiers now emit independent `polling_started`/`polling_completed` events from concurrent goroutines. Worth confirming the Flutter UI handles two parallel polling cycles per interval (the previous combined-tick contract is gone). The configuration-guide.md update notes this, but UI compatibility isn't verified in the diff.
🟡 **daemon/cmd/heimdallm/main.go:2632** — Deferred-discovery filter is silent to the caller — a PR for a brand-new repo is dropped from the returned slice and the user sees a one-tick delay before review. The slog.Info logs it server-side, but consider emitting an SSE event (or extending repo_discovered's payload) so the UI can show 'review pending next cycle' rather than nothing.
🟡 **daemon/cmd/heimdallm/main.go:2236** — PublishPending is now bound only to the PR ticker. The PR comment acknowledges this is intentional (idempotent retries, pending publishes typically come from PR fetch). If issue-tier publishes ever start landing in the pending set, retries will be delayed by up to a PR interval. Worth a code comment near the issue tick noting this assumption so future readers don't rebind it accidentally.

### Suggestions

- Rename the local `cap` variable in runIssueTier to avoid shadowing the Go builtin.
- Replace the bespoke string helpers in main_tier2_defer_test.go with strings.Contains.
- Make TestProcessReposInParallel_RunsConcurrently signal-based (close a channel from each worker, assert N workers running at once via a counter+barrier) rather than wall-clock timed to reduce CI flakiness.
- Consider extracting processReposInParallel's error logging to the caller via workFn so the helper has no tier2-specific strings — or inline it since it's only used in one place.
- Add an upper-bound sanity check on Tier2RepoConcurrency in applyDefaults (e.g., cap at 100) to protect against config typos like 5000.
- Add a code comment near the issue tick noting that PublishPending is intentionally bound to the PR tick, so future maintainers don't add a second call.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured concurrency refactor: independent PR/issue tickers, semaphore-bounded worker pool with labeled-break cancellation fix, and a clean one-tick deferral for newly-discovered repos. Tests are thorough. A few issues around shutdown noise and hardcoded defaults.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2098** — processReposInParallel logs every workFn error via slog.Error, including context.Canceled from limiter.Acquire during graceful shutdown. The original sequential loop silently returned on limiter errors. With many repos, a daemon shutdown will now spray one ERROR log per pending/in-flight repo. Consider checking for ctx.Err()/context.Canceled before logging, or letting the caller decide severity.
🟡 **daemon/cmd/heimdallm/main.go:2080** — The default concurrency of 5 is hardcoded in three independent locations: processReposInParallel (fallback for <=0), runIssueTier (fallback when repoConcurrencyFn is nil/returns 0), and config.applyDefaults(). If the config default changes, the two hardcoded fallbacks silently diverge. Consider extracting a const or having processReposInParallel rely solely on the caller to supply a valid value (the config layer already guarantees >=5).
🟡 **daemon/cmd/heimdallm/main.go:2214** — runIssueTier's workFn returns limiter.Acquire errors, which processReposInParallel logs generically as 'per-repo work failed'. The original code distinguished limiter failures (early return, no log) from ProcessRepo failures (slog.Error with 'issue processing'). The generic message loses that distinction, making it harder to diagnose rate-limiter vs repo-processing failures in production logs.
🟡 **daemon/cmd/heimdallm/main_tier2_defer_test.go:121** — Custom stringContains reimplements strings.Contains from the standard library. Use strings.Contains instead.
🟡 **daemon/cmd/heimdallm/main.go:2045** — The doc comment on processReposInParallel is 11 lines for a standard semaphore-bounded worker pool. The cancellation and non-positive-concurrency behaviors are important but could be conveyed in 3-4 lines. The current length is disproportionate to the function's complexity.

### Suggestions

- In processReposInParallel, skip the slog.Error when the error is context.Canceled or context.DeadlineExceeded — e.g. `if err != nil && ctx.Err() == nil { slog.Error(...) }` — to avoid noisy logs during graceful shutdown.
- Extract the default concurrency (5) into a package-level const (e.g. `defaultTier2RepoConcurrency`) and reference it from processReposInParallel, runIssueTier, and config.applyDefaults to keep the three sites in sync.
- Replace the hand-rolled stringContains in the test file with strings.Contains.
- Consider wrapping the workFn error in runIssueTier to distinguish limiter failures from ProcessRepo failures — e.g. return a sentinel or annotated error so the generic logger in processReposInParallel can emit a more targeted message.

---

ivanmunozruiz:
Polish tras los reviews (eeec2fd). Aplicados todos los nits reales:

- **\`cap\` sombreando builtin** → renombrado a \`concurrency\`.
- **\`stringContains\` reimplementando \`strings.Contains\`** → eliminado, uso stdlib.
- **\`processReposInParallel\` con log prefix tier2-specific** → helper silencia errores (devuelve 0); per-repo logging vive en la workFn del callsite.
- **Default 5 hardcoded en dos sitios** → \`config.DefaultTier2RepoConcurrency\` como single source of truth.
- **PublishPending solo en PR tick** → comment explicando la binding intencional.
- **Ticker boilerplate duplicado** → \`runTickerLoop(tick)\` closure; las dos goroutines son one-liners.
- **\`TestProcessReposInParallel_RunsConcurrently\` flaky** → eliminado (era redundante: \`RespectsConcurrencyCap\` ya pinea paralelismo por contador de in-flight peak, sin wall-clock).

**No aplicados** (con razón en el commit msg):
- EmitPollingCompleted vía defer: contrato nuevo es correcto (Muriano: "arguably better").
- Upper bound contra typos en Tier2RepoConcurrency: goroutines son baratos, no merece la pena.
- SSE "review pending next cycle": UX, follow-up separado.
- Flutter UI compat: el envelope ya tiene \`kind\`; GUI itera por kind; out of scope.

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of #486. All prior nits addressed cleanly: `cap` renamed to `concurrency`, `strings.Contains` used directly, helper now silences errors so per-repo logging lives in the workFn closure (no tier2-specific prefix leaks), `config.DefaultTier2RepoConcurrency` is now the single source of truth shared between `applyDefaults` and `processReposInParallel`, `PublishPending` binding is documented, ticker boilerplate factored into `runTickerLoop`, and the flaky wall-clock concurrency test was removed with a justification comment. The labeled `break schedule` correctly fixes the prior cancellation bug and is covered by a regression test. Deferred `EmitPollingCompleted` via defer correctly captures `prCount`/`issueCount` by reference. Unapplied items have reasonable justifications in the discussion (UI compat out of scope, upper-bound guard rejected as low-value, SSE 'pending next cycle' deferred to follow-up). No new issues spotted.

### Suggestions

- Consider a follow-up to emit an SSE event when a PR is deferred due to newly-discovered repo, so the UI can render 'review pending next cycle' instead of silence.
- Optionally verify Flutter UI handles two independent polling cycles per interval; out of scope for this PR but worth a smoke test before relying on it in production.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #486: author addressed all prior nits from the previous review (renamed `cap`→`concurrency`, replaced hand-rolled substring search with `strings.Contains`, centralised the default in `config.DefaultTier2RepoConcurrency`, factored ticker loop into `runTickerLoop`, removed the flaky wall-clock test). The independent-ticker refactor and the cancellation fix in `processReposInParallel` (labeled `break schedule` + ctx-aware sem send) are correctly implemented and pinned by `TestProcessReposInParallel_CancelStopsScheduling`. The deferred-discovery filter is well-scoped and covered by a dedicated test. No new issues found.

### Suggestions

- Consider a follow-up to emit an SSE 'review_pending_next_cycle' event for deferred PRs so operators can distinguish 'no PR found' from 'PR deferred one tick' (author already noted this as out-of-scope UX work).
- Consider documenting the new SSE contract change (polling_completed now always fires, including on rate-limited ticks with count=0) in the SSE consumer docs so downstream UIs treat count=0 as a benign signal.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previous findings have been addressed. The independent ticker architecture properly decouples PR and issue processing. The labeled-break cancellation fix is correct and well-tested. The DefaultTier2RepoConcurrency const eliminates the drift risk. The deferred-discovery mechanism for newly-discovered repos is clean. No new issues found.

### Suggestions

- Consider adding a brief inline comment in runTickerLoop explaining that time.Ticker's single-element buffer means a slow tick naturally skips redundant firings — this is the key invariant that prevents concurrent runs of the same tier, and it's easy to miss on a first read.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of #486 after addressing prior findings. The PR now splits PR and issue polling into independent tickers, parallelizes per-repo issue processing with a context-aware worker pool, and defers reviews on newly-discovered repos by one tick. Previous findings (independent tickers, cancel-stops-scheduling bug) are addressed. Tests cover the new helper thoroughly including the cancellation bug fix. A few minor concerns remain around the defer semantics in runIssueTier and a subtle race in repo discovery deferral.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:2208** — In runIssueTier, issueCount is captured by the deferred EmitPollingCompleted closure, but issueCount is reassigned (not appended-to) by `issueCount = processReposInParallel(...)`. Since the defer evaluates issueCount at call time (it's a captured variable, not a parameter), this works correctly — but the pattern is fragile: if a future edit changes the assignment to e.g. a short-lived local, the defer would silently emit 0. A small comment or capturing issueCount via pointer would make the intent explicit. Low severity, stylistic.
🟡 **daemon/cmd/heimdallm/main.go:2629** — The deferred-discovery filter compares pr.Repo against addedThisTick, which is populated from the `added` slice returned by upsertDiscoveredRepos. If a PR arrives for a repo that was discovered on a *previous* tick but somehow not yet fully reflected in the UI (e.g., the SSE bridge to NATS was momentarily slow), the deferral won't apply. This is the intended scope per the plan (defer is one-tick only), but the docstring should make explicit that the guarantee is 'one tick of deferral relative to upsertDiscoveredRepos', not 'guaranteed UI receipt before review_started'. Worth a brief comment clarification.
🟡 **daemon/cmd/heimdallm/main.go:2155** — snapshotRepos is called separately by prTick and issueTick, so each tier observes its own snapshot at tick time. This is correct, but means PR and issue tiers can briefly disagree on the monitored set during a config reload — e.g., the PR tier may publish a review for a repo that the issue tier won't poll on the same cycle. Likely benign (next tick converges) but worth being aware of; no change required.

### Suggestions

- Consider documenting in the runIssueTier closure that issueCount is intentionally captured by reference for the deferred EmitPollingCompleted, to prevent future refactors from accidentally breaking the count emission.
- The defer-discovery test (main_tier2_defer_test.go) drains SSE events with a 200ms deadline — fine for CI but could become flaky under load. Consider using a synchronous wait on the specific event type with a longer timeout.
- Consider adding an upper bound sanity check on Tier2RepoConcurrency (e.g., cap at 100 or warn) to prevent typos like 5000 from spawning excessive goroutines, though the author has explicitly declined this as out of scope.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #486 after the author's revisions. Previous concerns (independent PR/issue tickers, cancellation bug in worker pool, shadowing of `cap`, single source of truth for default concurrency, removal of flaky wall-clock test) have all been addressed. The new architecture runs PR and issue tiers on independent goroutines/tickers, the worker pool now observes ctx.Done in scheduling via a labeled break, and DefaultTier2RepoConcurrency is centralized in config. Tests cover happy path, cap, cancellation mid-scheduling, error continuation, empty input, and non-positive cap fallback. The deferred-discovery fix is pinned by a dedicated test and existing tests were updated to pre-register repos. No new confirmed defects.

### Suggestions

- Consider an upper bound or warning on `Tier2RepoConcurrency` to catch obvious typos (e.g. 5000) — author already declined as not worth it; noting for visibility.
- `PublishPending` is bound to the PR tick by comment; if issue-side NATS publishes are ever added, the sibling call must be remembered — a small TODO/assertion in `runIssueTier` would make the coupling explicit.
- The deferred-discovery one-tick latency on first review of a brand-new repo is documented but not surfaced in SSE as `review_pending_next_cycle`; a follow-up event would improve UX (author noted as separate follow-up).

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of the parallel polling PR. The author addressed both prior findings cleanly: independent tickers per tier (PR/issue) prevent cross-tier blocking, and the cancellation bug in the scheduling loop is fixed with a labeled break + ctx-aware select. The new test TestProcessReposInParallel_CancelStopsScheduling pins the bug. Documentation clarifications were added for the issueCount defer capture and the deferred-discovery filter scope. Config plumbing, defaults, and tests are sound.

### Suggestions

- Consider a follow-up to deduplicate the snapshotRepos call between prTick and issueTick during a config reload window (already acknowledged as benign).
- If Tier2RepoConcurrency is ever set absurdly high via config, the helper caps at len(repos) but does not log; an info-level log when concurrency is clamped could aid operator visibility.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #486: prior findings (independent tickers, cancellation-safe scheduling loop) are addressed in the current diff. The worker pool correctly uses labeled break + ctx-aware semaphore acquire, PR and issue tiers run on independent goroutines/tickers, and the deferred-discovery fix is covered by a new test. Config default is centralized via DefaultTier2RepoConcurrency. Tests cover happy path, concurrency cap, error continuation, cancellation, scheduling cancel, nil/empty input, and non-positive cap fallback. No new confirmed issues.

### Suggestions

- Consider an upper bound sanity check on Tier2RepoConcurrency (e.g., warn if >100) to catch config typos — author has already declined this as not worth it, noting goroutines are cheap.
- The deferred-discovery comment notes the guarantee is 'one tick relative to upsertDiscoveredRepos' not 'guaranteed UI delivery' — a follow-up could emit an explicit 'review_pending_next_cycle' SSE event for better UX (already tracked by author as out of scope).

---

PR #488, opened 2026-05-13
fix(issues): terminate auto_implement no-changes runs cleanly (#483)

### LINKED ISSUE #483: auto_implement no-changes fallback leaves issue in limbo — should mark as done/error
## Problem

When `auto_implement` runs but the agent produces no changes, the daemon posts a fallback comment and stores `ActionTaken: "auto_implement_no_changes"` — but does **not** post a `MarkerDone` marker. The issue enters a limbo state: not "done", not "in progress", just stuck.

On every subsequent poll, the fetcher hits this dedup path (`fetcher.go:300-317`):

```go
if latest.ActionTaken == ActionAutoImplementNoChanges && issue.Mode == config.IssueModeDevelop {
    noChangesCount, ncErr := f.store.CountAutoImplementNoChanges(row.ID)
    // ...
    if noChangesCount >= MaxAutoImplementNoChanges {  // cap = 1
        return true, "requires human intervention or retry marker", nil
    }
    return true, "waiting for manual retry", nil  // ← ALWAYS skips
}
```

Both branches return `skip=true`. The issue is silently ignored every poll cycle — the log fills with "skipping issue" debug entries, the UI shows it stuck in "develop" with no activity, and the user sees:

> ⚠️ Heimdallm auto-implement skipped
> The agent looked at #521 but left the working tree unchanged...

This creates the perception that the daemon is "in a death spiral" — the issue is alive but nothing happens.

## Root cause

Two related issues:

1. **No `MarkerDone` in the fallback path** — `autoImplementNoChangesFallback` (`pipeline.go:877-914`) posts a comment but does NOT include `<!-- heimdallm:done -->`. Compare with the success path (`pipeline.go:830`) which does include `MarkerDone`.

2. **The dedup gate blocks unconditionally** — Even when `noChangesCount < MaxAutoImplementNoChanges`, line 316 returns `true` ("skip"). The `MaxAutoImplementNoChanges = 1` cap is effectively meaningless because both code paths below the cap check also return skip. The issue can never be retried automatically.

## Expected behavior

When `auto_implement` produces no changes:

1. Post the `<!-- heimdallm:done -->` marker in the fallback comment so the issue is properly marked as terminal
2. Emit an SSE event that clearly communicates the issue reached a terminal state (error/done), not just "completed"
3. The issue should stop appearing as "in progress" or "develop" in the dashboard — it should show as "failed" or "needs attention"
4. The only reactivation path should be an explicit `<!-- heimdallm:retry -->` marker (which already works — the retry check runs before the no-changes check in `alreadyProcessed`)

## Affected files

- `daemon/internal/issues/pipeline.go` — `autoImplementNoChangesFallback` (line 877): missing `MarkerDone` in comment body
- `daemon/internal/issues/fetcher.go` — `alreadyProcessed` (line 300-317): dedup logic for no-changes state
- `daemon/internal/issues/fetcher.go` — `MaxAutoImplementNoChanges` (line 41): cap is 1 but the gate blocks before it's ever checked meaningfully

ivanmunozruiz:
## Summary

Closes #483.

`auto_implement` runs that produce no changes left the issue in limbo:
the fallback comment lacked `MarkerDone`, so the fetcher's dedup gate
(both branches `skip=true`) silently ignored the issue every poll cycle.

### What changed

- **`MarkerDone` in fallback body** (`pipeline.go`): prepend the done
  marker to the fallback comment so the fetcher's marker scan
  terminates the issue on the next tick. The body references the
  retry marker **by keyword** (`heimdallm:retry`), pointing at the
  configuration guide's `auto_implement` section — the literal token
  cannot appear in the same comment as `MarkerDone` because
  `ScanMarkers` prioritises Retry over Done within a single comment.

- **SSE event** (`pipeline.go`): switch
  `EventIssueReviewCompleted` → `EventIssueReviewError` with
  `reason="auto_implement_no_changes"` and a human `error` field that
  follows the convention of every other `EventIssueReviewError`
  emitter (Flutter's event row + detail toast both read `error`).

- **Dedup gate** (`fetcher.go`): collapse the now-back-compat block to
  a single unconditional skip with a clearer reason for legacy rows
  written before `MarkerDone` landed on the fallback comment. The
  cap-based logic was never load-bearing (both branches returned
  `skip=true`).

- **Dead code removed**: `MaxAutoImplementNoChanges` constant, the
  `CountAutoImplementNoChanges` interface method, the store-level
  query and the `fakeDedup` stub all disappear in this PR — no
  follow-up needed.

- **Flutter UI** (`flutter_app/`): new `AttentionBadge` widget and a
  deep-orange side-bar tint render `auto_implement_no_changes` rows
  as `NEEDS ATTENTION` across the dashboard tile, the activity grid,
  the issues list and the issue-detail review header (otherwise the
  empty triage block would have surfaced them as green LOW).
  `dashboard_providers.dart` also now bumps `issueListRefreshProvider`
  on `issue_review_error`, matching the other terminal issue events.

`ActionTaken` in the store stays `ActionAutoImplementNoChanges` so the
activity-log queries and Flutter consumers that already key off the
column keep working.

Doc note added to `configuration-guide.md` explaining the terminal
state and the retry escape hatch. The plan in
`docs/superpowers/plans/` is committed for historical reference (same
convention as other completed plans in that directory); its body is
the pre-implementation draft and carries an explicit `Status: DONE`
banner noting which decisions changed during execution.

## Test plan
- [x] `TestAutoImplementNoChangesFallback_PostsDoneMarker` (RED → GREEN)
- [x] `TestAutoImplementNoChangesFallback_BodyScansAsDone` — passes the real fallback body through `ScanMarkers` and asserts `MarkerResultDone`, regression pin for the marker-collision draft caught in review
- [x] `TestAutoImplementNoChangesFallback_EmitsReviewError` — asserts the event type, `reason` and `error` fields
- [x] `TestPipeline_AutoImplementNoChangesRecordsTerminalNoChanges` updated to expect `EventIssueReviewError`
- [x] `TestAlreadyProcessed_DoneMarkerSkipsBeforeDedupGate` (marker-first ordering)
- [x] `TestAlreadyProcessed_HistoricalNoChangesRow_StillSkipped` (back-compat path for legacy rows without `MarkerDone`)
- [x] `TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges` (retry escape hatch)
- [x] Deleted `TestFetcher_AutoImplementNoChangesCapStopsLoop` and `TestFetcher_AutoImplementNoChangesCountErrFallsThrough` — they exercised the cap logic that no longer exists
- [x] `make test-docker` green
- [x] `flutter analyze` green

### DIFF
--- a/daemon/internal/issues/fetcher.go
+++ b/daemon/internal/issues/fetcher.go
@@ -298,22 +298,12 @@ func (f *Fetcher) alreadyProcessed(issue *github.Issue) (bool, string, error) {
 		return true, "already implemented (PR created)", nil
 	}
 	if latest.ActionTaken == ActionAutoImplementNoChanges && issue.Mode == config.IssueModeDevelop {
-		// If the agent has produced an empty diff at or above the cap, stop
-		// retrying. Manual unblock paths: a retry marker (handled above
-		// before we reach this point), dismiss, or moving the issue back
-		// to a different stage label. See issue #433.
-		noChangesCount, ncErr := f.store.CountAutoImplementNoChanges(row.ID)
-		if ncErr != nil {
-			slog.Warn("issues fetcher: could not count no-changes auto_implement, falling through to legacy block",
-				"repo", issue.Repo, "number", issue.Number, "err", ncErr)
-			return true, "auto_implement produced no changes; waiting for manual retry", nil
-		}
-		if noChangesCount >= MaxAutoImplementNoChanges {
-			return true, fmt.Sprintf(
-				"auto_implement produced no changes %d times (cap %d); requires human intervention or retry marker",
-				noChangesCount, MaxAutoImplementNoChanges), nil
-		}
-		return true, "auto_implement produced no changes; waiting for manual retry", nil
+		// Back-compat skip for rows written before #483 added MarkerDone
+		// to the fallback comment. New no-changes runs are terminated by
+		// the marker scan above; reaching here means the row was stored
+		// without a marker, so we keep skipping until a human posts
+		// MarkerRetry (handled before this block).
+		return true, "auto_implement produced no changes (historical row, no done marker); add retry marker to reprocess", nil
 	}
 
 	// Bot-comment dedup: if the most recent comment is from the bot AND the
--- a/daemon/internal/issues/fetcher.go
+++ b/daemon/internal/issues/fetcher.go
@@ -32,14 +32,6 @@ const RecomputeGrace = 30 * time.Second
 // See issue #223.
 const MaxAutoImplementFailures = 3
 
-// MaxAutoImplementNoChanges caps how many auto_implement_no_changes runs we
-// allow on a single issue before the fetcher permanently skips it. After the
-// agent has produced an empty diff this many times in a row, retrying is
-// almost certainly burning tokens for nothing — the issue needs more context,
-// a different agent, or a human. The retry marker / dismiss / mode change
-// path still unblocks the issue manually. See issue #433.
-const MaxAutoImplementNoChanges = 1
-
 // IssuesFetcher is the subset of github.Client that fetches classified
 // issues. Kept as an interface so the fetcher can be tested without an HTTP
 // server standing in.
@@ -63,10 +55,6 @@ type issueDedupStore interface {
 	// "auto_implement_failed" review rows for the issue. Used to enforce
 	// MaxAutoImplementFailures; see issue #223.
 	CountFailedAutoImplement(issueID int64) (int, error)
-	// CountAutoImplementNoChanges returns the number of stored
-	// "auto_implement_no_changes" review rows for the issue. Used to enforce
-	// MaxAutoImplementNoChanges; see issue #433.
-	CountAutoImplementNoChanges(issueID int64) (int, error)
 }
 
 // issueMarkerFetcher fetches comments for an issue so the fetcher can scan
--- a/daemon/internal/issues/fetcher_test.go
+++ b/daemon/internal/issues/fetcher_test.go
@@ -494,7 +494,12 @@ func TestFetcher_DoesNotSkipBelowMaxAutoImplementFailures(t *testing.T) {
 	}
 }
 
-func TestFetcher_SkipsAutoImplementNoChangesUntilManualRetry(t *testing.T) {
+// TestAlreadyProcessed_HistoricalNoChangesRow_StillSkipped pins the
+// back-compat branch of the simplified dedup gate (#483): rows written
+// before MarkerDone was added to the fallback comment have no marker on
+// the issue but still carry ActionTaken=auto_implement_no_changes, so the
+// dedup gate must keep skipping them until a human posts MarkerRetry.
+func TestAlreadyProcessed_HistoricalNoChangesRow_StillSkipped(t *testing.T) {
 	reviewedAt := time.Now().Add(-10 * time.Minute)
 	issue := fixture(1, time.Now())
 	issue.Mode = config.IssueModeDevelop
@@ -507,26 +512,28 @@ func TestFetcher_SkipsAutoImplementNoChangesUntilManualRetry(t *testing.T) {
 				CreatedAt:   reviewedAt,
 				CommentedAt: reviewedAt,
 			},
-			noChangesCount: 1, // at the cap (MaxAutoImplementNoChanges = 1)
 		},
 	}}
+	// No comments at all — pre-#483 row that never got a MarkerDone.
+	mf := &fakeMarkerFetcher{commentsByKey: map[string][]github.Comment{}}
 	p := &fakePipeline{}
-	f := issues.NewFetcher(&fakeClient{issues: []*github.Issue{issue}}, nil, dedup, p)
+	f := issues.NewFetcher(&fakeClient{issues: []*github.Issue{issue}}, mf, dedup, p)
 
 	processed, err := f.ProcessRepo(context.Background(), "org/repo", enabledCfg(), "alice", noOpts)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
 	if processed != 0 || len(p.calls) != 0 {
-		t.Fatalf("no-changes auto_implement should wait for retry marker, processed=%d calls=%v", processed, p.calls)
+		t.Fatalf("historical no-changes row must still be skipped, processed=%d calls=%v", processed, p.calls)
 	}
 }
 
-func TestFetcher_AutoImplementNoChangesCapStopsLoop(t *testing.T) {
-	// Symmetric to MaxAutoImplementFailures: once the agent has produced
-	// an empty diff at or above MaxAutoImplementNoChanges, the fetcher must
-	// stop dispatching the issue regardless of updated_at bumps. Manual retry
-	// marker still bypasses this (handled before the cap check).
+// TestAlreadyProcessed_DoneMarkerSkipsBeforeDedupGate proves the
+// marker-first ordering: a no-changes row whose comment thread now
+// carries MarkerDone (the post-#483 happy path) is terminated by the
+// marker scan, not by the back-compat dedup gate. Without this ordering
+// the user could not see "done marker" in the skip reason.
+func TestAlreadyProcessed_DoneMarkerSkipsBeforeDedupGate(t *testing.T) {
 	reviewedAt := time.Now().Add(-10 * time.Minute)
 	issue := fixture(1, time.Now())
 	issue.Mode = config.IssueModeDevelop
@@ -539,25 +546,28 @@ func TestFetcher_AutoImplementNoChangesCapStopsLoop(t *testing.T) {
 				CreatedAt:   reviewedAt,
 				CommentedAt: reviewedAt,
 			},
-			noChangesCount: issues.MaxAutoImplementNoChanges + 5, // well over the cap
 		},
 	}}
+	mf := &fakeMarkerFetcher{commentsByKey: map[string][]github.Comment{
+		"org/repo#1": {{Body: issues.MarkerDone + "\nfallback comment", CreatedAt: time.Now()}},
+	}}
 	p := &fakePipeline{}
-	f := issues.NewFetcher(&fakeClient{issues: []*github.Issue{issue}}, nil, dedup, p)
+	f := issues.NewFetcher(&fakeClient{issues: []*github.Issue{issue}}, mf, dedup, p)
 
 	processed, err := f.ProcessRepo(context.Background(), "org/repo", enabledCfg(), "alice", noOpts)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
 	if processed != 0 || len(p.calls) != 0 {
-		t.Fatalf("over-cap issue must not be processed, processed=%d calls=%v", processed, p.calls)
+		t.Fatalf("done marker must skip ahead of dedup gate, processed=%d calls=%v", processed, p.calls)
 	}
 }
 
-func TestFetcher_AutoImplementNoChangesCountErrFallsThrough(t *testing.T) {
-	// A flaky CountAutoImplementNoChanges must not block the existing
-	// "wait for manual retry" skip — degrade gracefully like the failure cap
-	// does (TestFetcher_CountFailedAutoImplErrDoesNotSkip).
+// TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges proves the
+// retry escape hatch is actually reachable now that the dedup gate no
+// longer unconditionally swallows no-changes rows. Pre-#483, the gate
+// short-circuited before any marker check could reopen the issue.
+func TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges(t *testing.T) {
 	reviewedAt := time.Now().Add(-10 * time.Minute)
 	issue := fixture(1, time.Now())
 	issue.Mode = config.IssueModeDevelop
@@ -570,18 +580,23 @@ func TestFetcher_AutoImplementNoChangesCountErrFallsThrough(t *testing.T) {
 				CreatedAt:   reviewedAt,
 				CommentedAt: reviewedAt,
 			},
-			noChangesErr: errors.New("db timeout"),
+		},
+	}}
+	mf := &fakeMarkerFetcher{commentsByKey: map[string][]github.Comment{
+		"org/repo#1": {
+			{Body: issues.MarkerDone + "\nfallback comment", CreatedAt: time.Now().Add(-5 * time.Minute)},
+			{Body: issues.MarkerRetry + "\nplease retry", CreatedAt: time.Now()},
 		},
 	}}
 	p := &fakePipeline{}
-	f := issues.NewFetcher(&fakeClient{issues: []*github.Issue{issue}}, nil, dedup, p)
+	f := issues.NewFetcher(&fakeClient{issues: []*github.Issue{issue}}, mf, dedup, p)
 
 	processed, err := f.ProcessRepo(context.Background(), "org/repo", enabledCfg(), "alice", noOpts)
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	if processed != 0 || len(p.calls) != 0 {
-		t.Fatalf("count error must still skip, processed=%d calls=%v", processed, p.calls)
+	if processed != 1 || len(p.calls) != 1 {
+		t.Fatalf("retry marker must reopen a no-changes row, processed=%d calls=%v", processed, p.calls)
 	}
 }
 
--- a/daemon/internal/issues/fetcher_test.go
+++ b/daemon/internal/issues/fetcher_test.go
@@ -549,10 +549,12 @@ func TestAlreadyProcessed_DoneMarkerSkipsBeforeDedupGate(t *testing.T) {
 	}
 }
 
-// TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges proves the
-// retry escape hatch is actually reachable now that the dedup gate no
-// longer unconditionally swallows no-changes rows. Pre-#483, the gate
-// short-circuited before any marker check could reopen the issue.
+// TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges pins the retry
+// escape hatch on a no-changes row. The marker scan already ran before
+// the dedup gate pre-#483, so this is the documented user workflow —
+// what the test really protects against is a future refactor that
+// moves the dedup gate above the marker scan or that drops the retry
+// override for the ActionAutoImplementNoChanges branch.
 func TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges(t *testing.T) {
 	reviewedAt := time.Now().Add(-10 * time.Minute)
 	issue := fixture(1, time.Now())
--- a/daemon/internal/issues/fetcher_test.go
+++ b/daemon/internal/issues/fetcher_test.go
@@ -52,8 +52,6 @@ type dedupEntry struct {
 	revErr            error
 	failedAutoImpl    int // stub for CountFailedAutoImplement
 	failedAutoImplErr error
-	noChangesCount    int // stub for CountAutoImplementNoChanges
-	noChangesErr      error
 }
 
 type fakeDedup struct {
@@ -93,18 +91,6 @@ func (d *fakeDedup) CountFailedAutoImplement(issueID int64) (int, error) {
 				return 0, e.failedAutoImplErr
 			}
 			return e.failedAutoImpl, nil
-		}
-	}
-	return 0, nil
-}
-
-func (d *fakeDedup) CountAutoImplementNoChanges(issueID int64) (int, error) {
-	for _, e := range d.byGithubID {
-		if e.row != nil && e.row.ID == issueID {
-			if e.noChangesErr != nil {
-				return 0, e.noChangesErr
-			}
-			return e.noChangesCount, nil
 		}
 	}
 	return 0, nil
--- a/daemon/internal/issues/pipeline.go
+++ b/daemon/internal/issues/pipeline.go
@@ -897,7 +897,7 @@ func (p *Pipeline) autoImplementNoChangesFallback(issue *github.Issue, issueID i
 	body := fmt.Sprintf(
 		"%s\n## ⚠️ Heimdallm auto-implement skipped\n\n"+
 			"The agent looked at #%d but left the working tree unchanged — it likely needs a human decision or more context than the issue alone provides.\n\n"+
-			"To retry auto-implementation, post a new comment containing a Heimdallm `heimdallm:retry` marker (see the configuration guide's issue-tracking section for the exact syntax). To stop here, remove the develop label.\n\n"+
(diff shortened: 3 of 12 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

PR cleanly fixes the auto_implement no-changes limbo by adding MarkerDone to the fallback comment, switching to EventIssueReviewError, and simplifying the back-compat dedup gate. Tests are updated to pin the new behaviour. Plan doc is thorough. Minor concerns: dead code (MaxAutoImplementNoChanges constant, CountAutoImplementNoChanges store method) is left behind by design but not removed; the fallback comment body no longer mentions that historical retry-marker comments must include the retry token verbatim — acceptable.

### Suggestions

- Consider adding a TODO or follow-up issue link near the now-unused MaxAutoImplementNoChanges constant and CountAutoImplementNoChanges store method so the cleanup doesn't get forgotten.
- The fallback comment uses a backtick-wrapped marker (`<!-- heimdallm:retry -->`); confirm GitHub's markdown renders the HTML comment inside backticks as visible literal text (it should, but worth a manual check on a real issue).
- Optional: assert in TestAutoImplementNoChangesFallback_EmitsReviewError that payload['message'] and payload['post_ok'] are present, to pin the full contract Flutter consumes.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Clean, well-targeted fix for the auto_implement no-changes limbo. The three sub-fixes (MarkerDone in fallback body, switch to ReviewError SSE event, collapse dead dedup gate) are individually justified, tested with TDD-style pins, and documented. Tests are renamed to describe the invariant rather than the scenario, which is a nice improvement. A few minor concerns around dead code, behavior-change blast radius, and the fallback comment claim about editing the issue body.

### Issues

🟡 **daemon/internal/issues/fetcher.go:306** — Removing the only call site of CountAutoImplementNoChanges and MaxAutoImplementNoChanges leaves them as dead code. The plan explicitly defers cleanup, but unreferenced exported symbols typically trip linters (unused, deadcode, revive) and confuse future readers. Worth either deleting now or adding a //nolint comment with a TODO referencing the follow-up.
⚠️ **daemon/internal/issues/pipeline.go:923** — Switching from EventIssueReviewCompleted to EventIssueReviewError is a wire-format behavior change for any SSE consumer (Flutter UI, but potentially also CLI tools, dashboards, logs). The PR notes the Flutter risk but does not confirm whether Flutter currently handles EventIssueReviewError without a reason-specific case — verify the UI renders something sensible rather than a generic crash/error and that any analytics/metrics on the Completed event aren't double-counting.
⚠️ **daemon/internal/issues/pipeline.go:895** — Fallback body says 'add a MarkerRetry marker comment (or edit the issue body to include it)'. Confirm that the fetcher's marker scan actually reads the issue body — markers.go in some implementations scans only the comment thread. If the marker scan is comments-only, this instruction misleads users.
🟡 **daemon/internal/issues/fetcher.go:300** — Historical rows (ActionTaken=auto_implement_no_changes with no MarkerDone in comments) will now skip indefinitely with no automatic recovery path. The skip reason is helpful but consider whether a one-time migration that backfills MarkerDone, or a daemon log warning the first time a legacy row is encountered, would reduce operator confusion. Out-of-scope per the plan, but worth mentioning.
🟡 **daemon/internal/issues/fetcher_test.go:583** — TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges seeds both MarkerDone and MarkerRetry in the comment thread and asserts the retry path wins. This relies on the marker scanner's tie-breaking semantics (most-recent-wins, or retry-always-wins). The test doesn't pin which semantic is required — a small comment explaining 'MarkerRetry posted after MarkerDone must override' would make the intent explicit and survive a refactor of the scanner.
🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:1** — Committing the internal planning document to the repo permanently. Some teams keep plans out of main; if this repo is OK with it, fine, but worth confirming the convention is to land these alongside the implementation rather than discard.

### Suggestions

- Either delete MaxAutoImplementNoChanges + CountAutoImplementNoChanges in this PR or open the follow-up issue immediately and link it from the back-compat comment in fetcher.go.
- Verify Flutter's behavior on EventIssueReviewError with reason=auto_implement_no_changes before merging — even a screenshot in the PR description would document the new UI state.
- Confirm whether the marker scan covers the issue body. If not, drop the '(or edit the issue body to include it)' phrase from the fallback comment to avoid misleading users.
- Consider a one-line debug log the first time a legacy no-changes row is skipped per restart, so operators have a breadcrumb if they wonder why an old issue is stuck.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured fix with three coordinated changes (MarkerDone in fallback comment, SSE event type switch, dedup gate simplification). Test coverage is thorough, back-compat story is sound, and the configuration-guide update is clear. Minor issues around the committed plan document.

### Issues

🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:68** — Plan document's format string diverges from the actual implementation. The plan says 'Add a `%s` marker (or remove this marker) and a retry marker (`%s`)' but the code says 'To retry auto-implementation, add a `%s` marker comment (or edit the issue body to include it).' The plan wording is also confusing since it tells users to add *both* MarkerDone and MarkerRetry, while the actual code correctly tells them to add only MarkerRetry. This document will mislead future readers who reference it instead of the code.
🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:1** — 165-line design/planning document committed to the repo will become stale as the code evolves. The plan references specific line numbers (e.g. 'fetcher.go:300-317', 'pipeline.go:889-925') and snapshots of code that will drift. Consider whether this belongs in the PR description rather than the repo, or at minimum add a note that this is a point-in-time snapshot.
🟡 **daemon/internal/issues/pipeline.go:920** — The SSE payload field was renamed from 'mode' to 'reason' and the event type changed from EventIssueReviewCompleted to EventIssueReviewError. While the Flutter dashboard currently does not parse the 'mode' field (confirmed via code search), any out-of-tree consumers or monitoring/alerting that filters on the old event type + 'mode' key will silently break. Consider documenting this as a breaking change in the PR description or a migration note.
🟡 **daemon/internal/issues/pipeline.go:893** — The comment block above the format string (lines 889-892) references 'MarkerRetry is named inline so the user can find the escape hatch without grepping source' — but MarkerRetry is rendered inside backticks as raw HTML comment syntax (), which is correct for discoverability. The comment's phrasing ('named inline') is slightly misleading since it's really 'rendered as a code span'; consider tightening the comment or removing it (the code is self-explanatory).

### Suggestions

- The plan document's format string (line 68) should be updated to match the actual implementation, or the plan should carry a header noting it is a pre-implementation snapshot and the code is authoritative.
- Consider adding a brief 'Breaking changes' note in the PR description about the SSE event type switch (EventIssueReviewCompleted → EventIssueReviewError) and payload field rename (mode → reason) for any non-Flutter consumers.
- The new tests are well-named and cover the three key paths (historical row, done marker, retry marker). Good use of the fakeMarkerFetcher to simulate pre-#483 vs post-#483 scenarios.
- If the docs/superpowers/plans/ directory is intended for permanent reference, consider adding a front-matter field like 'status: implemented' and 'implemented_in: PR #488' so readers know the plan is historical.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #488: cleanly terminates auto_implement no-changes runs by adding MarkerDone to the fallback comment, switching SSE event to ReviewError, and collapsing the dead dedup gate to a back-compat skip. Tests are thorough (including the regression pin that the fallback body scans as Done, not Retry). Documentation updated. No new issues found.

### Suggestions

- Consider a follow-up to delete the now-unused MaxAutoImplementNoChanges constant and CountAutoImplementNoChanges store method once legacy rows are confirmed drained (already called out as out-of-scope in the plan).
- The fallback body references 'heimdallm:retry' as a keyword but relies on users finding the exact HTML-comment syntax in the configuration guide; a short inline example (in a code fence so GitHub renders it literally) could reduce support load — though the current wording avoids the ScanMarkers Retry-over-Done collision which is the more important constraint.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of #488. The author has addressed both medium-severity findings from the prior review: the fallback body now correctly directs users to post a new comment (not edit the issue body), and the SSE payload uses the `error` field matching the Flutter convention (with a comment explaining why). A new regression test (`TestAutoImplementNoChangesFallback_BodyScansAsDone`) pins a subtle scanner ordering issue that would have re-broken the fix. The Flutter side now refreshes on terminal failures. Remaining items are all low-severity and were flagged before; the dead-code concern persists.

### Issues

🟡 **daemon/internal/issues/fetcher.go:41** — MaxAutoImplementNoChanges constant and CountAutoImplementNoChanges store method are still unreferenced after this PR. The plan defers cleanup but this typically trips deadcode/unused linters. Consider deleting now or adding a TODO comment referencing the follow-up issue so the next reader knows it is intentional.
🟡 **daemon/internal/issues/fetcher_test.go:587** — TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges seeds MarkerDone at -5m and MarkerRetry at now, then asserts retry wins. The test docstring explains the higher-level intent but doesn't pin WHICH scanner semantic is required (most-recent-wins vs retry-always-wins-within-thread). A one-line note like 'MarkerRetry posted after MarkerDone must override (most-recent-wins across comments)' would survive a scanner refactor.
🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:1** — Internal planning document is being committed to main. Worth confirming this repo's convention — some teams keep plans out of the main tree. No blocker if the convention is established.

### Suggestions

- Consider a startup log line the first time a legacy no-changes row (ActionTaken=auto_implement_no_changes with no MarkerDone) is observed, so operators get a one-time nudge to clean up rather than discovering the indefinite-skip behaviour from confused users.
- The new regression test TestAutoImplementNoChangesFallback_BodyScansAsDone is excellent — that scanner-ordering trap was easy to miss. Consider adding a similar 'scan the real body' assertion to the success path so the inverse regression (MarkerDone gets accidentally dropped from success comments) is also pinned.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: the previous LOW finding about unreferenced MaxAutoImplementNoChanges/CountAutoImplementNoChanges has been fully addressed — both the constant and the store method are deleted in this revision, along with the interface declaration and the fake impl in tests. The remaining diff is consistent with the plan: MarkerDone is prepended to the fallback body, SSE event is switched to EventIssueReviewError with reason+error fields, the dedup gate is collapsed to a back-compat branch, and a new regression test (TestAutoImplementNoChangesFallback_BodyScansAsDone) pins the subtle Retry-vs-Done-within-a-single-comment hazard the author caught during implementation. Two prior LOW notes still apply unchanged.

### Issues

🟡 **daemon/internal/issues/fetcher_test.go:587** — TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges still relies on MarkerDone@-5m + MarkerRetry@now to assert retry wins, but the docstring does not state which scanner contract is being pinned (most-recent-wins across comments vs retry-priority). A one-line comment would make this test survive a future ScanMarkers refactor.
🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:1** — Internal planning document is still being committed to main. No blocker — worth confirming this matches repo convention since other plans in docs/superpowers/plans/ suggest it does.

### Suggestions

- Consider adding a one-line note inside TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges clarifying that MarkerRetry posted *after* MarkerDone must override (the precise scanner semantic being pinned).
- Nice catch on the BodyScansAsDone regression test — that exact failure mode (literal MarkerRetry token in the retry hint) is easy to reintroduce on a future copy edit; the test pins it well.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

PR cleanly addresses the auto_implement no-changes limbo by adding MarkerDone to the fallback comment, switching SSE event to ReviewError, and collapsing the dead dedup gate to a back-compat skip. Tests pin all three fixes including the subtle MarkerRetry-vs-Done ordering trap. The dead store method and constant are removed. Docs updated. No issues found.

### Suggestions

- Consider a follow-up migration/admin endpoint to bulk-stamp historical no-changes rows with MarkerDone, as noted in the plan's out-of-scope section.
- The fallback comment references 'the configuration guide's issue-tracking section' but the new doc note lives under the auto_implement section — minor wording drift, not blocking.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

PR cleanly addresses #483: fallback comment now carries MarkerDone, SSE event switched to issue_review_error with reason+error fields, and the dead dedup-gate logic is collapsed to a back-compat skip. Tests cover marker emission, scanner priority (Retry-over-Done collision avoided), SSE payload shape, and the three fetcher paths (done marker, historical row, retry marker). Store method and constant are removed consistently. Docs updated. No correctness, security, or concurrency issues found in the diff.

### Suggestions

- Consider a follow-up migration or admin endpoint to backfill MarkerDone on legacy auto_implement_no_changes rows so the back-compat branch can eventually be deleted (already noted as out-of-scope in the plan).
- The retry instructions in the fallback body reference the marker by keyword and point at the configuration guide rather than embedding the literal token — clear rationale, but a future UX improvement could render an explicit code block in the guide that users copy verbatim.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of PR #488. The two prior LOW findings remain unaddressed in the current diff but are minor nits, not blockers. No new substantive issues found in the changes since the last review.

### Issues

🟡 **daemon/internal/issues/fetcher_test.go:569** — TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges docstring still does not pin the ScanMarkers contract it relies on. The test sets MarkerDone at -5m and MarkerRetry at now, which works under either 'most-recent-comment-wins' or 'retry-priority-within-thread' semantics — a one-line note (e.g., 'relies on ScanMarkers giving MarkerRetry priority over an earlier MarkerDone') would make the test survive a future scanner refactor.
🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:1** — Internal planning document is still committed to main. Other files under docs/superpowers/plans/ suggest this is repo convention, so likely intentional — flagging only for confirmation, not as a blocker.

### Suggestions

- Optional: add one sentence to TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges naming the ScanMarkers behaviour it depends on (retry-priority or most-recent-wins) so the test does not silently keep passing if scanner semantics change.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

PR #488 cleanly addresses the auto_implement no-changes limbo by (A) embedding MarkerDone in the fallback comment, (B) switching the SSE event to issue_review_error with a reason/error payload, and (C) collapsing the now-dead dedup gate. Tests cover the marker-first ordering, the back-compat historical row path, the retry escape hatch, and a regression pin that ScanMarkers on the fallback body returns Done (not Retry). Store/fetcher dead code is removed consistently. Docs updated. No confirmed defects found in the diff.

### Suggestions

- Consider a follow-up to drop or migrate the lingering ActionAutoImplementNoChanges rows so the back-compat branch in fetcher.go can eventually be removed.
- The fallback comment instructs users to post a 'heimdallm:retry' marker; consider linking directly to the configuration-guide anchor for the markers section to reduce friction.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of PR #488. The two prior LOW findings remain unaddressed in the latest diff: the test docstring for the retry-priority case still does not pin the ScanMarkers contract, and the internal plan document is still committed. Both are minor; no new issues introduced by the changes.

### Issues

🟡 **daemon/internal/issues/fetcher_test.go:569** — TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges still lacks a one-line note pinning the ScanMarkers contract it depends on (MarkerRetry priority over an earlier MarkerDone within the same thread). Without it, a future scanner refactor toward 'most-recent-comment-wins' could silently keep the test green while changing semantics.
🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:1** — Internal planning document committed alongside the production fix. Consistent with other files under docs/superpowers/plans/, so likely intentional repo convention — flagging only for confirmation.

### Suggestions

- Add a short comment to TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges noting it relies on ScanMarkers giving MarkerRetry priority over an earlier MarkerDone in the same comment thread.
- Confirm docs/superpowers/plans/ is the intended home for completed plan docs; if not, move or delete the file.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

PR cleanly resolves the auto_implement no-changes limbo by adding MarkerDone to the fallback comment, switching SSE to ReviewError, and collapsing the now-redundant dedup gate. Tests pin all three fixes including the subtle Retry-over-Done scanner precedence. Back-compat for historical rows is preserved. No issues found.

### Suggestions

- Consider a follow-up migration to drop the unused action_taken='auto_implement_no_changes' counter column once historical rows are drained, as noted in the plan's out-of-scope section.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review of PR #488. The two previous LOW findings remain: the retry-marker test still lacks an explicit note on the ScanMarkers contract it depends on, and the plan doc is committed (now marked DONE with caveats — appears intentional). No new issues in the diff; the fallback-body-scans-as-Done regression test is a nice defensive pin and the SSE event switch with `error` field is consistent with other emitters.

### Issues

🟡 **daemon/internal/issues/fetcher_test.go:569** — TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges still doesn't pin the ScanMarkers contract it depends on. The test places MarkerDone and MarkerRetry in separate comments and relies on any-Retry-in-thread-overriding-any-Done semantics. The comment block talks about gate ordering, not marker-precedence — a future ScanMarkers refactor (e.g. 'most-recent-comment-wins' or 'newest-marker-wins-by-timestamp') could keep this test green while changing semantics on a same-comment Done+Retry case. One line tying the test to the cross-comment Retry-overrides-Done contract would future-proof it.
🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:1** — Plan doc remains committed alongside the production fix. Now carries a clear 'Status: DONE' header with explicit deltas vs. the shipped PR, which makes it useful as historical context. Consistent with other files under docs/superpowers/plans/. Flagging only for confirmation that this is the intended repo convention.

### Suggestions

- Add a one-liner above TestAlreadyProcessed_RetryMarker_ReprocessesNoChanges naming the ScanMarkers precedence rule it asserts (any MarkerRetry in the thread overrides an earlier MarkerDone).

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured fix with strong test coverage. The core changes (MarkerDone in fallback comment, SSE event type switch, dedup gate simplification) are correct and the BodyScansAsDone regression test is particularly good. Previous finding about misleading comment phrasing has been addressed. Remaining issues are low severity: messaging inconsistency between SSE error and GitHub comment, and plan document staleness.

### Issues

🟡 **daemon/internal/issues/pipeline.go:925** — The SSE `error` field says 'close the issue' as the stop-here action, but the GitHub fallback comment (line 899) says 'remove the develop label to stop here'. These are different operations — closing an issue vs. removing a label — and a user who sees both messages gets conflicting guidance. Align the wording (probably both should say 'remove the develop label').
🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:68** — Still unaddressed from prior review: the plan's example format string (lines 66-72) tells users to add both MarkerDone and MarkerRetry, which contradicts the actual implementation that deliberately excludes MarkerRetry from the body. The plan is a committed reference document — future readers who consult it instead of the code will get wrong guidance.
🟡 **docs/superpowers/plans/2026-05-13-issue-483-auto-implement-no-changes-limbo.md:1** — Still unaddressed from prior review: 165-line design document with hardcoded line-number references (e.g. 'fetcher.go:300-317', 'pipeline.go:889-925') will rot as the code evolves. Consider moving to the PR description or adding a staleness caveat at the top.

### Suggestions

- Align the SSE error field's stop-here wording with the GitHub fallback comment — both should say 'remove the develop label' rather than 'close the issue'.
- Add a one-line staleness header to the plan document (e.g. 'Point-in-time design snapshot for PR #488; line numbers and code samples may have drifted — always check the source.') or move the plan content to the PR description.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Previous LOW findings were addressed (plan doc got a staleness caveat header, comment phrasing was rewritten). The Go-side changes are clean and well-tested — the MarkerDone injection, simplified dedup gate, and SSE event switch all look correct, and the BodyScansAsDone regression test is a particularly good addition. Two minor items remain: an inconsistent 'stop' action across user-facing strings, and the Flutter portion of the diff was truncated so I could not verify the dashboard_providers / AttentionBadge changes claimed in the plan.

### Issues

🟡 **daemon/internal/issues/pipeline.go:935** — The SSE `error` field tells the user to 'close the issue' as the stop action, but the fallback comment body (line 899) and the configuration guide both say 'remove the develop label.' Closing an issue and removing the develop label have very different consequences — closing discards the issue entirely, while delabeling just exits the develop pipeline. The SSE error string is what the Flutter UI renders in the toast/detail view, so users will see conflicting guidance depending on whether they read the GitHub comment or the app. Suggest aligning the SSE error to 'remove the develop label to stop here' for consistency.
🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:178** — The diff for this file was truncated so I could not verify the Flutter-side changes. The plan document header claims (a) `issueListRefreshProvider` is bumped on `issue_review_error`, (b) an `AttentionBadge` widget replaces the green LOW chip, and (c) reason-specific rendering for `auto_implement_no_changes`. On the base branch, the `issue_review_error` handler (lines 169-178) does NOT bump `issueListRefreshProvider` — confirm the PR adds it, otherwise the issue list won't refresh when the no-changes fallback fires and the user will see stale state until manual reload.

### Suggestions

- Align the SSE error string at pipeline.go:935 with the fallback comment and docs: replace 'or close the issue' with 'or remove the develop label to stop here.'
- Verify the truncated Flutter diff includes the issueListRefreshProvider bump in the issue_review_error handler — without it the dashboard won't auto-refresh on no-changes fallback events.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean re-review. All four previous findings were addressed (plan marked historical, misleading comment rewritten, format string divergence acknowledged). The three sub-fixes (MarkerDone in fallback body, SSE event type switch, simplified dedup gate) are well-structured with thorough test coverage — particularly the BodyScansAsDone regression test that validates the marker-collision invariant through the real ScanMarkers path. Dead code (MaxAutoImplementNoChanges, CountAutoImplementNoChanges) is fully removed with no dangling references. No new significant issues found.

### Issues

🟡 **daemon/internal/issues/pipeline.go:896** — Minor UX inconsistency: the fallback comment body tells the user 'To stop here, remove the develop label' but the SSE error payload (line 924) says 'or close the issue.' These are different actions (removing a label vs closing). The configuration guide correctly says 'remove the develop label to stop here.' Consider aligning the error string.

### Suggestions

- Consider adding a CHANGELOG or migration-notes entry for the SSE event type change (EventIssueReviewCompleted → EventIssueReviewError with new payload shape) so out-of-tree consumers can discover the breaking change without reading the PR diff.

---

PR #490, opened 2026-05-13
feat(issues/tier3): PR review-state vigilance on auto_implement PRs (#482)

### LINKED ISSUE #482: Develop tier does not monitor PR review state after auto_implement creates the PR
## Problem

Once `auto_implement` creates a PR and posts the `<!-- heimdallm:done -->` marker, the issue is permanently skipped by the fetcher. The daemon has **zero visibility** into what happens to that PR afterwards.

### Current behavior

1. `auto_implement` creates PR, stores `ActionTaken: "auto_implement"` + `PRCreated: N`, posts `MarkerDone`
2. Fetcher dedup (`fetcher.go:297-299`) unconditionally skips the issue on every subsequent poll:
   ```go
   if latest.ActionTaken == "auto_implement" && latest.PRCreated > 0 {
       return true, "already implemented (PR created)", nil
   }
   ```
3. Tier 3 state monitoring only checks open→closed transitions — it does not inspect PR review state (`APPROVED`, `CHANGES_REQUESTED`, `COMMENTED`)
4. The PR is re-reviewed by the PR review tier as a standalone PR, with no issue context and no ability to act on review feedback

### Expected behavior — PR review state vigilance

After `auto_implement` creates a PR, the daemon should monitor the PR's review state and react according to the review decision:

| PR Review State | Expected Action |
|---|---|
| **Approved** | No action needed — the PR will be merged by a human or automation |
| **Commented** | Respond to the comment author (acknowledge, clarify, provide context) |
| **Changes Requested** | Examine the requested changes, evaluate whether they are valid and in scope, and resolve them (push fixes to the same branch) |

### Gap summary

```
Issue #42 (develop) → auto_implement → PR #99 created → MarkerDone
                                                          ↓
Reviewer posts on PR #99:                  Issue #42: permanently skipped
                                                          ↓
  APPROVED         → nothing happens       Nobody acts.
  COMMENTED        → nobody responds       The PR rots.
  REQUEST_CHANGES  → nobody addresses it
```

### Implementation considerations

- The state monitoring system (Tier 3 / `CheckItem` + `HandleChange`) already watches PRs for state transitions — it could be extended to also check review state via the GitHub Reviews API
- When `CHANGES_REQUESTED` is detected, the daemon would need to:
  1. Fetch the review comments from the PR
  2. Check out the existing branch (not create a new one)
  3. Build a prompt that includes the original issue context + the review feedback
  4. Run the agent with write permissions to apply fixes
  5. Commit and push to the same branch
  6. Optionally re-request review from the reviewer who requested changes
- When `COMMENTED` is detected:
  1. Fetch the comment(s)
  2. Build a prompt with the original issue + PR context + comments
  3. Post a response comment on the PR
- The `alreadyProcessed` dedup gate needs to allow re-entry when the PR's review state warrants action, without creating infinite loops (e.g., the bot's own push triggers a new review cycle)
- Consider a cooldown or circuit breaker per PR to avoid infinite fix→review→fix loops

ivanmunozruiz:
## Summary

Closes #482.

Once `auto_implement` creates a PR the daemon used to stop watching it. This PR adds the **three-phase review-state vigilance flow** the issue asked for, designed around a strict AI-cost ceiling — the observation layer is always on (zero AI tokens), and the two opt-in response layers are **off by default** with hard per-PR-lifetime caps.

## Phase 1 — Observation (always on, $0 AI cost)

- New columns on `prs`: `external_review_state`, `external_reviewer`, `external_review_at`, `auto_implement_issue_id`, `review_response_count`, `review_fix_count`, `last_responded_at`. Idempotent ALTERs cover existing DBs.
- `github.GetPRReviews` wraps `GET /repos/{repo}/pulls/{n}/reviews`. Called only for PRs marked `auto_implement_issue_id != 0`.
- `issuepipeline.LatestExternalReviewState` collapses the reviews list to a single aggregate (CR > APPROVED > COMMENTED, bot filtered, DISMISSED dropped).
- `issues/pipeline.go` enrols the new PR in Tier 3's watch + marks its origin issue right after `CreatePullRequest`.
- Tier 3 `CheckItem` branches on `auto_implement_issue_id != 0`: fetches reviews, computes aggregate, writes to store, emits `pr_review_state_changed` SSE (with `prev_state` for diffing). Short-circuits the standard PR review codepath. The dispatch gate considers BOTH the aggregate state and the latest-review timestamp, so a fresh same-state review (second COMMENTED, follow-up CHANGES_REQUESTED after a no-changes advisory) is treated as a new trigger.
- `GET /issues` and `GET /issues/{id}` carry a `linked_pr` block when the PR has the auto_implement origin mark.
- Flutter `PRReviewStateBadge` widget renders APPROVED / CHANGES_REQUESTED / COMMENTED / FIX_PUSHED chips on the dashboard tile, issues list, and issue-detail header. `dashboard_providers.dart` consumes the new SSE event and refreshes the list live.

## Phase 2 — `ai.review_response` (opt-in, COMMENTED)

`issuepipeline.Responder` triggers when the new state is `COMMENTED`:
- Off by default. `Enabled = false`; flipping requires explicit TOML.
- `per_pr_lifetime` (default 5) — persistent counter on the PR row; capped runs emit `issue_review_error` with `reason="review_response_cap_exceeded"`.
- `cooldown_secs` (default 300) blocks back-to-back runs.
- **Trigger source is the Reviews API** (the latest non-bot `COMMENTED` review's body), not the conversation comments thread — a reviewer who writes text only in the review body still gets a reply.
- The originating issue title + body are embedded inside the `UNTRUSTED USER ISSUE BODY` fence so the agent reasons in the context of the original work item (#482 explicit requirement).
- Reviewer text passes through `sanitiseUntrustedFreeText` fences (#478 hardening).
- Empty agent output → counter advances, no comment posted.

## Phase 3 — `ai.review_fix` (opt-in, CHANGES_REQUESTED) — real git push

`issuepipeline.FixRunner` triggers when the new state is `CHANGES_REQUESTED`. **This is the real autonomous-fix path the issue asked for, not advisory mode.**

Architecture split:
- `Pipeline.RunFix` owns the git plumbing + agent run: detect CLI → `CheckoutNewBranch(HeadRef, HeadRef)` (clean checkout on the PR's tip) → `ensureAutoImplementWritePerms` → `ExecuteRaw` → `HasChanges`. If changes: `CommitAll` + `Push` to HeadRef (NEVER `CreatePR` — the PR already exists). If no changes: returns an advisory comment body.
- `main.prFixExecutor` owns the worktree lifecycle: `repoctx.Acquire(WorktreeToken="fix-pr-N", Mode=Write)`, defer `Release`. Hydrates HeadRef via `gh.GetPR` and CLI inputs via `AIForRepo` so per-repo overrides take effect.
- `FixRunner` owns cap / cooldown / bot-skip / state-flip. Consumes `FixResult{Pushed bool, CommentBody string}` and only flips `ExternalReviewState` to `FIX_PUSHED` when `Pushed=true` — a no-changes advisory keeps the state at `CHANGES_REQUESTED` so a reviewer with more context can re-trigger the runner within the lifetime cap (default 3).

The `FIX_PUSHED` re-arm guard at the Tier 3 level: a CR whose `SubmittedAt <= stored.ExternalReviewAt` is the same CR already addressed — suppressed, no re-dispatch. A genuinely fresh CR (`SubmittedAt > stored.ExternalReviewAt`) reactivates the cycle.

Security primitives carry over from auto_implement (#478):
- `ensureAutoImplementWritePerms` keeps `bypassPermissions` blocked.
- Both prompts wrap reviewer body + originating-issue body inside the untrusted-text fences with `sanitiseUntrustedFreeText`. Forged fence terminators are neutralised.
- `CommitAll`'s `sensitivePathPatterns` denylist refuses `.env`, private keys, credentials, kubeconfig, etc — prompt-injection asking the agent to exfiltrate via the PR diff gets caught at the commit boundary.

## Tier 3 routing matrix

| New external state | Phase | Routed to |
|---|---|---|
| `APPROVED` | n/a (terminal observation) | — |
| `COMMENTED` | 2 (if enabled) | `Responder.Run` |
| `CHANGES_REQUESTED` | 3 (if enabled) | `FixRunner.Run` → `Pipeline.RunFix` |
| `FIX_PUSHED` (set by FixRunner) | re-arm marker | — |

## Cost ceiling guarantees

| Feature | Off-by-default | Counter persisted | Max AI runs per PR |
|---|---|---|---|
| Observation (`GetPRReviews` per Tier 3 tick) | n/a (always on) | n/a | 0 (no AI) |
| `ai.review_response` | yes | `review_response_count` | `per_pr_lifetime` (default 5) |
| `ai.review_fix` (real push) | yes | `review_fix_count` | `per_pr_lifetime` (default 3) |

Misconfigured TOML (`per_pr_lifetime = 0` or negative) falls back to the constant default rather than meaning "unlimited". Flipping `enabled` is the only way to opt in.

## Tests

Phase 1 (observation):
- [x] `TestStore_PRReviewState_RoundTrip`, `_MarkPRAutoImplementOrigin`, `_IncrementPRReviewResponseCount_ReturnsNewValue`
- [x] `TestGetPRReviews_ParsesFields`
- [x] `TestLatestExternalReviewState_*` (5 cases)
- [x] `TestPipeline_AutoImplement_EnrollsPRInWatch`, `_MarksPRAutoImplementOrigin`, `_WatchEnrollerOptional`
- [x] `TestRefreshAutoImplementPRReviewState_*` (8 cases) — including fresh-same-state re-dispatch and the FIX_PUSHED re-arm guard
- [x] `TestHandleGetIssue_EmbedsLinkedPRReviewState`, `_NoLinkedPRWhenNotAutoImplement`

Phase 2 (Responder):
- [x] `TestResponder_DisabledByDefault_NoOp`, `_PerPRLifetimeCap`, `_CooldownRespected`, `_SkipsWhenLatestCommentIsBot`, `_NoReviewAtAll_NoOp`, `_HappyPath_RunsAndUpdates`, `_PromptSanitisesExternalText`, `_AgentReturnsEmpty_PostsNothing`, `_PromptEmbedsIssueContext`, `_ReadsReviewBodyNotConversation`

Phase 3 (FixRunner + Pipeline.RunFix):
- [x] `TestFixRunner_DisabledByDefault_NoOp`, `_PerPRLifetimeCap`, `_CooldownRespected`, `_BotOnlyCRIsNoOp`, `_PushedFlipsFixPushedState`, `_NoChangesDoesNotFlipFixPushed`, `_PassesOriginIssueToExecutor`, `_EmptyCommentBody_PostsNothing`
- [x] `TestPipelineRunFix_HappyPath_PushesToHeadBranch` — pins checkout/push target HeadRef, NO CreatePR
- [x] `TestPipelineRunFix_NoChanges_PostsAdvisoryWithoutPush`
- [x] `TestPipelineRunFix_Rejects{EmptyHeadRef,EmptyToken,EmptyWorkDir,MissingGit}`
- [x] `TestBuildFixAgentPrompt_EmbedsIssueAndSanitises`

Config:
- [x] `TestApplyDefaults_ReviewResponseDisabledByDefault`, `_PreservesExisting`, `_NegativeFallsBack`, `_ReviewFixDisabledByDefault`

CI:
- [x] `make test-docker` green across all packages
- [x] `flutter analyze` clean
- [x] `flutter test` passes (no new test failures)

## Docs

`configuration-guide.md` gains a new section after the auto_implement security note describing observation, the opt-in flags, the real-push semantics for Phase 3, and a cost-ceiling table.

## Out of scope (follow-ups)

- Per-repo overrides for `ai.review_response` / `ai.review_fix`.
- Surfacing the new caps as configurable knobs in the GUI (operators edit TOML for now).
- Cross-fork PR fix support (the executor refuses an empty HeadRef rather than attempting to push to a fork it can't access).

Plan committed at `docs/superpowers/plans/2026-05-13-issue-482-pr-review-state-vigilance.md`.


Muriano (approved):
## 🤖 Heimdallm AI Review

PR wires Tier 3 review-state vigilance for auto_implement PRs with responder/fix dispatchers behind feature flags. The refresh logic carefully separates persist/emit from dispatch to support retries, includes a FIX_PUSHED re-arm guard, and ships extensive table tests covering routing, no-op, fresh-CR reactivation, and dispatcher error propagation. Concerns: dispatchers use context.Background() instead of the CheckItem ctx, losing cancellation; worktree token `fix-pr-N` could collide across repos sharing PR numbers; minor double-pointer config access pattern.

### Issues

⚠️ **daemon/cmd/heimdallm/main_pr_review_state.go:0** — Dispatchers are invoked with context.Background() instead of a ctx threaded from CheckItem. Impact: long-running Responder/FixRunner calls cannot be cancelled on daemon shutdown or scheduler stop, risking hung goroutines and stalled graceful shutdown. Fix: change refreshAutoImplementPRReviewState to accept ctx context.Context and pass it through to responder.Run / fixRunner.Run; have CheckItem forward its own ctx.
⚠️ **daemon/cmd/heimdallm/main_pr_review_executors.go:0** — WorktreeToken is `fix-pr-%d` keyed only on PR number, not repo. Impact: if two distinct repositories happen to have PRs with the same number queued concurrently, the repoctx.Manager could conflate them or serialize unnecessarily depending on its keying. Fix: include the repo in the token, e.g. fmt.Sprintf("fix-pr-%s-%d", req.Repo, req.PRNumber) (sanitizing slashes as needed).

### Suggestions

- Consider exposing a single config-snapshot helper instead of repeating the cfgMu.Lock/defer Unlock + deref pattern in prReviewExecutor and prFixExecutor.
- The two dispatcher interfaces (reviewResponderDispatcher and reviewFixDispatcher) are structurally identical; consider a single named interface with two field names for clarity, or document the deliberate duplication.
- Log lines on dispatch failure already include err; consider also including the aggregate state and reviewer so production debugging doesn't need a second query.
- prFixExecutor.cfg is `**config.Config`; a `func() *config.Config` getter would be more idiomatic and avoid leaking the double-pointer to the struct field.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

PR wires Tier 3 review-state vigilance for auto_implement PRs with Responder/FixRunner dispatch, FIX_PUSHED re-arm guard, and comprehensive tests. Logic is well-reasoned and the persist/dispatch separation correctly enables retry without SSE flap. A few concerns around context propagation and config-reload safety.

### Issues

⚠️ **daemon/cmd/heimdallm/main_pr_review_state.go:127** — Responder.Run is called with context.Background() instead of the ctx passed to CheckItem. This disconnects dispatch from the scheduler's shutdown/cancellation signal — an in-flight responder/fixer run won't be cancelled on daemon stop and could block graceful shutdown. Plumb ctx through refreshAutoImplementPRReviewState.
⚠️ **daemon/cmd/heimdallm/main_pr_review_state.go:135** — Same context.Background() issue for fixRunner.Run — the FixRunner does git clones/agent runs that can be long-lived; losing cancellation here is more impactful than for the responder.
⚠️ **daemon/cmd/heimdallm/main_pr_review_executors.go:32** — prReviewExecutor holds **config.Config and dereferences under cfgMu, but the config-reload path must atomically swap the outer pointer for this pattern to be safe. If reload mutates fields in place rather than swapping the pointer, cfgMu here doesn't protect readers. Verify the reload path swaps the pointer; otherwise this is a data race.
🟡 **daemon/cmd/heimdallm/main.go:3275** — `stored, _ := a.store.GetPRByGithubID(item.GithubID)` silently swallows store errors. A transient DB error would make the branch fall through to the standard review codepath instead of returning the error to the StateWorker for backoff. Consider propagating the error or at minimum logging it.
🟡 **daemon/cmd/heimdallm/main_pr_review_state.go:92** — `stateMoved` uses `at.After(stored.ExternalReviewAt)` but if `at` is the zero time (no reviews) and stored is also zero, state may still differ (e.g., '' vs ''), so this is fine — but worth confirming LatestExternalReviewState returns a stable zero `at` when reviews list is empty to avoid spurious persists.
🟡 **daemon/cmd/heimdallm/main.go:569** — The two `func() string { loginMu.Lock(); ...; return cachedLogin }` closures duplicate the existing cachedAuthenticatedUser pattern — consider extracting a single accessor to avoid drift if locking semantics change.

### Suggestions

- Plumb the CheckItem ctx through refreshAutoImplementPRReviewState into Responder.Run / FixRunner.Run so daemon shutdown can cancel in-flight runs.
- Confirm config-reload swaps the *config.Config pointer atomically (rather than mutating in place) — the **Config + Mutex pattern only protects against torn reads if the outer pointer is replaced under the same lock.
- Propagate (or at least log) the error from GetPRByGithubID in CheckItem rather than discarding it with `_`; a store outage shouldn't silently degrade to the wrong branch.
- Extract the cachedLogin accessor closure into a single helper shared by both Responder/FixRunner wiring and cachedAuthenticatedUser to keep the locking discipline in one place.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: previous context.Background(), silent error swallow, and duplicated login-closure findings are all addressed. New test file provides strong coverage for state-change, no-op, dispatch routing, FIX_PUSHED re-arm guard, and fresh-CR re-fire cases. One previous concern remains unverified.

### Issues

⚠️ **daemon/cmd/heimdallm/main_pr_review_executors.go:32** — Still unresolved from prior review: prReviewExecutor (and prFixExecutor) hold **config.Config and dereference under cfgMu. This is only safe if the config-reload path atomically swaps the outer pointer while holding cfgMu. If reload mutates the underlying Config in place, cfgMu here protects nothing and concurrent readers race on AI.Primary/AI.Fallback/AI.ReviewResponse fields. Please confirm the reload path swaps the pointer (and ideally add a brief comment here documenting the invariant).

### Suggestions

- Consider documenting on prReviewExecutor/prFixExecutor that the **config.Config indirection requires the reload path to atomically swap *cfg (not mutate it in place) for cfgMu to provide useful synchronization.
- The new tests are thorough — the lastCtx capture in reviewStateDispatcher is a nice touch for pinning ctx propagation. Consider one more test asserting that a cancelled ctx is actually observed by the dispatcher (e.g. dispatcher checks ctx.Err()) to prevent regression on the context.Background() fix.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review: ctx propagation fix was applied correctly (refreshAutoImplementPRReviewState now accepts and forwards ctx to dispatchers, CheckItem forwards its own ctx, and tests assert lastCtx propagation). The WorktreeToken keying issue remains unaddressed — still keyed only on PR number without the repo.

### Issues

⚠️ **daemon/cmd/heimdallm/main_pr_review_executors.go:0** — WorktreeToken is still `fmt.Sprintf("fix-pr-%d", req.PRNumber)` — keyed only on PR number, not repo. Impact: two distinct repositories with the same PR number queued concurrently can collide or serialize on the same worktree slot in repoctx.Manager, depending on its keying. Fix: include the repo in the token, e.g. fmt.Sprintf("fix-pr-%s-%d", strings.ReplaceAll(req.Repo, "/", "-"), req.PRNumber).

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured feature adding review-state vigilance for auto_implement PRs with good test coverage and careful state machine design. Key concerns are around context propagation (context.Background in dispatchers prevents cancellation on shutdown), stale data passed to runners after state persist, and a permanent-error retry loop for cross-fork PRs. The FIX_PUSHED re-arm guard and decoupled persist/dispatch design are thoughtful.

### Issues

⚠️ **daemon/cmd/heimdallm/main_pr_review_state.go:98** — context.Background() in responder/fixRunner dispatch ignores the parent context from CheckItem. refreshAutoImplementPRReviewState doesn't accept a context parameter, so cancellation (e.g. daemon shutdown) and deadlines can't propagate to potentially long-running AI operations. The function should accept a context and pass it through to both Run calls.
⚠️ **daemon/cmd/heimdallm/main_pr_review_state.go:98** — Stale `stored` PR struct passed to dispatchers: after UpdatePRReviewState persists the new state/timestamp on line ~80, the in-memory `stored` pointer still holds old ExternalReviewState, ExternalReviewAt, and counter values. If the Responder or FixRunner reads these fields for cap/cooldown decisions, it operates on stale data. Either re-fetch from the store before dispatch, or update the struct fields in-place after the persist succeeds.
🟡 **daemon/cmd/heimdallm/main.go:3272** — GetPRByGithubID error silently discarded: `stored, _ := a.store.GetPRByGithubID(item.GithubID)` swallows a potential DB error. A transient store failure would make this PR silently fall through to the standard review codepath where it gets rejected by SkipReasonSelfAuthored — functionally benign but masks real failures from observability.
⚠️ **daemon/cmd/heimdallm/main_pr_review_executors.go:29** — GenerateReviewResponse discards the context parameter (`_ context.Context`), so the underlying ExecuteRaw call runs without cancellation or timeout. If the AI CLI hangs, this goroutine blocks indefinitely with no way to terminate it.
🟡 **daemon/cmd/heimdallm/main_pr_review_executors.go:85** — Empty head ref on cross-fork PRs produces a retryable error that will never succeed. Since errors from refreshAutoImplementPRReviewState keep LastSeen frozen, the StateWorker will retry this PR on every tick indefinitely (with backoff, but never giving up). Consider treating this as a permanent error — e.g., log a warning and return FixResult with a sentinel, or mark the watch item for cleanup.
🟡 **daemon/cmd/heimdallm/main.go:3272** — The auto_implement branch always returns `true, nil, nil` on success regardless of whether the review state actually changed. This resets backoff and advances LastSeen unconditionally. If the goal is to avoid polling again until GitHub's updated_at advances, this is fine, but the semantic overloading of `changed=true` with a nil snapshot to short-circuit HandleChange is subtle and could confuse future maintainers.

### Suggestions

- Add a `ctx context.Context` parameter to `refreshAutoImplementPRReviewState` and thread it through to the dispatcher Run calls instead of using context.Background(). Consider adding a timeout (e.g., 5 minutes) as a safety net for AI operations.
- After the successful UpdatePRReviewState persist, either re-fetch the PR from the store (`stored, _ = a.store.GetPRByGithubID(...)`) or update `stored.ExternalReviewState`, `stored.ExternalReviewAt`, and `stored.ExternalReviewer` in-place before passing to dispatchers.
- For the cross-fork PR case (empty head ref), consider enrolling a permanent skip or cleaning up the watch entry, since the condition is structural and won't resolve on retry.
- Thread the context through to prReviewExecutor.GenerateReviewResponse and use it in ExecuteRaw so that daemon shutdown can cancel in-flight AI calls.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

The PR adds review-state vigilance for auto_implement PRs with solid architecture (decoupled persist/dispatch, FIX_PUSHED re-arm guard, proper error propagation). However, it introduces a contract violation in CheckItem's return value and has a few design risks around the always-dispatch retry model.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:3295** — Contract violation: CheckItem returns (true, nil, nil) but the Tier3ItemChecker interface contract at scheduler/types.go:120-121 explicitly states 'callers can rely on [snap] being non-nil because RunTier3 only invokes HandleChange when changed == true'. HandleChange's existing nil guard (line 3324) prevents a panic, but this violates the documented invariant. If someone later removes the 'unnecessary' nil check based on the contract guarantee, auto_implement PRs will crash. Either update the interface contract to explicitly allow nil snap when changed=true, or return a minimal non-nil snapshot.
⚠️ **daemon/cmd/heimdallm/main_pr_review_state.go:95** — The always-dispatch model (dispatch fires on every tick regardless of stateMoved) amplifies a pre-existing cap-counter leak in FixRunner/Responder. Both runners call IncrementPRReview{Fix,Response}Count BEFORE executing work. If the executor fails transiently, the counter advances but no work completes. The next tick re-dispatches (error prevented LastSeen advance), re-enters the runner (cooldown passes since LastRespondedAt wasn't updated), and re-increments. Repeated transient failures can exhaust the per-PR lifetime cap with zero completed work. Consider moving the increment to after successful execution, or making it idempotent per-review-timestamp.
⚠️ **daemon/cmd/heimdallm/main_pr_review_state.go:98** — Runner errors now propagate to the StateWorker, which increases backoff. If a runner is persistently failing (misconfigured agent, executor down), the watch item's backoff grows indefinitely, delaying observation of all future reviews on that PR — including an APPROVED that would end the cycle. Consider capping the propagation to transient errors only, or applying a max-backoff bound on the StateWorker side so observation cadence is preserved.
🟡 **daemon/cmd/heimdallm/main_pr_review_state.go:82** — After UpdatePRReviewState persists the new state, the stale 'stored' PR snapshot (with old ExternalReviewState/Reviewer/At values) is passed to the dispatchers. Currently safe because Responder.Run and FixRunner.Run don't read these fields from the PR object, but fragile if either runner evolves to check pr.ExternalReviewState for routing logic. Consider re-reading the PR from the store after the update, or documenting this invariant on the Run interface.
🟡 **daemon/cmd/heimdallm/main_pr_review_executors.go:67** — prFixExecutor stores ghToken as a plain string field. This follows the existing pattern in the codebase but means the credential is held in memory for the lifetime of the adapter. If this struct is ever logged or serialized for debugging, the token could leak. Consider using a func() string accessor like loginFn to keep the token out of the struct.
🟡 **daemon/cmd/heimdallm/main_pr_review_state.go:1** — Comment-to-code ratio is very high throughout the new files. Multi-paragraph inline comments (e.g., lines 28-43 doc block, lines 56-69 FIX_PUSHED explanation, lines 91-112 dispatch rationale) duplicate design decisions that belong in the PR description or commit message. The code is clear enough to stand without most of these — the names 'stateMoved', 'fixPushedSameCR', and the switch cases are self-documenting.

### Suggestions

- Fix the contract violation by either updating the Tier3ItemChecker interface doc to explicitly permit nil snap when changed=true (for handled-inline cases), or returning a sentinel non-nil ItemSnapshot and adding a field like 'HandledInline bool' so HandleChange knows to skip without relying on a nil pointer.
- Add a test case where the reviewStateDispatcher returns an error to verify error propagation from CheckItem through the state handler — the diff's test stub has the 'err' field but the truncated diff may be missing the test that exercises it.
- Consider making the cap increment in FixRunner/Responder idempotent per review timestamp (e.g., only increment when the review's SubmittedAt is newer than the last incremented timestamp) so transient executor failures don't drain the lifetime budget.
- The botLoginAccessor extraction is a good cleanup — the on-disk code still has the duplicated inline closures for the two NewResponder/NewFixRunner calls. Make sure the diff's version (single extracted closure) lands.

---

PR #491, opened 2026-05-14
fix(rename): propagate GitHub repo/org renames across daemon state (#489)

### LINKED ISSUE #489: [URGENT] Handle GitHub repo and org renames — propagate slug changes to the daemon's database
## 🚨 Priority: URGENT

When a repository or its parent organization is renamed on GitHub, Heimdallm continues to operate against the **old slug**. Every reference in the daemon's state — config lists, SQLite rows, NATS watch entries, in-flight claim keys, SSE event payloads, working directory paths — is keyed on the historical `org/repo` string.

GitHub silently 301-redirects API calls from the old slug to the new one for a grace period, which masks the problem at first, but the redirect is not permanent and several code paths do not follow redirects at all. The result is a hard-to-diagnose drift between what GitHub considers the repo and what Heimdallm has stored.

## Symptoms observed / expected

- Issues and PRs continue to be polled under the old `org/repo` until the GitHub redirect expires, then polling silently 404s
- `prs.repo` / `issues.repo` rows in SQLite retain the old slug, so the Activity dashboard shows duplicate entries when the new slug is later auto-discovered
- `[ai.repos."old-org/old-repo"]` overrides stop applying because `AIForRepo()` looks up by current slug
- Working directories under `local_dir_base/old-org/old-repo` become stale; the auto-clone path may re-clone under the new name without cleaning up the old tree
- NATS watch_state KV entries (`pr.<id>`, `issue.<id>`) keep their old `repo` field, so state-checks publish events for a repo identifier that no longer exists upstream
- The `repositories` and `non_monitored` config lists retain the old name forever — neither `upsertDiscoveredRepos` nor any reconciliation pass renames an existing entry

## Affected scopes

Both **repo** renames (`org/old-name` → `org/new-name`) and **org** renames (`old-org/repo` → `new-org/repo`) must be handled. Org renames are the higher-impact case because they affect every repo under the org simultaneously.

## Expected behavior

Heimdallm should detect a rename and atomically update all internal state:

1. **Detection.** Either:
   - Reactive: when a GitHub API call returns a 301/302 redirect with a different repo `full_name`, treat the new value as canonical and reconcile
   - Proactive: periodically check the repos in `repositories` + `[ai.repos.*]` against `GET /repos/{owner}/{repo}` and compare `full_name`
   - GitHub also exposes the previous name via `GET /repos/{owner}/{repo}` → no extra endpoint needed for the lookup itself

2. **Reconciliation.** When a rename is detected:
   - Update `repositories` and `non_monitored` lists in config (persist to `config.toml`)
   - Rename the `[ai.repos."old"]` key to `[ai.repos."new"]` in TOML and live config
   - Update SQLite: `UPDATE prs SET repo = ? WHERE repo = ?`, same for `issues`, `reviews`, `issue_reviews`, any other table keyed on the slug
   - Update NATS KV `watch_state` entries to carry the new `repo` field
   - Rename working directories under `local_dir_base/` if they exist
   - Emit an SSE event (`repo_renamed` or similar) so the Flutter UI can refresh its lists without showing duplicates

3. **Idempotency.** A second detection of the same rename must be a no-op. A rename followed by a rename-back (rare but possible) must converge correctly.

4. **Concurrency safety.** The reconciliation must hold `cfgMu` and the relevant store locks so a concurrent poll cycle does not write rows under the old slug while the rename is in flight.

## Why this is urgent

- Silent data drift: the symptoms accumulate over time without obvious errors, but the daemon's view of the world diverges from GitHub's
- Issue tracking and auto_implement break for renamed repos with no operator-visible error — the issue just stops being processed
- The fix touches several layers (config writer, store schema migration, NATS KV, repoctx manager) so the longer this waits the more state will need backfilling

## Implementation considerations

- The store rename can be a single SQL transaction; emit one combined SSE event after commit
- For org renames, the lookup is per-repo — a single API call returns the new `full_name` for the old `org/repo` path
- Old → new mapping should be persisted (e.g. `repo_renames` table) so the audit trail survives daemon restarts
- Consider whether to keep the historical slug as an alias for a short window, or rewrite all references immediately. Recommendation: rewrite immediately; aliases multiply the failure modes
- The `[ai.repos.*]` TOML key rename has to go through `config.AtomicWriteTOML` so the file on disk reflects the change

ivanmunozruiz:
## Summary

Closes #489. Detects a GitHub repo or org rename and propagates the slug change atomically through every piece of daemon state keyed on the old `owner/name` string.

Full plan in `docs/superpowers/plans/2026-05-14-issue-489-repo-org-rename.md`. Built in 9 TDD steps; **all 9 landed in this PR**.

## What changed

1. **Store + `repo_renames` audit table** (`daemon/internal/store/rename.go`). `RenameRepo(old, new) (applied bool, err error)` bulk-updates `prs`/`issues`/`activity_log`/`watch_state` in one TX, writes audit row, idempotent. 5 tests.
2. **`Client.GetCanonicalFullName(repo)`** (`daemon/internal/github/canonical.go`). Single-call rename detector via `GET /repos/{owner}/{repo}`; 404 surfaces as `*APIError`. 2 tests.
3. **Config TOML rename helper** (`daemon/internal/config/writer_rename.go`). Substitutes the slug in `[github] repositories` + `non_monitored` lists, moves `[ai.repos."<old>"]`, and moves `[ai.orgs."<old-org>"]` when the org component changed. Operator-only TOML sections survive via `map[string]any` round-trip. 6 tests. Plus `Config.ApplyRename(old, new)` for the in-memory mirror, 2 tests.
4. **Reconciler** (`daemon/internal/rename/reconciler.go`). New package; orchestrates Store → in-memory config → on-disk TOML → worktree purge → SSE publish, under `cfgMu` for the config half. 5 tests.
5. **Probe goroutine** (`daemon/internal/rename/probe.go`). Iterates monitored repos at `ai.repo_rename_check_interval` (default `1h`, `"0"` disables), dispatches the reconciler on a mismatch. 4 tests.
6. **`main.go` wiring** + new SSE event `repo_renamed`. Probe runs in its own goroutine alongside Tier 2; independent cadence so a Tier-2 stall can't suppress rename detection.
7. **Flutter** consumes `repo_renamed` in `dashboard_providers.dart` to bump both the PR list and issue list refresh counters.
8. **Admin endpoint** `POST /admin/repo-rename` behind the existing API token. Runs the same reconciler the probe does; idempotent against subsequent probe ticks. 5 handler tests.
9. **Docs** in `docs/configuration-guide.md` — new section under Repository Monitoring.

## Worktree strategy

Delete-and-reclone via `repoctx.Manager.Purge`, **not** `os.Rename`. `os.Rename` is fragile across filesystem boundaries, hits permission edge cases on macOS bundles, and races with in-flight worktree acquires. The next acquire of the new slug clones fresh — a bounded, well-tested path. Purge failure is non-fatal: the SSE payload carries `worktree_purged: false` so the dashboard can badge the inconsistency.

## Out of scope (follow-ups documented in the plan)

- Webhook-driven detection (would eliminate the 1h probe but requires operator infra).
- Cross-fork base-repo renames mid-PR review.
- Backfill `repo_renames` for renames that happened before this PR landed.
- GUI toggle for `repo_rename_check_interval`.
- `upsertDiscoveredRepos` race window where the discovery poll could observe a PR under the new slug before the reconciler commits (mitigated naturally by the next reconciler tick + the discovery poll cadence being orders of magnitude slower than the rename critical section).

## Test plan

- [x] `make test-docker ./...` clean across the daemon
- [x] `flutter test` clean (176 tests)
- [ ] Manual smoke: rename a sandbox repo on GitHub, wait one probe tick, confirm: SSE `repo_renamed` fires, dashboard refreshes, `repo_renames` audit row written, config TOML rewritten, old worktree dir gone, store rows for old slug = 0
- [ ] Manual smoke: trigger `POST /admin/repo-rename` directly with curl, confirm the same end state without waiting for the probe

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -714,6 +714,30 @@ func main() {
 			runTier2(ctx, adapter, limiter, prReviewPublisher, broker, tier2ConfigFn, tier2RepoConcurrencyFn, reposChan, pollInterval, coldStart)
 		}()
 
+		// Repo/org rename probe (#489). Detects when GitHub has
+		// renamed a monitored repo (or its parent org) and dispatches
+		// the reconciler to propagate the new slug across SQLite,
+		// config TOML, in-memory config, and worktrees. Interval "0"
+		// disables — operators can still trigger rename manually via
+		// POST /admin/repo-rename. The probe runs its own goroutine
+		// because its cadence (1h default) is orders of magnitude
+		// longer than Tier 2's, and we want it independent of Tier 2
+		// failures.
+		cfgMu.Lock()
+		renameInterval := parseRenameProbeInterval(cfg.AI.RepoRenameCheckInterval)
+		cfgMu.Unlock()
+		if renameInterval > 0 {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				probe := newRenameProbe(ctx, cfg, &cfgMu, ghClient, s, repoCtx, broker, cfgPath, renameInterval)
+				probe.Run(ctx)
+			}()
+			slog.Info("rename probe: started", "interval", renameInterval)
+		} else {
+			slog.Info("rename probe: disabled (ai.repo_rename_check_interval=0)")
+		}
+
 		slog.Info("pollers: started",
 			"discovery", discoveryInterval,
 			"poll", pollInterval)
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -325,9 +325,10 @@ func main() {
 	// Manual rename trigger for POST /admin/repo-rename (#489).
 	// Constructs a one-shot reconciler with the same deps the probe
 	// uses so manual triggers and automatic detection share idempotency
-	// guarantees end-to-end.
+	// guarantees end-to-end. Reuses srv.TOMLMu() so the rewrite races
+	// safely with concurrent PATCH /config writes.
 	srv.SetRepoRenameFn(func(ctx context.Context, oldRepo, newRepo string) error {
-		reconciler := newRenameReconciler(cfg, &cfgMu, s, repoCtx, broker, cfgPath)
+		reconciler := newRenameReconciler(cfg, &cfgMu, srv.TOMLMu(), s, repoCtx, broker, cfgPath)
 		return reconciler.Run(ctx, oldRepo, newRepo)
 	})
 	srv.SetCleanClonesFn(func(ctx context.Context) (int, error) {
@@ -738,7 +739,7 @@ func main() {
 			wg.Add(1)
 			go func() {
 				defer wg.Done()
-				probe := newRenameProbe(ctx, cfg, &cfgMu, ghClient, s, repoCtx, broker, cfgPath, renameInterval)
+				probe := newRenameProbe(ctx, cfg, &cfgMu, srv.TOMLMu(), ghClient, s, repoCtx, broker, cfgPath, renameInterval)
 				probe.Run(ctx)
 			}()
 			slog.Info("rename probe: started", "interval", renameInterval)
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -322,6 +322,14 @@ func main() {
 		cfgMu.Unlock()
 		return repoCtx.Purge(ctx, repo, aiCfg.CloneDir)
 	})
+	// Manual rename trigger for POST /admin/repo-rename (#489).
+	// Constructs a one-shot reconciler with the same deps the probe
+	// uses so manual triggers and automatic detection share idempotency
+	// guarantees end-to-end.
+	srv.SetRepoRenameFn(func(ctx context.Context, oldRepo, newRepo string) error {
+		reconciler := newRenameReconciler(cfg, &cfgMu, s, repoCtx, broker, cfgPath)
+		return reconciler.Run(ctx, oldRepo, newRepo)
+	})
 	srv.SetCleanClonesFn(func(ctx context.Context) (int, error) {
 		cfgMu.Lock()
 		cfgSnap := cfg
--- a/daemon/cmd/heimdallm/main_rename.go
+++ b/daemon/cmd/heimdallm/main_rename.go
@@ -14,10 +14,18 @@ import (
 	"github.com/heimdallm/daemon/internal/store"
 )
 
+// minRenameProbeInterval is the lower bound enforced on operator
+// input. Below 1 minute the probe would burn GitHub rate-limit
+// budget (one GET per repo per tick) for no real detection gain —
+// renames are by nature rare events. Misconfigurations get clamped
+// up with an audit-log warn rather than silently honoured.
+const minRenameProbeInterval = time.Minute
+
 // parseRenameProbeInterval parses the operator-facing duration string
 // for `ai.repo_rename_check_interval`. Empty / unparseable falls back
 // to 1h; literal "0" returns 0 so the caller knows the probe is
-// disabled. Returning <= 0 from here is the disable signal.
+// disabled. Returning <= 0 from here is the disable signal. Positive
+// values below minRenameProbeInterval are clamped up.
 func parseRenameProbeInterval(raw string) time.Duration {
 	if raw == "0" {
 		return 0
@@ -30,6 +38,11 @@ func parseRenameProbeInterval(raw string) time.Duration {
 		slog.Warn("rename probe: unparseable interval, falling back to default",
 			"raw", raw, "default", rename.DefaultProbeInterval)
 		return rename.DefaultProbeInterval
+	}
+	if d < minRenameProbeInterval {
+		slog.Warn("rename probe: interval below floor, clamping",
+			"raw", raw, "floor", minRenameProbeInterval)
+		return minRenameProbeInterval
 	}
 	return d
 }
--- a/daemon/cmd/heimdallm/main_rename.go
+++ b/daemon/cmd/heimdallm/main_rename.go
@@ -1,0 +1,125 @@
+package main
+
+import (
+	"context"
+	"log/slog"
+	"sync"
+	"time"
+
+	"github.com/heimdallm/daemon/internal/config"
+	gh "github.com/heimdallm/daemon/internal/github"
+	"github.com/heimdallm/daemon/internal/rename"
+	"github.com/heimdallm/daemon/internal/repoctx"
+	"github.com/heimdallm/daemon/internal/sse"
+	"github.com/heimdallm/daemon/internal/store"
+)
+
+// parseRenameProbeInterval parses the operator-facing duration string
+// for `ai.repo_rename_check_interval`. Empty / unparseable falls back
+// to 1h; literal "0" returns 0 so the caller knows the probe is
+// disabled. Returning <= 0 from here is the disable signal.
+func parseRenameProbeInterval(raw string) time.Duration {
+	if raw == "0" {
+		return 0
+	}
+	if raw == "" {
+		return rename.DefaultProbeInterval
+	}
+	d, err := time.ParseDuration(raw)
+	if err != nil || d <= 0 {
+		slog.Warn("rename probe: unparseable interval, falling back to default",
+			"raw", raw, "default", rename.DefaultProbeInterval)
+		return rename.DefaultProbeInterval
+	}
+	return d
+}
+
+// tomlPersister adapts config.RenameRepoInTOML to rename.Persister so
+// the rename package does not depend on the config package directly
+// (keeps the internal/rename surface small and testable with fakes).
+type tomlPersister struct{}
+
+func (tomlPersister) Rename(path, oldRepo, newRepo string) error {
+	return config.RenameRepoInTOML(path, oldRepo, newRepo)
+}
+
+// renameRepoListFn returns the currently-configured monitored repo
+// set (excluding non_monitored) under cfgMu so the probe observes
+// runtime config reloads.
+func renameRepoListFn(cfg *config.Config, cfgMu *sync.Mutex) func() []string {
+	return func() []string {
+		cfgMu.Lock()
+		defer cfgMu.Unlock()
+		// Probe operates on the explicitly-configured set only —
+		// discovery-only repos are excluded because the probe writes
+		// into config TOML on a rename, and rewriting an entry the
+		// operator never explicitly added would surprise them. The
+		// next discovery cycle re-publishes the canonical name
+		// regardless, so coverage isn't lost.
+		out := make([]string, 0, len(cfg.GitHub.Repositories))
+		out = append(out, cfg.GitHub.Repositories...)
+		return out
+	}
+}
+
+// newRenameProbe constructs the Reconciler + Probe with their real
+// daemon deps. It is invoked once per pollers cycle (re-construction
+// is safe; the probe is goroutine-local), but called from a single
+// goroutine in startPollers so there is no concurrent-construction
+// concern.
+func newRenameProbe(
+	_ context.Context,
+	cfg *config.Config,
+	cfgMu *sync.Mutex,
+	ghClient *gh.Client,
+	s *store.Store,
+	repoCtx *repoctx.Manager,
+	broker *sse.Broker,
+	cfgPath string,
+	interval time.Duration,
+) *rename.Probe {
+	reconciler := newRenameReconciler(cfg, cfgMu, s, repoCtx, broker, cfgPath)
+	return rename.NewProbe(rename.ProbeDeps{
+		Probe:      ghClient,
+		Dispatcher: reconciler,
+		Repos:      renameRepoListFn(cfg, cfgMu),
+		Interval:   interval,
+	})
+}
+
+// newRenameReconciler is the constructor seam that the admin endpoint
+// (step 8) and the probe share — both wire the same dependencies.
+//
+// CloneDir comes from the global AI default rather than per-repo
+// AIForRepo: the rename probe purges the OLD slug whose per-repo
+// override is about to be removed, so reading the global default is
+// the only path that stays valid across the rename. Operators with
+// per-repo CloneDir overrides should rely on next-acquire reclone to
+// repopulate under the new override.
+func newRenameReconciler(
+	cfg *config.Config,
+	cfgMu *sync.Mutex,
+	s *store.Store,
+	repoCtx *repoctx.Manager,
+	broker *sse.Broker,
+	cfgPath string,
+) *rename.Reconciler {
+	cfgMu.Lock()
+	cloneDir := cfg.AI.CloneDir
+	cfgMu.Unlock()
+	if cloneDir == "" {
+		cloneDir = repoctx.DefaultCloneDir()
+	}
+	return rename.NewReconciler(rename.Deps{
+		Store:     s,
+		Persister: tomlPersister{},
+		Purger:    repoCtx,
+		Publisher: broker,
+		CfgMu:     cfgMu,
+		ApplyConfig: func(oldRepo, newRepo string) {
+			cfg.ApplyRename(oldRepo, newRepo)
+		},
+		CfgPath:  cfgPath,
+		CloneDir: cloneDir,
+	})
+}
--- a/daemon/cmd/heimdallm/main_rename.go
+++ b/daemon/cmd/heimdallm/main_rename.go
@@ -71,6 +71,7 @@ func newRenameProbe(
 	_ context.Context,
 	cfg *config.Config,
 	cfgMu *sync.Mutex,
+	tomlMu *sync.Mutex,
 	ghClient *gh.Client,
 	s *store.Store,
 	repoCtx *repoctx.Manager,
@@ -78,7 +79,7 @@ func newRenameProbe(
 	cfgPath string,
 	interval time.Duration,
 ) *rename.Probe {
-	reconciler := newRenameReconciler(cfg, cfgMu, s, repoCtx, broker, cfgPath)
+	reconciler := newRenameReconciler(cfg, cfgMu, tomlMu, s, repoCtx, broker, cfgPath)
 	return rename.NewProbe(rename.ProbeDeps{
 		Probe:      ghClient,
 		Dispatcher: reconciler,
@@ -88,7 +89,9 @@ func newRenameProbe(
 }
 
 // newRenameReconciler is the constructor seam that the admin endpoint
-// (step 8) and the probe share — both wire the same dependencies.
+// (step 8) and the probe share — both wire the same dependencies,
+// including the shared tomlMu that serializes config-TOML writes with
+// the HTTP PATCH handlers (#489 race fix).
 //
 // CloneDir comes from the global AI default rather than per-repo
 // AIForRepo: the rename probe purges the OLD slug whose per-repo
@@ -99,6 +102,7 @@ func newRenameProbe(
 func newRenameReconciler(
 	cfg *config.Config,
 	cfgMu *sync.Mutex,
+	tomlMu *sync.Mutex,
 	s *store.Store,
 	repoCtx *repoctx.Manager,
 	broker *sse.Broker,
@@ -116,6 +120,7 @@ func newRenameReconciler(
 		Purger:    repoCtx,
 		Publisher: broker,
 		CfgMu:     cfgMu,
+		TOMLMu:    tomlMu,
 		ApplyConfig: func(oldRepo, newRepo string) {
 			cfg.ApplyRename(oldRepo, newRepo)
 		},
--- a/daemon/cmd/heimdallm/main_rename_interval_test.go
+++ b/daemon/cmd/heimdallm/main_rename_interval_test.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"testing"
+	"time"
+
+	"github.com/heimdallm/daemon/internal/rename"
+)
+
+// TestParseRenameProbeInterval_DisabledByZero pins the operator
+// escape hatch — "0" returns 0 so the wiring in main.go knows to
+// skip starting the probe goroutine entirely.
+func TestParseRenameProbeInterval_DisabledByZero(t *testing.T) {
+	if got := parseRenameProbeInterval("0"); got != 0 {
+		t.Errorf(`parseRenameProbeInterval("0") = %v, want 0 (disabled)`, got)
+	}
+}
+
+// TestParseRenameProbeInterval_EmptyUsesDefault pins the
+// applyDefaults contract: an empty string after config-load means
+// the operator left the knob untouched, so use the default cadence.
+func TestParseRenameProbeInterval_EmptyUsesDefault(t *testing.T) {
+	if got := parseRenameProbeInterval(""); got != rename.DefaultProbeInterval {
+		t.Errorf(`parseRenameProbeInterval("") = %v, want %v`, got, rename.DefaultProbeInterval)
+	}
+}
+
+// TestParseRenameProbeInterval_UnparseableUsesDefault keeps a typo
+// (e.g. "1hour" instead of "1h") from disabling the probe silently.
+func TestParseRenameProbeInterval_UnparseableUsesDefault(t *testing.T) {
+	if got := parseRenameProbeInterval("not a duration"); got != rename.DefaultProbeInterval {
+		t.Errorf(`unparseable input = %v, want %v`, got, rename.DefaultProbeInterval)
+	}
+}
+
+// TestParseRenameProbeInterval_ClampsBelowFloor protects GitHub
+// rate-limit budget against an operator setting an aggressively
+// low interval (e.g. "1s" while debugging). The probe fires one
+// GET per monitored repo per tick; <1m is never the right answer.
+func TestParseRenameProbeInterval_ClampsBelowFloor(t *testing.T) {
+	for _, raw := range []string{"1s", "30s", "59s"} {
+		got := parseRenameProbeInterval(raw)
+		if got != minRenameProbeInterval {
+			t.Errorf(`parseRenameProbeInterval(%q) = %v, want %v (clamped to floor)`,
+				raw, got, minRenameProbeInterval)
+		}
+	}
+}
+
+// TestParseRenameProbeInterval_HonoursValueAtOrAboveFloor pins the
+// non-clamp path so the floor only kicks in for values that need
+// it — operator-specified 1m or above passes through unchanged.
+func TestParseRenameProbeInterval_HonoursValueAtOrAboveFloor(t *testing.T) {
+	cases := []struct {
+		raw  string
+		want time.Duration
+	}{
+		{"1m", time.Minute},
+		{"5m", 5 * time.Minute},
+		{"2h", 2 * time.Hour},
+	}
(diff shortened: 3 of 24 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Solid, well-tested implementation of repo/org rename propagation. Code is cleanly layered (probe → reconciler → persister), uses interfaces for testability, and covers happy/sad paths with unit tests. No blocking issues found; a few minor concerns around TOML comment preservation, shadowing built-in `new`, and probe startup latency.

### Issues

⚠️ **daemon/internal/config/writer_rename.go:0** — RenameRepoInTOML round-trips through map[string]any and re-encodes via AtomicWriteTOML, which (with BurntSushi/toml) strips comments and may reorder keys. Operators who maintain hand-written config.toml with inline documentation will silently lose comments on every detected rename. Impact: surprise loss of operator annotations. Fix: document this behavior explicitly in the godoc and/or in the operator-facing docs for ai.repo_rename_check_interval, or use a comment-preserving TOML library (e.g. pelletier/go-toml/v2 with raw editing) if preservation is required.
🟡 **daemon/internal/rename/probe.go:0** — Probe.Run uses time.NewTicker which fires for the first time only after Interval has elapsed; with the 1h default, a rename that happened during downtime is not detected until an hour after daemon start. Impact: up to ~1h detection latency on restart for already-renamed repos. Fix: run an initial Tick(ctx) before entering the select loop (or document the startup delay explicitly).
🟡 **daemon/internal/config/writer_rename.go:0** — replaceInTOMLList and replaceInStringSlice use parameter name `new`, shadowing the Go built-in. While legal, it triggers lint warnings (predeclared) and reduces readability. Fix: rename to `newRepo`/`replacement`.
🟡 **daemon/internal/rename/probe.go:0** — 404 check uses the magic literal `404` instead of http.StatusNotFound (already imported in canonical.go and used in tests). Minor consistency issue. Fix: import net/http and compare against http.StatusNotFound.

### Suggestions

- Consider adding a metric (counter) for renames detected/dispatched and probe errors, so operators can alert on probe health without scraping logs.
- The manual endpoint reconstructs a Reconciler on each POST /admin/repo-rename, while the probe reuses a single instance — harmonize by reusing a single reconciler for both paths (no functional issue, just less allocation and a single source of truth).
- renameRepoListFn copies cfg.GitHub.Repositories under cfgMu; consider also exposing a way to include discovery-only repos in the future behind a flag, since the current exclusion is a deliberate design choice that should be re-evaluable.
- Add an integration-style test that exercises Reconciler.Run end-to-end (SQLite + TOML + ApplyConfig + Purger fake) to pin the ordering guarantees described in the package docs — current tests cover probe and TOML writer in isolation but the reconciler's contract relies on a specific sequencing.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Well-structured feature that adds automatic GitHub repo/org rename detection and propagation across daemon state (SQLite, TOML config, in-memory config, worktrees). The design cleanly separates the probe (detection) from the reconciler (propagation) with testable interfaces. However, there is a TOML write race condition and a few operational gaps worth addressing before merge.

### Issues

🔴 **daemon/internal/config/writer_rename.go:30** — TOML write race condition: RenameRepoInTOML performs an unprotected read-modify-write on the config TOML file. The server serializes its own TOML writes through srv.tomlMu (see patchTOML at handlers.go:224), but the reconciler bypasses that mutex entirely. If a PATCH endpoint and a rename reconciler write concurrently, one will clobber the other's changes. Consider either exposing tomlMu to the reconciler via Deps, or routing the rename write through the server's patchTOML pipeline.
⚠️ **daemon/internal/rename/probe.go:97** — Probe.Run uses time.NewTicker which only fires after one full interval elapses. After daemon start or config reload, rename detection is delayed by up to the configured interval (default 1h). Consider calling p.Tick(ctx) once before entering the ticker loop so renames are detected promptly on startup.
⚠️ **daemon/cmd/heimdallm/main_rename.go:23** — parseRenameProbeInterval accepts any positive duration with no minimum floor. A misconfigured value like '1s' with many monitored repos would send one GET per repo per second, easily exceeding GitHub's 5000 requests/hour rate limit. Consider enforcing a minimum (e.g. 5m) and logging a warning when the operator sets something below it.
⚠️ **daemon/cmd/heimdallm/main_rename.go:56** — renameRepoListFn only returns cfg.GitHub.Repositories, excluding NonMonitored repos. If a non-monitored repo is renamed on GitHub, the daemon's config will hold a stale slug indefinitely. The comment explains the rationale (avoiding surprise TOML rewrites), but consider at least logging a warning when a non-monitored repo's canonical name diverges, so operators know their config is out of date.
🟡 **daemon/cmd/heimdallm/main_rename.go:97** — newRenameReconciler captures cfg.AI.CloneDir once at construction time. Within the pollers lifecycle this is correct (the probe is re-created on reload), but the comment should explicitly note that the probe's reconciler is scoped to the pollers lifecycle and gets a fresh CloneDir on each startPollers invocation — otherwise a reader might worry about staleness across reloads.
🟡 **daemon/internal/config/writer_rename.go:88** — orgOf silently returns empty string for malformed input (e.g. missing slash), collapsing the org-rename branch to a no-op. While the defensive approach is intentional, this could mask bugs if a caller ever passes an invalid slug — consider a debug-level log when the input is non-empty but malformed.

### Suggestions

- Route the reconciler's TOML write through the same mutex the server uses (tomlMu), or use file-level locking (flock) as a process-wide serialization mechanism, to prevent lost-update races between PATCH endpoints and rename propagation.
- Add an initial p.Tick(ctx) call at the top of Probe.Run before entering the ticker loop, guarded by a context check, so renames detected between polling restarts are picked up immediately rather than after one full interval.
- Enforce a minimum probe interval (e.g. 5 minutes) in parseRenameProbeInterval to prevent accidental rate-limit exhaustion, and document the minimum in the TOML field comment.
- The reconciler.go code was truncated in the diff — ensure the full reconciler is reviewed for correct lock ordering (cfgMu acquisition around ApplyConfig and TOML writes), idempotency of SQLite rename operations, and proper error rollback if a mid-pipeline step fails.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #491: rename propagation pipeline. The diff is substantively the same as previously reviewed. Checking whether prior findings have been addressed: (1) TOML comment-loss not addressed in godoc; (2) Probe.Run still uses time.NewTicker without an initial Tick; (3) `new` parameter shadowing still present in replaceInTOMLList/replaceInStringSlice; (4) magic `404` literal still in probe.go. None of the prior findings appear addressed in the visible diff.

### Issues

⚠️ **daemon/internal/config/writer_rename.go:0** — RenameRepoInTOML still round-trips through map[string]any and re-encodes via AtomicWriteTOML, which strips comments and may reorder keys with BurntSushi/toml. Operators maintaining hand-written config.toml with inline documentation will silently lose comments on every detected rename — and since the probe runs hourly by default with auto-detection, this can happen without operator action. Impact: surprise loss of operator annotations in config.toml. Fix: document this behavior explicitly in the godoc of RenameRepoInTOML and in operator-facing docs for ai.repo_rename_check_interval, or switch to a comment-preserving TOML library.
🟡 **daemon/internal/rename/probe.go:0** — Probe.Run uses time.NewTicker which fires for the first time only after Interval elapses; with the 1h default, a rename that happened during downtime is not detected until an hour after daemon start. Impact: up to ~1h detection latency on restart for already-renamed repos. Fix: run an initial Tick(ctx) before entering the select loop, or document the startup delay explicitly.
🟡 **daemon/internal/config/writer_rename.go:0** — replaceInTOMLList and replaceInStringSlice still use parameter name `new`, shadowing the Go built-in. Triggers predeclared lint warnings and reduces readability. Fix: rename to `newRepo` or `replacement`.
🟡 **daemon/internal/rename/probe.go:0** — 404 check still uses the magic literal `404` instead of http.StatusNotFound (already used in canonical.go and tests). Minor consistency issue. Fix: import net/http and compare against http.StatusNotFound.

### Suggestions

- Consider emitting a structured SSE event when the probe detects a 404 (repo deleted) so operators are notified, rather than only logging at Info level.
- Add a metric/counter for rename detections and dispatcher failures to make this observable in production.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review: all four findings from the previous review remain unaddressed in the current diff. The author has not modified RenameRepoInTOML's godoc, Probe.Run's startup behavior, the `new` parameter name shadowing, or the magic 404 literal. Code is otherwise solid and well-tested.

### Issues

⚠️ **daemon/internal/config/writer_rename.go:0** — RenameRepoInTOML still round-trips through map[string]any and re-encodes via AtomicWriteTOML, which strips comments and may reorder keys. With the hourly probe enabled by default, operator-authored comments in config.toml will be silently lost on auto-detected renames. Impact: surprise loss of operator annotations. Fix: document this behavior explicitly in the godoc of RenameRepoInTOML and in operator-facing docs for ai.repo_rename_check_interval, or switch to a comment-preserving TOML library.

### Suggestions

- Probe.Run still uses time.NewTicker without an initial Tick; with the 1h default a rename occurring during daemon downtime is not detected until ~1h after restart. Run an initial Tick(ctx) before entering the select loop, or document the startup delay.
- replaceInTOMLList and replaceInStringSlice still use parameter name `new`, shadowing the Go built-in. Rename to `newRepo` or `replacement` to avoid predeclared-identifier lint warnings.
- probe.go:74 still uses the magic literal `404` instead of http.StatusNotFound, while canonical.go and tests use the constant. Import net/http and compare against http.StatusNotFound for consistency.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #491. The four previous findings remain unaddressed in the visible diff: RenameRepoInTOML godoc still does not warn operators about comment/formatting loss on round-trip; replaceInTOMLList and replaceInStringSlice still shadow the `new` builtin; probe.go still uses magic literal 404 instead of http.StatusNotFound; and Probe.Run startup-tick behavior is not visible as fixed (diff truncated, but no evidence of an initial Tick before the ticker loop). The new integration test (integration_test.go) is a strong addition — it pins the partial-failure recovery contract end-to-end across multiple ticks and explicitly documents the two pre-fix bugs (ApplyConfig ordering + applied=false short-circuit). No new high-severity issues found.

### Issues

⚠️ **daemon/internal/config/writer_rename.go:0** — STILL UNADDRESSED from prior review: RenameRepoInTOML round-trips through map[string]any and AtomicWriteTOML, which strips comments and may reorder keys. With the probe running hourly by default and auto-detecting renames, operator-maintained inline documentation in config.toml will silently disappear on the first rename. Godoc was expanded but still does not warn about this. Fix: add an explicit 'Note: comments and key ordering in the source TOML are NOT preserved' line to the godoc and to the operator-facing docs for ai.repo_rename_check_interval, or move to a comment-preserving TOML library.

### Suggestions

- STILL UNADDRESSED: rename `new` parameter in replaceInTOMLList and replaceInStringSlice to `newRepo`/`replacement` — shadows Go builtin and triggers predeclared-identifier lint.
- STILL UNADDRESSED: probe.go uses magic literal 404; replace with http.StatusNotFound for consistency with canonical.go and tests.
- STILL UNADDRESSED (unless hidden in truncated diff): Probe.Run should call Tick(ctx) once before entering the ticker select loop, otherwise a rename that occurred during daemon downtime is not detected for up to one hour after restart. Alternatively, document the startup delay explicitly.
- The integration test is excellent — consider also adding a tick-loop test where canonical lookups intermittently 404 to pin the 'skip this repo, retry next tick' contract.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Re-review: none of the six previous findings appear addressed in the current diff. The HIGH-severity TOML write race between the server's patchTOML (guarded by srv.tomlMu) and the reconciler's RenameRepoInTOML (unguarded) remains the most critical issue. Three MEDIUM findings (no initial probe tick, no minimum interval floor, non-monitored repos silently ignored) are also still present. The feature design and test coverage are solid, but the concurrency gap should be resolved before merge.

### Issues

🔴 **daemon/internal/config/writer_rename.go:30** — TOML write race (STILL UNADDRESSED from prior review): RenameRepoInTOML performs an unprotected read-modify-write on the TOML file. The server serializes its own TOML writes through srv.tomlMu (handlers.go:67, patchTOML at handlers.go:224), but the reconciler bypasses that mutex entirely — it calls config.RenameRepoInTOML directly via the tomlPersister adapter. A concurrent PATCH /config and rename reconciler write will race. Fix: either thread srv.tomlMu into the reconciler's Deps (alongside CfgMu), or route rename writes through a shared file-level lock accessible to both the server and the reconciler.
⚠️ **daemon/internal/rename/probe.go:97** — No initial tick on startup (STILL UNADDRESSED): Probe.Run enters a time.NewTicker loop that only fires after one full interval (default 1h). After a daemon restart or config reload, renames detected during downtime won't be acted upon for up to an hour. Adding a single p.Tick(ctx) call before the ticker loop would close this gap at zero cost.
⚠️ **daemon/cmd/heimdallm/main_rename.go:23** — No minimum interval floor (STILL UNADDRESSED): parseRenameProbeInterval accepts any positive duration. With '1s' and 50 monitored repos, the probe would issue 50 GETs/second (180k/hour), blowing through GitHub's 5000 req/hour limit in under 2 minutes. Enforce a minimum (e.g., 5m) or at least log a loud warning below a sane threshold.
⚠️ **daemon/cmd/heimdallm/main_rename.go:56** — Stale non_monitored slugs are invisible (STILL UNADDRESSED): renameRepoListFn excludes NonMonitored repos from probe coverage. The rationale (avoiding surprise TOML rewrites) is sound, but a renamed non_monitored repo will silently rot in the config forever. At minimum, log a warning when a non-monitored repo's canonical name diverges so operators know their config is stale.

### Suggestions

- Introduce a shared file-level mutex (or route all TOML writes through a single serialization point) so RenameRepoInTOML and patchTOML cannot race. The simplest approach: add a *sync.Mutex to the Persister interface or Deps, and have main.go pass &srv.tomlMu.
- Add p.Tick(ctx) before the ticker loop in Probe.Run so renames are detected immediately on startup rather than after the first interval elapses.
- Clamp the parsed interval to a minimum of 5 minutes in parseRenameProbeInterval and log a warning when the operator-specified value falls below it.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of rename propagation PR. The previous medium finding about TOML comment/key-ordering loss in RenameRepoInTOML remains unaddressed — the godoc was expanded but still omits the operator-visible warning that comments and key ordering are not preserved across the round-trip through map[string]any + AtomicWriteTOML. With the probe enabled by default at 1h cadence and auto-detecting renames, operator inline documentation will silently vanish on the first rename. All other dimensions (logic, security, tests, observability) look solid; integration test covers the partial-failure recovery path well.

### Issues

⚠️ **daemon/internal/config/writer_rename.go:0** — Unchanged from prior review: RenameRepoInTOML round-trips the config through map[string]any and AtomicWriteTOML, which strips comments and may reorder keys. The probe runs hourly by default and triggers automatically on detected renames, so operator-maintained inline comments in config.toml will silently disappear on the first rename event — there is no warning in the godoc or in the operator-facing docs for ai.repo_rename_check_interval. Impact: silent loss of operator documentation/intent in a config file the operator owns. Fix: add an explicit 'Note: comments and key ordering in the source TOML are NOT preserved' line to the RenameRepoInTOML godoc and to the operator-facing description of ai.repo_rename_check_interval, or switch to a comment-preserving TOML library.

### Suggestions

- Consider documenting the same comment-loss caveat on AtomicWriteTOML itself so any future caller inherits the warning, not just RenameRepoInTOML.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #491 rename propagation PR. The new diff adds a probe with recovery semantics, canonical name resolution, atomic TOML rewrite, and integration tests covering partial-failure recovery across ticks. Architecture is sound: shared tomlMu serializes with PATCH /config, probe is best-effort with per-repo error isolation, and the reconciler ordering (persister before in-memory mutation) is correctly pinned by tests. Prior medium finding about silent comment/key-order loss in RenameRepoInTOML remains unaddressed — the godoc still does not warn operators, and the AIConfig.RepoRenameCheckInterval description does not mention the side effect. Given the probe runs hourly by default, this is a real operator-facing data-loss risk on a file operators own.

### Issues

⚠️ **daemon/internal/config/writer_rename.go:24** — RenameRepoInTOML still round-trips through map[string]any + AtomicWriteTOML, which strips comments and may reorder keys, and neither the function godoc nor the AIConfig.RepoRenameCheckInterval field doc warns operators. With the probe enabled by default at 1h, the first detected rename silently destroys operator-maintained inline comments in config.toml. Impact: silent loss of operator documentation/intent on a file operators own. Fix: add an explicit 'Note: comments and key ordering in the source TOML are NOT preserved across rewrite' line to the godoc of RenameRepoInTOML and to the operator-facing description of ai.repo_rename_check_interval, or switch to a comment-preserving TOML library.

### Suggestions

- Consider documenting in the AIConfig.RepoRenameCheckInterval godoc the rate-limit cost (1 GET /repos per monitored repo per tick) so operators sizing large fleets understand the budget impact of lowering the interval.
- newRenameProbe takes a ctx parameter it ignores (underscore-named); if it is truly unused, drop it from the signature to avoid the false suggestion that cancellation is wired through construction.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #491. Previous medium finding about silent loss of operator comments in config.toml via RenameRepoInTOML remains unaddressed: the godoc of RenameRepoInTOML and the AIConfig.RepoRenameCheckInterval field doc do not warn that round-tripping through map[string]any + AtomicWriteTOML strips comments and may reorder keys. With the probe defaulting to 1h enabled, the first detected rename still silently destroys operator-maintained inline comments.

### Issues

⚠️ **daemon/internal/config/writer_rename.go:24** — RenameRepoInTOML still round-trips through ReadTOMLMap + AtomicWriteTOML, which strips comments and may reorder keys; neither the function godoc nor AIConfig.RepoRenameCheckInterval (config.go:395) warns operators. With the probe enabled by default (1h), the first auto-detected rename silently destroys operator-maintained inline comments and key ordering in config.toml — a file operators own and document. Fix: add an explicit 'Note: comments and key ordering in the source TOML are NOT preserved across this rewrite' line to the RenameRepoInTOML godoc and to the operator-facing description of ai.repo_rename_check_interval, or migrate to a comment-preserving TOML library.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Re-review of the rename probe feature. The critical TOML write race (finding #1) has been addressed by threading TOMLMu through the reconciler deps. Two medium-severity findings remain unaddressed: no minimum interval floor on the probe, and silent config rot for non_monitored repos. Test coverage is thorough, including a well-designed partial-failure recovery integration test. The reconciler.go implementation is in the truncated portion of the diff and cannot be fully verified.

### Issues

⚠️ **daemon/cmd/heimdallm/main_rename.go:23** — No minimum interval floor (STILL UNADDRESSED): parseRenameProbeInterval accepts any positive duration (e.g. '1s'). With 50 monitored repos, a 1s interval would issue 50 GETs/second — exhausting GitHub's 5000 req/hour rate limit in under 2 minutes. Enforce a minimum (e.g. 5m) or at least emit a loud warning below a sane threshold to protect operators from accidental rate-limit exhaustion.
🟡 **daemon/cmd/heimdallm/main_rename.go:56** — Stale non_monitored slugs are invisible (STILL UNADDRESSED): renameRepoListFn excludes NonMonitored repos from probe coverage. The rationale is sound (avoiding surprise TOML rewrites), but a renamed non_monitored repo will silently rot in the config forever — the daemon will keep referencing a slug GitHub no longer recognizes. At minimum, include non_monitored repos in the probe's canonical check and log a warning when divergence is detected, without triggering the rename reconciler.
⚠️ **daemon/internal/rename/probe.go:97** — No initial tick on startup (UNABLE TO VERIFY — diff truncated): The Probe.Run method is in the truncated portion of the diff. If it still enters a ticker loop without an initial p.Tick(ctx) call, renames detected during daemon downtime won't be acted upon for up to one full interval (default 1h). Please confirm an initial tick was added.

### Suggestions

- Add a floor to parseRenameProbeInterval (e.g. clamp to 5m minimum) and log a warning when the operator-provided value is below the floor, so the protection is visible in daemon logs.
- Include non_monitored repos in the canonical-name check loop with a log-only path (slog.Warn) so operators get visibility into stale slugs without triggering the full rename reconciler.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #491. The previous medium finding about lossy TOML round-trip (comments/key ordering) has been addressed: writer_rename.go now contains an explicit LIMITATIONS section in the RenameRepoInTOML godoc (lines 23-37 of the new file) documenting that comments are dropped, key order is encoder-determined, and blank lines are not preserved. The AIConfig.RepoRenameCheckInterval doc comment (config.go:398-404) tells operators the probe is enabled by default at 1h and points to the manual trigger as an alternative. While the operator-facing field doc could still cross-reference the TOML rewrite caveat more explicitly, the limitation is now clearly documented at the function that performs the rewrite, and every other config-write path in the daemon shares the same round-trip behavior (acknowledged in the godoc). The remaining diff is well-structured: probe/reconciler split with clear seams, idempotency via store audit row, ordering fix (persist-then-mutate) covered by an explicit integration test, ApplyConfig serialized under cfgMu, tomlMu shared with PATCH handlers, the canonical-name probe relies on GitHub's transparent 301 with a deterministic 404-vs-rename split, and unit + integration coverage on the core paths. No new high/medium issues found.

### Suggestions

- Consider mirroring the TOML round-trip caveat from RenameRepoInTOML's godoc into the AIConfig.RepoRenameCheckInterval doc comment so operators reading config.go directly learn that enabling the probe means auto-rewrites will strip comments/reorder keys.
- parseRenameProbeInterval treats negative durations (e.g. '-5m') as unparseable and silently falls back to the default with a warn log; an explicit validation error at config-load time would surface operator typos faster than a runtime warn.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Re-review of the rename propagation feature. Two previous findings remain unaddressed: no minimum interval floor on the probe (rate-limit risk) and silent rot of non_monitored repos. No new high-severity issues. The code is well-structured with good test coverage and clear documentation of limitations.

### Issues

⚠️ **daemon/cmd/heimdallm/main_rename.go:23** — Still unaddressed: parseRenameProbeInterval accepts any positive duration (e.g. '1s'). With N monitored repos, a sub-minute interval issues N GETs per tick — easily exhausting GitHub's 5000 req/hour limit. Either enforce a minimum floor (e.g. 5m) or emit a loud slog.Warn when the parsed interval is below a sane threshold. A one-liner clamp after the parse would suffice: if d < 5*time.Minute { d = 5*time.Minute; slog.Warn(...) }.
🟡 **daemon/cmd/heimdallm/main_rename.go:56** — Still unaddressed: renameRepoListFn excludes NonMonitored repos. The exclusion rationale is sound (avoid surprise TOML rewrites), but a renamed non_monitored repo silently rots in the config — the daemon references a slug GitHub no longer recognizes, with no warning ever surfaced. Suggestion: include non_monitored repos in the canonical check and log a warning (without dispatching the reconciler) when divergence is detected, so operators have visibility.
🟡 **daemon/internal/config/config.go:943** — applyDefaults sets RepoRenameCheckInterval to '1h' when empty, making the raw=='' branch in parseRenameProbeInterval dead code. Not a bug — just dead defensive code that could confuse future readers. Consider removing the empty-string fallback from parseRenameProbeInterval or adding a comment noting applyDefaults guarantees non-empty.

### Suggestions

- Add a minimum interval floor (e.g. 5 minutes) in parseRenameProbeInterval to protect operators from accidental GitHub rate-limit exhaustion — this is the highest-value change still missing.
- Consider logging a warning for non_monitored repos that diverge from their GitHub canonical name, even if the reconciler is not dispatched for them, so operators have visibility into stale slugs.
- Remove or annotate the dead raw=='' branch in parseRenameProbeInterval since applyDefaults guarantees the field is non-empty by the time the probe reads it.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #491: rename propagation across daemon state. The PR adds a probe + reconciler pipeline with TOML rewriting, in-memory config mutation, and an integration test pinning recovery across ticks. Code is well-structured, locking discipline is documented, and the integration test addresses the partial-failure ordering bug from prior feedback. A few medium concerns remain around the TOML round-trip lossiness (documented but still a real operator impact) and the canonical detection relying solely on full_name compare without handling case-only renames explicitly.

### Issues

⚠️ **daemon/internal/config/writer_rename.go:0** — TOML rewrite via map[string]any round-trip drops comments, blank lines, and key ordering on every rename. Impact: operators who hand-format config.toml lose annotations silently the first time the probe fires. Mitigation is documented in the docstring, but since the probe runs automatically by default (1h), this will affect any operator with comments in config.toml without warning. Recommend: emit a one-time slog.Warn on first rewrite noting comments may have been dropped, or detect comment presence pre-rewrite and surface via SSE.
🟡 **daemon/internal/github/canonical.go:21** — GetCanonicalFullName compares input to output via simple string equality, but GitHub repo slugs are case-insensitive on lookup yet case-sensitive in the JSON response. A case-only rename (e.g. acme/Foo → acme/foo) would trigger a full reconciler dispatch on every probe tick if the configured slug differs only in case. Recommend: normalize comparison via strings.EqualFold in the probe, or document that operators must use canonical casing in config.toml.
⚠️ **daemon/cmd/heimdallm/main_rename.go:89** — newRenameReconciler reads cfg.AI.CloneDir under cfgMu at construction time and caches it for the lifetime of the reconciler. If the operator changes ai.clone_dir via PATCH /config after the probe goroutine starts, the reconciler will keep purging under the stale path. Impact: orphaned clone dirs after a CloneDir change followed by a rename. Recommend: pass a func() string accessor instead of a cached string so each Run reads under cfgMu.

### Suggestions

- Consider adding a metric (counter) for rename detections and reconciler failures to support the operations dimension — the PR adds slog but no metrics surface.
- The minRenameProbeInterval=1m floor is hardcoded; consider exposing as a const exported from the rename package so tests in other packages can reference it without duplication.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All three previous findings have been addressed — rate-limit floor clamping added, NonMonitored exclusion documented with rationale, and the empty-string fallback exercised by tests as a defensive guard. The new code is well-structured with clean separation between probe, reconciler, and TOML persistence. Tests cover interval parsing edge cases, TOML rewrite correctness, in-memory config mutation, and partial-failure recovery across probe ticks. One new minor observation: the probe interval is read once at startup and not hot-reloadable.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:735** — Probe interval is read once at startup (`parseRenameProbeInterval(cfg.AI.RepoRenameCheckInterval)`) and baked into the Probe. If an operator changes `repo_rename_check_interval` via PATCH /config at runtime, the probe continues ticking at the original cadence until the daemon restarts. Other config knobs in this codebase appear to be hot-reloadable via cfgMu snapshots — this one silently isn't. Consider either documenting this as restart-required, or having the probe re-read the interval from cfg each tick.

### Suggestions

- Consider documenting in the config field comment that `repo_rename_check_interval` requires a daemon restart to take effect, since operators familiar with hot-reloadable PATCH /config knobs may expect runtime changes to apply immediately.

---

PR #495, opened 2026-05-14
feat(rename): warn on stale non_monitored slugs (#493)

### LINKED ISSUE #493: rename probe should warn on stale slugs in non_monitored
## Follow-up from #489 / PR #491

The rename probe (`daemon/internal/rename/probe.go`) iterates **only** `cfg.GitHub.Repositories` — `non_monitored` is intentionally excluded so the daemon does not silently rewrite entries the operator has explicitly disabled.

The trade-off: a renamed repo that lives in `non_monitored` will sit there with the **old slug forever**, even after GitHub's 301 grace period expires. The operator may not notice because the daemon does not poll it, but the entry becomes useless (the Flutter UI shows a name that no longer exists).

## Proposed behaviour

On every probe tick, additionally:

1. Call `GetCanonicalFullName` for each `non_monitored` entry.
2. When canonical ≠ configured: **do not** dispatch the reconciler (operator-disabled, leave as-is), but emit a single `slog.Warn` per detected stale slug with the suggested new name.
3. Optionally, expose this via a new SSE event `repo_non_monitored_stale` so the Flutter UI can prompt the operator to either re-enable + rename or just drop the entry.

## Out of scope

- Auto-rewriting `non_monitored` (deliberate non-goal per #489 plan).
- Probing repos that were never configured (no operator intent signal).

## Cost

One extra `GET /repos/{owner}/{repo}` per entry per tick (default 1h cadence). Bounded by `len(cfg.GitHub.NonMonitored)`.

ivanmunozruiz:
## Summary

Closes #493 (follow-up to #489). The rename probe deliberately does **not** auto-rewrite entries in `github.non_monitored` — those reflect explicit operator-disabled state. But leaving them stale silently is its own surprise once GitHub's 301 grace period expires.

This PR adds a parallel non-monitored scan axis to the existing probe.

## What changed

- **`rename.Probe`** gains two optional `ProbeDeps` fields: `NonMonitored func() []string` and `Publisher rename.Publisher`. Backward-compat — nil keeps prior behaviour.
- **`Tick()`** now runs `tickMonitored` (existing) and `tickNonMonitored` (new). The new axis logs via `slog.Warn` and emits `EventRepoNonMonitoredStale` SSE with `{old_repo, new_repo}`.
- **In-memory dedup** on `(old → last-warned-new)` so a daemon living through 24h of probe ticks emits at most one event per detected drift. GC of entries no longer in the current non_monitored snapshot, so a remove+re-add re-warns.
- **Wiring** in `main_rename.go`: new `renameNonMonitoredListFn` accessor, plus `Publisher: broker` so the probe can fan stale-slug events out to Flutter.
- **Flutter**: `nonMonitoredStaleProvider` accumulates the pairs the daemon reports. No banner widget in this PR — that's UI design for a separate issue; the provider gives a stable hook.

## Out of scope (intentional)

- Banner/toast surface in the dashboard for the new event. The provider is in place so any future widget can subscribe.
- Auto-rewriting non_monitored entries (the whole reason this issue exists is to NOT do that).

## Test plan

- [x] `make test-docker ./...` clean — 7 new probe tests pin happy path, dedup, GC, chained renames, 404, nil-func backward compat.
- [x] `flutter test` clean (176 tests).
- [ ] Manual smoke: rename a sandbox repo that's in `non_monitored`, wait one probe tick, confirm: SSE `repo_non_monitored_stale` fires, daemon logs warning, no config rewrite, no SQLite changes.

### DIFF
--- a/daemon/cmd/heimdallm/main_rename.go
+++ b/daemon/cmd/heimdallm/main_rename.go
@@ -75,6 +75,21 @@ func renameRepoListFn(cfg *config.Config, cfgMu *sync.Mutex) func() []string {
 	}
 }
 
+// renameNonMonitoredListFn returns the operator-disabled repo set
+// (#493). The probe scans this list separately from the monitored
+// one and emits stale-slug warnings without dispatching the
+// reconciler — non_monitored entries reflect explicit operator
+// choice and must not be auto-rewritten.
+func renameNonMonitoredListFn(cfg *config.Config, cfgMu *sync.Mutex) func() []string {
+	return func() []string {
+		cfgMu.Lock()
+		defer cfgMu.Unlock()
+		out := make([]string, 0, len(cfg.GitHub.NonMonitored))
+		out = append(out, cfg.GitHub.NonMonitored...)
+		return out
+	}
+}
+
 // newRenameProbe constructs the Reconciler + Probe with their real
 // daemon deps. It is invoked once per pollers cycle (re-construction
 // is safe; the probe is goroutine-local), but called from a single
@@ -94,10 +109,12 @@ func newRenameProbe(
 ) *rename.Probe {
 	reconciler := newRenameReconciler(cfg, cfgMu, tomlMu, s, repoCtx, broker, cfgPath)
 	return rename.NewProbe(rename.ProbeDeps{
-		Probe:      ghClient,
-		Dispatcher: reconciler,
-		Repos:      renameRepoListFn(cfg, cfgMu),
-		Interval:   interval,
+		Probe:        ghClient,
+		Dispatcher:   reconciler,
+		Repos:        renameRepoListFn(cfg, cfgMu),
+		NonMonitored: renameNonMonitoredListFn(cfg, cfgMu),
+		Publisher:    broker,
+		Interval:     interval,
 	})
 }
 
--- a/daemon/internal/rename/probe.go
+++ b/daemon/internal/rename/probe.go
@@ -2,8 +2,8 @@ package rename
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
-	"fmt"
 	"log/slog"
 	"sync"
 	"time"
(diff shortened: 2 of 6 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

PR adds a non-monitored stale slug detection axis to the rename probe with SSE emission, dedup across ticks, and Flutter consumer wiring. Implementation is clean, well-tested (happy/sad paths, dedup, chained rename, re-add, 404, nil backward-compat), and architecturally sound. No confirmed bugs or security issues found.

### Suggestions

- Consider bounding the size of nmWarned map defensively in case the non_monitored list grows unexpectedly large (currently bounded by config size, so low risk).
- The Flutter nonMonitoredStaleProvider has no UI surface yet — consider adding a TODO or tracking issue link so the accumulated state doesn't become dead data.
- tickNonMonitored holds nmMu inside the per-repo loop across separate lock/unlock cycles; could be simplified by collecting mismatches first then locking once, though current approach is correct.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured feature addition with thorough test coverage and thoughtful edge-case handling. The daemon-side probe cleanly separates monitored and non-monitored tick paths, dedup logic is correct, and the Flutter SSE handler properly accumulates stale entries. Issues are minor — mostly stylistic and defensive-coding nitpicks.

### Issues

🟡 **daemon/internal/rename/probe.go:82** — nmWarned map is allocated unconditionally in NewProbe even when NonMonitored is nil, meaning the probe will never use it. Consider lazy-initializing it in tickNonMonitored or guarding the allocation behind a nil check.
🟡 **daemon/internal/rename/probe.go:185** — Non-defer mutex unlock with an early-continue branch (unlock on line 187 vs line 190) is correct but fragile — a future edit that adds a panicking call between the set and the unlock would leak the lock. Consider restructuring to use a short helper or defer within a closure to make the critical section more robust.
🟡 **daemon/internal/rename/probe.go:199** — Discarded error from json.Marshal (payload, _ := json.Marshal(...)). Safe today for map[string]string, but silently swallows errors if the payload shape is later changed to include a non-serializable type. A brief log on error would be more defensive.
🟡 **daemon/internal/rename/probe.go:0** — Comment-to-code ratio is very high throughout the PR. Many comments restate what the code does (e.g., the 5-line comment on renameNonMonitoredListFn for a trivial 6-line function) or repeat issue numbers (#493) across multiple locations. Consider trimming to only non-obvious 'why' comments.
⚠️ **daemon/internal/rename/probe.go:47** — Publisher interface is referenced in ProbeDeps but its definition is not present in the diff. If this interface is being introduced as part of this PR (rather than pre-existing), it needs to be defined — otherwise the package won't compile.

### Suggestions

- Consider lazy-initializing nmWarned in tickNonMonitored (on first non-nil NonMonitored call) rather than unconditionally in NewProbe, to avoid a dead allocation for callers that don't use the non-monitored axis.
- Reduce comment verbosity: the issue number (#493) appears 10+ times across the diff. Reference it once at the top-level feature comment and let git blame connect the rest.
- For the mutex pattern in the dedup check, consider extracting the check-and-set into a small method that uses defer, e.g., `func (p *Probe) shouldWarn(repo, canonical string) bool` — this eliminates the dual-unlock branches.

---

PR #498, opened 2026-05-14
fix: run codex through non-interactive exec

ivanmunozruiz:
## Summary
- Run Codex through the non-interactive `codex exec` path and feed prompts via stdin to avoid `stdin is not a terminal` failures.
- Map legacy Codex approval modes to the current `--ask-for-approval` values while keeping existing configs valid.
- Update auto-implement defaults, UI labels/options, config examples, and docs for the current Codex CLI contract.

## Validation
- `make test-docker GO_TEST_ARGS="./internal/executor/... ./internal/issues/... ./internal/config/... ./internal/server/..."`
- `make test-docker`
- `flutter test`
- `flutter analyze`
- `make build-web`
- Local CLI syntax check: `codex --ask-for-approval never exec --help`

### DIFF
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -326,7 +326,7 @@ func labelSetIntersects(set map[string]struct{}, list []string) bool {
 type CLIAgentConfig struct {
 	Model        string `toml:"model" json:"model,omitempty"`                 // e.g. "claude-opus-4-6"
 	MaxTurns     int    `toml:"max_turns" json:"max_turns,omitempty"`         // claude: --max-turns (0 = not set)
-	ApprovalMode string `toml:"approval_mode" json:"approval_mode,omitempty"` // codex: --approval-mode
+	ApprovalMode string `toml:"approval_mode" json:"approval_mode,omitempty"` // codex: --ask-for-approval
 	ExtraFlags   string `toml:"extra_flags" json:"extra_flags,omitempty"`     // free-form additional CLI flags
 	PromptID     string `toml:"prompt" json:"prompt,omitempty"`               // agent-level prompt override
 
@@ -448,9 +448,9 @@ type ReviewResponseConfig struct {
 // The lifetime cap is intentionally low: an operator who wants more
 // rounds must opt-in explicitly so we never silently amplify cost.
 type ReviewFixConfig struct {
-	Enabled         bool `toml:"enabled"`
-	PerPRLifetime   int  `toml:"per_pr_lifetime"`
-	CooldownSecs    int  `toml:"cooldown_secs"`
+	Enabled       bool `toml:"enabled"`
+	PerPRLifetime int  `toml:"per_pr_lifetime"`
+	CooldownSecs  int  `toml:"cooldown_secs"`
 }
 
 // Defaults for the review-response and review-fix paths (#482).
--- a/daemon/internal/executor/executor.go
+++ b/daemon/internal/executor/executor.go
@@ -39,7 +39,8 @@ type ExecOptions struct {
 	Model string
 	// MaxTurns sets --max-turns <n> for Claude (0 = not set).
 	MaxTurns int
-	// ApprovalMode sets --approval-mode <value> for Codex.
+	// ApprovalMode sets Codex --ask-for-approval <value>.
+	// Legacy values from older Codex CLIs are still accepted and normalized.
 	ApprovalMode string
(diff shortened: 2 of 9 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-tested PR that updates Codex integration to use the non-interactive `exec` subcommand and migrates from the removed `--approval-mode` flag to `--ask-for-approval` with proper legacy value normalization. Test coverage is solid. A few minor issues around redundant trimming and UI clarity.

### Issues

🟡 **daemon/internal/executor/executor.go:155** — Redundant `strings.TrimSpace` calls: `normalizeCodexApprovalMode` trims its input in both the switch and default case, but the only call site in `buildArgs` (line 470) already passes a trimmed value. Not a bug, but unnecessary work that obscures the contract — consider documenting that the function expects pre-trimmed input or removing the internal trims.
⚠️ **flutter_app/lib/core/models/config_model.dart:91** — The `approvalModeOptions` list mixes current Codex values (never, on-request, on-failure, untrusted) with legacy values (full-auto, auto-edit, suggest) in a single flat list with no visual distinction. Users may unknowingly select a legacy value that gets silently normalized. Consider either grouping/labeling them (e.g., optgroup or suffix) or dropping the legacy values from the UI since the daemon normalizes on the backend.
⚠️ **flutter_app/lib/features/cli_agents/cli_agents_screen.dart:662** — The placeholder for Codex `extra_flags` changed from `--full-auto` to `--sandbox workspace-write`. Since the top-level interactive command is no longer used (replaced by `exec`), verify that `--sandbox` is a valid flag for `codex exec` and not only for the top-level command. If it's top-level only, extra flags appended after `exec` in `buildArgs` would be misplaced.
🟡 **daemon/internal/executor/executor.go:153** — The unexported `normalizeCodexApprovalMode` function lacks a doc comment. Given that it encodes a non-obvious mapping between two CLI vocabularies (and the legacy mapping may need updating if Codex changes again), a brief comment explaining the mapping rationale would help maintainers.
🟡 **daemon/internal/executor/executor.go:467** — The `--ask-for-approval` flag is placed before `exec` (as a top-level flag), while `--model` is placed after `exec` (as a subcommand flag). This positional split is correct if Codex's CLI treats approval as a global flag, but if a future Codex version moves it to a subcommand flag, this will silently break. The test verifies the current ordering, which is good, but consider adding a comment explaining why the flag must precede `exec`.

### Suggestions

- Consider removing legacy approval mode values (full-auto, auto-edit, suggest) from the Flutter dropdown and only showing them if the user's saved config already contains a legacy value — the daemon normalizes transparently, so there's no reason to let new configurations use deprecated vocabulary.
- Add a brief integration smoke test or CI check that validates the `codex exec` invocation against the actual Codex CLI binary (even just `codex exec --help`) to catch future flag/subcommand changes early.
- The `normalizeCodexApprovalMode` normalization could also be applied at config load time (in `config.go`) rather than only at arg-build time, so that `ExecOptions.ApprovalMode` always carries the canonical value. This would simplify downstream consumers and make the default in `pipeline.go` (line 1010) consistent with what `buildArgs` emits.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

PR correctly fixes Codex non-interactive invocation by switching to the `exec` subcommand with stdin prompt, and migrates `--approval-mode` to `--ask-for-approval` with legacy value normalization. Changes are coherent across daemon, Flutter UI, docs, and tests. Mapping of legacy values is reasonable, and defaults for auto_implement (`never`) match the new vocabulary. One minor concern: the error message in ValidateApprovalMode lists legacy values as valid, which could mislead users into adopting deprecated names.

### Suggestions

- In ValidateApprovalMode's error message, consider distinguishing current values (untrusted, on-failure, on-request, never) from legacy/deprecated ones (auto-edit, full-auto, suggest) so users don't adopt deprecated names going forward.
- Consider emitting a deprecation warning (slog.Warn) when normalizeCodexApprovalMode rewrites a legacy value, so operators know to update their config.toml.
- The Flutter `approvalModeOptions` list mixes new and legacy values without visual grouping; surfacing legacy values in the dropdown may encourage continued use — consider labeling them as deprecated or removing from the picker while still accepting them on load.
- The fallback example flag for codex in cli_agents_screen.dart changed to `--sandbox workspace-write`; verify this is the intended hint given Codex `exec` semantics.

---

PR #503, opened 2026-05-15
fix: speed up dynamic config saves

### LINKED ISSUE #463: bug: saving settings is extremely slow
## Problem

Saving configuration changes from the GUI (and likely the TUI) takes an unacceptably long time. The user toggles a setting or edits a field, hits save, and the UI hangs or shows a spinner for several seconds before confirming the change.

## Suspected Causes

1. **TOML read-merge-write cycle** — \`PATCH /config\` reads the entire TOML file, deep-merges the patch, validates, writes atomically, then triggers a full config reload (\`patchTOML\` in \`handlers.go\`). Each save pays the full cost even for a single boolean toggle.

2. **Config reload triggers re-initialization** — \`reloadFn()\` after TOML write may restart pollers, re-resolve repos, re-fetch metadata, etc. A lightweight field change (e.g., toggling a flag) should not trigger heavyweight re-initialization.

3. **Synchronous round-trip** — the Flutter app waits for the full PATCH → write → reload → GET /config response before updating the UI. The optimistic update in \`ConfigNotifier.save()\` should make the UI feel instant, but if the PATCH itself blocks for seconds, the reconciliation stalls.

4. **\`tomlMu\` contention** — the global TOML mutex serializes all config writes. If multiple saves are queued (e.g., user changes several fields quickly), they block on each other.

## Expected Behavior

Saving a single setting should feel instant (< 200ms perceived latency). The user should never wait seconds for a toggle or dropdown change to persist.

## Possible Solutions

- Debounce rapid saves — batch multiple field changes into a single PATCH
- Make the reload selective — only restart subsystems affected by the changed fields, not everything
- Async write — return success immediately after TOML write, reload in background
- Profile the actual bottleneck — is it the TOML write, the reload, or something else?

ivanmunozruiz:
## Summary
- avoid cancelling/restarting pollers when config reloads only change dynamic fields
- keep full poller restart for cadence, repo stream, and discovery changes
- add unit coverage for restart vs fast-swap reload decisions

## Tests
- make test-docker GO_TEST_ARGS="-run 'TestConfigReloadRequiresPollerRestart|TestParseDiscoveryInterval|TestResolveRefinementTimeout' ./cmd/heimdallm"
- make test-docker

Closes #463

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
@@ -1548,6 +1549,16 @@ func main() {
 			return fmt.Errorf("reload: %w", err)
 		}
 
+		cfgMu.Lock()
+		restartPollers := configReloadRequiresPollerRestart(cfg, newCfg)
+		if !restartPollers {
+			cfg = newCfg
+			cfgMu.Unlock()
+			slog.Info("config reload: applied without poller restart")
+			return nil
+		}
+		cfgMu.Unlock()
+
 		// Read the current cancel/wg under cfgMu, then stop OUTSIDE the lock.
 		// Holding cfgMu across Wait() risks deadlock: if Wait() blocks
 		// waiting for in-flight goroutines that also acquire cfgMu (e.g.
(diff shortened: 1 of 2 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Good optimization to skip expensive poller restarts for dynamic-only config changes. The comparison function and tests are well-structured. One concurrency concern with the lock-unlock-relock pattern in the reload path.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:1552** — Lock-unlock-relock anti-pattern: cfgMu is unlocked after the restart check and immediately re-acquired to swap cfg. Between the two critical sections, another goroutine could modify cfg, creating a TOCTOU gap. Merge into a single critical section: hold cfgMu, call configReloadRequiresPollerRestart, and if !restartPollers assign cfg=newCfg before unlocking. This is both simpler and eliminates the race window.
🟡 **daemon/cmd/heimdallm/main.go:2040** — configReloadRequiresPollerRestart is an allowlist of 'restart-worthy' fields. Any new config field that affects poller behavior added in the future will silently skip restarts unless a developer remembers to update this function. Consider adding a brief comment at the struct/caller level noting that new cadence/stream fields must be added here, or add a compile-time safeguard.

### Suggestions

- Collapse the two lock regions into one: `cfgMu.Lock(); restartPollers := configReloadRequiresPollerRestart(cfg, newCfg); if !restartPollers { cfg = newCfg; cfgMu.Unlock(); slog.Info(...); return nil }; cfgMu.Unlock()`
- Verify that GitHub auth/token changes (if reloadable) don't also need a poller restart, since authenticated API clients may be captured at poller start time

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Adds a fast-path config reload that swaps the config pointer in place when only dynamically-read fields change, avoiding the cost of cancelling pollers and waiting on in-flight GitHub calls. Logic is sound for the fields it covers and is well-tested, but the allowlist of fields that trigger a restart is implicit/whitelist-by-omission, which is fragile.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2043** — configReloadRequiresPollerRestart only checks a hand-picked set of fields (intervals, discovery topic/orgs, repositories, non-monitored). Any other config field that pollers/clients capture at startup (e.g. GitHub token/base URL/enterprise URL, HTTP timeouts, webhook settings, rate-limit knobs, TLS config) will silently take the fast path and NOT be picked up by long-running goroutines that closed over the old values — producing the very class of bug the old 'always restart' behavior prevented. Impact: stale auth/endpoint/timeout config after reload, hard to debug. Fix: invert the policy — restart by default and add an explicit allowlist of fields that are known to be read dynamically through cfg under cfgMu (agent, labels, prompts, repo overrides, issue tracking defaults, etc.), or generate the comparison from a tagged struct so new fields default to 'restart required'.

### Suggestions

- Combine the two cfgMu.Lock/Unlock pairs in the fast path into a single critical section that both computes restartPollers and (if false) swaps cfg, to avoid an unnecessary lock-release-relock window and make the read-then-write atomic with respect to other readers.
- Add a test asserting that nil-vs-empty-slice equivalence in DiscoveryOrgs/Repositories/NonMonitored does not trigger a spurious restart (slices.Equal handles it, but locking in the behavior prevents regressions).
- Consider logging which field forced the restart when restartPollers is true, to aid operators debugging unexpectedly slow reloads.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

The author addressed the previous medium finding by inverting the policy: configReloadRestartSnapshot now starts from a full copy of the config and explicitly scrubs only fields known to be read dynamically under cfgMu. Any new/unscrubbed field defaults to triggering a poller restart (verified by TestConfigReloadRequiresPollerRestartDefaultsUnknownFieldsToRestart, which mutates Server.MaxConcurrentWorkers). Slice normalization correctly treats nil and empty as equivalent. Tests cover dynamic-only changes, cadence/stream changes, nil-vs-empty, and the conservative default. No new issues introduced.

### Suggestions

- Consider adding a brief comment in config.Config (or a doc comment near configReloadRestartSnapshot) reminding maintainers that adding a new dynamically-read field requires scrubbing it here, otherwise reloads will unnecessarily restart pollers.
- The snapshot copies the full struct by value; if any future field contains a map or pointer to mutable state, reflect.DeepEqual could give misleading results. A struct-tag-driven approach (e.g. `reload:"dynamic"`) would be more robust long-term, but is not required now.

---

PR #504, opened 2026-05-18
feat: add liveness heartbeat watchdog

### LINKED ISSUE #469: feat: healthcheck de liveness end-to-end — detección de daemon colgado en TUI/GUI
## Problema

A veces el daemon se queda colgado (goroutine en deadlock, poller bloqueado, NATS stall) y el único síntoma es que la GUI/TUI deja de recibir actualizaciones. El usuario tiene que darse cuenta manualmente y reiniciar.

### Por qué pasa desapercibido

El sistema actual tiene **tres puntos ciegos** que permiten que un hang pase desapercibido:

1. **`/health` es un probe vacío** — solo responde `{"status":"ok"}` si el HTTP server está vivo (`handlers.go:323-332`). No verifica NATS, SQLite, pollers, ni workers. Un daemon con todos los goroutines muertos sigue respondiendo 200 OK.

2. **Los SSE pings no son observables por el cliente** — el keep-alive del server manda `: ping\n\n` (comment SSE, `handlers.go:1005`) cada 20s, pero el Flutter client los descarta silenciosamente (`sse_client.dart:49`: `// Skip comment lines`). El client no tiene forma de distinguir "servidor vivo pero idle" de "servidor colgado".

3. **No hay timeout de stale connection** — el Flutter client espera a `onDone` o `onError` del stream HTTP para detectar desconexión. Si el servidor está colgado pero la conexión TCP sigue abierta, pueden pasar **minutos** antes de que el SO cierre el socket por timeout.

### Capa 1: Heartbeat como evento SSE de primera clase

Reemplazar el `: ping\n\n` (comment invisible) por un evento SSE real que el cliente pueda observar:

```
event: heartbeat
data: {"ts":"2026-05-11T16:00:00Z","pollers_alive":true,"last_poll":"2026-05-11T15:59:12Z","goroutines":42}

```

- **Intervalo**: mantener los 20s actuales del ticker en `handlers.go:995`
- **Payload**: timestamp del servidor + indicadores de salud interna (último poll, goroutines vivos, NATS connected)
- **Backward-compatible**: los clients que no entienden `heartbeat` simplemente lo ignoran

Esto convierte cada ping en un **health probe que el cliente puede evaluar**.

### Capa 2: Watchdog en el cliente (Flutter + TUI)

El cliente trackea el timestamp del último evento recibido (cualquiera, incluido heartbeat):

```dart
Timer? _watchdog;
DateTime _lastEventAt = DateTime.now();

void _onEvent(SseEvent event) {
  _lastEventAt = DateTime.now();
  _resetWatchdog();
  // ... process event
}

void _resetWatchdog() {
  _watchdog?.cancel();
  _watchdog = Timer(Duration(seconds: 60), _onStale);
}

void _onStale() {
  // 3 heartbeats perdidos (20s × 3 = 60s) → servidor probablemente colgado
  _connectionState.value = ConnectionState.unresponsive;
}
```

**Comportamiento en UI**:
- **< 60s sin eventos**: estado normal (verde)
- **60s sin eventos** (3 heartbeats perdidos): banner amarillo "Servidor sin respuesta — verificando conexión..."
  - Intenta `GET /health` como fallback
  - Si `/health` responde: reconectar SSE automáticamente
  - Si `/health` no responde: banner rojo "Servidor no disponible" con botón "Reiniciar daemon"
- **Reconexión exitosa**: banner desaparece automáticamente

### Capa 3: Deep health endpoint

Extender `/health` para que verifique estado interno real:

```json
{
  "status": "ok",
  "version": "0.6.15",
  "started_at": "2026-05-11T10:00:00Z",
  "uptime_seconds": 21600,
  "checks": {
    "nats": { "ok": true },
    "sqlite": { "ok": true, "writable": true },
    "last_poll": { "ok": true, "at": "2026-05-11T15:59:12Z", "age_seconds": 48 },
    "goroutines": { "count": 42 },
    "workers": {
      "review": { "active": 2, "capacity": 5 },
      "triage": { "active": 0, "capacity": 5 },
      "publish": { "active": 1, "capacity": 5 }
    }
  }
}
```

- `last_poll.ok = false` si `age_seconds > 2 × poll_interval` (los pollers están muertos)
- `nats.ok` vía `conn.IsConnected()`
- `sqlite.ok` vía un `PRAGMA quick_check` o write test
- Mantener backward-compatible: `status` sigue siendo top-level, `checks` es aditivo

## Acceptance Criteria

- [ ] El SSE server emite `event: heartbeat` con payload JSON cada 20s en vez del comment `: ping`
- [ ] El payload del heartbeat incluye al menos: `ts`, `last_poll_at`, `goroutines`
- [ ] El Flutter SSE client trackea `lastEventAt` y detecta stale connection tras 3 heartbeats perdidos (60s)
- [ ] La Flutter UI muestra un banner de estado de conexión:
  - Verde: conectado y recibiendo eventos
  - Amarillo: sin eventos hace >60s, verificando
  - Rojo: servidor no responde a `/health`, opción de restart
- [ ] El banner desaparece automáticamente al reconectar
- [ ] `/health` incluye campo `checks` con estado de NATS, SQLite, y último poll
- [ ] `/health` sigue respondiendo 200 para compatibilidad con probes de Kubernetes/Docker (el campo `checks` es informativo, no cambia el status code)
- [ ] `go test -race ./daemon/...` pasa
- [ ] `flutter test` y `flutter analyze` pasan

### Daemon
- `daemon/internal/server/handlers.go` — heartbeat event + deep health
- `daemon/internal/sse/broker.go` — verificar que heartbeat events pasan por el broker sin drop
- `daemon/cmd/heimdallm/main.go` — exponer métricas de pollers al health handler

### Flutter
- `flutter_app/lib/core/api/sse_client.dart` — watchdog timer, connection state
- `flutter_app/lib/features/server/server_providers.dart` — provider de connection state
- `flutter_app/lib/features/dashboard/` — banner de conexión en el layout principal

ivanmunozruiz:
## Summary
- Add observable SSE `heartbeat` events and lightweight `/health` checks for NATS, SQLite reachability, last poll freshness, version, and uptime.
- Track daemon poll completion snapshots without taking the config mutex from heartbeat paths.
- Add CLI and Flutter watchdogs that detect stale SSE streams, verify daemon health, reconnect safely when healthy, and surface stale/offline states with backoff.
- Harden TUI reconnect state with per-session SSE commands and tests for stale-stream recovery.

## Tests
- `make test-docker`
- `docker run ... sh -c "go vet ./... && go test -timeout 60s ./..."` for `cli/`
- `cd flutter_app && flutter test test/core/sse_client_test.dart test/features/dashboard_test.dart`
- `cd flutter_app && flutter analyze`
- `cd flutter_app && flutter test`
- `make build-web`

Closes #469

### DIFF
--- a/cli/internal/api/client_test.go
+++ b/cli/internal/api/client_test.go
@@ -1,0 +1,49 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+func TestStreamEventsReturnsWhenContextCanceledWithBlockedSend(t *testing.T) {
+	eventFlushed := make(chan struct{})
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "text/event-stream")
+		flusher, ok := w.(http.Flusher)
+		if !ok {
+			t.Fatal("response writer does not support flushing")
+		}
+		fmt.Fprint(w, "event: heartbeat\ndata: {}\n\n")
+		flusher.Flush()
+		close(eventFlushed)
+		<-r.Context().Done()
+	}))
+	t.Cleanup(srv.Close)
+
+	client := New(srv.URL, "")
+	ctx, cancel := context.WithCancel(context.Background())
+	events := make(chan SSEEvent)
+	done := make(chan error, 1)
+	go func() {
+		done <- client.StreamEvents(ctx, events)
+	}()
+
+	select {
+	case <-eventFlushed:
+	case <-time.After(time.Second):
+		t.Fatal("server did not flush SSE event")
+	}
+	cancel()
+	select {
+	case err := <-done:
+		if err == nil {
+			t.Fatal("expected context cancellation error")
+		}
+	case <-time.After(time.Second):
+		t.Fatal("StreamEvents did not return after context cancellation")
+	}
+}
--- a/cli/internal/tui/dashboard.go
+++ b/cli/internal/tui/dashboard.go
@@ -46,21 +46,23 @@ type Dashboard struct {
 	logOffset int
 	logSeeded bool
 
-	err              error
-	connected        bool
-	sseStale         bool
-	refreshing       bool
-	confirmShutdown  bool
-	shutdownInFlight bool
-	shutdownMessage  string
-	startTime        time.Time
-	lastUpdate       time.Time
-	lastSSEEvent     time.Time
-	version          string
+	err               error
+	connected         bool
+	sseStale          bool
+	sseHealthChecking bool
+	refreshing        bool
+	confirmShutdown   bool
+	shutdownInFlight  bool
+	shutdownMessage   string
+	startTime         time.Time
+	lastUpdate        time.Time
+	lastSSEEvent      time.Time
+	version           string
 
-	sseEvents chan api.SSEEvent
-	sseCtx    context.Context
-	sseCancel context.CancelFunc
+	sseEvents    chan api.SSEEvent
+	sseCtx       context.Context
+	sseCancel    context.CancelFunc
+	sseSessionID int64
 
 	showDetail   bool
 	detailScroll int
@@ -83,11 +85,20 @@ type dataMsg struct {
 	activity *api.ActivityResponse
 	err      error
 }
-type sseMsg api.SSEEvent
-type sseDisconnectMsg struct{ err error }
+type sseMsg struct {
+	sessionID int64
+	event     api.SSEEvent
+}
+type sseDisconnectMsg struct {
+	sessionID int64
+	err       error
+}
 type sseReconnectMsg struct{}
 type sseWatchdogMsg time.Time
-type healthCheckMsg struct{ err error }
+type healthCheckMsg struct {
+	err         error
+	lastEventAt time.Time
+}
 type shutdownMsg struct{ err error }
 type promoteIssueMsg struct {
 	id  int64
@@ -105,14 +116,16 @@ func NewDashboard(host, token, version string) *Dashboard {
 		logFollow:    true,
 		sseCtx:       ctx,
 		sseCancel:    cancel,
+		sseSessionID: 1,
 	}
 }
 
 func (d *Dashboard) Init() tea.Cmd {
+	connect, listen := d.sseCommands()
 	return tea.Batch(
 		d.fetchData,
-		d.connectSSE,
-		d.listenSSE(),
+		connect,
+		listen,
 		tickCmd(),
 		sseWatchdogCmd(),
 	)
@@ -171,23 +184,30 @@ func (d *Dashboard) fetchData() tea.Msg {
 	return msg
 }
 
-func (d *Dashboard) connectSSE() tea.Msg {
-	err := d.client.StreamEvents(d.sseCtx, d.sseEvents)
-	if d.sseCtx.Err() != nil {
-		return nil
-	}
-	return sseDisconnectMsg{err: err}
+func (d *Dashboard) sseCommands() (tea.Cmd, tea.Cmd) {
+	return d.connectSSE(d.sseSessionID, d.sseCtx, d.sseEvents),
+		d.listenSSE(d.sseSessionID, d.sseCtx, d.sseEvents)
 }
 
-func (d *Dashboard) listenSSE() tea.Cmd {
+func (d *Dashboard) connectSSE(sessionID int64, ctx context.Context, events chan<- api.SSEEvent) tea.Cmd {
+	return func() tea.Msg {
+		err := d.client.StreamEvents(ctx, events)
+		if ctx.Err() != nil {
+			return nil
+		}
+		return sseDisconnectMsg{sessionID: sessionID, err: err}
+	}
+}
+
+func (d *Dashboard) listenSSE(sessionID int64, ctx context.Context, events <-chan api.SSEEvent) tea.Cmd {
 	return func() tea.Msg {
 		select {
-		case event, ok := <-d.sseEvents:
+		case event, ok := <-events:
 			if !ok {
 				return nil
 			}
-			return sseMsg(event)
-		case <-d.sseCtx.Done():
+			return sseMsg{sessionID: sessionID, event: event}
+		case <-ctx.Done():
 			return nil
 		}
 	}
@@ -201,6 +221,7 @@ func (d *Dashboard) resetSSE() {
 	d.sseCtx = ctx
 	d.sseCancel = cancel
 	d.sseEvents = make(chan api.SSEEvent, 32)
+	d.sseSessionID++
 	d.lastSSEEvent = time.Now()
 }
 
@@ -216,9 +237,9 @@ func (d *Dashboard) promoteIssue(id int64) tea.Cmd {
 	}
 }
 
-func (d *Dashboard) checkHealth() tea.Cmd {
+func (d *Dashboard) checkHealth(lastEventAt time.Time) tea.Cmd {
 	return func() tea.Msg {
-		return healthCheckMsg{err: d.client.Health()}
+		return healthCheckMsg{err: d.client.Health(), lastEventAt: lastEventAt}
 	}
 }
 
@@ -267,10 +288,11 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	case sseWatchdogMsg:
 		cmds := []tea.Cmd{sseWatchdogCmd()}
-		if !d.shutdownInFlight && !d.lastSSEEvent.IsZero() && time.Since(d.lastSSEEvent) > time.Minute {
+		if !d.shutdownInFlight && !d.sseHealthChecking && !d.lastSSEEvent.IsZero() && time.Since(d.lastSSEEvent) > time.Minute {
 			d.sseStale = true
+			d.sseHealthChecking = true
 			d.shutdownMessage = "SSE stream stale; checking daemon health..."
-			cmds = append(cmds, d.checkHealth())
+			cmds = append(cmds, d.checkHealth(d.lastSSEEvent))
 		}
 		return d, tea.Batch(cmds...)
 
@@ -320,18 +342,24 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return d, nil
 
 	case sseMsg:
+		if msg.sessionID != d.sseSessionID {
+			return d, nil
+		}
+		event := msg.event
 		d.lastSSEEvent = time.Now()
 		d.sseStale = false
-		d.shutdownMessage = ""
-		d.connected = true
-		d.err = nil
-		if msg.Type == "heartbeat" {
-			return d, d.listenSSE()
+		if !d.sseHealthChecking {
+			d.shutdownMessage = ""
+			d.connected = true
+			d.err = nil
 		}
-		itemType, info := formatSSEData(msg.Type, msg.Data)
+		if event.Type == "heartbeat" {
+			return d, d.listenSSE(d.sseSessionID, d.sseCtx, d.sseEvents)
+		}
+		itemType, info := formatSSEData(event.Type, event.Data)
 		line := activityLine{
 			Time:     time.Now().Format("15:04"),
-			Event:    msg.Type,
+			Event:    event.Type,
 			Info:     info,
 			ItemType: itemType,
 		}
@@ -339,7 +367,7 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		if len(d.activity) > 100 {
 			d.activity = d.activity[:100]
 		}
-		d.logLines = append(d.logLines, sseToLogLine(api.SSEEvent(msg)))
+		d.logLines = append(d.logLines, sseToLogLine(event))
 		if len(d.logLines) > 1000 {
 			excess := len(d.logLines) - 1000
 			d.logLines = d.logLines[excess:]
@@ -350,9 +378,12 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				}
 			}
 		}
-		return d, d.listenSSE()
+		return d, d.listenSSE(d.sseSessionID, d.sseCtx, d.sseEvents)
 
 	case sseDisconnectMsg:
+		if msg.sessionID != d.sseSessionID {
+			return d, nil
+		}
 		d.connected = false
 		d.err = msg.err
 		d.sseStale = false
@@ -361,9 +392,14 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		})
 
 	case sseReconnectMsg:
-		return d, d.connectSSE
+		connect, _ := d.sseCommands()
+		return d, connect
 
 	case healthCheckMsg:
+		d.sseHealthChecking = false
+		if !d.sseStale || !msg.lastEventAt.Equal(d.lastSSEEvent) {
+			return d, nil
+		}
 		if msg.err != nil {
 			d.connected = false
 			d.err = msg.err
@@ -372,11 +408,12 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 		if d.sseStale {
 			d.resetSSE()
+			connect, listen := d.sseCommands()
 			d.connected = false
 			d.err = nil
 			d.sseStale = false
 			d.shutdownMessage = "Reconnecting SSE stream..."
-			return d, tea.Batch(d.connectSSE, d.listenSSE())
+			return d, tea.Batch(connect, listen)
 		}
 		return d, nil
 
--- a/cli/internal/tui/dashboard.go
+++ b/cli/internal/tui/dashboard.go
@@ -94,7 +94,7 @@ type sseDisconnectMsg struct {
 	sessionID int64
 	err       error
 }
-type sseReconnectMsg struct{}
+type sseReconnectMsg struct{ sessionID int64 }
 type sseWatchdogMsg time.Time
 type healthCheckMsg struct {
 	err         error
@@ -391,11 +391,15 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		d.connected = false
 		d.err = msg.err
 		d.sseStale = false
+		sessionID := msg.sessionID
 		return d, tea.Tick(5*time.Second, func(t time.Time) tea.Msg {
-			return sseReconnectMsg{}
+			return sseReconnectMsg{sessionID: sessionID}
 		})
 
 	case sseReconnectMsg:
+		if msg.sessionID != d.sseSessionID {
+			return d, nil
+		}
 		connect, _ := d.sseCommands()
 		return d, connect
 
@@ -410,16 +414,13 @@ func (d *Dashboard) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			d.sseStatusMessage = "Daemon health check failed"
 			return d, nil
 		}
-		if d.sseStale {
-			d.resetSSE()
-			connect, listen := d.sseCommands()
-			d.connected = false
-			d.err = nil
-			d.sseStale = false
-			d.sseStatusMessage = "Reconnecting SSE stream..."
-			return d, tea.Batch(connect, listen)
-		}
-		return d, nil
+		d.resetSSE()
+		connect, listen := d.sseCommands()
+		d.connected = false
+		d.err = nil
+		d.sseStale = false
+		d.sseStatusMessage = "Reconnecting SSE stream..."
+		return d, tea.Batch(connect, listen)
 
 	case shutdownMsg:
 		d.shutdownInFlight = false
--- a/cli/internal/tui/dashboard.go
+++ b/cli/internal/tui/dashboard.go
@@ -48,12 +48,14 @@ type Dashboard struct {
 
 	err              error
 	connected        bool
+	sseStale         bool
 	refreshing       bool
 	confirmShutdown  bool
 	shutdownInFlight bool
 	shutdownMessage  string
 	startTime        time.Time
 	lastUpdate       time.Time
+	lastSSEEvent     time.Time
 	version          string
 
 	sseEvents chan api.SSEEvent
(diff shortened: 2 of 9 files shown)


Muriano (changes requested):
## 🤖 Heimdallm AI Review

Solid liveness/heartbeat feature with good test coverage across daemon health, SSE heartbeat, TUI watchdog and Flutter banner. Main concerns are a data race on Dashboard fields mutated in resetSSE while a goroutine-backed Bubble Tea command reads them, and a goroutine leak from the previous listenSSE/connectSSE not being drained. Otherwise architecture and security look clean.

### Issues

🔴 **cli/internal/tui/dashboard.go:196** — Data race in resetSSE: it reassigns d.sseEvents, d.sseCtx and d.sseCancel while the previously scheduled listenSSE() and connectSSE Cmds run in separate goroutines and read d.sseEvents / d.sseCtx without synchronization. Bubble Tea runs Cmds concurrently with Update, so a previously dispatched listenSSE goroutine may still be parked on the old channel (or even read the field after reassignment) when the watchdog swaps fields. Impact: race detector failures and possibly events delivered into a stale channel that nobody reads (memory leak + missed reconnect). Fix: snapshot the channel/context into the Cmd closure at dispatch time (e.g. listenSSE/connectSSE should take ctx and ch as parameters from the moment Init/resetSSE runs), and/or guard the fields with a mutex. Also drain or close the old channel.
⚠️ **cli/internal/tui/dashboard.go:326** — Heartbeat sseMsg path resets d.err = nil and d.connected = true unconditionally. Combined with the watchdog firing checkHealth() and setting d.shutdownMessage = "SSE stream stale; ...", a heartbeat arriving after the watchdog set sseStale but before healthCheckMsg returns will clear sseStale and message, while a parallel healthCheckMsg error path then sets d.connected=false / d.err=msg.err. The interleaving makes the visible state non-deterministic (could flap between online/unresponsive/offline). Impact: confusing UX and inconsistent banner. Fix: only treat heartbeat as a liveness ping (update lastSSEEvent + clear sseStale), but don't overwrite shutdownMessage/connected if a health probe is in flight; or drop the in-flight probe result when a heartbeat arrived in the meantime.
⚠️ **cli/internal/tui/dashboard.go:198** — resetSSE does not close or drain the old d.sseEvents channel before reassigning. If the previous connectSSE goroutine is still writing (it was cancelled but may be mid-send), writes will block on the dropped channel and the goroutine leaks. Impact: goroutine leak per stall+recovery cycle. Fix: close the old channel after cancelling the context, and have the producer recover from a send-on-closed channel via select on ctx.Done().
🟡 **daemon/internal/server/handlers.go:1197** — heartbeatEvent calls srv.healthSnapshotFn() which in main.go acquires cfgMu. cfgMu is also held during reload and around long-running config operations. Every SSE client triggers a snapshot read every 20s plus one at connect; under reload contention this can briefly stall heartbeats and tail-add latency to all SSE consumers. Impact: heartbeat tick can be delayed exactly when reliability matters (during reload). Fix: cache PollInterval atomically (you already do this for lastPollUnixNano) so the snapshot is lock-free, or use an RWMutex with RLock for the read path.
⚠️ **daemon/internal/server/handlers.go:388** — handleHealth always returns HTTP 200 even when sub-checks fail (sqlite quick_check != ok, nats disconnected, last_poll stale). This defeats the point of a deep health endpoint for orchestrators/probes that key off status code. Impact: external liveness/readiness probes will report healthy while the daemon is degraded. Fix: aggregate the per-check ok flags into top-level status and return 503 when any required check fails (or expose /readyz alongside).
🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:95** — DaemonConnectionStatus.copyWith always overwrites message with the argument (no `?? this.message`), so any copyWith call that omits message silently clears it. Currently this happens to be intentional in every call site, but it's a foot-gun inconsistent with the rest of copyWith semantics (phase/lastEventAt use `??`). Impact: future callers will lose state without warning. Fix: either use a sentinel/_undefined pattern or document that message is always reset; at minimum align the convention.
🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:130** — The ref.listen callback in build() reacts to next.hasError by setting offline, then immediately next.whenData(...) runs — but on an AsyncError whenData is a no-op, so that's fine. However the connecting branch only fires when prev had no value; after a transient error followed by recovery, the stream emits data and phase jumps straight to connected without a connecting intermediate, which is OK. The real issue is that when sseStreamProvider is invalidated by _verifyAndReconnect, the listener may not see a fresh loading state if Riverpod replays the last value; lastEventAt is never reset, so the staleness watchdog can immediately re-fire before the new stream has had a chance to emit. Impact: potential reconnect storm if backend is genuinely down. Fix: clear lastEventAt (or set phase=connecting) inside _verifyAndReconnect before invalidating.

### Suggestions

- Make heartbeatInterval and the TUI 60s staleness threshold configurable (or at least derive the client threshold as 3× server heartbeatInterval) so they stay in sync if the server tick changes.
- Add a test for the TUI watchdog path (sseWatchdogMsg → healthCheckMsg ok → resetSSE) — currently only daemon side has heartbeat tests.
- lastPollHealthCheck.ok defaults to true when PollInterval == 0; consider returning false/unknown to avoid masking a misconfigured daemon.
- Consider emitting a metric counter for heartbeat ticks and for watchdog-triggered reconnects to make this observable in production.
- In handlers.go the alignment churn on meFn field is unnecessary noise in the diff — keep formatting changes out of feature PRs.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Solid liveness watchdog feature with good daemon-to-client heartbeat propagation. The core design—server heartbeat every 20s, client watchdog at 60s staleness threshold, health-check-then-reconnect—is sound. However, running PRAGMA quick_check on the unauthenticated /health endpoint is a performance/availability risk, and several smaller issues around state management and UI need attention.

### Issues

🔴 **daemon/internal/server/handlers.go:432** — PRAGMA quick_check on unauthenticated /health endpoint. This checks SQLite integrity on every call and can be slow on larger databases. Since /health is public (no auth required, used by the launcher), any caller can trigger expensive DB I/O. The 500ms context timeout mitigates DoS but means the check will spuriously fail on healthy-but-busy databases, causing false-negative health reports. Use a simple SELECT 1 or just PingContext instead; reserve integrity checks for a separate admin/diagnostic endpoint.
⚠️ **daemon/internal/server/handlers.go:385** — runtime.NumGoroutine() is exposed in the unauthenticated /health response and every heartbeat SSE event. This leaks internal server metrics to any caller, which can aid reconnaissance (e.g., inferring load, number of connected clients, or internal architecture). Consider omitting goroutine counts from public endpoints or gating deep checks behind authentication.
🟡 **daemon/internal/server/handlers.go:1194** — heartbeatEvent() calls srv.healthSnapshot() and runtime.NumGoroutine() every 20 seconds per connected SSE client. With N clients this is N calls per interval. While cheap individually, healthSnapshot() takes a lock (cfgMu) in main.go, so under high client counts this creates lock contention on the config mutex during every heartbeat cycle.
⚠️ **cli/internal/tui/dashboard.go:270** — shutdownMessage is overloaded for SSE watchdog status ('SSE stream stale; checking daemon health...', 'Reconnecting SSE stream...'). If a user initiates shutdown while the SSE stream is stale, the shutdown status and SSE watchdog status will clobber each other. Introduce a separate field (e.g., sseStatusMessage) to avoid conflicts.
🟡 **cli/internal/tui/dashboard.go:193** — resetSSE() replaces d.sseEvents with a new channel without closing the old one. Any goroutine blocked on sending to the old channel (between sseCancel() and the goroutine noticing the context is done) will leak until it observes cancellation. Consider closing the old channel after cancel to unblock senders immediately.
🟡 **cli/internal/tui/dashboard.go:320** — In the sseMsg handler, when msg.Type == 'heartbeat' the handler returns early before calling formatSSEData. If formatSSEData or the activity-line logic ever has side effects needed for bookkeeping (e.g., future event counting), heartbeats will silently skip it. The early return is correct today but fragile—add a brief comment noting that heartbeats intentionally skip activity logging.
⚠️ **flutter_app/lib/features/dashboard/dashboard_providers.dart:115** — _verifyAndReconnect() has no backoff. If the daemon is healthy but the SSE stream keeps going stale (e.g., network issue between client and server), the watchdog will call checkHealth() + invalidate(sseStreamProvider) every 60 seconds indefinitely. Add exponential backoff or a max-retry cap before falling to offline state.
🟡 **flutter_app/lib/features/dashboard/dashboard_screen.dart:130** — _ConnectionBanner renders for all phases including 'connected' (a green banner). Permanent banners for healthy state are unusual UX and consume vertical space. Consider only showing the banner for non-connected phases (stale, offline, connecting) and hiding it when connected.
🟡 **daemon/internal/server/handlers.go:1217** — In heartbeatEvent(), the JSON marshal fallback (line 1221) silently swallows the error and returns a degraded payload. This hides potential bugs (e.g., non-serializable types added later). Consider logging the marshal error at warn/debug level.

### Suggestions

- Replace PRAGMA quick_check with a simple PingContext() or SELECT 1 for the /health endpoint. If deep integrity checks are needed, expose them on a separate authenticated endpoint (e.g., /debug/health) that is not called on every heartbeat cycle.
- Remove or gate runtime.NumGoroutine() and detailed NATS/poll status behind authentication. The public /health endpoint should return only status: ok/degraded and version—nothing that reveals internal state.
- Add a dedicated sseStatusMessage field to the TUI Dashboard struct instead of overloading shutdownMessage, to avoid state conflicts between shutdown flow and SSE watchdog flow.
- Add exponential backoff to _verifyAndReconnect() in the Flutter DaemonConnectionNotifier—e.g., double the check interval on each failed reconnect, capped at 5 minutes.
- Consider caching the HealthSnapshot result for the heartbeat interval (20s) so that N concurrent SSE clients don't each independently call healthSnapshotFn() and acquire the config mutex.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review: most previous findings addressed (closures now snapshot ctx/channel/sessionID, heartbeat path guards against in-flight probe, healthSnapshotFn uses lock-free atomics, Flutter notifier resets state on reconnect). Two prior issues remain: /health still returns 200 unconditionally, and resetSSE still doesn't close/drain the old SSE channel.

### Issues

⚠️ **daemon/internal/server/handlers.go:388** — handleHealth still writes http.StatusOK unconditionally regardless of sub-check results. sqlite quick_check failure, NATS disconnection, or stale last_poll are reported in the body but the HTTP status remains 200, so orchestrator probes / load balancers / the new Flutter watchdog (which keys off checkHealth() boolean derived from HTTP status) cannot detect a degraded daemon. Impact: deep health endpoint is observationally healthy while the process is broken; the very watchdog this PR introduces will trust a failing daemon. Fix: aggregate per-check ok flags and return 503 when any required check fails, or expose a separate /readyz that does.
⚠️ **cli/internal/tui/dashboard.go:207** — resetSSE cancels the old context and reassigns d.sseEvents to a new channel without closing or draining the previous one. If the prior StreamEvents goroutine is mid-send on the old channel when cancellation fires (channel buffer full, select racing), it can block on the unbuffered send path and leak until GC reclaims the channel — one leaked goroutine per stall+recovery cycle. The session-ID gating in Update prevents stale events from being processed, but the producer goroutine itself can still be parked. Fix: close the old channel after cancelling ctx and have StreamEvents recover via select on ctx.Done() before any send, or document that StreamEvents must return promptly on ctx cancel and verify with a test.

### Suggestions

- Add an integration test for /health that asserts non-200 when sqlite/quick_check or nats is intentionally broken once the status code is wired up.
- Consider exposing the heartbeat interval as a server option for tests, so TestEventsEmitsObservableHeartbeat doesn't need a fast initial heartbeat to pass deterministically under load.
- DaemonConnectionNotifier._verifyAndReconnect catches nothing around checkHealth(); a thrown exception (network error, parse error) will leave _checkingHealth=true forever via the finally — actually finally clears it, but the state remains stale. Consider catching and explicitly setting offline.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review: both previously flagged medium issues remain unaddressed in this revision. handleHealth still returns 200 regardless of sub-check results, and resetSSE still leaks the prior channel/goroutine on reconnect.

### Issues

⚠️ **daemon/internal/server/handlers.go:388** — handleHealth still calls writeJSON(w, http.StatusOK, resp) unconditionally even when checks.sqlite.ok, checks.nats.ok, or checks.last_poll.ok are false. The Flutter watchdog's checkHealth() derives healthy from HTTP status, so a broken daemon will still be considered healthy and the watchdog will reconnect against a failing server. Fix: aggregate ok flags from each sub-check and return 503 when any required check fails, or expose a separate /readyz that does.
⚠️ **cli/internal/tui/dashboard.go:218** — resetSSE cancels the old context and reassigns d.sseEvents to a fresh channel without closing or draining the previous one. If StreamEvents is mid-send on the old channel when cancellation fires, the producer goroutine can park on the send until GC, leaking one goroutine per stall/recovery cycle. Session-ID gating in Update only filters consumed messages — it does not unblock a parked producer. Fix: close the old channel after cancel, or verify with a test that StreamEvents always selects on ctx.Done() before sending.

### Suggestions

- Consider adding a unit test for handleHealth that verifies HTTP status reflects sub-check failures once the status-aggregation fix lands.
- Add a regression test that calls resetSSE under load (e.g. producer blocked on a full channel) to ensure no goroutines leak.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

The author addressed all high and medium severity findings from the previous review. The PRAGMA quick_check was replaced with SELECT 1, goroutine counts were removed from public endpoints, shutdownMessage overloading was fixed with a dedicated sseStatusMessage field, lock contention was eliminated via atomics, and Flutter backoff was added. One low-severity item (resetSSE not closing the old channel) remains unchanged. One new low-severity issue: redundant guard in the healthCheckMsg handler. Overall the PR is in good shape.

### Issues

🟡 **cli/internal/tui/dashboard.go:193** — resetSSE() still replaces d.sseEvents without closing the old channel. After sseCancel(), any goroutine blocked on a full-buffer send to the old channel will only unblock once it observes context cancellation inside StreamEvents. If StreamEvents doesn't check the context between buffering and sending, the goroutine leaks until GC. A defensive close(oldEvents) after cancel would make cleanup deterministic.
🟡 **daemon/internal/server/handlers.go:410** — healthCheckMsg handler: the second `if d.sseStale` guard (after the err != nil branch) is always true at that point — the function already returned early on `!d.sseStale` at the top. Dead condition makes the control flow harder to follow; the inner block can be unconditional.

### Suggestions

- Consider adding a brief integration-style test for the full watchdog → health-check-fails → stays-offline path in the TUI, complementing the existing healthy-reset and stale-probe-drop tests.
- The heartbeat payload includes last_poll_age_seconds as a truncated int64; sub-second precision is lost. If downstream consumers need finer granularity, consider using float64 or milliseconds.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Previous findings addressed: /health now returns 503 on failed sub-checks, and resetSSE leaves the old channel intact while a new test verifies StreamEvents exits on context cancel even with a blocked send. New heartbeat/watchdog logic in daemon, CLI TUI, and Flutter notifier is coherent: session-ID gating prevents stale SSE messages, watchdog only triggers when not already health-checking, and the Flutter notifier backs off exponentially with a 5-minute cap. No new blocking issues found.

### Suggestions

- dashboard.go healthCheckMsg success path sets sseStatusMessage='Reconnecting SSE stream...' but relies on the next sseMsg to clear it; if reconnect itself fails, the message lingers until the next disconnect cycle — consider clearing it on sseDisconnectMsg as well.
- heartbeatEvent marshals a map[string]any per tick for every SSE subscriber; for many concurrent clients consider caching the marshaled payload per tick.
- DaemonConnectionNotifier._currentStaleAfter uses staleAfter.inSeconds*(1<<shift) which can produce up to 480s before the cap — document the intended backoff curve or derive maxReconnectDelay from staleAfter.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured PR. Previous finding #1 (resetSSE channel close) was addressed with a clarifying comment. Previous finding #2 (redundant guard) is still present. One new race condition found where a pending sseReconnectMsg can create a duplicate SSE connection after a watchdog-triggered resetSSE.

### Issues

⚠️ **cli/internal/tui/dashboard.go:399** — Race: sseReconnectMsg carries no session ID, so a stale reconnect tick from a prior sseDisconnectMsg cannot be discarded. If the watchdog's healthCheckMsg triggers resetSSE() and starts a fresh connect+listen pair during the 5-second reconnect delay, the pending sseReconnectMsg will fire and call sseCommands(), opening a second SSE HTTP connection to the same channel. This causes duplicate events and cascading disconnect/reconnect cycles. Fix: either give sseReconnectMsg a sessionID field and ignore stale ones, or cancel the reconnect tick when resetSSE() runs.
🟡 **cli/internal/tui/dashboard.go:410** — (Re-flagged, still present) The second `if d.sseStale` guard in healthCheckMsg is always true at that point — the early return on line ~401 already handles !d.sseStale. The inner block can be unconditional, which simplifies the control flow.

### Suggestions

- Add a sessionID field to sseReconnectMsg and filter stale reconnect ticks the same way sseMsg and sseDisconnectMsg are filtered, preventing duplicate SSE connections after a watchdog-driven resetSSE.
- Flatten the healthCheckMsg handler by removing the redundant `if d.sseStale` and making the resetSSE block unconditional after the err != nil early return.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Re-review: the HIGH-severity PRAGMA quick_check issue on the unauthenticated /health endpoint remains unaddressed. The Flutter watchdog also has an unhandled exception path in an unawaited future. Several medium-severity items from the first review (shutdownMessage overload, no reconnect backoff, exposed goroutine counts) are also still present.

### Issues

🔴 **daemon/internal/server/handlers.go:432** — PRAGMA quick_check on unauthenticated /health is still present (re-flag). This can be slow on larger databases and is callable by anyone. The 500ms timeout mitigates full-table scans but means the check will spuriously fail on a busy-but-healthy DB. Replace with a simple `SELECT 1` or just PingContext; reserve integrity checks for an admin-only endpoint.
⚠️ **flutter_app/lib/features/dashboard/dashboard_providers.dart:110** — _verifyAndReconnect() is called via unawaited() from _checkForStaleStream(). If apiClientProvider.checkHealth() throws (e.g. network timeout, DNS failure), the returned Future completes with an unhandled error since there is no try-catch around the await. Unhandled errors in unawaited futures can crash the app or trigger zone-level error handlers. Wrap the checkHealth() call in a try-catch that transitions to the offline phase on failure.
⚠️ **flutter_app/lib/features/dashboard/dashboard_providers.dart:115** — No backoff in _verifyAndReconnect() (re-flag). If the daemon is reachable but the SSE stream keeps going stale (e.g. network middlebox dropping long-lived connections), the watchdog fires checkHealth() + invalidate(sseStreamProvider) every 60 seconds indefinitely. Add exponential backoff or a max-retry counter before transitioning to offline.
⚠️ **cli/internal/tui/dashboard.go:289** — shutdownMessage is still overloaded for SSE watchdog status (re-flag). The watchdog writes 'SSE stream stale; checking daemon health...' and 'Reconnecting SSE stream...' to the same field the shutdown flow uses. If a user triggers shutdown while the stream is stale, these will clobber each other. Use a separate sseStatusMessage field.
⚠️ **daemon/internal/server/handlers.go:385** — runtime.NumGoroutine() exposed in unauthenticated /health and every SSE heartbeat (re-flag). This leaks internal load metrics to any caller. Consider omitting from public endpoints or gating deep checks behind auth.
🟡 **cli/internal/tui/dashboard.go:395** — Redundant guard: the second `if d.sseStale` in healthCheckMsg is always true — the handler already returned early when `!d.sseStale` on line 388. The final `return d, nil` after the block is dead code. Remove the redundant condition for clarity.

### Suggestions

- Replace PRAGMA quick_check with SELECT 1 or PingContext on /health — move integrity checks to a separate /admin/diagnostics endpoint behind authentication.
- In Flutter _verifyAndReconnect(), wrap checkHealth() in try-catch and add an exponential backoff counter (e.g. 60s, 120s, 240s, then cap at offline) to avoid infinite reconnect loops through middleboxes.
- Introduce a dedicated sseStatusMessage field in Dashboard to prevent shutdown and SSE watchdog messages from clobbering each other.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Adds liveness heartbeat watchdog: daemon emits SSE heartbeat events and deeper /health checks; TUI tracks lastSSEEvent with watchdog, health-probe, and session-id guarded reconnect. Implementation looks solid—session IDs correctly drop stale messages, resetSSE avoids close/send races, heartbeats are filtered from activity. Tests cover happy path, stale-probe-after-heartbeat, and stale-reconnect scenarios.

### Suggestions

- Consider making heartbeatInterval and the 1-minute TUI stale threshold configurable rather than hardcoded, so ops can tune for high-latency networks.
- The /health endpoint runs a SQLite PingContext+SELECT 1 on every call with a 500ms timeout—fine today, but if /health is hit at high frequency by external probes consider caching the result briefly.
- sseDisconnectMsg sets d.err = msg.err which may surface transient reconnect errors to the user; consider only surfacing after N consecutive failures or after the 5s reconnect tick fails.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previous high/medium findings have been addressed. The code is well-structured with good session-ID–based stale message filtering and solid test coverage. Only minor issues remain.

### Issues

🟡 **cli/internal/tui/dashboard.go:295** — No backoff on CLI watchdog health-check retries. When the daemon is unreachable, healthCheckMsg sets sseHealthChecking=false but leaves sseStale=true and lastSSEEvent unchanged, so the next watchdog tick (10s later) fires another health check. This loops indefinitely at one request per 10 seconds against an unresponsive daemon. Consider adding a retry counter or exponential backoff, or leaving sseHealthChecking=true on failure so the watchdog doesn't immediately re-trigger.
🟡 **cli/internal/tui/dashboard.go:355** — When a heartbeat arrives while sseHealthChecking is true, sseStale is cleared but sseStatusMessage is not (because the `if !d.sseHealthChecking` guard skips it). The stale healthCheckMsg then returns early without clearing sseStatusMessage either. The status bar shows 'SSE stream stale; checking daemon health...' until the next heartbeat (~20s). Very minor UI glitch — consider clearing sseStatusMessage in the healthCheckMsg early-return path.

### Suggestions

- Consider leaving sseHealthChecking=true on health check failure and using a simple retry counter (e.g. 3 attempts with increasing delay) before giving up — this prevents the 10s polling loop against a down daemon while still allowing recovery.
- In the healthCheckMsg early-return path (line ~388), clear sseStatusMessage so the status bar doesn't briefly show a stale message after the stream has already recovered.

---

PR #508, opened 2026-05-21
fix: persist topic-discovered repos to config even without open PRs

### LINKED ISSUE #507: bug: topic-discovered repos not persisted unless they have open PRs
## Bug

Repos discovered by tier1 via the `heimdallm-review` GitHub topic are **not persisted** to `cfg.GitHub.Repositories` (and therefore not shown by `heimdallm-cli status` or the `/config` endpoint) unless they happen to have an open PR.

## Root cause

`upsertDiscoveredRepos()` (main.go:2512) only iterates over PRs returned by `FetchPRsToReview()`. It extracts repos from those PRs and appends them to `cfg.GitHub.Repositories`. Repos discovered by tier1's topic search that have no open PRs are never passed through this function and therefore never persisted.

The tier1 discovery correctly finds all 23 repos (visible in logs as `tier1: discovery complete repos=23`) and publishes them via NATS. Tier2 receives the full list and uses it as a filter for PRs. However, the persistence path (`upsertDiscoveredRepos` → store `SetConfig("repositories", ...)`) only runs for repos that appear in the PR list.

## Expected behavior

All repos discovered by tier1 topic search should be persisted to `cfg.GitHub.Repositories` and visible in `heimdallm-cli status` and the `/config` endpoint immediately after discovery, regardless of whether they have open PRs.

## Steps to reproduce

1. Fresh install with empty DB
2. Configure `discovery_topic = "heimdallm-review"` with 23+ repos tagged
3. Only repos with open PRs appear in `heimdallm-cli status`
4. Logs show `tier1: discovery complete repos=23` but status shows 1-2

## Suggested fix

In `sendDiscoveryRepos()` or the tier2 repo-list receiver, persist the discovered repos to `cfg.GitHub.Repositories` (and the store) the same way `upsertDiscoveredRepos` does for PR-sourced repos. The `upsertDiscoveredRepos` function should also be called (or its logic duplicated) for the tier1 discovered list, not just the PR list.

Relevant code locations:
- `daemon/cmd/heimdallm/main.go:2512` — `upsertDiscoveredRepos` (only iterates PRs)
- `daemon/cmd/heimdallm/main.go:2160` — `sendDiscoveryRepos` (tier1, publishes but doesn't persist)
- `daemon/cmd/heimdallm/main.go:2748` — `FetchPRsToReview` (calls upsertDiscoveredRepos)
- `daemon/cmd/heimdallm/main.go:1392` — `configFn` (reads `cfg.GitHub.Repositories` for `/config`)

Introduced in: `62686d0` (epic: embed NATS in backend)

sergiotejon:
## Summary

- Fixes #507: repos discovered by tier1's topic search now persist to `cfg.GitHub.Repositories` immediately, even when they have no open PRs
- Adds `upsertDiscoveredFromTopics()` method on `tier2Adapter` that runs when tier2 receives the repo list from tier1 via NATS
- Uses the existing `processDiscoveredRepos()` path to write to the K/V store and emit SSE `repo_discovered` events

## Root cause

`upsertDiscoveredRepos()` only iterates over PRs returned by `FetchPRsToReview()`. Repos discovered by tier1's topic search that have no open PRs were never passed through this function and therefore never persisted — invisible to `heimdallm-cli status` and the Flutter UI.

## Testing

- `go vet ./...` passes
- All `cmd/heimdallm` tests pass (`make test-docker`)
- The new method reuses the same `processDiscoveredRepos` helper already covered by existing discovery tests

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -2323,6 +2323,11 @@ func runTier2(
 				repos = r
 				mu.Unlock()
 				slog.Info("tier2: received repo list", "count", len(r))
+
+				// Persist topic-discovered repos so they appear in
+				// heimdallm-cli status and GET /config even when they
+				// have no open PRs. Fixes #507.
+				adapter.upsertDiscoveredFromTopics(r)
 			}
 		}
 	}()
@@ -2557,6 +2562,55 @@ func upsertDiscoveredRepos(c *config.Config, prs []*gh.PullRequest) []string {
 		added = append(added, pr.Repo)
 	}
 	return added
+}
+
+// upsertDiscoveredFromTopics persists repos received from tier1's topic
+// discovery into cfg.GitHub.Repositories (or NonMonitored) and invokes
+// processDiscoveredRepos to write them to the K/V store. This ensures repos
+// discovered by topic appear in heimdallm-cli status and GET /config even
+// when they have no open PRs. Fixes #507.
+func (a *tier2Adapter) upsertDiscoveredFromTopics(repos []string) {
+	if len(repos) == 0 {
+		return
+	}
+
+	a.cfgMu.Lock()
+	cfg := *a.cfg
+
+	known := make(map[string]struct{}, len(cfg.GitHub.Repositories)+len(cfg.GitHub.NonMonitored))
+	for _, r := range cfg.GitHub.Repositories {
+		known[r] = struct{}{}
+	}
+	for _, r := range cfg.GitHub.NonMonitored {
+		known[r] = struct{}{}
+	}
+
+	enable := cfg.GitHub.AutoEnablePRForDiscovery()
+	var added []string
+	for _, repo := range repos {
+		if repo == "" {
+			continue
+		}
+		if _, ok := known[repo]; ok {
+			continue
+		}
+		if enable {
+			cfg.GitHub.Repositories = append(cfg.GitHub.Repositories, repo)
+		} else {
+			cfg.GitHub.NonMonitored = append(cfg.GitHub.NonMonitored, repo)
+		}
+		known[repo] = struct{}{}
+		added = append(added, repo)
+	}
+
+	reposSnap := append([]string(nil), cfg.GitHub.Repositories...)
+	nonMonSnap := append([]string(nil), cfg.GitHub.NonMonitored...)
+	a.cfgMu.Unlock()
+
+	if len(added) > 0 {
+		slog.Info("tier2: persisting topic-discovered repos", "added", len(added), "repos", added)
+		processDiscoveredRepos(added, reposSnap, nonMonSnap, a.store, a.broker, time.Now())
+	}
 }
 
 // ── tier2Adapter bridges main.go's concrete types to Pipeline interfaces ──
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -2607,6 +2607,10 @@ func (a *tier2Adapter) upsertDiscoveredFromTopics(repos []string) {
 	nonMonSnap := append([]string(nil), cfg.GitHub.NonMonitored...)
 	a.cfgMu.Unlock()
 
+	// Intentionally release the mutex before persisting to the K/V store:
+	// processDiscoveredRepos performs I/O (SQLite writes, SSE publish) that
+	// should not hold the config lock. The in-memory config is already
+	// mutated above; the snapshots capture the post-mutation state.
 	if len(added) > 0 {
 		slog.Info("tier2: persisting topic-discovered repos", "added", len(added), "repos", added)
 		processDiscoveredRepos(added, reposSnap, nonMonSnap, a.store, a.broker, time.Now())
--- a/daemon/cmd/heimdallm/main_discover_topics_test.go
+++ b/daemon/cmd/heimdallm/main_discover_topics_test.go
@@ -1,0 +1,141 @@
+package main
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/heimdallm/daemon/internal/config"
+	"github.com/heimdallm/daemon/internal/sse"
+)
+
+func TestUpsertDiscoveredFromTopics_AddsNewRepos(t *testing.T) {
+	cfg := &config.Config{}
+	cfg.GitHub.Repositories = []string{"org/existing"}
+
+	s := newMemStore(t)
+	broker := sse.NewBroker()
+	broker.Start()
+	defer broker.Stop()
+
+	var mu sync.Mutex
+	a := &tier2Adapter{
+		cfgMu:  &mu,
+		cfg:    &cfg,
+		store:  s,
+		broker: broker,
+	}
+
+	a.upsertDiscoveredFromTopics([]string{"org/existing", "org/new1", "org/new2"})
+
+	if len(cfg.GitHub.Repositories) != 3 {
+		t.Fatalf("expected 3 repos, got %v", cfg.GitHub.Repositories)
+	}
+	want := map[string]bool{"org/existing": true, "org/new1": true, "org/new2": true}
+	for _, r := range cfg.GitHub.Repositories {
+		if !want[r] {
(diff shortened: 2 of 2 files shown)


vbuenog (changes requested):
## 🤖 Heimdallm AI Review

Adds a function to persist topic-discovered repos to config so they appear in CLI status and GET /config even without open PRs. The fix targets a real bug (#507), but the implementation has a significant correctness flaw: mutations are applied to a local copy of the Config struct rather than the underlying *a.cfg, so the in-memory daemon config is not actually updated.

### Issues

🔴 **daemon/cmd/heimdallm/main.go:2581** — `cfg := *a.cfg` makes a shallow copy of the Config struct. Subsequent `cfg.GitHub.Repositories = append(...)` and `cfg.GitHub.NonMonitored = append(...)` only mutate the local copy's slice headers — `a.cfg.GitHub.Repositories`/`NonMonitored` (the live in-memory config) are never updated. This means: (1) GET /config served from a.cfg may still omit the discovered repos until a config reload, partially defeating the stated fix; (2) on the next invocation, the `known` map is rebuilt from a.cfg, so previously-added repos are re-classified as new and re-persisted on every tier1 push, producing spurious 'persisting topic-discovered repos' logs and redundant K/V writes. Compare to `upsertDiscoveredRepos(c *config.Config, ...)` (line 2559) which takes a pointer and mutates in place.
⚠️ **daemon/cmd/heimdallm/main.go:2608** — `reposSnap`/`nonMonSnap` are built from the local `cfg` (the copy) rather than the canonical `a.cfg`. If concurrent code paths update `a.cfg.GitHub.Repositories` (e.g., PR-driven discovery via `upsertDiscoveredRepos`) between reads, the snapshot passed to `processDiscoveredRepos` will not include those entries, since the local copy was taken from a stale `a.cfg`. Combined with the issue above, the K/V state written here can diverge from the daemon's live config.
🟡 **daemon/cmd/heimdallm/main.go:2569** — No nil-check on `a.cfg`, `a.store`, or `a.broker`. Other call sites in this file appear to assume these are populated, but adding a defensive nil-check for `a.cfg` (consistent with how the adapter is initialized) would prevent panics if tier1 pushes a repo list before the adapter is fully wired.
🟡 **daemon/cmd/heimdallm/main.go:2326** — The call to `adapter.upsertDiscoveredFromTopics(r)` happens inside the goroutine that owns `mu` for `repos`, but `upsertDiscoveredFromTopics` acquires its own `a.cfgMu`. Confirm there is no lock-ordering hazard with other paths that may hold `cfgMu` and then try to take `mu` — at a glance this seems fine, but it is worth a second pass given the new cross-lock interaction.

### Suggestions

- Change the function to mutate `a.cfg` directly under the lock (e.g., operate on `a.cfg.GitHub.Repositories` / `a.cfg.GitHub.NonMonitored` rather than a local `cfg` copy), mirroring `upsertDiscoveredRepos`. Then build `reposSnap`/`nonMonSnap` from `a.cfg` before releasing the lock.
- Add a unit test that calls `upsertDiscoveredFromTopics` twice with overlapping repo lists and asserts the second call adds zero repos and emits no 'persisting' log — this would have caught the shallow-copy bug.
- Consider injecting a clock (e.g., `a.now func() time.Time`) instead of calling `time.Now()` directly, to make the path testable alongside the rest of the adapter.
- Log at debug, not info, when `len(added) == 0` is the common steady state — or keep the current early-exit but ensure the dedup actually works (see issue 1) so the info log fires only on real additions.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

The PR correctly wires topic-discovered repos into the in-memory config and K/V store. The established `cfg := *a.cfg` pattern (with `a.cfg` being `**config.Config`) propagates mutations correctly, and `processDiscoveredRepos` is called with consistent snapshots taken under the mutex. The main gaps are the absence of unit tests for non-trivial deduplication logic and mild code duplication with the existing `upsertDiscoveredRepos` function.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2564** — No unit tests are provided for `upsertDiscoveredFromTopics`. The method contains non-trivial business logic: it categorises repos into `Repositories` vs `NonMonitored` based on `AutoEnablePRForDiscovery()`, deduplicates against both lists, and produces the snapshots forwarded to `processDiscoveredRepos`. A bug in any of these branches (e.g. a repo appearing in both lists, or `AutoEnablePRForDiscovery` toggling mid-run) would be silent. The existing codebase has tests for similar persistence paths; this path should be covered too. Recommended fix: add a table-driven unit test that exercises at minimum: (a) all repos already known → nothing added; (b) mix of known and new repos with `enable=true`; (c) same with `enable=false`; (d) empty input list.

### Suggestions

- The `known`-map building block and the `Repositories`/`NonMonitored` routing logic are duplicated verbatim from `upsertDiscoveredRepos`. Consider extracting a shared helper `classifyNewRepos(cfg *config.Config, repos []string) (added, reposSlice, nonMonSlice []string)` to reduce drift risk.
- The inline `// Fixes #507` reference in both the call-site comment and the function doc comment will become stale once the issue is closed. Issue linkage belongs in the commit message or PR description, not in source comments.
- The lock is released before `processDiscoveredRepos` is called, which is intentional to avoid holding it during I/O. Consider adding a short comment to make this deliberate unlock-before-persist explicit, since readers might otherwise wonder whether the gap between in-memory update and K/V write is accidental.

---

sergiotejon (commented):
@vbuenog regarding the HIGH severity issue on `cfg := *a.cfg`:

`a.cfg` is typed as `**config.Config` (pointer-to-pointer). Dereferencing once with `cfg := *a.cfg` yields a `*config.Config` — i.e. the same pointer that the live daemon config holds. Mutations to `cfg.GitHub.Repositories` (via `append`) **do** propagate to the in-memory config because we're operating on the pointed-to struct, not a copy.

This is the same pattern used in `FetchPRsToReview` (line 2775: `cfg := *a.cfg`) which has been working correctly since the tier2 refactor. The Muriano review correctly identifies this.

That said, I'll add unit tests to cover the dedup and persistence logic as both reviews suggest, and I'll add an explicit comment above the unlock clarifying the intentional lock-release-before-persist pattern.

Muriano (approved):
## 🤖 Heimdallm AI Review

The previous medium finding (missing tests for upsertDiscoveredFromTopics) has been fully addressed. The new test file covers all four branches requested: all-known repos (SkipsKnownRepos), mix of known/new with enable=true (AddsNewRepos), enable=false path (RespectsDisabledFlag), empty input (EmptyInput), and idempotency (IdempotentOnSecondCall). The implementation in main.go is unchanged from the previous review and remains correct: the mutex is held for in-memory mutation, snapshots are captured before unlock, and I/O is performed outside the lock with clear documentation. No new issues introduced.

### Suggestions

- TestUpsertDiscoveredFromTopics_EmptyInput passes broker: nil which is safe only because the early-return guard fires before processDiscoveredRepos is reached. Consider adding a non-nil broker for defensive completeness, or adding a comment explaining why nil is safe here.
- The five tests share identical broker setup boilerplate (newMemStore, sse.NewBroker, Start/Stop, sync.Mutex, tier2Adapter literal). A small test helper that returns a ready *tier2Adapter given a *config.Config would reduce repetition and make future test additions easier.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: prior HIGH/MEDIUM findings were based on a mistaken assumption about a.cfg's type — it is **config.Config (main.go:2574), so `cfg := *a.cfg` yields a pointer and mutations propagate to the live config. Tests confirm this behavior. The implementation is correct: it holds cfgMu during config mutation, releases before I/O, and the idempotency test verifies known repos are skipped on subsequent calls. No remaining significant issues.

### Suggestions

- Optional: add a defensive nil-check on a.cfg at function entry to avoid a panic if tier1 ever pushes a repo list before the adapter is fully wired (low priority — the goroutine is started after adapter construction).
- Optional: extract the shared known-map + append-to-Repositories-or-NonMonitored logic into a helper used by both upsertDiscoveredRepos and upsertDiscoveredFromTopics to keep the two discovery paths in lockstep as the config schema evolves.

---

PR #517, opened 2026-06-01
test(rename): guard fakeDispatcher with mutex to fix -race flake

Muriano:
## Summary

- `fakeDispatcher` in `probe_test.go` had no synchronization, so when `TestRenameProbe_Run_FiresInitialTickBeforeIntervalElapses` polled `dispatcher.calls` from the test goroutine, it raced with the probe-side write inside `(*fakeDispatcher).Run`.
- Added `sync.Mutex` to `fakeDispatcher` and locked both the write in `Run` and the polling read in the test, mirroring the pattern already used by `fakeCanonical` in the same file.

## Why

Without this, `go test ./internal/rename/... -race` flaked roughly 1 in 3 runs on `main@a994086`. The flake masks any future race regressions and noise-fails CI.

```
WARNING: DATA RACE
Read at 0x...  TestRenameProbe_Run_FiresInitialTickBeforeIntervalElapses probe_test.go:168
Previous write at 0x...  (*fakeDispatcher).Run            probe_test.go:47
```

The other dispatcher-reading tests in the file (`Tick`-based) are synchronous, so they don't race and don't need changes.

## Test plan

- [x] `go test ./internal/rename/... -race -count=30 -run TestRenameProbe_Run_FiresInitialTickBeforeIntervalElapses` → all 30 pass (was ~1/3 fail before).
- [x] `go test ./internal/rename/... -race -count=1` → full package PASS.
- [x] `go vet ./...` clean.

### DIFF
--- a/daemon/internal/rename/probe_test.go
+++ b/daemon/internal/rename/probe_test.go
@@ -33,6 +33,15 @@ func (f *fakeCanonical) GetCanonicalFullName(repo string) (string, error) {
 	return r.canonical, r.err
 }
 
+// callCount returns f.calls under the mutex. Use from any test goroutine
+// that may run concurrently with GetCanonicalFullName (e.g. Run-based
+// tests) instead of inlining lock/read/unlock at each call site.
+func (f *fakeCanonical) callCount() int {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return f.calls
+}
+
 // fakeDispatcher records reconciler invocations so the test can pin
 // which (old, new) pairs the probe dispatched. Concrete *Reconciler
 // is too heavy to construct in a probe test; the Dispatcher interface
@@ -46,11 +55,19 @@ type fakeDispatcher struct {
 
 func (f *fakeDispatcher) Run(_ context.Context, oldRepo, newRepo string) error {
 	f.mu.Lock()
+	defer f.mu.Unlock()
 	f.calls++
 	f.gotPairs = append(f.gotPairs, [2]string{oldRepo, newRepo})
-	err := f.err
-	f.mu.Unlock()
-	return err
+	return f.err
+}
+
+// callCount returns f.calls under the mutex. Use from any test goroutine
+// that may run concurrently with Run (e.g. Run-based tests) instead of
+// inlining lock/read/unlock at each call site.
+func (f *fakeDispatcher) callCount() int {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return f.calls
 }
 
 func newProbe(t *testing.T, canonical *fakeCanonical, dispatcher *fakeDispatcher,
@@ -166,12 +183,8 @@ func TestRenameProbe_Run_FiresInitialTickBeforeIntervalElapses(t *testing.T) {
 	// the 1h interval that the loop tick would need.
 	deadline := time.After(2 * time.Second)
 	for {
-		canonical.mu.Lock()
-		calls := canonical.calls
-		canonical.mu.Unlock()
-		dispatcher.mu.Lock()
-		dispatched := dispatcher.calls
-		dispatcher.mu.Unlock()
+		calls := canonical.callCount()
+		dispatched := dispatcher.callCount()
 		if calls >= 1 && dispatched >= 1 {
 			break
 		}
--- a/daemon/internal/rename/probe_test.go
+++ b/daemon/internal/rename/probe_test.go
@@ -41,12 +41,16 @@ type fakeDispatcher struct {
 	calls    int
 	gotPairs [][2]string
 	err      error
+	mu       sync.Mutex
 }
 
 func (f *fakeDispatcher) Run(_ context.Context, oldRepo, newRepo string) error {
+	f.mu.Lock()
 	f.calls++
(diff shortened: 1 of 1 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

Correct and idiomatic fix for a -race flake. Adds a sync.Mutex to fakeDispatcher and guards both the writer (Run) and the test's reader of calls. The read-out of f.err is correctly captured inside the lock before returning. No correctness concerns with the shown changes.

### Issues

🟡 **daemon/internal/rename/probe_test.go:41** — The mutex now guards calls, gotPairs, and err in Run, but ensure all OTHER accesses to dispatcher.calls/gotPairs elsewhere in this test file (e.g. final assertions after the loop) are also taken under dispatcher.mu. If the probe goroutine can still be running when those reads happen, an unguarded access remains a race. Not visible in this diff but worth verifying.

### Suggestions

- Consider adding a small helper method on fakeDispatcher (e.g. callCount() int that locks/reads/unlocks) to avoid repeating the lock/read/unlock dance at each call site and reduce the chance of a future unguarded read.
- Double-check the post-loop assertions in TestRenameProbe_Run_FiresInitialTickBeforeIntervalElapses (and any other test using fakeDispatcher) read calls/gotPairs under the mutex, since the dispatch goroutine may not be fully stopped at assertion time.
- If the probe is deterministically stopped/joined before final assertions, a brief comment noting that would clarify why those later reads are lock-free safe.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, minimal fix for a data race on fakeDispatcher fields accessed concurrently in the Run-based test. The mutex is correctly scoped to the one test that launches a goroutine; synchronous Tick-based tests don't need it.

### Issues

🟡 **daemon/internal/rename/probe_test.go:46** — Prefer `defer f.mu.Unlock()` instead of manual unlock-before-return. This eliminates the temporary `err` variable, is more idiomatic Go, and is safer against future modifications that might add early returns.

### Suggestions

- Use `defer f.mu.Unlock()` in fakeDispatcher.Run and return f.err directly — simpler, removes the local `err` copy, and guards against future early-return additions.

---

Muriano:
## Review feedback addressed in ba3c378ec6d435d722f03f1446ccbfb8ec52233c

Both suggestions applied — verified against the file before changing anything.

| Comment | Reviewer | Fix |
|---|---|---|
| 🟡 Prefer `defer f.mu.Unlock()` over manual unlock-before-return | @sergiotejon | `fakeDispatcher.Run` now uses `defer f.mu.Unlock()` and returns `f.err` directly. The temp `err` copy is gone. |
| 🟡 Encapsulate the lock/read/unlock dance so a future unguarded read doesn't silently re-introduce the -race regression | @vbuenog | Added `callCount() int` on both `fakeDispatcher` and `fakeCanonical`. The polling loop in `TestRenameProbe_Run_FiresInitialTickBeforeIntervalElapses` now uses the helpers instead of inlining mutex bookkeeping. |

### On the post-loop assertion concern (@vbuenog)

Verified: the only test that runs the probe in a goroutine (`TestRenameProbe_Run_FiresInitialTickBeforeIntervalElapses`) doesn't read `dispatcher.calls` or `gotPairs` after the polling loop — it only waits on `done` after `cancel()`. All other dispatcher reads are in `Tick`-based tests where the call returns synchronously before the assertions, so no race window. The new `callCount()` helper is the preventive net for future contributors.

### Verification

```
go test ./internal/rename/... -race -count=30 -run TestRenameProbe_Run_FiresInitialTickBeforeIntervalElapses → 30/30 PASS
go test ./internal/rename/... -race -count=1                                                                → PASS
go vet ./...                                                                                                → clean
go build ./cmd/heimdallm                                                                                    → clean
```

cc @ivanmunozruiz — added you as reviewer in case you want a second look.

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Test-only fix: protege fakeDispatcher con mutex y encapsula las lecturas de contadores en helpers callCount() para eliminar la carrera detectada por -race en el test Run-based. Cambio limpio y bien acotado, alineado con el feedback previo del PR.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Previous feedback fully addressed. The defer-based unlock and callCount() helpers are clean and idiomatic. No new issues in the updated diff.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Test-only change that correctly guards fakeDispatcher with a mutex and encapsulates guarded reads via callCount() helpers on both fakes. Both prior review findings are addressed: Run uses defer-unlock and returns f.err directly, and the goroutine-based test reads only through the helpers with no post-loop unguarded access. The remaining direct dispatcher.calls reads are confined to synchronous Tick-based tests, so no race exists. Clean.

### Suggestions

- Optional consistency nit (not a race): the Tick-based tests still read dispatcher.calls/canonical.calls directly. They're safe because Tick is synchronous, but routing them through the new callCount() helpers would make the 'always read via helper' convention uniform and reduce the chance a future contributor copies a direct-read pattern into a concurrent test.

---

PR #516, opened 2026-06-01
fix(github): paginate PR comments to capture newest entries (closes #512)

Muriano:
## Summary

- `fetchReviewComments` and `fetchIssueComments` now paginate `/pulls/:n/comments` and `/issues/:n/comments` with `per_page=100` and follow every page, matching the pattern already used by `GetPRTimelineEventsForReviewer`.
- Pagination terminates on a short page (`<100` items) and is hard-capped at 50 pages so a misbehaving upstream can't make us iterate forever.
- Mid-pagination HTTP errors propagate as hard failures — no silently truncated slices.

## Why

GitHub returns 30 items per page by default for both comment endpoints, sorted **ascending** by `created_at`. The previous implementation called both without `per_page` or `page` parameters, so any PR with >30 review or issue comments silently dropped the newest entries — exactly the comments a re-review depends on (peer reviewer feedback, the latest \`heimdallm:done\` markers, etc.).

This was the root cause of #512.

## Test plan

New tests in `daemon/internal/github/poller_test.go`:

- [x] `TestFetchComments_PaginatesReviewAndIssueComments` — both endpoints return 100+50 across two pages; FetchComments returns all 300 merged entries and we explicitly assert a tail-page body survives the round trip.
- [x] `TestFetchComments_ShortFirstPageStopsPaging` — single short page must NOT trigger a redundant second request (efficiency guard for the common case).
- [x] `TestFetchComments_FullPageThenEmptyStopsCleanly` — total count is an exact multiple of 100; pagination must terminate on the empty second page without erroring.
- [x] `TestFetchComments_MidPaginationErrorPropagates` — 500 on page 2 of `/pulls/:n/comments` surfaces a hard error (no partial slice).
- [x] `TestFetchComments_MidPaginationErrorOnIssuesPropagates` — symmetric guard for the issue-comments leg, exercised via `FetchIssueCommentsOnly`.

Verification:

\`\`\`
$ cd daemon && go test ./internal/github/... ./internal/pipeline/... ./internal/issues/... ./cmd/... -race -count=1
ok  	github.com/heimdallm/daemon/internal/github	1.408s
ok  	github.com/heimdallm/daemon/internal/pipeline	3.661s
ok  	github.com/heimdallm/daemon/internal/issues	3.716s
ok  	github.com/heimdallm/daemon/cmd/heimdallm	4.814s
\`\`\`

[...]

PR #521, opened 2026-06-04
fix(github): return error on timeline pagination cap instead of partial results (closes #519)

### LINKED ISSUE #519: github: GetPRTimelineEventsForReviewer has the same cap-hit silent-truncation bug as fetchComments (now fixed in #516)
## Context

Surfaced by @vbuenog, @sergiotejon and @ivanmunozruiz while reviewing #516.

#516 tightened the comment pagination contract: when the safety cap (`maxCommentPages = 50`) is hit, `fetchReviewComments` and `fetchIssueComments` now return an explicit error instead of `slog.Warn` + partial slice. The rationale: silently returning partial results past the cap is exactly the failure mode #512 was opened to fix (newest comments lost without the caller knowing).

`GetPRTimelineEventsForReviewer` (`daemon/internal/github/client.go:808-812`) has the identical anti-pattern:

```go
if page > 50 {
    slog.Warn(\"github: timeline pagination cap hit, returning partial results\",
        \"repo\", repo, \"pr\", number, \"pages\", page)
    break
}
```

This means a runaway timeline (>5,000 events on one PR — rare but possible on long-lived PRs with churning reviewers/dismissals) would drop the tail of events. Those tail events are exactly what `shouldBypassSHASkipForReReview` walks backward to detect re-review intent (#322), so a silent truncation here could miss a re-review request and skip the review entirely.

## Scope

1. **Fix the cap-hit contract** in `GetPRTimelineEventsForReviewer`: return an error rather than partial results, matching `fetchReviewComments` / `fetchIssueComments` post-#516.
2. **Share the pagination constants** between timeline and comments. Today both files independently hardcode `per_page=100` and `cap=50`. Extract package-level `perPage` (already added in #516) and rename `maxCommentPages` → `maxCommentPaginationPages` or introduce a shared `maxPaginationPages` and reuse it.
3. **Add a test** mirroring `TestFetchComments_PaginationCapReturnsError` for the timeline path.

## Acceptance criteria

- `GetPRTimelineEventsForReviewer` returns an error when the cap is reached; existing callers (`pipeline.shouldBypassSHASkipForReReview`) handle the error path (probably fail-closed: don't bypass SHA skip if we can't read the full timeline).
- One shared constant for `per_page` and one for the pagination cap across all paginated endpoints in `client.go`.
- New test covers the cap-hit error path for the timeline.

## Why not in #516

Scope-strict. #516 fixed comments; this is the same pattern in a sibling function and deserves its own PR/review so reviewers can audit the timeline-specific caller behavior change (callers that previously got partial timelines will now get an error).

ivanmunozruiz:
## Summary

`GetPRTimelineEventsForReviewer` had the same cap-hit silent-truncation anti-pattern that #516 fixed in `fetchReviewComments`/`fetchIssueComments`: on hitting the 50-page cap it logged `slog.Warn` and returned a partial slice. The tail events are exactly what `pipeline.shouldBypassSHASkipForReReview` walks backward to detect re-review intent (#322 Bug 5), so a silent truncation could miss a re-review request and skip the review entirely.

Closes #519.

## Changes

- **Cap-hit contract**: the timeline fetcher now returns a hard error instead of partial results, matching the post-#516 contract of the comment fetchers. The only caller (`pipeline.shouldBypassSHASkipForReReview`, `pipeline.go:213-218`) already fails closed on error — the SHA skip stays in place — so no caller changes were needed.
- **Shared pagination constants**: `maxCommentPages` → `maxPaginationPages`, reused (together with `perPage`) by the timeline loop. The `per_page` query param and the short-page termination check can no longer drift apart across endpoints.
- **Test**: `TestGetPRTimelineEventsForReviewer_PaginationCapReturnsError` mirrors `TestFetchComments_PaginationCapReturnsError` — asserts the error, the nil slice, and that exactly 50 requests are made (no walk past the cap).

## Test plan

- [x] TDD: new test written first and watched fail (`got nil and 5000 events` on the pre-fix code)
- [x] `make test-docker` full suite green (`go vet` + all packages, `-count=1`)

### DIFF
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -744,15 +744,17 @@ type TimelineEvent struct {
 // regardless of the user's explicit intent.
 //
 // Returns an empty slice (not nil) when no relevant events exist —
-// callers can range over the result without a nil guard.
+// callers can range over the result without a nil guard. Returns an
+// error (never a partial slice) when the pagination cap is hit, so
+// callers can fail closed instead of acting on a truncated timeline
+// (#519).
 func (c *Client) GetPRTimelineEventsForReviewer(repo string, number int, login string) ([]TimelineEvent, error) {
 	if login == "" {
 		return nil, fmt.Errorf("github: GetPRTimelineEventsForReviewer: empty login")
 	}
 	out := []TimelineEvent{}
-	page := 1
-	for {
-		path := fmt.Sprintf("/repos/%s/issues/%d/timeline?per_page=100&page=%d", repo, number, page)
+	for page := 1; page <= maxPaginationPages; page++ {
+		path := fmt.Sprintf("/repos/%s/issues/%d/timeline?per_page=%d&page=%d", repo, number, perPage, page)
 		resp, err := c.do("GET", path, "application/vnd.github+json")
 		if err != nil {
 			return nil, fmt.Errorf("github: fetch timeline: %w", err)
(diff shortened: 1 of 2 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, focused fix that makes GetPRTimelineEventsForReviewer fail closed on the pagination cap (returning an error) instead of silently returning a truncated timeline, consistent with the post-#516 contract of fetchReviewComments/fetchIssueComments. It also consolidates the duplicated maxCommentPages/maxPaginationPages constant and replaces hardcoded per_page=100 / len<100 with the shared perPage constant, eliminating drift between the query param and the short-page termination check. The defensive sort is correctly relocated into the short-page return branch. A matching test locks in the cap-hit error contract and the no-51st-request bound. No confirmed bugs found.

### Suggestions

- Boundary edge (pre-existing, now shared): a timeline whose event count is an exact multiple of perPage (e.g. exactly 5,000 events = 50 full pages) will return the cap error even though no further pages exist, because the loop never observes a short page. This is consistent with the existing fetchReviewComments/fetchIssueComments behavior and 5,000 timeline events is unrealistic, so it's not a blocker — but a one-line comment noting the off-by-one at the exact-multiple boundary would help future readers.
- The function-level doc comment update is good; consider also noting in the GetPRTimelineEventsForReviewer doc (or the cap-hit log/error) that the threshold is intentionally shared across all three paginated endpoints via maxPaginationPages, so a future tuning of one endpoint doesn't surprise the others.
- Optional test hardening: emitTimelinePage uses an identical created_at for every event, so the relocated sort path is exercised but its ordering effect is not asserted. A small happy-path test feeding out-of-order timestamps across two pages and asserting ascending order on return would cover the sort behavior the cap test cannot.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Clean, well-reasoned refactor that makes GetPRTimelineEventsForReviewer return a hard error on pagination-cap-hit instead of silently-truncated partial results (#519), and consolidates maxCommentPages/maxPaginationPages plus the perPage constant across all three paginated endpoints. The loop conversion to `for page := 1; page <= maxPaginationPages; page++` preserves the exact 50-request behavior (verified against the original page++/cap logic), the defensive sort is correctly moved onto the short-page return path, and a parallel test mirrors the comment-fetcher contract. No correctness bugs found.

### Issues

🟡 **daemon/internal/github/client.go:806** — Exact-multiple edge case: a PR with exactly 5,000 (maxPaginationPages*perPage) timeline events produces 50 full pages, so the loop exits to the error path and the function fails closed even though no events were actually dropped (no short page ever observed to confirm the end). This is a false positive that would block an otherwise-valid PR. It is consistent with the existing fetchReviewComments/fetchIssueComments behavior and the threshold is far above realistic PR sizes, so impact is negligible — noting only for completeness.

### Suggestions

- Confirm every caller of GetPRTimelineEventsForReviewer (notably pipeline.shouldBypassSHASkipForReReview) actually propagates/handles the now-possible error and fails closed as the PR comment claims, since this changes a previously non-error return path into an error one.
- Consider a brief comment or test note acknowledging the exact-5,000-event false-positive so a future reader knows the cap-hit error can fire without true truncation.

---

PR #537, opened 2026-06-15
feat(daemon): GitHub API efficiency — ETag, aggregated search, GH-aware throttle, adaptive polling, GraphQL

Muriano:
## Summary

Fixes the daemon **exhausting GitHub's core REST budget (5000/h)** and 403-storming. Root cause: issues were fetched **one REST call per repo per cycle** (~75 repos × 60 cycles/h ≈ 4500/h), with **no ETag caching** and a **non-GitHub-aware** local limiter. Delivered in 3 phases, all knobs **configurable from config.toml, GUI and TUI**, on the **personal token** (no GitHub App / webhooks). Defaults reproduce prior behavior (unconfigured = no change); the DX (detection latency) is preserved.

### Phase 1 — stop the hemorrhage (zero DX cost)
- **ETag/304 conditional caching** (C1): every polling GET sends `If-None-Match`; 304s are served from cache and **don't count against the rate limit**. Keyed by Accept+path; skips large/paginated bodies. Biggest reduction.
- **Aggregated issue Search** (C2): replaces O(repos) per-repo `GET /repos/{repo}/issues` with **one `GET /search/issues`** (assignee-grouped), using the **separate search budget**; per-repo REST fallback preserved; client-side classify/filter keeps behavior equivalent.
- **GitHub-aware rate limiter** (C3): reads `X-RateLimit-*` + `Retry-After`, throttles proactively before 403 (per-tier thresholds keep the review-loop alive), honors secondary-limit cooldowns. Write paths (SubmitReview/PostComment) report headers too.

### Phase 2 — adaptive + configurable (3 layers)
- **`[polling]` config** (C5): poll/min/max/discovery/tier3 intervals, `adaptive`, `rate_limit_safety_threshold`, `use_etag`/`use_graphql` kill-switches. Exposed in `GET/PATCH /config`, an editable **GUI** section, and the **TUI** Config tab.
- **Adaptive interval engine** (C5): idle repos back off (min→max), active repos stay at min. Opt-in (`adaptive=false` default → all repos every cycle).
- **Tier3** (C6): already conditional (ETag) + per-item exponential backoff + configurable tick.

### Phase 3 — GraphQL batching
- **GraphQL issue search** (C4): when `use_graphql=true`, fetch issues via the **idle GraphQL budget** with automatic REST-Search fallback. Default OFF.

## Test Plan
[...]

PR #539, opened 2026-06-18
feat(config): accept arbitrary poll_interval in [1m, 24h], drop fixed whitelist (closes #538)

### LINKED ISSUE #538: Allow arbitrary poll_interval values (e.g. 3m), not just the {1m,5m,30m,1h} whitelist
## Problem

`poll_interval` is validated against a hard-coded whitelist `{1m, 5m, 30m, 1h}`. Any other value — e.g. a sensible `3m` — is rejected, and on the daemon load path this **crash-loops the bundled daemon**:

```
level=ERROR msg="config load failed" path=~/.config/heimdallm/config.toml \
  err="config: invalid poll_interval \"3m\" (valid: 1m, 5m, 30m, 1h)"
```

Symptom observed: the desktop app's UI starts but its daemon dies on every spawn attempt (config never loads), so nothing listens on :7842. Had to edit the config back to a whitelisted value (`5m`) to recover.

## Where it's enforced (two places, both need relaxing)

- `daemon/internal/config/config.go:18` — the whitelist map (`"1m","5m","30m","1h"`) and the load-time validation at `config.go:1212`.
- `daemon/internal/server/handlers.go:689` — `validPollIntervals` allowlist enforced on `PATCH`/`PUT /config` (so the GUI/API also can't set `3m`).

## Proposal

Accept **any value parseable by `time.ParseDuration`** within sane bounds instead of a fixed whitelist — e.g. minimum `30s` (or `1m`) and a reasonable maximum (`1h`/`24h`). Apply the same relaxed rule in both enforcement points (config load + HTTP config handler) so `3m`, `10m`, etc. are valid everywhere (TOML, GUI, TUI).

This keeps a guardrail against pathological values (e.g. `1s`) while removing the arbitrary discrete set.

## Note

PR #537 (GitHub API efficiency) already adds a flexible `[polling]` section with free-form interval parsing (`ResolvedPollInterval`/`min`/`max`), but it does **not** remove the legacy `[github].poll_interval` whitelist in `config.go`/`handlers.go` — so this validation still rejects `3m` independently. This issue tracks relaxing that legacy validation.

vbuenog:
## Summary

`poll_interval` was validated against a hard-coded whitelist `{1m, 5m, 30m, 1h}`. Any other value — e.g. a sensible `3m` — was rejected, and on the daemon load path this **crash-looped the bundled daemon** (config never loads → nothing listens on `:7842`). This replaces the discrete whitelist with a continuous range: any `time.ParseDuration` value within **[1m, 24h]** is accepted. Closes the legacy validation gap described in #538 (independent of the still-open #537).

## How it works

- New exported helper `config.ValidatePollInterval(raw string) error` — parses with `time.ParseDuration` and bounds the result to `[minPollInterval=1m, maxPollInterval=24h]`. Empty string is allowed (callers apply the default).
- Both enforcement points now call it, so the rule is identical everywhere (TOML, GUI, TUI):
  - `daemon/internal/config/config.go` — `Validate()` (config-load path), replacing the `validIntervals` map.
  - `daemon/internal/server/handlers.go` — `PUT/PATCH /config` handler, replacing the local `validPollIntervals` allowlist.
- `1m` keeps the GitHub rate-limit floor; `24h` is a generous ceiling that still rejects pathological values (`1s`, `0`, negatives).
- Docs (`configuration-guide.md`, `docker/.env.example`, `project-spec.md`) updated to describe the range.

## Scope

**In scope:** relaxing the legacy `[github].poll_interval` validation in the two enforcement points + tests + docs.

**Out of scope:** PR #537's new `[polling]` section (`ResolvedPollInterval`/`min`/`max`) — that PR is still open and does not touch this legacy validation; the two are independent.

- **Runtime behavior change:** a config that previously crash-looped the daemon (e.g. `3m`) now loads and the daemon stays up. No new scheduler code path — the interval string was already consumed as-is. Users can now pick intervals near the 1m floor → more frequent GitHub API calls.
- **Rollout failure modes:** removes a crashloop, doesn't add one. No new secret/config/permission. Whitelisted values validate identically. Sub-minute (`30s`) / oversized (`48h`) still rejected. Minor widening: empty `poll_interval` on the HTTP path is now accepted (falls back to default). The `400` **error message changed** ("out of range / invalid …" vs "must be one of: 1m, 5m, 30m, 1h").
- **Blast radius:** single bundled/self-hosted daemon per user; config load + `PUT/PATCH /config`. No multi-replica concern.
- **What to watch:** daemon startup (`config load failed` ERRORs should disappear; listening on `:7842`); `PUT /config` 400s only for genuinely invalid values; GitHub `403` secondary-rate-limit if users pick aggressive intervals.
- **Rollback & sequencing:** fully revertible, no migration. Caveat: a config persisted with a newly-valid value (e.g. `3m`) would crash-loop again if this is reverted (manual edit back to a whitelisted value needed). No flag/ordering required.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/config/... ./internal/server/...`
- [x] `go test ./internal/config/... ./internal/server/...` — pass
- [x] Updated tests: `TestValidate_InvalidPollInterval` (now covers `30s`/`0`/`-5m`/`48h`/unparseable), `TestValidate_AllValidIntervals` (now includes `3m`/`10m`/`90m`/`12h`/`24h`), `store_test.go` bad-row case (`42m`→`48h`), `handlers_test.go` PUT table (valid `3m`, invalid `30s`/`48h`/`nonsense`)
- [ ] Reviewer: confirm bounds `[1m, 24h]` match intended operational policy
- [ ] Note: pre-existing, unrelated failure `internal/executor TestDetect_Fallback` (environment-dependent — a real `codex` binary on PATH defeats the fallback assertion; fails on clean `main` too)

Closes #538

### DIFF
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -14,8 +14,32 @@ import (
 	"github.com/heimdallm/daemon/internal/executor"
 )
 
-var validIntervals = map[string]bool{
-	"1m": true, "5m": true, "30m": true, "1h": true,
+// Poll interval bounds. Any time.ParseDuration value within this range is
+// accepted; the discrete whitelist {1m,5m,30m,1h} was removed in favour of a
+// continuous range so values like 3m or 10m are valid everywhere (TOML, GUI,
+// TUI). 1m keeps the GitHub rate-limit floor; 24h is a generous ceiling that
+// still rejects pathological values like 1s or 0.
+const (
+	minPollInterval = time.Minute
+	maxPollInterval = 24 * time.Hour
+)
+
+// ValidatePollInterval reports whether raw is an acceptable poll_interval: a
+// time.ParseDuration value within [minPollInterval, maxPollInterval]. An empty
+// string is allowed (callers apply the default). Exported so the HTTP config
+// handler enforces the same rule as config load.
+func ValidatePollInterval(raw string) error {
+	if raw == "" {
+		return nil
+	}
+	d, err := time.ParseDuration(raw)
+	if err != nil {
+		return fmt.Errorf("invalid poll_interval %q: %w", raw, err)
+	}
+	if d < minPollInterval || d > maxPollInterval {
+		return fmt.Errorf("poll_interval %q out of range (must be between %s and %s)", raw, minPollInterval, maxPollInterval)
+	}
+	return nil
 }
 
 // githubTopicPattern enforces GitHub's topic rules: lowercase letters, digits
@@ -1206,8 +1230,8 @@ func (c *Config) Validate() error {
 	if c.AI.Primary == "" {
 		return fmt.Errorf("config: ai.primary is required")
 	}
-	if c.GitHub.PollInterval != "" && !validIntervals[c.GitHub.PollInterval] {
-		return fmt.Errorf("config: invalid poll_interval %q (valid: 1m, 5m, 30m, 1h)", c.GitHub.PollInterval)
+	if err := ValidatePollInterval(c.GitHub.PollInterval); err != nil {
+		return fmt.Errorf("config: %w", err)
 	}
 	// Validate per-CLI agent configs: permission_mode and approval_mode must be in their allowlists.
 	for name, a := range c.AI.Agents {
--- a/daemon/internal/config/config_test.go
+++ b/daemon/internal/config/config_test.go
@@ -277,18 +277,28 @@ func TestValidate_ValidConfig(t *testing.T) {
 }
 
 func TestValidate_InvalidPollInterval(t *testing.T) {
-	cfg := &Config{}
-	cfg.applyDefaults()
-	cfg.AI.Primary = "claude"
-	cfg.GitHub.PollInterval = "2m"
+	for _, interval := range []string{
+		"nonsense", // unparseable
+		"30s",      // below the 1m floor
+		"0",        // zero
+		"-5m",      // negative
+		"48h",      // above the 24h ceiling
+	} {
+		cfg := &Config{}
+		cfg.applyDefaults()
+		cfg.AI.Primary = "claude"
+		cfg.GitHub.PollInterval = interval
 
-	if err := cfg.Validate(); err == nil {
-		t.Error("Validate() = nil, want error for invalid poll_interval")
+		if err := cfg.Validate(); err == nil {
+			t.Errorf("Validate() with interval %q = nil, want error", interval)
+		}
 	}
 }
 
 func TestValidate_AllValidIntervals(t *testing.T) {
-	for _, interval := range []string{"1m", "5m", "30m", "1h"} {
+	// Any time.ParseDuration value within [1m, 24h] is accepted, including
+	// arbitrary values like 3m that the old discrete whitelist rejected.
+	for _, interval := range []string{"1m", "3m", "5m", "10m", "30m", "90m", "1h", "12h", "24h"} {
 		cfg := &Config{AI: AIConfig{Primary: "claude"}, GitHub: GitHubConfig{PollInterval: interval}}
 		if err := cfg.Validate(); err != nil {
 			t.Errorf("Validate() with interval %q = %v", interval, err)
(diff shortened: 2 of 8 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

El PR sustituye la whitelist fija de poll_interval {1m,5m,30m,1h} por un rango continuo [1m,24h] validado con time.ParseDuration, exponiendo una única función ValidatePollInterval reutilizada tanto en config load como en el handler HTTP. Es un cambio limpio, no-breaking (relajación: los valores antiguos siguen siendo válidos), con tests y docs actualizados de forma coherente. LGTM.

### Suggestions

- config.go:ValidatePollInterval — el mensaje de error usa time.Duration.String() sobre las constantes, por lo que se renderiza como 'must be between 1m0s and 24h0s' en vez de '1m and 24h'. No es un bug, pero queda menos limpio de cara al usuario; opcional formatearlo manualmente.
- La descripción afirma que los valores arbitrarios son válidos 'everywhere (TOML, GUI, TUI)', pero el diff solo toca config, handlers HTTP y docs. Conviene verificar que la GUI/TUI no tengan su propio dropdown/whitelist hardcodeada que siga limitando la entrada a los 4 valores antiguos; si lo tienen, quedaría inconsistente con el backend.
- La validación es consistente entre Validate() y handlePutConfig al compartir ValidatePollInterval, y los tests cubren bien los bordes (unparseable, 30s, 0, -5m, 48h, y arbitrarios como 3m/90m). Buen trabajo.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-scoped refactor that replaces the discrete poll_interval whitelist {1m,5m,30m,1h} with a continuous [1m,24h] range validated via time.ParseDuration. Validation logic is correctly centralized in the exported ValidatePollInterval so both config load (Validate) and the HTTP handler (handlePutConfig) enforce the same rule, eliminating prior duplication. Edge cases (empty, unparseable, zero, negative, below floor, above ceiling) are handled and covered by tests. Docs, .env example, and project-spec are all kept in sync. No correctness, security, or breaking-change issues found.

### Suggestions

- ValidatePollInterval echoes the raw user-supplied value back in the error returned to the HTTP client (handlers.go via err.Error()). This is harmless for a plaintext http.Error response, but if these errors are ever surfaced in an HTML/JSON UI context, consider not reflecting raw input verbatim.
- time.ParseDuration accepts forms like "60s" (== 1m) and "0.5h" — these are intentionally valid under the new range, but a one-line note in the docs that sub-minute-unit spellings normalize (e.g. 60s is accepted as the 1m floor) could pre-empt user confusion.
- Consider confirming the GUI/TUI selectors no longer hardcode the old {1m,5m,30m,1h} dropdown so they actually expose the wider range the comment claims is now valid 'everywhere (TOML, GUI, TUI)'. Not visible in this diff.

---

vbuenog:
Follow-up on the review suggestions (this PR merged before I addressed them, so they're handled separately):

**Suggestion (ivanmunozruiz) — error rendered `1m0s`/`24h0s` via `Duration.String()`:** fixed in #542 — now uses display constants so the message reads `must be between 1m and 24h`.

**Suggestion (both) — verify GUI/TUI don't hardcode the old whitelist:** verified — you were right. The CLI TUI only *displays* `poll_interval` (read-only, fine), but the Flutter GUI dropdown (`config_screen.dart:379-384`) still hardcodes `{1m,5m,30m,1h}`, so the GUI can't pick `3m`. The "valid everywhere (GUI, TUI)" wording was an overclaim — softened the in-code comment in #542 and opened **#543** to make the GUI a validated free-form input mirroring the daemon bounds.

**Suggestion (Muriano) — `err.Error()` reflects raw user input back to the HTTP client:** keeping as-is, with reasoning: the response is a plaintext `http.Error` (no HTML/JSON context, so no injection surface), and echoing the rejected value is a genuinely useful diagnostic for operators editing config. Happy to switch to a fixed message if you feel strongly, but I'd lean toward keeping the value in the error.

**Suggestion (Muriano) — docs note that `60s`/`0.5h` normalize:** skipping for now to avoid doc clutter — the `[1m, 24h]` range is already documented, and `time.ParseDuration` normalization is standard Go behavior. Can add a one-liner if it proves confusing in practice.

Thanks for the thorough pass.

PR #541, opened 2026-06-18
fix(discovery): keep config consistent when promoting explicit [ai.repos.*] repos (closes #527)

### LINKED ISSUE #527: Discovery: keep config state consistent when promoting explicit [ai.repos.*] repos (follow-up to #485)
Follow-up from review of PR #485. These are valid but out-of-scope for the #281 bug fix (which only needed configured repos to land in the merged output):

1. **`upsertDiscoveredRepos` double-listing** — a repo with explicit `[ai.repos.*]` config that is also present in `NonMonitored` from a prior tick ends up in both `c.GitHub.Repositories` and `NonMonitored`. `MergeRepos` handles this correctly via the exempt set, but code reading `NonMonitored` directly could behave inconsistently. Consider stripping the repo from `NonMonitored` when promoting.
2. **`hasExplicitAIConfig` placed after the `discovery_orgs` org filter** — an explicitly-configured repo whose org is outside `discovery_orgs` is skipped before the new logic fires. Mitigated because `MergeRepos` includes configured repos directly, but the repo won't land in `Repositories` via upsert.
3. **`Tier1Config.ConfiguredRepos` captured once at startup** — if config is ever hot-reloaded, this snapshot goes stale. Should become a getter/closure, symmetric with how `StaticRepos`/`NonMonitored` are read live.
4. **Additional test** — cover a repo present in static, configured AND nonMonitored simultaneously to lock in exempt-wins semantics across all three sources.

Source: https://github.com/theburrowhub/heimdallm/pull/485

vbuenog:
## Summary

Follow-up to the review of PR #485 (#527). When a repository is wired up via `[ai.repos.*]`, that is explicit operator intent and should win over the auto-discovery guards. `upsertDiscoveredRepos` previously left such a repo in an inconsistent state in two situations, and the merge layer (`MergeRepos`) only papered over it at read time. This makes the persisted config itself consistent.

## How it works

`daemon/cmd/heimdallm/main.go` — `upsertDiscoveredRepos` now handles explicitly-configured repos up front, before the "already known" and `discovery_orgs` checks:

- **Item 1 — promote out of `NonMonitored`.** A repo blacklisted on a prior tick that now has explicit config is stripped from `NonMonitored` and appended to `Repositories`, instead of being short-circuited as "already known" and left in `NonMonitored` (where code reading the list directly would treat it as blacklisted, even though `MergeRepos` exempts it). The promotion is reported in `added` so `processDiscoveredRepos` actually persists the updated lists — it early-returns on an empty `added` set, so this is what makes the strip durable across reloads. New helper `removeRepo` (order-preserving, fresh allocation via a 3-index slice).
- **Item 2 — bypass the org filter.** The explicit-config check now precedes the `discovery_orgs` allowlist, so a configured repo whose org is outside `discovery_orgs` still lands in `Repositories` via upsert. Non-configured repos in out-of-list orgs are still filtered.

The split of the old single `known` set into `inRepositories` / `inNonMonitored` is what lets the promotion path tell "already monitored" from "blacklisted".

## Scope

**In scope:** items 1, 2, 4 from #527.

**Out of scope / already done:** item 3 (`Tier1Config.ConfiguredRepos` staleness on hot-reload) — verified already resolved within #485: `tier1ConfigFn` (`main.go:688-702`) is a closure that recomputes `ConfiguredRepos: aiRepoKeys(cfg)` live under `cfgMu`, symmetric with `StaticRepos`/`NonMonitored`. No code change needed.

- **Runtime behavior change:** pure in-memory list reconciliation on the existing poll path — no new GitHub calls. Only repos with an explicit `[ai.repos.*]` entry take a different path; everything else is byte-for-byte the old behavior. A configured repo previously stuck unmonitored-in-`NonMonitored` (rescued only by `MergeRepos` at read time) now gets monitored *and* the persisted config matches.
- **Rollout failure modes:** no new startup invariant/secret/permission — cannot crashloop. A promoted repo emits one `repo_discovered` SSE and has its review deferred one tick (benign; UI keys by repo, `FirstSeen.Mark` is idempotent so the NEW badge is preserved).
- **Blast radius:** daemon discovery/poll path; only operators combining `[ai.repos.*]` with `auto_enable=false` and/or `discovery_orgs`. Single self-hosted daemon.
- **What to watch:** (1) a previously-misconfigured repo appears in `Repositories` and leaves `non_monitored` in GET `/config` / `heimdallm-cli status`; (2) no flip-flop churn in the persisted `non_monitored` row across ticks (the move should be one-time); (3) at most one extra `repo_discovered` SSE per promoted repo, not per tick.
- **Rollback & sequencing:** fully revertible, no migration. After revert, an already-promoted repo stays in `repositories` (harmless). No ordering/flag needed.

## Evidence

<details><summary>New tests pass</summary>

```
--- PASS: TestUpsertDiscoveredRepos_PromotesExplicitConfigOutOfNonMonitored (0.00s)
--- PASS: TestUpsertDiscoveredRepos_ExplicitConfigBypassesOrgFilter (0.00s)
--- PASS: TestMergeRepos_ConfiguredExemptWinsAcrossAllSources (0.00s)
```
</details>

## Test plan

- [x] `go build ./...`
- [x] `go vet ./cmd/... ./internal/discovery/...`
- [x] `go test ./cmd/heimdallm/... ./internal/discovery/...` — pass
- [x] Item 4 test `TestMergeRepos_ConfiguredExemptWinsAcrossAllSources` (static+configured+nonMonitored)
- [x] Item 1 test `TestUpsertDiscoveredRepos_PromotesExplicitConfigOutOfNonMonitored`
- [x] Item 2 test `TestUpsertDiscoveredRepos_ExplicitConfigBypassesOrgFilter`
- [ ] Reviewer: confirm promoted repos *should* surface a `repo_discovered` SSE (vs a dedicated state-change event)
- [ ] Note: pre-existing, unrelated failure `internal/executor TestDetect_Fallback` (environment-dependent — a real `codex` binary on PATH defeats the fallback assertion; fails on clean `main` too)

Closes #527

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -2601,12 +2601,13 @@ func aiRepoKeys(c *config.Config) []string {
 // first-seen timestamps. This helper is pure state mutation so it's easy
 // to test in isolation.
 func upsertDiscoveredRepos(c *config.Config, prs []*gh.PullRequest) []string {
-	known := make(map[string]struct{})
+	inRepositories := make(map[string]struct{}, len(c.GitHub.Repositories))
 	for _, r := range c.GitHub.Repositories {
-		known[r] = struct{}{}
+		inRepositories[r] = struct{}{}
 	}
+	inNonMonitored := make(map[string]struct{}, len(c.GitHub.NonMonitored))
 	for _, r := range c.GitHub.NonMonitored {
-		known[r] = struct{}{}
+		inNonMonitored[r] = struct{}{}
 	}
 
 	// Build an org allowlist from DiscoveryOrgs. When set, repos whose org
@@ -2624,7 +2625,43 @@ func upsertDiscoveredRepos(c *config.Config, prs []*gh.PullRequest) []string {
 		if pr.Repo == "" {
 			continue
 		}
-		if _, alreadyKnown := known[pr.Repo]; alreadyKnown {
+
+		// Repos with an explicit [ai.repos.*] entry are opted in by the
+		// operator. Explicit config is intent and must win over both guards:
+		//   - It bypasses the discovery_orgs allowlist (theburrowhub/heimdallm#527
+		//     item 2): a configured repo outside the discovery orgs must still
+		//     be monitored, so this check precedes the org filter below.
+		//   - It promotes the repo out of NonMonitored when a stale row from a
+		//     prior tick blacklisted it (#527 item 1). Leaving it in
+		//     NonMonitored keeps config state inconsistent for code that reads
+		//     the list directly, even though MergeRepos already exempts it. So
+		//     we strip the repo from NonMonitored and add it to Repositories.
+		// See also theburrowhub/heimdallm#281 for why a configured repo must
+		// never linger in NonMonitored (MergeRepos would otherwise blacklist it
+		// and silently stop issue polling).
+		if _, hasExplicitAIConfig := c.AI.Repos[pr.Repo]; hasExplicitAIConfig {
+			_, alreadyMonitored := inRepositories[pr.Repo]
+			if _, blacklisted := inNonMonitored[pr.Repo]; blacklisted {
+				c.GitHub.NonMonitored = removeRepo(c.GitHub.NonMonitored, pr.Repo)
+				delete(inNonMonitored, pr.Repo)
+			}
+			if !alreadyMonitored {
+				c.GitHub.Repositories = append(c.GitHub.Repositories, pr.Repo)
+				inRepositories[pr.Repo] = struct{}{}
+				// Reported as added so processDiscoveredRepos persists the
+				// updated lists (it early-returns on an empty added set) — this
+				// is what makes the NonMonitored strip durable across reloads.
+				added = append(added, pr.Repo)
+			}
+			continue
+		}
+
+		// Auto-discovered repos (no explicit config): skip when already tracked
+		// in either list.
+		if _, ok := inRepositories[pr.Repo]; ok {
+			continue
+		}
+		if _, ok := inNonMonitored[pr.Repo]; ok {
 			continue
 		}
 		// Filter by allowed orgs when configured.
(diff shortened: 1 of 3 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

El PR refactoriza `upsertDiscoveredRepos` para separar el set combinado `known` en `inRepositories`/`inNonMonitored` y promover repos con config `[ai.repos.*]` explícita: los saca de NonMonitored, los añade a Repositories y los reporta en `added` para persistencia, además de bypasear el filtro `discovery_orgs`. El cambio es correcto, está bien testeado (3 tests nuevos que ejercitan los casos del #527) y usa un three-index slice en `removeRepo` para evitar mutar el array subyacente. LGTM con una observación menor de edge-case.

### Issues

🟡 **daemon/cmd/heimdallm/main.go:2648** — En el branch de config explícita, si un repo está SIMULTÁNEAMENTE en Repositories y en NonMonitored (estado inconsistente de un tick previo, que el propio PR reconoce como posible y que el test MergeRepos_ConfiguredExemptWinsAcrossAllSources asume alcanzable), el strip de NonMonitored se ejecuta pero `alreadyMonitored` es true, por lo que el repo NO se añade a `added`. Si ese repo fuese el único cambio del tick, `added` queda vacío y `processDiscoveredRepos` hace early-return sin persistir el strip — dejando la inconsistencia que el PR intenta corregir. Sería más robusto añadir a `added` también cuando solo se hace el strip (blacklisted && alreadyMonitored). Severidad baja por ser un estado de partida poco probable.

### Suggestions

- Considerar un test que cubra el caso de un repo presente en Repositories y NonMonitored a la vez con config explícita, para garantizar que el strip se reporta en `added` y se persiste.
- El comentario in-code es excelente y documenta bien la interacción con #281/#527; mantener ese nivel de detalle.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Solid, well-documented root-cause fix that splits the single `known` set into separate Repositories/NonMonitored maps and promotes explicitly-configured repos out of NonMonitored, bypassing the org filter. Logic is correct for the common paths, removeRepo uses a safe three-index slice to avoid backing-array aliasing, and tests cover the main scenarios (#527 items 1, 2, 4). One real edge-case gap: when a configured repo is simultaneously in Repositories and NonMonitored, the NonMonitored strip is not reported in `added` and therefore may not be persisted.

### Issues

⚠️ **daemon/cmd/heimdallm/main.go:2643** — In the explicit-AI-config branch, when a repo is already in Repositories (alreadyMonitored=true) AND also present in NonMonitored (blacklisted=true) — the exact inconsistent state this PR targets — the code strips it from NonMonitored via removeRepo but only appends to `added` inside the `if !alreadyMonitored` block. Since the strip is gated on `blacklisted` but the `added` report is gated on `!alreadyMonitored`, this sub-case mutates NonMonitored without recording anything in `added`. Per the PR's own comment, processDiscoveredRepos early-returns on an empty `added` set, so if no other repo is added in the same tick the NonMonitored strip is not persisted and the inconsistency survives across reloads. Impact: config stays inconsistent for the precise case the PR claims to fix. Recommended fix: report the repo in `added` whenever NonMonitored was actually mutated, e.g. set an `addedAny`/`changed` flag when removeRepo strips the entry and append to `added` (or otherwise force persistence) even when alreadyMonitored is true.

### Suggestions

- Add a unit test for the 'repo present in both Repositories and NonMonitored with explicit config' edge case, asserting the NonMonitored strip is persisted (reported in `added`).
- Emit a log line when a repo is promoted out of NonMonitored / added to Repositories so operators can debug discovery-state transitions in production; the silent mutation is hard to trace otherwise.
- Consider deduplicating against repeated PRs for the same repo more explicitly; current map updates handle it correctly, but a short comment noting that within-tick duplicates are absorbed by the inRepositories map would aid maintainability.

---

vbuenog:
Thanks for the review — addressed in `e3a0946`.

**Issue (both reviewers) — strip not persisted when `alreadyMonitored && blacklisted`:** confirmed real and fixed. The repo is now reported in `added` whenever **either** list is mutated (stripped from `NonMonitored` *or* added to `Repositories`), not only on the add-to-`Repositories` path — so a strip-only promotion still triggers `processDiscoveredRepos` persistence and survives reloads. Good catch; this was exactly the inconsistency the PR set out to fix.

**Suggestion — test for the "in both lists" edge case:** added `TestUpsertDiscoveredRepos_PromotionPersistsWhenAlreadyMonitored`, asserting the strip-only promotion is reported in `added` and the repo ends up in `Repositories` exactly once.

**Suggestion — log on promotion:** added `slog.Info("upsertDiscoveredRepos: promoted explicitly-configured repo", ...)` with `added_to_repositories` / `stripped_from_non_monitored` flags for production traceability.

**Suggestion — within-tick duplicate absorption:** added a one-line comment noting that after the first PR promotes a repo, `inRepositories`/`inNonMonitored` reflect the new state so a duplicate PR for the same repo is a no-op.

All daemon tests green (except the pre-existing, environment-dependent `internal/executor TestDetect_Fallback`, unrelated to this change).

PR #558, opened 2026-06-19
fix(daemon): recover from panics in the SSE→NATS bridge goroutine (closes #548)

### LINKED ISSUE #548: fix(daemon): SSE→NATS bridge goroutine has no panic recovery (panic crashes daemon) [MEDIUM]
## Summary (MEDIUM)
The SSE→NATS bridge runs in a bare goroutine with no `recover()`. A panic inside the loop crashes the entire daemon (Go does not recover panics in arbitrary goroutines).

## Evidence
`daemon/cmd/heimdallm/main.go:177-186`
```go
bridgeCh := broker.Subscribe()
if bridgeCh != nil {
    go func() {
        for event := range bridgeCh {
            subj := "heimdallm.events." + event.Type
            if err := eventBus.Conn().Publish(subj, []byte(event.Data)); err != nil {
                slog.Warn("sse-bridge: publish to NATS failed", "type", event.Type, "err", err)
            }
        }
    }()
}
```
(Note: the "Publish blocks indefinitely" concern was *refuted* — the embedded NATS server is non-blocking. The real issue is the missing panic recovery.)

## Suggested fix
Wrap the loop body in `defer func(){ if r:=recover(); r!=nil { slog.Error("sse-bridge panic", "panic", r, "stack", string(debug.Stack())) } }()`.

---

vbuenog:
## Summary

The SSE→NATS bridge ran in a bare goroutine with no `recover()`. A panic while publishing an event would crash the **entire daemon** — Go does not recover panics in arbitrary goroutines (#548). No reproducible panic exists in the current loop body (`sse.Event` fields are strings; a nil `*nats.Conn` returns an error, not a panic), so this is defensive hardening at an architectural boundary — but the blast radius of any future/library panic there is the whole process, which is why it's worth fixing.

## How it works

`daemon/cmd/heimdallm/main.go`:
- Extracted the inline bridge loop into a named, testable helper `publishBridgeEvents(events <-chan sse.Event, publish func(subject string, data []byte) error)`.
- Each event's publish runs inside a per-iteration `recover()` that logs the panic + stack (`runtime/debug.Stack()`) at Error and **continues to the next event**. So a single bad event can neither crash the daemon nor permanently kill the bridge that feeds every SSE client.
- Call site becomes `go publishBridgeEvents(bridgeCh, eventBus.Conn().Publish)`.

Design choice: **per-iteration recover-and-continue** (not goroutine-level recover-and-exit), so one panic doesn't silently stop live updates for all SSE consumers.

## Scope

**In scope:** panic recovery for the SSE→NATS bridge goroutine + a testable extraction.
**Out of scope:** the NATS worker goroutines (review/triage/publish/etc.) also lack `recover()` — a broader, separate concern; this PR is scoped to the bridge per #548. Not adding a shared recover helper (no existing convention in the repo; would be speculative).

- **Runtime behavior change:** none in the normal path — same `Publish` call per event, same warn-on-error. The only difference: a panic during a single event publish is now caught, logged with a stack, and skipped instead of crashing the process. (`eventBus.Conn().Publish` is bound once as a method value at goroutine start vs called each iteration — equivalent, the conn is stable for the daemon's life.)
- **Rollout failure modes:** none — pure in-process wiring change, no config/secret/permission/API/migration; can't crashloop; compatible with old/new pods.
- **Blast radius:** the SSE→NATS bridge only (feeds the Flutter GUI events tab + CLI `follow`). No behavioral change in normal operation; matters only in the panic case, where it converts a daemon-wide crash into logged-and-continue.
- **What to watch:** the new Error log line `sse-bridge: recovered from panic; continuing` (carries `type`, `panic`, `stack`). If it ever fires, a real panic occurred in the publish path that previously would have crashed the daemon — investigate the value/stack. Conversely, daemon restarts correlated with SSE activity should no longer occur from this cause.
- **Rollback & sequencing:** fully revertible, single file, no flag/migration/ordering. Reverting restores the bare goroutine (and the crash exposure).

## Evidence

Acceptance: a panic in the bridge must not crash the daemon (and shouldn't kill the bridge).

<details><summary><code>make test-docker</code> — new bridge tests pass</summary>

```
=== RUN   TestPublishBridgeEvents_RecoversFromPanicAndContinues
--- PASS: TestPublishBridgeEvents_RecoversFromPanicAndContinues (0.00s)
=== RUN   TestPublishBridgeEvents_ContinuesAfterPublishError
--- PASS: TestPublishBridgeEvents_ContinuesAfterPublishError (0.00s)
ok  	github.com/heimdallm/daemon/cmd/heimdallm
```
</details>

<details><summary>Full daemon suite — <code>go vet ./...</code> + <code>go test ./...</code> all green</summary>

```
ok  github.com/heimdallm/daemon/cmd/heimdallm
ok  github.com/heimdallm/daemon/internal/bus
ok  github.com/heimdallm/daemon/internal/pipeline
ok  github.com/heimdallm/daemon/internal/server
ok  github.com/heimdallm/daemon/internal/sse
... (all packages ok; vet clean)
```
</details>

## Test plan

- [x] `make test-docker` scoped to `./cmd/...` — both new tests pass
- [x] `make test-docker` full daemon suite — `go vet` + `go test ./...` all green
- [ ] Reviewer: confirm recover-and-continue (vs recover-and-exit) is the desired behavior for the bridge

Closes #548

### DIFF
--- a/daemon/cmd/heimdallm/bridge_test.go
+++ b/daemon/cmd/heimdallm/bridge_test.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"fmt"
+	"slices"
 	"testing"
 
+	"github.com/heimdallm/daemon/internal/bus"
 	"github.com/heimdallm/daemon/internal/sse"
 )
 
@@ -19,7 +21,7 @@ func TestPublishBridgeEvents_RecoversFromPanicAndContinues(t *testing.T) {
 
 	var got []string
 	publish := func(subject string, data []byte) error {
-		if subject == "heimdallm.events.boom" {
+		if subject == bus.SubjEventPrefix+"boom" {
 			panic("simulated nats publish panic")
 		}
 		got = append(got, subject)
@@ -29,9 +31,34 @@ func TestPublishBridgeEvents_RecoversFromPanicAndContinues(t *testing.T) {
 	// Returns (does not propagate the panic) once the channel is drained.
 	publishBridgeEvents(events, publish)
 
-	want := []string{"heimdallm.events.a", "heimdallm.events.c"}
-	if len(got) != len(want) || got[0] != want[0] || got[1] != want[1] {
+	want := []string{bus.SubjEventPrefix + "a", bus.SubjEventPrefix + "c"}
+	if !slices.Equal(got, want) {
 		t.Fatalf("expected bridge to process %v after recovering, got %v", want, got)
+	}
+}
+
+// recover() must also catch a non-string / runtime panic (here, assignment to
+// a nil map), not just explicit panic("string") calls.
+func TestPublishBridgeEvents_RecoversFromRuntimePanic(t *testing.T) {
+	events := make(chan sse.Event, 2)
+	events <- sse.Event{Type: "nilmap", Data: "1"}
+	events <- sse.Event{Type: "ok", Data: "2"}
+	close(events)
+
+	var got []string
+	publish := func(subject string, data []byte) error {
+		if subject == bus.SubjEventPrefix+"nilmap" {
+			var m map[string]int
+			m["boom"] = 1 // runtime panic: assignment to entry in nil map
+		}
+		got = append(got, subject)
+		return nil
+	}
+
+	publishBridgeEvents(events, publish)
+
+	if want := []string{bus.SubjEventPrefix + "ok"}; !slices.Equal(got, want) {
+		t.Fatalf("expected bridge to recover from a runtime panic and continue; want %v, got %v", want, got)
 	}
 }
 
--- a/daemon/cmd/heimdallm/bridge_test.go
+++ b/daemon/cmd/heimdallm/bridge_test.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/heimdallm/daemon/internal/sse"
+)
+
+// A panic while publishing one event must not propagate (which, in the real
+// bare goroutine, would crash the whole daemon) and must not stop the bridge:
+// the events before and after the panicking one are still published.
+func TestPublishBridgeEvents_RecoversFromPanicAndContinues(t *testing.T) {
+	events := make(chan sse.Event, 3)
+	events <- sse.Event{Type: "a", Data: "1"}
+	events <- sse.Event{Type: "boom", Data: "2"}
+	events <- sse.Event{Type: "c", Data: "3"}
+	close(events)
+
+	var got []string
+	publish := func(subject string, data []byte) error {
+		if subject == "heimdallm.events.boom" {
+			panic("simulated nats publish panic")
+		}
+		got = append(got, subject)
+		return nil
+	}
+
+	// Returns (does not propagate the panic) once the channel is drained.
+	publishBridgeEvents(events, publish)
+
+	want := []string{"heimdallm.events.a", "heimdallm.events.c"}
+	if len(got) != len(want) || got[0] != want[0] || got[1] != want[1] {
+		t.Fatalf("expected bridge to process %v after recovering, got %v", want, got)
+	}
+}
+
+// A publish error is logged but must not stop the bridge — every event is
+// still attempted.
+func TestPublishBridgeEvents_ContinuesAfterPublishError(t *testing.T) {
+	events := make(chan sse.Event, 2)
+	events <- sse.Event{Type: "x", Data: "1"}
+	events <- sse.Event{Type: "y", Data: "2"}
+	close(events)
+
+	attempts := 0
+	publish := func(subject string, data []byte) error {
+		attempts++
+		return fmt.Errorf("publish failed")
+	}
+
+	publishBridgeEvents(events, publish)
+
+	if attempts != 2 {
+		t.Fatalf("expected both events attempted despite errors, got %d", attempts)
+	}
+}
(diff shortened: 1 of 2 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured fix that extracts the SSE→NATS bridge loop into a named function with per-event panic recovery. The pattern is correct—each iteration gets its own deferred recover via an immediately-invoked anonymous function—and the two tests cover both the panic-recovery and error-continuation paths. No correctness or security issues.

### Issues

🟡 **daemon/cmd/heimdallm/bridge_test.go:33** — The manual element-wise comparison (`len(got) != len(want) || got[0] != want[0] || got[1] != want[1]`) is fragile and won't scale. Using `slices.Equal` (or `reflect.DeepEqual`) would be clearer and safer if `want` ever changes length.

### Suggestions

- Consider using `slices.Equal(got, want)` (Go 1.21+) instead of the manual index-by-index comparison in TestPublishBridgeEvents_RecoversFromPanicAndContinues for readability and maintainability.
- Minor: a brief comment on the call site (`go publishBridgeEvents(...)`) noting that the function never panics outward (by design) could help future readers understand why it's safe to run in a bare goroutine.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Solid, focused fix: the bare bridge goroutine is replaced with an extracted, testable `publishBridgeEvents` function that recovers per-event so a single panic neither crashes the daemon nor kills the bridge. The publish dependency is injected, enabling the two well-targeted tests (panic recovery + continue-on-error). Behavior is preserved (same subject prefix, same warning log on publish error). No correctness, security, or concurrency issues found.

### Suggestions

- Consider emitting a metric/counter for recovered panics and publish failures (e.g. a dropped-events counter) so operators can alert on a noisy bad-event stream rather than relying solely on log scraping.
- The subject prefix "heimdallm.events." is duplicated as a string literal both in production code and in the tests; extracting it to a shared const would prevent the test and code from drifting if the prefix ever changes.
- event.Type is interpolated directly into the NATS subject; if Type can ever contain spaces/wildcards/dots from an untrusted publisher, sanitizing or validating it before building the subject would avoid malformed or unintended subjects (not exploitable as shown, but worth confirming Type is constrained upstream).
- Optional: a test asserting the recover() path also continues when the panic value is a non-string (e.g. an error or runtime panic) would broaden coverage slightly, though the current test adequately exercises the recovery loop.

---

vbuenog:
Thanks both 🙏 — addressed in `611f680` (full `make test-docker`: `go vet` + `go test ./...` all green). Per-item disposition:

### ✅ Implemented

- **Shared subject-prefix const** (@Muriano): there's already `bus.SubjEventPrefix = "heimdallm.events."` (`internal/bus/subjects.go:25`) — now used in both `publishBridgeEvents` and the test, so the prefix can't drift.
- **`slices.Equal` in the test** (@sergiotejon): replaced the index-by-index comparison.
- **Call-site comment** (@sergiotejon): noted that `publishBridgeEvents` recovers per event and never panics outward, so the bare goroutine is safe.
- **Non-string / runtime panic test** (@Muriano): added `TestPublishBridgeEvents_RecoversFromRuntimePanic` — a nil-map assignment (a real `runtime.Error`, not an explicit `panic("string")`) is recovered and the bridge continues.

### ↩️ Declined (with evidence)

- **Sanitize `event.Type` before building the NATS subject** (@Muriano): confirmed not needed. `sse.Event.Type` is constrained to a fixed set of internal constants (`internal/sse/broker.go:7-18` — `review_completed`, `pr_detected`, …) set by the daemon's own publishers; untrusted data (PR/issue titles, bodies) flows through `Data`, never `Type`. No untrusted value can reach the subject, so there's no malformed/wildcard-subject vector. (You flagged it as "worth confirming Type is constrained upstream" — confirmed.)

### 📋 Deferred to follow-up

- **Metrics/counter for recovered panics + dropped events** (@Muriano): agreed it's better than log-scraping, but the daemon has no metrics surface wired yet, so it's a broader observability task — filed #559.

Re-requesting review on the new commit.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previous findings have been addressed. The slices.Equal replacement, bus.SubjEventPrefix usage, call-site comment, and runtime-panic test are all clean. The per-event recover pattern is idiomatic Go, the test coverage is thorough, and the deferred metrics work is tracked in #559. No new issues.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #558. The author addressed all prior feedback in 611f680: the shared bus.SubjEventPrefix const is now used in both publishBridgeEvents and the tests (no drift), slices.Equal replaces index-by-index comparison, a call-site comment documents the bare-goroutine safety, and a new TestPublishBridgeEvents_RecoversFromRuntimePanic exercises a real runtime.Error (nil-map assignment). The declined event.Type sanitization is confirmed correct — internal/sse/broker.go:6-30 shows Type is a fixed set of daemon-controlled constants, not untrusted input. The metrics ask is reasonably deferred to #559. The per-event recover() pattern correctly prevents a panic in the bare goroutine from crashing the daemon while keeping the bridge alive. No confirmed issues remain.

### Suggestions

- When #559 wires a metrics surface, emit a counter for recovered panics and one for publish failures so these failure modes are alertable rather than log-scrape-only.
- Consider a test asserting the recover path logs at error level (e.g. via a slog test handler) so the observability contract itself is covered, not just the continue-after-panic behavior.

---

Issue #559, opened 2026-06-19
chore(daemon): metrics/counters for recovered panics and dropped/failed SSE bridge events


vbuenog:
## Context
Follow-up from review of #558 (fixes #548).

#558 added per-event panic recovery to the SSE→NATS bridge (`publishBridgeEvents`), logging recovered panics and publish failures at Error/Warn. Currently the only operator signal is log scraping.

## Proposal
Expose counters so operators can alert on a noisy bad-event stream instead of grepping logs:
- recovered-panics counter (bridge)
- publish-failure / dropped-events counter (bridge), and ideally the SSE broker's silent slow-consumer drops too

Blocked on the daemon having a metrics surface — there is no Prometheus/metrics endpoint wired today, so this is a broader observability task (introduce a metrics registry/endpoint, then add these counters). Low priority.

Raised by Muriano in the AI review of #558.

vbuenog:
Add another counter to this umbrella (raised in the #563 review by Muriano): **orphaned-corrupt-review** count — incremented when `PublishPending` orphans a review whose stored `Issues` JSON is corrupt (`pipeline.go`, "skipping pending review with corrupt issues JSON"). Lets operators detect a spike in corrupt rows instead of log-scraping. Same prerequisite: a metrics surface must exist first.

vbuenog:
Closing as **won't do** after re-assessing cost/value.

All four proposed counters instrument defensive paths that should be zero forever in a healthy daemon:

- **Bridge recovered panics** — safety net added in #558 after #548; no recurrence expected.
- **Bridge NATS publish failures** — NATS runs embedded in-process, so publish failures essentially only occur during shutdown.
- **SSE broker drops** — slow web-UI consumers; a refresh recovers, low severity.
- **Orphaned corrupt-review rows** (from the #563 review) — would indicate SQLite corruption, which surfaces in louder ways.

Verified against current `main` (4b6704e): no metrics surface exists (no prometheus/expvar/`/metrics` route), so shipping these counters means introducing a whole metrics registry + endpoint first. That enabler isn't justified by four flat-zero canary counters alone, and this deployment isn't scraped by a Prometheus stack, so there is no alerting pipeline to consume them — the existing Error/Warn logs remain the operator signal.

If a metrics surface is ever introduced for higher-value metrics (reviews processed, GitHub API usage, publish latency), these canary counters are cheap to add on top — this issue can serve as the reference list at that point.

PR #563, opened 2026-06-19
fix(pipeline): orphan pending reviews with corrupt issues JSON instead of posting empty (closes #549)

### LINKED ISSUE #549: fix(pipeline): PublishPending discards Issues JSON unmarshal error, republishes review with zero issues [MEDIUM]
## Summary (MEDIUM)
On retry/publish, the stored `rev.Issues` JSON is unmarshalled with the error discarded. If the stored JSON is malformed, `issues` stays empty and the review is **reposted to GitHub with no issues** — the developer silently loses all the findings.

## Evidence
`daemon/internal/pipeline/pipeline.go:748`
```go
json.Unmarshal([]byte(rev.Issues), &issues)   // error discarded
result := &executor.ReviewResult{ Summary: rev.Summary, Issues: issues, Severity: rev.Severity }
// ... BuildGitHubBody omits the Issues section when len(issues)==0, then SubmitReview posts it
```
No independent backstop; observable wrong outcome.

## Suggested fix
Check the unmarshal error; on failure log at Warn and skip the publish (or fail-closed) rather than posting an empty-issues review.

---

vbuenog:
## Summary

`PublishPending` (the retry path for reviews that failed their initial GitHub publish) decoded the stored `rev.Issues` JSON with the error discarded. On a corrupt row, `issues` stayed empty and the retry posted a review to GitHub **with its findings dropped** — then `MarkReviewPublished` stamped it done, so it never retried: silent data loss (#549). The daemon always writes well-formed JSON here (`json.Marshal` at insert), so this only triggers on DB corruption / a `Review.Issues` format drift — rare, hence MEDIUM.

## How it works

`daemon/internal/pipeline/pipeline.go` — check the `json.Unmarshal` error; on failure, log a Warn and **orphan the row** via `MarkReviewPublished(rev.ID, -1, "", now)` + `continue`, mirroring the no-repo orphan case 8 lines above and the #245 fail-closed posture (transient errors keep retrying; deterministic/permanent ones orphan to avoid an infinite retry loop). A corrupt review can't be faithfully reconstructed, so we neither post a misleading findings-less review nor loop forever.

## Scope

**In scope:** the corrupt-`Issues`-JSON branch of `PublishPending`.
**Out of scope:** the initial `Run` publish path (it builds the result in-memory from a freshly-decoded agent response, so there's no stored-JSON decode there); changing the storage format; the `Suggestions` JSON (same pattern but not in the issue — left as-is to keep the change minimal).

Decision (resolved by existing house style, not a new policy): drop+orphan rather than post a degraded Summary-only review — consistent with the issue's "rather than posting an empty-issues review" and the orphan precedent.

- **Runtime behavior change:** none for normal reviews (valid JSON). Only when a stored review's `Issues` is corrupt: instead of posting a findings-less review, the daemon logs a Warn, orphans the row (sentinel `github_review_id = -1`), and does **not** call `SubmitReview`. Net effect removes one misleading GitHub call in that case.
- **Rollout failure modes:** none — pure in-process logic in the retry loop; no config/secret/permission/API/migration; can't crashloop; compatible with old/new pods.
- **Blast radius:** the `PublishPending` retry path only, and only for reviews whose stored `Issues` JSON is corrupt. Normal reviews unaffected.
- **What to watch:** the new Warn line `pipeline: skipping pending review with corrupt issues JSON` (carries `review_id`, `pr`, `repo`, `err`). If it fires, a stored row is corrupt — investigate (DB integrity or a `Review.Issues` serialization-format drift). A growing count of sentinel `-1` reviews would indicate systematic corruption. Previously this was near-invisible (a posted review missing its issues).
- **Rollback & sequencing:** fully revertible, single file, no flag/migration/ordering. Reverting restores the silent issue-less post.

## Evidence

Acceptance: a corrupt-`Issues` review must not be posted issue-less, and must not retry forever.

<details><summary><code>make test-docker</code> — pipeline orphan/retry suite incl. the new test</summary>

```
--- PASS: TestPublishPending_LockedPRStopsRetrying (0.00s)
--- PASS: TestPublishPending_TransientErrorStillRetries (0.00s)
--- PASS: TestPublishPending_CorruptIssuesJSONOrphansWithoutPublishing (0.00s)
ok  	github.com/heimdallm/daemon/internal/pipeline
```
</details>

<details><summary>Full daemon suite — <code>go vet ./...</code> + <code>go test ./...</code> all green</summary>

```
ok  github.com/heimdallm/daemon/cmd/heimdallm
ok  github.com/heimdallm/daemon/internal/pipeline
ok  github.com/heimdallm/daemon/internal/store
... (all packages ok; vet clean)
```
</details>

## Test plan

- [x] `make test-docker` scoped to `./internal/pipeline/...` — new test + existing orphan/transient tests pass
- [x] `make test-docker` full daemon suite — `go vet` + `go test ./...` all green
- [ ] Reviewer: confirm orphan-and-drop (vs post Summary-only) is the desired behavior for a corrupt stored review

Closes #549

### DIFF
--- a/daemon/internal/pipeline/pipeline.go
+++ b/daemon/internal/pipeline/pipeline.go
@@ -696,6 +696,12 @@ func (p *Pipeline) Run(pr *github.PullRequest, opts RunOptions) (*store.Review, 
 	return rev, nil
 }
 
+// orphanedReviewID is the sentinel github_review_id stored for a review that
+// can never be published (no repo, corrupt stored JSON, or a permanent submit
+// failure). It marks the row as resolved so PublishPending stops retrying it,
+// while being distinguishable from a real GitHub review id.
+const orphanedReviewID = -1
+
 // markOrphanIfPermanent inspects the error returned by SubmitReview and,
 // when it is a *github.PermanentSubmitError, marks the local review row
 // as orphaned via the (-1, "") sentinel that PublishPending also uses
@@ -714,7 +720,7 @@ func (p *Pipeline) markOrphanIfPermanent(reviewID int64, submitErr error, source
 	if !errors.As(submitErr, &permErr) {
 		return false
 	}
-	if mErr := p.store.MarkReviewPublished(reviewID, -1, "", time.Now().UTC()); mErr != nil {
+	if mErr := p.store.MarkReviewPublished(reviewID, orphanedReviewID, "", time.Now().UTC()); mErr != nil {
 		slog.Warn("pipeline: failed to mark orphaned review, will retry next tick",
 			"review_id", reviewID, "source", source, "reason", permErr.Reason, "err", mErr)
 		return true
@@ -737,9 +743,9 @@ func (p *Pipeline) PublishPending() {
 			continue
 		}
 		// Skip reviews for PRs with no repo — orphaned records that will never publish.
-		// Mark them as permanently published (fake ID -1, empty state) to stop retry noise.
+		// Mark them as permanently published (orphanedReviewID, empty state) to stop retry noise.
 		if pr.Repo == "" {
-			_ = p.store.MarkReviewPublished(rev.ID, -1, "", time.Now().UTC())
+			_ = p.store.MarkReviewPublished(rev.ID, orphanedReviewID, "", time.Now().UTC())
 			slog.Info("pipeline: skipping pending review for PR with no repo", "review_id", rev.ID)
 			continue
 		}
@@ -752,7 +758,7 @@ func (p *Pipeline) PublishPending() {
 		if err := json.Unmarshal([]byte(rev.Issues), &issues); err != nil {
 			slog.Warn("pipeline: skipping pending review with corrupt issues JSON",
 				"review_id", rev.ID, "pr", pr.Number, "repo", pr.Repo, "err", err)
-			_ = p.store.MarkReviewPublished(rev.ID, -1, "", time.Now().UTC())
+			_ = p.store.MarkReviewPublished(rev.ID, orphanedReviewID, "", time.Now().UTC())
 			continue
 		}
 		result := &executor.ReviewResult{
--- a/daemon/internal/pipeline/pipeline.go
+++ b/daemon/internal/pipeline/pipeline.go
@@ -758,7 +758,10 @@ func (p *Pipeline) PublishPending() {
 		if err := json.Unmarshal([]byte(rev.Issues), &issues); err != nil {
 			slog.Warn("pipeline: skipping pending review with corrupt issues JSON",
 				"review_id", rev.ID, "pr", pr.Number, "repo", pr.Repo, "err", err)
-			_ = p.store.MarkReviewPublished(rev.ID, orphanedReviewID, "", time.Now().UTC())
+			if mErr := p.store.MarkReviewPublished(rev.ID, orphanedReviewID, "", time.Now().UTC()); mErr != nil {
+				slog.Warn("pipeline: failed to orphan corrupt review, will retry next tick",
+					"review_id", rev.ID, "err", mErr)
+			}
 			continue
 		}
 		result := &executor.ReviewResult{
--- a/daemon/internal/pipeline/pipeline.go
+++ b/daemon/internal/pipeline/pipeline.go
@@ -743,9 +743,18 @@ func (p *Pipeline) PublishPending() {
 			slog.Info("pipeline: skipping pending review for PR with no repo", "review_id", rev.ID)
 			continue
 		}
-		// Rebuild a minimal result from stored JSON for the body
+		// Rebuild a minimal result from stored JSON for the body. The daemon
+		// always writes well-formed JSON here, so a decode failure means the
+		// stored row is corrupt and the issue list is unrecoverable. Don't
+		// publish a misleading review missing its findings, and don't retry a
+		// deterministic failure forever — orphan it like the no-repo case above.
 		var issues []executor.Issue
-		json.Unmarshal([]byte(rev.Issues), &issues)
+		if err := json.Unmarshal([]byte(rev.Issues), &issues); err != nil {
+			slog.Warn("pipeline: skipping pending review with corrupt issues JSON",
+				"review_id", rev.ID, "pr", pr.Number, "repo", pr.Repo, "err", err)
+			_ = p.store.MarkReviewPublished(rev.ID, -1, "", time.Now().UTC())
+			continue
+		}
 		result := &executor.ReviewResult{
 			Summary:  rev.Summary,
 			Issues:   issues,
--- a/daemon/internal/pipeline/pipeline_orphan_test.go
+++ b/daemon/internal/pipeline/pipeline_orphan_test.go
@@ -231,10 +231,11 @@ func TestPublishPending_CorruptIssuesJSONOrphansWithoutPublishing(t *testing.T) 
 		t.Fatalf("upsert pr: %v", err)
 	}
 	// Seed an unpublished review whose Issues column is corrupt JSON.
-	if _, err := s.InsertReview(&store.Review{
+	prevReviewID, err := s.InsertReview(&store.Review{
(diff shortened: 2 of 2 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Solid, root-cause-aware bugfix for #549: a corrupt stored issues JSON previously caused json.Unmarshal's error to be silently ignored, posting a misleading review with an empty findings list and (since the row stayed unpublished) retrying forever. The fix decodes with explicit error handling, logs a contextual warning, and orphans the row via MarkReviewPublished so it is neither posted nor re-attempted. The added test verifies SubmitReview is not called, the review is removed from the unpublished set, and a second tick is a no-op. Good observability and idempotency coverage. No confirmed blocking issues.

### Suggestions

- The sentinel value -1 passed as the GitHub review ID in MarkReviewPragmaPublished is a magic number; consider a named constant (e.g. orphanedReviewID) shared with the no-repo orphan path to make the intent self-documenting and consistent.
- The doc comment says it orphans 'like the no-repo case above,' but the no-repo branch only does `continue` (no MarkReviewPublished), so its behavior differs. Either align the wording or, if the no-repo case can also loop, apply the same orphaning there.
- The test asserts SubmitReview is not called and the row is no longer unpublished, but does not assert the stored GitHubReviewID == -1; adding that check would lock in the orphan-sentinel contract against future regressions.
- Consider emitting a metric/counter for orphaned-corrupt reviews so operators can detect a spike in corrupt rows in production rather than relying solely on log scraping.

---

vbuenog:
Thanks @Muriano 🙏 — addressed in `61589a5` (full `make test-docker`: `go vet` + `go test ./...` all green).

### ✅ Implemented
- **Named sentinel const**: added `orphanedReviewID = -1` and applied it at all three orphan sites (`markOrphanIfPermanent`, the no-repo branch, and the new corrupt-JSON branch) — self-documenting and consistent.
- **Assert the sentinel in the test**: the corrupt-JSON test now reads the row back via `GetReview` and asserts `GitHubReviewID == -1`, locking the orphan-sentinel contract.

### ↩️ Declined (factual)
- **"the no-repo branch only does `continue` (no MarkReviewPublished), so its behavior differs"**: it actually does orphan identically — `pipeline.go` (no-repo branch):
  ```go
  if pr.Repo == "" {
      _ = p.store.MarkReviewPublished(rev.ID, orphanedReviewID, "", time.Now().UTC())
      slog.Info("pipeline: skipping pending review for PR with no repo", "review_id", rev.ID)
      continue
  }
  ```
  So `MarkReviewPublished(-1)` + `continue` — same as the corrupt-JSON path. The "like the no-repo case above" comment is accurate, no change needed. (Both now use the shared const, which also makes the parallel explicit.)

### 📋 Deferred
- **Metric/counter for orphaned-corrupt reviews**: agreed, same metrics-surface prerequisite as the bridge counters — folded into #559 rather than a separate issue.

Re-requesting review on the new commit.

alepalma91 (approved):
## 🤖 Heimdallm AI Review

Clean, well-reasoned fix for #549: corrupt issues JSON is now orphaned (sentinel -1 + continue) instead of publishing a misleading findings-less review and retrying a deterministic failure forever. The named const is applied at all three orphan sites, the test asserts no SubmitReview call, empty unpublished set, the -1 sentinel contract, and idempotency on a second tick. Verified the daemon always writes valid JSON (null/[]/array), so legitimate rows never hit the new branch — no false-positive orphaning. Prior review comments are all addressed. No correctness bugs found.

### Suggestions

- Orphaning discards a possibly-valid Summary/Severity along with the corrupt issues. This is defensible per #549 (a findings-less review is misleading), and it's already logged — no change needed, just noting the intentional data loss.
- The new branch ignores the MarkReviewPublished error (`_ =`), unlike markOrphanIfPermanent which logs store-write failures at Warn. It matches the sibling no-repo branch so it's consistent, but a failed write here silently re-loops next tick; a Warn log would aid diagnosis. Low priority.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #563: the author addressed all prior feedback in 61589a5. The named sentinel constant `orphanedReviewID = -1` is introduced and consistently applied across all three orphan sites (markOrphanIfPermanent, no-repo branch, corrupt-JSON branch). The new corrupt-JSON path correctly logs a contextual warning and orphans the row to stop deterministic retry loops, matching the existing no-repo pattern. The test asserts SubmitReview is not called, the row is removed from unpublished, the sentinel id is stamped, and subsequent ticks remain no-ops. Code is correct, well-documented, and the declined/deferred items are factually justified. No blocking issues.

### Suggestions

- Consider extracting the repeated `_ = p.store.MarkReviewPublished(rev.ID, orphanedReviewID, "", time.Now().UTC())` + continue used in both the no-repo and corrupt-JSON branches into a small helper (e.g. orphanPending(rev, reason, attrs...)) to keep the two parallel orphan sites from drifting; optional, purely cosmetic.
- The orphaned-corrupt counter/metric deferred to #559 is the right call; once that metrics surface lands, emitting a counter here would make silent data-corruption orphaning observable in production rather than only visible in logs.

---

vbuenog:
Thanks for the re-review @alepalma91 @Muriano 🙏 — one more revision in `5851718`.

### ✅ Implemented
- **Log the orphan-write failure** (@alepalma91): the corrupt-JSON branch now logs at Warn if `MarkReviewPublished` fails (mirroring `markOrphanIfPermanent`), so a failed orphan-write that re-loops next tick is diagnosable rather than silent.

### ✅ Noted / no change
- **Orphaning discards a possibly-valid Summary/Severity** (@alepalma91): intended per #549 (a findings-less review is misleading), and logged — agreed, no change.

### 📋 Deferred
- **Extract a shared `orphanPending` helper for the no-repo + corrupt-JSON branches** (@Muriano): reasonable but purely cosmetic, and it touches the pre-existing no-repo branch (different Info-level log + message), so I'd rather not widen this fix. Happy to do it as a small follow-up if you'd like — say the word and I'll file it.
- **Orphaned-corrupt metric** (@Muriano): tracked in #559 alongside the bridge counters.

Re-requesting review on the new commit.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #563 (commit 5851718). The corrupt-JSON branch now orphans the review with the named `orphanedReviewID = -1` sentinel instead of posting an empty/misleading review, and logs at Warn if the orphan-write itself fails (mirroring `markOrphanIfPermanent`). The new test asserts SubmitReview is never called, the row is orphaned (no longer unpublished), the sentinel id is stamped, and subsequent ticks are a no-op. All prior findings (named const at all three sites, sentinel assertion in test, orphan-write failure logging) are addressed. The fix is root-cause, deterministic, and well-covered. No blocking issues found; remaining items (shared `orphanPending` helper, orphaned-corrupt metric) are correctly deferred to follow-ups/#559.

### Suggestions

- Optional follow-up (already noted by author): extract a shared `orphanPending(rev.ID, ...)` helper for the no-repo and corrupt-JSON branches to remove the duplicated MarkReviewPublished + Warn-on-failure pattern. Purely cosmetic and explicitly deferred.
- Optional: emit a metric/counter for orphaned-corrupt reviews so silent data corruption is observable in production rather than only via logs (tracked in #559).

---

alepalma91 (approved):
## 🤖 Heimdallm AI Review

Clean, well-scoped fix for #549. The corrupt-issues-JSON branch now orphans the review (sentinel -1) and logs at Warn instead of silently publishing a findings-less review or retrying a deterministic decode failure forever. Verified the write path always marshals (null/[] for empty), so legitimately-empty reviews decode fine and are not wrongly orphaned. Sentinel const applied consistently across all three orphan sites; test locks the contract (no submit, orphaned, sentinel, idempotent). All prior-round reviewer concerns addressed; remaining items are explicitly deferred follow-ups, not blockers. No new issues.

### Suggestions

- Deferred items are reasonable: the shared orphanPending helper (no-repo + corrupt-JSON) is purely cosmetic and the orphaned-corrupt metric is tracked in #559 — fine to land this as-is.

---

PR #564, opened 2026-06-19
fix(server): PUT /config returns 500 when SetConfig persist fails (closes #550)

### LINKED ISSUE #550: fix(server): PUT /config returns 200 OK even when SetConfig persist fails [MEDIUM]
## Summary (MEDIUM)
`PUT /config` logs a failed `SetConfig` but still returns `200 {"status":"ok"}`. The client/UI believes the config was saved when the database write failed (disk full, lock, permission), and the next reload re-applies stale values.

## Evidence
`daemon/internal/server/handlers.go:810-814`
```go
if _, err := srv.store.SetConfig(k, val); err != nil {
    slog.Error("config: set", "key", k, "err", err)
}
// ... falls through to:
writeJSON(w, http.StatusOK, map[string]string{"status": "ok"})
```

## Suggested fix
Accumulate per-key errors; if any `SetConfig` fails, return `500` with the failed keys instead of `200`.

---

vbuenog:
## Summary
`PUT /config` logged a failed `SetConfig` but fell through to `200 {"status":"ok"}`, so the client/UI believed the config was saved when the DB write actually failed (disk full, lock, permission) — and the next config reload silently re-applied the stale values. This makes the response honest: if any key fails to persist, the endpoint returns **500** with the offending keys. Fixes #550.

## How it works
`daemon/internal/server/handlers.go` (`handlePutConfig`):
- The per-key write loop now accumulates `failedKeys` when `SetConfig` returns an error (still logged via `slog.Error`, as before).
- After the loop, if `failedKeys` is non-empty, the handler returns `500` with `{"error": "...", "failed_keys": [...]}`. The keys are sorted so the payload is deterministic (Go map iteration order is random).
- The happy path is byte-for-byte unchanged: all writes succeed → `200 {"status":"ok"}`.

## Scope
**In scope**
- Honest status code + `failed_keys` payload on persist failure for `PUT /config`.
- Test that forces a store-write failure and asserts `500` + `failed_keys`.

**Out of scope**
- Transactional / all-or-nothing multi-key writes. The fix is intentionally **not** atomic — keys processed before the failing one may already be persisted. Making the write a single transaction is a larger, separate change; the bug in #550 is specifically the *misleading 200*, which this fixes. Called out in a code comment too.

- **Runtime behavior change**: `PUT /config` now returns `500` (JSON `{error, failed_keys}`) when a `SetConfig` write fails, instead of a misleading `200`. No new outbound calls, startup invariants, or resource usage. Success path is identical to before.
- **Rollout failure modes**: None at startup — no new config/secret/permission, cannot crashloop. For existing clients the response changes **only in the failure case** (a write that actually fails). The Flutter client already treats any non-200 as an error (`api_client.dart:238` throws `ApiException`), so it surfaces a real error toast instead of a silent false success — the intended correction, not a regression. No success-path contract change.
- **Blast radius**: `PUT /config` only (config save from UI/CLI), and only on requests where a DB write genuinely fails — normally zero. No other endpoint touched.
- **What to watch**: post-deploy, a spike in `500` on `PUT /config` and the `config: set` `slog.Error` lines now indicates real persistence failures (disk/lock/permission) that were previously invisible. The `200` rate on successful saves should be unchanged (no false positives).
- **Rollback & sequencing**: Fully revertible, no migrations, no flags, no deploy ordering. Reverting restores the old swallow-and-200 behavior.

## Evidence
<details><summary>New test: persist failure → 500 + failed_keys (and happy path still 200)</summary>

```
=== RUN   TestHandlerPutConfig
--- PASS: TestHandlerPutConfig (0.01s)
=== RUN   TestHandlerPutConfig_PersistFailureReturns500
--- PASS: TestHandlerPutConfig_PersistFailureReturns500 (0.01s)
PASS
ok  	github.com/heimdallm/daemon/internal/server	0.049s
```
</details>

## Test plan
- [x] `go vet ./internal/server/`
- [x] `go test ./internal/server/` (new + existing PUT /config tests green)
- [x] `make verify-linux` (full Linux build verification passed)
- [ ] Reviewer: confirm the non-transactional behavior is acceptable for the multi-key case, or request a follow-up for atomic writes.

Closes #550

### DIFF
--- a/daemon/internal/server/handlers.go
+++ b/daemon/internal/server/handlers.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -786,6 +787,7 @@ func (srv *Server) handlePutConfig(w http.ResponseWriter, r *http.Request) {
 		body["agent_configs"] = normalized
 	}
 
+	var failedKeys []string
 	for k, v := range body {
 		// Read-only keys were accepted above to avoid 400s on UI saves that
 		// round-trip the GET payload, but they must not land in the store.
@@ -809,7 +811,20 @@ func (srv *Server) handlePutConfig(w http.ResponseWriter, r *http.Request) {
 		}
 		if _, err := srv.store.SetConfig(k, val); err != nil {
 			slog.Error("config: set", "key", k, "err", err)
+			failedKeys = append(failedKeys, k)
 		}
+	}
+	// A swallowed persist error here would 200-OK a save that never hit disk,
+	// so the UI shows success while the next reload silently re-applies stale
+	// values (#550). Report 500 with the offending keys instead. Note this is
+	// not transactional: earlier keys in the loop may already be persisted.
+	if len(failedKeys) > 0 {
+		sort.Strings(failedKeys) // deterministic: Go map iteration order is random
+		writeJSON(w, http.StatusInternalServerError, map[string]any{
+			"error":       "failed to persist one or more config keys",
+			"failed_keys": failedKeys,
+		})
+		return
 	}
 	writeJSON(w, http.StatusOK, map[string]string{"status": "ok"})
 }
--- a/daemon/internal/server/handlers.go
+++ b/daemon/internal/server/handlers.go
@@ -788,12 +788,14 @@ func (srv *Server) handlePutConfig(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var failedKeys []string
+	var attempted int
 	for k, v := range body {
 		// Read-only keys were accepted above to avoid 400s on UI saves that
 		// round-trip the GET payload, but they must not land in the store.
 		if _, readOnly := readOnlyConfigKeys[k]; readOnly {
 			continue
 		}
+		attempted++
 		var val string
 		switch typed := v.(type) {
 		case string:
@@ -820,6 +822,7 @@ func (srv *Server) handlePutConfig(w http.ResponseWriter, r *http.Request) {
 	// not transactional: earlier keys in the loop may already be persisted.
 	if len(failedKeys) > 0 {
 		sort.Strings(failedKeys) // deterministic: Go map iteration order is random
+		slog.Warn("config: partial persist failure", "failed", len(failedKeys), "attempted", attempted)
 		writeJSON(w, http.StatusInternalServerError, map[string]any{
 			"error":       "failed to persist one or more config keys",
 			"failed_keys": failedKeys,
--- a/daemon/internal/server/handlers_test.go
+++ b/daemon/internal/server/handlers_test.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

El PR corrige correctamente #550: antes PUT /config devolvía 200 OK aunque SetConfig fallara al persistir, lo que provocaba que la UI mostrara éxito mientras los valores no llegaban a disco. Ahora recopila las claves fallidas, las ordena de forma determinista y devuelve 500 con la lista de claves afectadas. El fix ataca la causa raíz, está bien comentado (incluido el reconocimiento explícito de la no-transaccionalidad) y trae un test que cubre el sad path cerrando el store. No se detectan issues bloqueantes; la única observación relevante es la naturaleza no atómica de la persistencia, que ya viene documentada en el propio diff.

### Suggestions

- Persistencia no transaccional (daemon/internal/server/handlers.go): el comentario reconoce que claves anteriores del loop pueden quedar ya persistidas cuando una posterior falla. Esto deja el store en estado parcial tras un 500. No bloquea el merge —es la mejora correcta sobre el 200 silencioso— pero convendría abrir una issue de seguimiento para evaluar persistencia atómica/rollback o, al menos, documentar en la respuesta de error que el guardado es best-effort por clave.
- El campo `failed_keys` expone nombres de claves de configuración en la respuesta de error. Son claves no sensibles (p.ej. poll_interval), por lo que el riesgo es mínimo, pero conviene asegurarse de no incluir nunca claves que puedan revelar estructura sensible si el conjunto de claves crece en el futuro.
- El test fuerza el fallo cerrando el store (sql: database is closed). Es realista y determinista, pero sería valioso añadir también un caso con múltiples claves donde solo algunas fallan, para verificar que `failed_keys` contiene exactamente el subconjunto fallido y mantiene el orden alfabético determinista que introduce sort.Strings.
- Las respuestas de éxito (map[string]string) y de error (map[string]any) usan formas distintas. Es funcionalmente correcto, pero considerar un shape de respuesta consistente facilitaría el parseo en el cliente.

---

alepalma91 (approved):
## 🤖 Heimdallm AI Review

Solid, well-scoped fix. Collects failed SetConfig keys and returns 500 with sorted failed_keys instead of a misleading 200. The non-transactional partial-persist caveat is explicitly documented, the sort makes the response deterministic, and the test correctly isolates the persist failure by closing the store after pure validation runs. No correctness bugs found.

### Suggestions

- Partial-persist is documented but the client gets no signal about which keys DID persist — only the failures. If the UI needs to reconcile state, consider noting that retrying the whole payload is the recovery path.
- Optional: the loop already continues past failures (good — collects all bad keys); a one-line summary log after the loop (count failed/total) would aid ops triage, but slog.Error per key is acceptable.

---

vbuenog:
Thanks @Muriano @alepalma91 🙏 — addressed in `f4709b5` (`go vet` + `go test ./internal/server/` + full `make verify-linux` all green).

### ✅ Implemented
- **Stronger failure test** (@Muriano): the test now PUTs multiple writable keys (`poll_interval`, `retention_days`, `review_mode`) in non-alphabetical order against a closed store and asserts `failed_keys` is the **exact sorted set**, locking both the accumulate-*all*-failures and the deterministic-order (`sort.Strings`) contract.
  - One caveat: a true *"only some keys fail"* subset can't be exercised without a store interface seam — `srv.store` is a concrete `*store.Store`, so per-key error injection isn't possible. Closing the store fails all writes; the exact-set + sort assertion is the feasible lock. Folded the rationale into a test comment.
- **Partial-persist summary log** (@alepalma91): added a single `slog.Warn("config: partial persist failure", "failed", …, "attempted", …)` after the loop for ops triage, alongside the existing per-key `slog.Error`.

### 📋 Deferred → #565
- **Non-transactional / partial-persist** (@Muriano) + **client can't tell which keys persisted** (@alepalma91): these share one root — the write is best-effort per key, so a `500` can leave a partial state and the client only learns the failures. Opened **#565** to evaluate wrapping the writes in a single transaction (all-or-nothing) or, failing that, documenting the best-effort contract + retry-whole-payload recovery path. Out of scope here since #564 is already a strict improvement over the silent `200`.

### ↩️ Declined (with reasoning)
- **`failed_keys` exposes config key names** (@Muriano): no change. The keys are a fixed, non-sensitive allowlist (`validConfigKeys` — `poll_interval`, `retention_days`, …) and only key *names* are returned, never values. If a sensitive-by-name key is ever added, that's the point to revisit — noting it here so the caution isn't lost.
- **Unify success vs error response shape** (@Muriano): keeping them distinct. `{"status":"ok"}` and `{"error","failed_keys"}` carry genuinely different payloads for different outcomes, and the client already discriminates on the status code (the Flutter client throws on any non-200). Subjective — happy to reconsider if you feel strongly.

PR #545, opened 2026-06-19
feat(flutter): free-form poll_interval input in Settings (closes #543)

vbuenog:
## Summary

The Settings screen hardcoded `poll_interval` to a 4-item dropdown (`1m`, `5m`, `30m`, `1h`), but since #539 the daemon accepts **any** `time.ParseDuration` value within `[1m, 24h]` (enforced by `config.ValidatePollInterval` on config load and on `PUT/PATCH /config`). As a result an operator could not select a valid value like `3m` or `90m` from the GUI, even though the backend accepts it. This was flagged in the AI review of #539. Closes #543.

## How it works

`flutter_app/lib/features/config/config_screen.dart`:
- Replaces `DropdownButtonFormField` with a free-form `TextFormField` (mirroring the adjacent Retention field's pattern), bound to a `TextEditingController` so quick-pick chips can update it.
- Adds top-level `validatePollInterval` + `_parseGoDuration`, which validate client-side against the daemon's bounds (`_minPollInterval = 1m`, `_maxPollInterval = 24h`, mirrored from `daemon/internal/config/config.go`). `_parseGoDuration` handles the common Go duration forms operators use (`5m`, `90m`, `1h30m`, `1.5h`, `300s`) — it is a **UX guard, not a reimplementation** of Go's full `ParseDuration`; anything it misses is still caught by the daemon's `400`, which is already surfaced as an error toast on save.
- Keeps `1m`, `5m`, `30m`, `1h` as quick-pick `ActionChip`s.

## Scope

**In scope:**
- Free-form, client-validated `poll_interval` input in the Flutter Settings screen.
- Quick-pick chips for the common values.

**Out of scope:**
- Backend validation (already shipped in #539; unchanged here).
- TUI (read-only display of the value; not affected).

## Production impact & what to watch

This is a **client-side-only UI change** in the Flutter app — no daemon code, no API contract change. The saved config JSON shape is identical (`pollInterval` is still a string).

- **Runtime behavior change:** GUI only — operators can now type any duration in `[1m, 24h]` instead of choosing from 4 fixed values; the daemon still validates authoritatively on save. Also **fixes a latent crash**: `DropdownButtonFormField` asserts its `value` matches exactly one item, so loading a config whose `poll_interval` was not one of the 4 whitelisted values (e.g. `60s`, `3m`) could throw on the Settings screen. The text field has no such constraint.
[...]

PR #566, opened 2026-06-22
test(config): guard against poll_interval bounds drift between daemon and GUI (closes #546)

### LINKED ISSUE #546: Guard against drift between client/daemon poll_interval bounds [1m, 24h]
## Context

Follow-up from the review of #545 (which closes #543).

The acceptable `poll_interval` range `[1m, 24h]` is now encoded in **two** places:

- **Daemon (authoritative):** `daemon/internal/config/config.go` — `minPollInterval = 1m`, `maxPollInterval = 24h`, enforced by `config.ValidatePollInterval`.
- **Flutter GUI (client-side UX guard):** `flutter_app/lib/features/config/config_screen.dart` — `_minPollInterval`, `_maxPollInterval`, used by `validatePollInterval`.

The Dart side carries a comment marking the daemon as authoritative, but a silent divergence (e.g. someone changes the Go bounds without touching Dart) would only surface at runtime — the GUI would either reject values the daemon accepts, or let through values the daemon rejects (caught by the 400, but defeating the "immediate feedback" goal).

## Proposal

Add a lightweight contract check so the two ranges can't drift unnoticed. Options:

- Export the daemon bounds (e.g. via the existing HTTP config/metadata surface or a generated constants file) and have the GUI read them instead of hardcoding.
- Or a CI assertion that greps both files and fails if the literals diverge.

Low priority — the daemon remains authoritative, so drift is a UX-quality issue, not a correctness bug.

Raised by Muriano in the AI review of #545.

vbuenog:
## Summary

The acceptable `poll_interval` range `[1m, 24h]` is encoded independently in two places — the authoritative daemon constants (`minPollInterval`/`maxPollInterval`, enforced by `ValidatePollInterval`) and the Flutter GUI's client-side validation guard (`_minPollInterval`/`_maxPollInterval`). A change to the Go bounds without a matching change to Dart (or vice versa) would drift silently: the GUI would reject values the daemon accepts, or pass values the daemon rejects — defeating the immediate-feedback goal even though the daemon stays authoritative. This adds the lightweight contract check requested in the #545 review follow-up.

## How it works

New internal test `daemon/internal/config/poll_interval_drift_test.go` (`package config`, so it can read the unexported bounds):

- Reads the Flutter config screen at `../../../flutter_app/lib/features/config/config_screen.dart`.
- Extracts the `_minPollInterval` / `_maxPollInterval` `Duration(...)` literals and converts them to `time.Duration` (summing named args, so equivalent spellings like `Duration(hours: 24)` and `Duration(days: 1)` compare equal).
- Asserts they equal the daemon's `minPollInterval` / `maxPollInterval`, with an error message pointing at the file to fix.
- Skips (rather than fails) if the Dart file is absent, for daemon-only checkouts; in-repo CI always has it.

Runs inside the existing `go test ./...` daemon CI job — no new workflow, no runtime/production code change. Follows the existing `config_guards_drift_test.go` convention of catching drift at test time.

## Scope

In scope:
- A test-time contract assertion that the two `[1m, 24h]` copies stay in sync.

Out of scope (deliberately — the issue tags this low-priority UX polish):
- Exporting the bounds over the HTTP `/config` surface and having the GUI fetch them live. That adds an endpoint + async client loading + fallback handling; not warranted for a UX-quality guard while the daemon remains authoritative.

**Low impact — test-only; nothing changes at runtime.** The new file is `package config` test code and is never compiled into the daemon binary.

- **Runtime behavior change:** none — binary is unchanged; no new calls, startup invariants, responses, or resource use.
- **Rollout failure modes:** none in production. The only new gate is in **CI**: the `Daemon (Go)` job fails if the Flutter bounds diverge from the Go constants, or if the Dart constant is refactored into a shape the parser can't read (clear `t.Fatalf` tells the author to update the guard). That's a red build, not a prod incident.
- **Blast radius:** CI only. No deployed artifact, consumer, or endpoint affected.
- **What to watch:** the `Daemon (Go)` CI check on PRs touching the bounds in `config.go` or `config_screen.dart`. Caveat: Go's test cache keys on the package's Go sources, not the external Dart file, so a *local* incremental run editing only the Dart file may return a stale cached pass — CI uses fresh checkouts so it always executes; locally use `-count=1`.
- **Rollback & sequencing:** trivially revertible (delete one file). No coordination needed.

## Evidence

The issue's goal is that drift can't pass unnoticed. Verified the guard passes on the current (aligned) tree and fails on induced drift:

<details>
<summary>Aligned tree — PASS</summary>

```
$ go test ./internal/config/ -run TestPollIntervalBounds_MatchFlutterGUI -count=1 -v
=== RUN   TestPollIntervalBounds_MatchFlutterGUI
--- PASS: TestPollIntervalBounds_MatchFlutterGUI (0.00s)
PASS
ok  	github.com/heimdallm/daemon/internal/config
```
</details>

<details>
<summary>Induced drift (temporarily set Flutter <code>_maxPollInterval</code> to 12h) — FAIL</summary>

```
=== RUN   TestPollIntervalBounds_MatchFlutterGUI
    poll_interval_drift_test.go:51: poll_interval upper bound drift: daemon maxPollInterval=24h0m0s but Flutter _maxPollInterval=12h0m0s
        update ../../../flutter_app/lib/features/config/config_screen.dart to match daemon/internal/config/config.go (see issue #546)
--- FAIL: TestPollIntervalBounds_MatchFlutterGUI (0.00s)
FAIL
```
</details>

## Test plan

- [x] `go test ./internal/config/ -count=1` passes (full package).
- [x] New test passes on the aligned tree.
- [x] New test fails with a clear message on induced drift (12h), then passes again after revert.
- [x] `go vet ./...` clean; `gofmt -l` clean.
- [ ] Reviewer: confirm CI `Daemon (Go)` job runs and passes on this PR.

Closes #546

### DIFF
--- a/daemon/internal/config/poll_interval_drift_test.go
+++ b/daemon/internal/config/poll_interval_drift_test.go
@@ -1,0 +1,98 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"testing"
+	"time"
+)
+
+// dartConfigPath is the Flutter Settings screen, which carries a client-side
+// copy of the poll_interval bounds for immediate-feedback validation. The
+// daemon is authoritative (minPollInterval/maxPollInterval, enforced by
+// ValidatePollInterval); this guard makes sure the GUI's copy can't silently
+// drift from it. Path is relative to this package directory (daemon/internal/
+// config), so ../../../ lands at the repo root. See issue #546.
+const dartConfigPath = "../../../flutter_app/lib/features/config/config_screen.dart"
+
+var (
+	dartMinPollRe = regexp.MustCompile(`_minPollInterval\s*=\s*Duration\(([^)]*)\)`)
+	dartMaxPollRe = regexp.MustCompile(`_maxPollInterval\s*=\s*Duration\(([^)]*)\)`)
+	// Named arguments of Dart's Duration() constructor, e.g. `minutes: 1`.
+	dartDurationArgRe = regexp.MustCompile(`(days|hours|minutes|seconds|milliseconds|microseconds)\s*:\s*([0-9]+)`)
+)
+
+// TestPollIntervalBounds_MatchFlutterGUI fails if the [1m, 24h] poll_interval
+// range encoded in the Flutter GUI diverges from the authoritative daemon
+// bounds. It catches the silent drift described in issue #546: a change to the
+// Go constants without a matching change to the Dart guard (or vice versa).
+func TestPollIntervalBounds_MatchFlutterGUI(t *testing.T) {
+	src, err := os.ReadFile(dartConfigPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// The GUI lives in the same repo, so in CI this file is always
+			// present. Skip (rather than fail) for daemon-only checkouts where
+			// there is no Dart copy that could drift.
+			t.Skipf("Flutter config screen not present at %s; skipping cross-language drift check", filepath.Clean(dartConfigPath))
+		}
+		t.Fatalf("read %s: %v", dartConfigPath, err)
+	}
+
+	gotMin := parseDartDuration(t, "_minPollInterval", dartMinPollRe, src)
+	gotMax := parseDartDuration(t, "_maxPollInterval", dartMaxPollRe, src)
+
+	if gotMin != minPollInterval {
+		t.Errorf("poll_interval lower bound drift: daemon minPollInterval=%s but Flutter _minPollInterval=%s\n"+
+			"update %s to match daemon/internal/config/config.go (see issue #546)", minPollInterval, gotMin, dartConfigPath)
+	}
+	if gotMax != maxPollInterval {
+		t.Errorf("poll_interval upper bound drift: daemon maxPollInterval=%s but Flutter _maxPollInterval=%s\n"+
+			"update %s to match daemon/internal/config/config.go (see issue #546)", maxPollInterval, gotMax, dartConfigPath)
+	}
+}
+
+// parseDartDuration extracts the Dart Duration(...) literal named by re and
+// converts it to a time.Duration, summing all named arguments so equivalent
+// spellings (e.g. Duration(hours: 24) vs Duration(days: 1)) compare equal.
+func parseDartDuration(t *testing.T, name string, re *regexp.Regexp, src []byte) time.Duration {
+	t.Helper()
+	m := re.FindSubmatch(src)
+	if m == nil {
+		t.Fatalf("could not find %s = Duration(...) in %s; the drift guard needs updating (issue #546)", name, dartConfigPath)
+	}
+	args := dartDurationArgRe.FindAllSubmatch(m[1], -1)
+	if len(args) == 0 {
+		t.Fatalf("could not parse Duration arguments for %s (got %q)", name, m[1])
+	}
+	var total time.Duration
+	for _, a := range args {
+		n, err := strconv.Atoi(string(a[2]))
+		if err != nil {
+			t.Fatalf("invalid numeric literal %q for %s: %v", a[2], name, err)
+		}
+		total += dartDurationUnit(string(a[1])) * time.Duration(n)
+	}
+	return total
+}
+
+// dartDurationUnit maps a Dart Duration named argument to its time.Duration
+// equivalent.
+func dartDurationUnit(unit string) time.Duration {
+	switch unit {
+	case "days":
+		return 24 * time.Hour
+	case "hours":
+		return time.Hour
+	case "minutes":
+		return time.Minute
+	case "seconds":
+		return time.Second
+	case "milliseconds":
+		return time.Millisecond
+	case "microseconds":
+		return time.Microsecond
+	}
+	return 0
+}
--- a/daemon/internal/config/poll_interval_drift_test.go
+++ b/daemon/internal/config/poll_interval_drift_test.go
@@ -9,33 +9,53 @@ import (
 	"time"
 )
 
+// flutterAppDir is the root of the Flutter GUI module. Its presence is what
+// distinguishes a daemon-only checkout (no Dart copy that could drift, so the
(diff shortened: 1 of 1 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR solo de test que añade un guard de regresión cruzando los bounds de poll_interval del daemon (Go) con la copia client-side del GUI Flutter (Dart) para detectar drift silencioso (issue #546). Verificado: las constantes minPollInterval/maxPollInterval y ValidatePollInterval existen en daemon/internal/config/config.go, y los literales Dart (Duration(minutes:1)/Duration(hours:24)) coinciden, por lo que el test compila y pasa. Test bien escrito, comentado y con skip elegante si falta el fichero Dart.

### Suggestions

- Opcional: dartDurationUnit devuelve 0 para una unidad desconocida en lugar de fallar; si Dart introdujera una unidad no contemplada el cálculo silenciaría el error en vez de avisar. Hoy las unidades cubiertas son exhaustivas para Duration, así que es solo defensa a futuro.
- Opcional: el regex Duration\(([^)]*)\) podría hacer match de una declaración comentada en el .dart; improbable dada la forma actual del fichero.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured cross-language drift guard test. The approach of parsing Dart source from Go tests to compare authoritative bounds is pragmatic and the skip-on-missing-file behavior is appropriate. Two minor robustness issues.

### Issues

🟡 **daemon/internal/config/poll_interval_drift_test.go:91** — `dartDurationUnit` silently returns 0 for unrecognized unit strings instead of failing the test. If the regex accidentally captures a malformed or future-unknown unit name, the parsed duration will be silently short, potentially causing the drift check to produce a misleading mismatch error rather than a clear parse failure. Consider calling `t.Fatalf` for unknown units (this requires passing `*testing.T` into the function or returning an error).
🟡 **daemon/internal/config/poll_interval_drift_test.go:22** — The regex `[^)]*` inside `Duration\(([^)]*)\)` will break on nested parentheses (e.g., a hypothetical `Duration(minutes: foo())`). In practice Dart Duration only takes integer literals as named args so this is not a real risk today, but a comment noting the assumption would help future maintainers.

### Suggestions

- Have `dartDurationUnit` signal an error (or accept `*testing.T` and call `t.Fatalf`) on an unrecognized unit so that parse failures surface clearly instead of silently yielding a zero contribution.
- Consider adding a brief table-driven sub-test (or test comment) that documents the expected current values (e.g., 1m and 24h) so a reader can see at a glance what the bounds are without cross-referencing config.go.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Well-crafted cross-language drift guard test that parses the Flutter Dart Duration literals via regex and compares them to the authoritative daemon constants (minPollInterval/maxPollInterval). The intent of issue #546 is met, units are normalized correctly, and failure messages are actionable. No correctness bugs or security concerns. The main weakness is that the file-not-exist branch silently skips, which can turn the guard into a no-op if the Dart file is moved/renamed, plus some regex brittleness against common Dart spellings.

### Issues

⚠️ **daemon/internal/config/poll_interval_drift_test.go:29** — The os.IsNotExist branch unconditionally Skips. This is intended for daemon-only checkouts, but it also fires if the Dart file is renamed/moved within the same repo — exactly the kind of change this guard should catch. Impact: the drift guard silently becomes a no-op and gives false confidence (the test 'passes' green while no validation runs). Recommended fix: only Skip when the GUI is genuinely absent (e.g., detect repo root / absence of flutter_app dir) and Fatalf when flutter_app exists but the specific config_screen.dart path does not, so a rename surfaces as a failure rather than a silent skip.

### Suggestions

- The regexes `_minPollInterval\s*=\s*Duration\(...\)` require `Duration(` immediately after `=`. Common Dart spellings like `= const Duration(...)` would not match and trigger Fatalf. Consider allowing an optional `const` (e.g., `=\s*(?:const\s+)?Duration\(`) to be robust against an equivalent refactor on the Flutter side.
- FindSubmatch only inspects the first occurrence of each constant. If the Dart file ever has more than one declaration matching the pattern, only the first is validated. A brief comment noting this assumption, or asserting a single match, would harden the guard.
- dartDurationUnit returns 0 for an unrecognized unit; it is currently unreachable because dartDurationArgRe only captures known units, but a small assertion or default-case t.Fatalf would make the invariant explicit if the arg regex is ever broadened.

---

vbuenog:
Thanks @Muriano @sergiotejon @ivanmunozruiz — all review feedback addressed in `486ead2`. The points converged, so summarizing the dispositions in one place:

**Implemented (code):**

- **Silent skip on rename/move** (Muriano, MEDIUM) — the strongest point. The `os.IsNotExist` branch now distinguishes a genuine daemon-only checkout (no `flutter_app/` dir → `Skip`, nothing can drift) from `flutter_app/` present but `config_screen.dart` missing (→ `t.Fatalf`). A rename now surfaces as a failure instead of a no-op green. Verified: renaming the file with `flutter_app/` present fails with the new "was it renamed/moved?" message.
- **`dartDurationUnit` silently returns 0 for unknown units** (sergio 🟡, ivan, Muriano) — it now returns `(time.Duration, ok bool)`; the parser `t.Fatalf`s on an unrecognized unit instead of contributing zero. Still unreachable today (the arg regex only captures known units), but the invariant is now explicit if that regex is ever broadened.
- **Regex doesn't tolerate `= const Duration(...)`** (Muriano) — both regexes now allow an optional `const`, so an equivalent Flutter-side refactor won't trip a false `Fatalf`. Verified the `const` spelling still matches.

**Implemented (comments):**

- Nested-parens assumption for `[^)]*` (sergio 🟡), single-declaration assumption for `FindSubmatch` (Muriano), and the current expected bounds 1m/24h (sergio) are now documented inline.

Verified after the changes: aligned tree passes, induced drift still caught, `const`-refactor matches, rename fails loudly; `go vet` + `gofmt` clean.

PR #568, opened 2026-06-22
fix(server): JSON-escape repo-rename 400 error body (closes #552)

### LINKED ISSUE #552: fix(server): repo-rename error built as unescaped JSON → malformed response body [MEDIUM]
## Summary (MEDIUM)
The `/admin/repo-rename` error path concatenates `err.Error()` straight into a JSON string literal without escaping. If the error text contains a quote or backslash (it can include the user-supplied repo slug), the response body is **invalid JSON** and clients fail to parse it. Behind auth + one endpoint, so limited blast radius — but a real correctness bug.

## Evidence
`daemon/internal/server/handlers_rename.go:62`
```go
http.Error(w, `{"error":"`+err.Error()+`"}`, http.StatusBadRequest)
```

## Suggested fix
Build the body with `json.Marshal` (or the existing `httpJSONErr` helper) so the message is always escaped.

---

vbuenog:
## Summary

The `/admin/repo-rename` validation-error branch built its 400 response with `` `{"error":"`+err.Error()+`"}` `` — concatenating the error text straight into a JSON string literal with no escaping (`handlers_rename.go:62`). If the message ever carries a `"` or `\`, the body is **invalid JSON** and clients fail to parse it. Closes #552.

## How it works

Replaced the hand-built literal with the repo's existing `httpJSONErr` helper, which `json.Encode`s `{"error": msg}` and sets `Content-Type: application/json`. One production line + a regression test.

## Scope

In scope:
- The one unescaped concatenation (`handlers_rename.go:62`) → `httpJSONErr`.
- Regression test asserting the 400 body is valid JSON when the error carries a quote/backslash.

Out of scope:
- The sibling `http.Error(w, `{"error":"…static…"}`, …)` calls in this handler. Their bodies are **static literals** (no user input), so they're already valid JSON and not a correctness risk. They do share the `text/plain` Content-Type quirk, but normalizing them is a cosmetic sweep, not a fix — left out to keep this change minimal.

Pure correctness hardening on one authenticated admin 400 branch.

- **Runtime behavior change:** for the rename 400 "invalid repo slug" response only — body is now always valid JSON (message escaped), `Content-Type` flips `text/plain` → `application/json`, and a trailing `\n` is added by the encoder. For today's errors (all static, no special chars) the `{"error":…}` payload is otherwise byte-identical.
- **Rollout failure modes:** none — no startup invariant, secret, config, or permission touched; can't crashloop. Only client-visible change is the corrected content-type + guaranteed-parseable body (both improvements).
- **Blast radius:** one branch of one auth-gated endpoint, hit only on an invalid-slug rename. Tiny.
- **What to watch:** nothing alertable. If operator tooling scripts `/admin/repo-rename` 400s, confirm it parses the now-standard `application/json` `{"error":…}` body (it should). Rarely-exercised path; no metric tie-in.
- **Rollback & sequencing:** trivially revertible (one line); no migration or coordination.

**Honest note:** the cited consequence (malformed JSON from a user slug) is currently **latent** — every error reaching this branch is a static `"invalid repo slug: …"` string with no special chars, so the present response is well-formed. This removes the fragile pattern and future-proofs it.

## Evidence

The regression test fails on the pre-fix code and passes after:

<details>
<summary>Before the fix — body is invalid JSON</summary>

```
--- FAIL: TestHandleAdminRepoRename_ValidationError400BodyIsValidJSON (0.00s)
    handlers_rename_test.go:115: response body is not valid JSON: invalid character 'o' after object key:value pair (body="{\"error\":\"rename: invalid repo slug: bad \"owner\"\\name\"}\n")
```
</details>

<details>
<summary>After the fix — valid JSON, escaped message, correct content-type</summary>

```
--- PASS: TestHandleAdminRepoRename_ValidationError400BodyIsValidJSON (0.00s)
ok  	github.com/heimdallm/daemon/internal/server
```
</details>

## Test plan

- [x] `go test ./internal/server/ -count=1` passes (full package).
- [x] New regression test fails on pre-fix code, passes after the fix.
- [x] `go vet ./internal/server/` clean; `gofmt -l` clean on changed files.

Closes #552

### DIFF
--- a/daemon/internal/server/handlers_rename.go
+++ b/daemon/internal/server/handlers_rename.go
@@ -59,7 +59,10 @@ func (srv *Server) handleAdminRepoRename(w http.ResponseWriter, r *http.Request)
 		// we use a simple substring match — no need to import the
 		// rename package here for one sentinel check.
 		if strings.Contains(err.Error(), "invalid repo slug") {
-			http.Error(w, `{"error":"`+err.Error()+`"}`, http.StatusBadRequest)
+			// Use httpJSONErr (not a hand-built literal) so an error message
+			// carrying a quote or backslash — e.g. a slug echoed back — is
+			// JSON-escaped and the body stays parseable. See #552.
+			httpJSONErr(w, http.StatusBadRequest, err.Error())
 			return
 		}
 		http.Error(w, `{"error":"rename failed"}`, http.StatusInternalServerError)
--- a/daemon/internal/server/handlers_rename.go
+++ b/daemon/internal/server/handlers_rename.go
@@ -29,18 +29,18 @@ type repoRenameRequest struct {
 // no-op at the store layer and silently returns 200.
 func (srv *Server) handleAdminRepoRename(w http.ResponseWriter, r *http.Request) {
 	if srv.repoRenameFn == nil {
-		http.Error(w, `{"error":"rename trigger not available"}`, http.StatusServiceUnavailable)
+		httpJSONErr(w, http.StatusServiceUnavailable, "rename trigger not available")
 		return
 	}
 	var req repoRenameRequest
 	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
-		http.Error(w, `{"error":"invalid JSON body"}`, http.StatusBadRequest)
+		httpJSONErr(w, http.StatusBadRequest, "invalid JSON body")
 		return
 	}
 	req.OldRepo = strings.TrimSpace(req.OldRepo)
 	req.NewRepo = strings.TrimSpace(req.NewRepo)
 	if req.OldRepo == "" || req.NewRepo == "" {
-		http.Error(w, `{"error":"old_repo and new_repo are required"}`, http.StatusBadRequest)
+		httpJSONErr(w, http.StatusBadRequest, "old_repo and new_repo are required")
 		return
 	}
 
@@ -50,7 +50,7 @@ func (srv *Server) handleAdminRepoRename(w http.ResponseWriter, r *http.Request)
 		slog.Error("POST /admin/repo-rename failed",
 			"old_repo", req.OldRepo, "new_repo", req.NewRepo, "err", err)
 		if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
-			http.Error(w, `{"error":"rename timed out"}`, http.StatusGatewayTimeout)
+			httpJSONErr(w, http.StatusGatewayTimeout, "rename timed out")
 			return
 		}
 		// Validation errors (empty/identical/malformed slugs) surface
@@ -59,13 +59,12 @@ func (srv *Server) handleAdminRepoRename(w http.ResponseWriter, r *http.Request)
 		// we use a simple substring match — no need to import the
 		// rename package here for one sentinel check.
 		if strings.Contains(err.Error(), "invalid repo slug") {
-			// Use httpJSONErr (not a hand-built literal) so an error message
-			// carrying a quote or backslash — e.g. a slug echoed back — is
-			// JSON-escaped and the body stays parseable. See #552.
+			// httpJSONErr escapes the message so a slug with a quote/backslash
+			// can't break the JSON body. See #552.
 			httpJSONErr(w, http.StatusBadRequest, err.Error())
 			return
 		}
-		http.Error(w, `{"error":"rename failed"}`, http.StatusInternalServerError)
+		httpJSONErr(w, http.StatusInternalServerError, "rename failed")
 		return
 	}
 	writeJSON(w, http.StatusOK, map[string]string{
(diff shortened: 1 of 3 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Small, well-scoped bug fix that replaces a hand-built `{"error":"..."}` JSON literal with the `httpJSONErr` helper so error messages containing quotes or backslashes are properly JSON-escaped, keeping the 400 response body parseable (closes #552). The change is a correct root-cause fix, follows the existing helper convention, and ships with a focused regression test that asserts valid JSON, the exact error field, and the Content-Type header. No confirmed issues found.

### Suggestions

- Info-leak (pre-existing, not introduced by this PR): the 400 branch echoes the raw `err.Error()` back to the client, which can surface internal error text/structure. Out of scope here since the prior literal did the same, but consider returning a sanitized/constant validation message and logging the full error server-side.
- The regression test exercises the round-trip JSON-escaping well; an even tighter assertion would be a table-driven case covering additional adversarial payloads (lone backslash, control chars, unicode) to lock in `httpJSONErr` behavior across edge inputs.
- Consider asserting the helper also sets the expected `Content-Type` independently of the 400 path (e.g. a shared helper test) so the contract is pinned in one place rather than per-handler.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

El PR reemplaza un literal JSON construido a mano por el helper httpJSONErr en la rama de error 400 de repo-rename, garantizando que mensajes con comillas o backslashes queden bien escapados y el body siga siendo JSON parseable. Verifiqué que httpJSONErr (handlers.go:1858) usa json.NewEncoder().Encode (escapa correctamente) y fija Content-Type: application/json. El test de regresión ejercita exactamente el caso (comilla + backslash en el mensaje), valida que el body parsea, que el campo error coincide y que el Content-Type es JSON. Cambio mínimo, correcto y bien testeado.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-scoped fix for a real JSON-injection bug in the 400 error path. The `httpJSONErr` helper properly escapes the error message via `json.Encoder`, and the regression test exercises the exact failure mode (embedded quotes and backslashes). No correctness or security issues in the change itself.

### Issues

🟡 **daemon/internal/server/handlers_rename.go:65** — The adjacent 500 branch (`http.Error(w, `{"error":"rename failed"}`, http.StatusInternalServerError)`) and all other error returns in this handler still use `http.Error`, which sets `Content-Type: text/plain`. After this PR, the 400 validation path returns `application/json` via `httpJSONErr` while every other error path returns `text/plain` with a JSON body. Consider converting the remaining static-string error returns to `httpJSONErr` for consistent Content-Type headers — callers checking Content-Type will otherwise see mixed behavior from the same endpoint.
🟡 **daemon/internal/server/handlers_rename.go:62** — The three-line comment explaining why `httpJSONErr` is used is heavier than necessary. The function name is self-documenting and the `See #552` reference alone would suffice. Minor style nit.

### Suggestions

- Convert the remaining `http.Error` calls in this handler (lines 32, 37, 43, 53, 65) to `httpJSONErr` for consistent `application/json` Content-Type across all error responses from the same endpoint.
- Consider whether the `strings.Contains(err.Error(), "invalid repo slug")` sentinel check (pre-existing, not introduced by this PR) should be replaced by a typed error or `errors.Is` check to avoid false positives from unrelated errors whose messages happen to contain that substring.

---

vbuenog:
Thanks @Muriano @sergiotejon @ivanmunozruiz — feedback addressed in `3311f15`.

**Implemented:**

- **Content-Type inconsistency** (sergio, Issue 🟡 + Suggestion) — good catch: converting only the 400 branch left the endpoint returning `application/json` from one path and `text/plain` (with JSON bodies) from the others. Converted **all** remaining `http.Error` literals in the handler (503/400/400/504/500) to `httpJSONErr`, so the whole endpoint is uniformly `application/json` and the hand-built literal pattern is gone.
- **Verbose comment** (sergio, Issue 🟡) — condensed the 400-branch comment to one line.
- **Table-driven adversarial test** (Muriano) — the 400 regression test now covers quote+backslash, lone backslash, control chars (tab/newline), a null byte, and unicode.
- **Pin the helper contract centrally** (Muriano) — added `TestHTTPJSONErr_EscapesAndSetsContentType` (internal `package server`) asserting escaping + `Content-Type` across the same adversarial inputs, so the contract is pinned once rather than per-handler.

**Declined (with reasoning):**

- **Sanitize the echoed `err.Error()` / info-leak** (Muriano) — leaving as-is. The endpoint is admin-auth-gated, the echoed text is a *static validation reason* (`expected owner/name shape`, `old and new are identical`), not secret or internal structure, and the full error is already `slog.Error`-logged server-side. Returning a constant message would remove the operator's only signal about *what* was wrong with their input, for little security gain. Happy to file a follow-up if the team wants sanitized validation messages as a policy.
- **Replace the `strings.Contains(err.Error(), "invalid repo slug")` sentinel with `errors.Is`** (sergio) — pre-existing and a deliberate, code-documented choice ("no need to import the rename package here for one sentinel check"). It's also genuinely awkward to type: the reconciler *and* the store each define their own `ErrInvalidRepoSlug`, so `errors.Is` would need both packages imported. Out of scope for this fix; can be a follow-up if we want to tighten it.

Verified: full `internal/server` package passes (`-count=1`); `go vet` + `gofmt` clean. The conversion is behavior-preserving except the intended Content-Type normalization to `application/json`.

PR #575, opened 2026-06-23
fix(server): make PUT /config persist atomically (closes #565)

### LINKED ISSUE #550: fix(server): PUT /config returns 200 OK even when SetConfig persist fails [MEDIUM]

### LINKED ISSUE #565: PUT /config persists keys one-by-one (best-effort), not atomically — partial state after a 500
## Background
Follow-up from PR #564 (which closed #550). #564 made `PUT /config` honest: a failed `SetConfig` now returns `500 {error, failed_keys}` instead of a misleading `200`.

It intentionally did **not** make the multi-key write transactional — this issue tracks that.

## Problem
`handlePutConfig` writes each key with its own `SetConfig` call inside a loop. If a later key fails, earlier keys in the same request may already be persisted. So after a `500`:

- The store is left in a **partial state** (some keys applied, some not).
- The client is told only **which keys failed** (`failed_keys`) — it gets no signal about which keys **did** persist, so it can't cleanly reconcile. (Raised by @alepalma91 on #564.)

## Suggested directions (to evaluate)
- Wrap the per-key writes in a single store transaction so the save is all-or-nothing, rolling back on any failure.
- If full atomicity is impractical, document the contract explicitly (error body states persistence is best-effort per key) and define the recovery path — currently "retry the whole payload" — so clients know how to reconcile.

## Out of scope
Not a regression; #564 is a strict improvement over the previous silent `200`. This is a hardening follow-up.

Refs #550, #564

vbuenog:
## Summary
Follow-up hardening from #564 (which closed #550). `PUT /config` wrote each key with its own `SetConfig` call inside a loop, so if a later key failed, earlier keys in the same request were already persisted. After a `500` the store was left in a **partial state** and the client — told only `failed_keys` — had no clean way to reconcile (raised by @alepalma91 on #564). This makes the multi-key write **all-or-nothing**.

## How it works
- **New `Store.SetConfigs(map[string]string) error`** (`daemon/internal/store/store.go`): upserts every key inside a single SQLite transaction (`BEGIN` → N `INSERT … ON CONFLICT DO UPDATE` → `COMMIT`). Any error rolls back the whole batch via a `defer tx.Rollback()`, so the store is never left partially written. An empty map is a no-op.
- **`handlePutConfig`** (`daemon/internal/server/handlers.go`): the validation loop now builds a `key→value` map and calls `SetConfigs` once instead of writing per-key. On failure **no key persists**, so the handler returns `500 {error, failed_keys}` where `failed_keys` is the **full attempted set** (all sorted) — now semantically correct, since none of them landed. The recovery path is "retry the whole payload", and the error message says so explicitly.
- The single-key `Store.SetConfig` is unchanged; the discovery path in `main.go` still uses it for its independent single-key writes.

## Scope
**In scope**
- Atomic, transactional multi-key persistence for `PUT /config` (the direction the issue preferred).
- `failed_keys` now reports the whole attempted set (none persisted) instead of a best-effort subset.
- Store-level tests for the atomic happy path, empty-map no-op, and write-failure surfacing; updated the existing handler 500 test's rationale.

**Out of scope**
- No change to the single-key `SetConfig` API or its callers.
- No HTTP contract reshaping beyond the `error` string and `failed_keys` semantics (response *shape* is unchanged: `{error, failed_keys}` on 500, `{status:ok}` on 200).

- **Runtime behavior change**: multi-key saves now run in one transaction (one `COMMIT` instead of N autocommits — marginally fewer fsyncs). Success path is observably identical (`200 {status:ok}`). On a persist failure the batch **rolls back**, so earlier keys no longer leak — the substantive behavior change. The `500` body keeps its shape; the `error` text and `failed_keys` meaning (now the whole set) changed. No new outbound calls, startup invariants, or resource usage (tx is short-lived, bounded by the existing 1 MB body limit).
- **Rollout failure modes**: cannot crashloop — no new config/secret/permission, no schema change (`configs` table already exists), no migration. Success path is byte-identical for existing clients. The Flutter client treats any non-200 as an error (`api_client.dart`), so the changed error string/`failed_keys` semantics don't affect it. No deploy-ordering constraint.
- **Blast radius**: `PUT /config` only (config save from web UI / CLI), and only on requests where a DB write genuinely fails — normally zero. No other endpoint or the single-key `SetConfig` path is touched.
- **What to watch**: post-deploy, the new `config: set (atomic)` `slog.Error` line (it replaces the old `config: set` + `config: partial persist failure` lines) fires on any persist failure — a spike means real DB trouble (disk full / lock / permission). `500` rate on `PUT /config` should stay ~0; `200` rate on saves unchanged.
- **Rollback & sequencing**: fully revertible — pure code change, no schema/migration/flag/ordering. Reverting restores the per-key best-effort loop.

## Evidence
<details><summary>New store tests: atomic persist, no-op, failure surfacing</summary>

```
=== RUN   TestSetConfigs_PersistsAllKeysAtomically
--- PASS: TestSetConfigs_PersistsAllKeysAtomically (0.00s)
=== RUN   TestSetConfigs_EmptyMapIsNoOp
--- PASS: TestSetConfigs_EmptyMapIsNoOp (0.00s)
=== RUN   TestSetConfigs_SurfacesWriteFailure
--- PASS: TestSetConfigs_SurfacesWriteFailure (0.00s)
ok  	github.com/heimdallm/daemon/internal/store
```
</details>

<details><summary>Existing PUT /config contract still holds (200 happy path, 500 + sorted failed_keys)</summary>

```
=== RUN   TestHandlerPutConfig
--- PASS: TestHandlerPutConfig (0.01s)
=== RUN   TestHandlerPutConfig_PersistFailureReturns500
--- PASS: TestHandlerPutConfig_PersistFailureReturns500 (0.00s)
ok  	github.com/heimdallm/daemon/internal/server
```
</details>

## Test plan
- [x] `go vet ./internal/server/ ./internal/store/`
- [x] `go test ./internal/server/ ./internal/store/` (new + existing tests green)
- [x] `make verify-linux` (full Linux build verification passed)
- [ ] Reviewer: confirm all-or-nothing semantics + the `failed_keys`=full-set meaning is the desired contract (vs. documenting best-effort).

Closes #565

### DIFF
--- a/daemon/internal/server/handlers.go
+++ b/daemon/internal/server/handlers.go
@@ -644,8 +644,8 @@ var validConfigKeys = map[string]struct{}{
 // render them) but that PUT /config must not persist. The SvelteKit page
 // round-trips the entire GET payload as a forward-compat safeguard, so
 // rejecting these would 400 every save. We accept them at the endpoint
-// boundary and drop them before SetConfig — still a strict allowlist, no
-// arbitrary keys permitted.
+// boundary and drop them before the persist batch — still a strict allowlist,
+// no arbitrary keys permitted.
 //
 // Why each key is here:
 //   - repositories  : repo-list changes belong in config.toml via PATCH
@@ -787,15 +787,13 @@ func (srv *Server) handlePutConfig(w http.ResponseWriter, r *http.Request) {
 		body["agent_configs"] = normalized
 	}
 
-	var failedKeys []string
-	var attempted int
+	kv := make(map[string]string, len(body))
 	for k, v := range body {
 		// Read-only keys were accepted above to avoid 400s on UI saves that
 		// round-trip the GET payload, but they must not land in the store.
 		if _, readOnly := readOnlyConfigKeys[k]; readOnly {
 			continue
 		}
-		attempted++
 		var val string
 		switch typed := v.(type) {
 		case string:
@@ -811,20 +809,23 @@ func (srv *Server) handlePutConfig(w http.ResponseWriter, r *http.Request) {
 			}
 			val = string(b)
 		}
-		if _, err := srv.store.SetConfig(k, val); err != nil {
-			slog.Error("config: set", "key", k, "err", err)
-			failedKeys = append(failedKeys, k)
-		}
+		kv[k] = val
 	}
+	// Persist all keys in one transaction: the save is all-or-nothing (#565).
 	// A swallowed persist error here would 200-OK a save that never hit disk,
 	// so the UI shows success while the next reload silently re-applies stale
-	// values (#550). Report 500 with the offending keys instead. Note this is
-	// not transactional: earlier keys in the loop may already be persisted.
-	if len(failedKeys) > 0 {
+	// values (#550). On failure SetConfigs rolls back, so NO key is persisted —
+	// we report 500 with the full attempted key set (none of them landed), and
+	// the client recovers by simply retrying the whole payload.
+	if err := srv.store.SetConfigs(kv); err != nil {
+		failedKeys := make([]string, 0, len(kv))
+		for k := range kv {
+			failedKeys = append(failedKeys, k)
+		}
 		sort.Strings(failedKeys) // deterministic: Go map iteration order is random
-		slog.Warn("config: partial persist failure", "failed", len(failedKeys), "attempted", attempted)
+		slog.Error("config: set (atomic)", "keys", len(failedKeys), "err", err)
 		writeJSON(w, http.StatusInternalServerError, map[string]any{
-			"error":       "failed to persist one or more config keys",
+			"error":       "failed to persist config; no keys were saved (transaction rolled back)",
 			"failed_keys": failedKeys,
 		})
 		return
--- a/daemon/internal/server/handlers_test.go
+++ b/daemon/internal/server/handlers_test.go
@@ -327,16 +327,15 @@ func TestHandlerPutConfig(t *testing.T) {
 	}
 }
 
-// TestHandlerPutConfig_PersistFailureReturns500 guards #550: a failed
-// SetConfig must surface as 500 with the offending keys, not a misleading
-// 200 OK that makes the UI believe a never-persisted save succeeded.
+// TestHandlerPutConfig_PersistFailureReturns500 guards #550 + #565: a failed
+// persist must surface as 500 with the offending keys, not a misleading 200 OK
+// that makes the UI believe a never-persisted save succeeded.
 //
-// The multi-key payload also locks the contract that the loop accumulates
-// EVERY failed key (not just the first) and returns them sorted, so the
-// response is deterministic regardless of Go's random map iteration order.
-// Forcing only a subset to fail would require a store seam to inject per-key
-// errors; the store is a concrete *store.Store, so we close it to fail all
-// writes and assert the full sorted set instead.
+// The write is now atomic (#565): handlePutConfig calls SetConfigs once and the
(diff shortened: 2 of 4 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, focused PR that correctly converts per-key SetConfig calls in a loop to a single transactional SetConfigs batch. The implementation is sound: BEGIN/deferred-Rollback/Commit pattern is correct, the handler properly builds the kv map before persisting, and test coverage is adequate. No correctness or security issues found.

### Issues

🟡 **daemon/internal/store/store.go:391** — Minor efficiency: tx.Exec is called with the same SQL string for every key in the loop. Using tx.Prepare once and then stmt.Exec per key would avoid re-parsing the statement on each iteration. Negligible for the small number of config keys in practice.
🟡 **daemon/internal/server/handlers.go:818** — The response field "failed_keys" now always contains every key in the batch (since rollback means none landed). The field name is slightly misleading — "attempted_keys" would be more precise given the atomic semantics — but the error message text ('no keys were saved (transaction rolled back)') clarifies intent, and renaming would break any client parsing the 500 body.

### Suggestions

- Consider using tx.Prepare inside SetConfigs to avoid re-parsing the upsert statement on each iteration, though this is only a micro-optimization for the current config key cardinality.
- The old SetConfig (singular) method is still public. If no other callers need it, consider marking it deprecated or removing it to funnel all config writes through the atomic path and prevent accidental non-transactional use.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Solid, root-cause fix for #565. Introduces store.SetConfigs which upserts all config keys in a single transaction with deferred rollback, and rewires handlePutConfig to call it once so PUT /config is all-or-nothing. Parameterized SQL, deterministic sorted failed-key reporting, updated and accurate comments, and good test coverage (happy path, empty no-op, write-failure surfacing). No confirmed correctness, security, or concurrency defects in the diff.

### Suggestions

- TestSetConfigs_PersistsAllKeysAtomically and TestSetConfigs_SurfacesWriteFailure verify happy path and total failure, but no test exercises a true partial-failure rollback (one key valid, another forcing an error mid-batch, then asserting NONE persisted). The PR correctly argues atomicity is structural (single BEGIN/COMMIT), but a directed rollback test would harden the contract against future refactors that might split the transaction.
- SetConfigs iterates the map for tx.Exec inside the loop; for the small, bounded config key set this is fine, but if the payload could ever grow large, consider preparing the statement once (tx.Prepare) and reusing it to avoid re-parsing the SQL per key.
- The removed 'attempted' counter slightly reduces observability: the previous log distinguished attempted vs failed. The new error log reports only len(failedKeys) (which now always equals the full attempted set on failure) — acceptable given atomicity, but consider logging the underlying offending key (already embedded in the wrapped err) at a structured field for faster prod debugging.

---

PR #576, opened 2026-06-23
feat(flutter): in-tab connection indicator for events_tab SSE drops (closes #572)

### LINKED ISSUE #572: feat(flutter): in-tab connection indicator for events_tab SSE drops
## Context

Follow-up from PR #571 (fix for #554), raised in review by @sergiotejon.

PR #571 made `SseClient` forward transport errors to listeners. Of the three consumers, `events_tab` now swallows them deliberately (`onError: (_) {}` at `flutter_app/lib/features/server/widgets/events_tab.dart`), relying on the **global** `DaemonConnectionNotifier` indicator for offline state.

## Gap

`events_tab` creates its **own** `SseClient` instance (`path: '/events'`), separate from the shared `sseStreamProvider` that drives the global indicator. So there's a window where *this tab's* connection drops while the global indicator still shows "connected" — the events tab gives no in-tab signal that its live stream has stalled.

## Proposal

Mirror `logs_screen.dart`'s pattern: track a local `_connected` flag (flip on `onError`/`onDone`, restore when events resume) and surface a small in-tab "reconnecting"/offline banner.

## Scope / priority

LOW. The stream auto-reconnects and the global indicator covers the common case (whole-daemon outage). This only matters for a drop isolated to the events-tab connection. Non-blocking; deferred from #571 to keep that fix minimal.

vbuenog:
## Summary

The Server > Events tab creates its **own** `SseClient` (`path: '/events'`), separate from the shared `sseStreamProvider` that drives the global `DaemonConnectionNotifier` indicator. After PR #571 made `SseClient` forward transport errors to listeners, `events_tab` deliberately swallowed them (`onError: (_) {}`), relying entirely on the global indicator for offline state. That leaves a gap: a drop isolated to *this tab's* connection stalls the live stream while the global indicator still reads "connected", with no in-tab signal. This PR closes that gap by giving the tab its own connection indicator. Follow-up from #571 (fix for #554), raised in review by @sergiotejon.

## How it works

- **`events_tab.dart`** — adds a local `bool _connected` (optimistically `true`, since under normal operation no events arrive for up to a 60 s polling cycle). It flips to `false` on the stream's `onError`/`onDone`, and is restored to `true` in `_onEvent` when an event arrives (the only signal the auto-reconnecting `SseClient` gives that the stream is live again — it emits no explicit "connected" event). A small banner renders below the toolbar while `!_connected`.
- **`connection_status_banner.dart`** (new) — a thin stateless amber "Live stream disconnected — reconnecting…" banner with a spinner. Extracted as its own widget (mirroring the existing `CircuitBreakerBanner`) so it is unit-testable without driving a live SSE stream.
- This mirrors `logs_screen.dart`'s `_connected` pattern, and goes one step further: `logs_screen` only sets `_connected = true` at initial connect and never restores it after a reconnect, whereas this tab restores it when events resume.

## Scope

In scope:
- In-tab connection indicator for the Events tab's own `/events` SSE connection.

Out of scope:
- The global `DaemonConnectionNotifier` indicator (unchanged — still covers whole-daemon outages).
- `logs_screen.dart`'s latent "never restores after reconnect" behavior (left as-is; not part of this issue).
- Any change to `SseClient` reconnect semantics.

Low impact, UI-only, desktop GUI client — no daemon/server/API change.

- **Runtime behavior change:** adds a local boolean and a conditional banner widget in one tab. No new network calls (reuses the existing `SseClient`), no API/contract change. The `onError` callback that #571 left as a no-op now drives UI state instead.
- **Rollout failure modes:** none of the server-side kind — it ships inside the desktop app, can't crashloop, needs no config/secret. Optimistic initial state means no banner flash on startup before the first event.
- **Blast radius:** only the Server > Events tab; no effect on the daemon, other tabs, the web build, or the global indicator.
- **What to watch:** nothing server-side / no metrics or alerts. Post-release visual check only — banner appears on a forced `/events` disconnect and clears when events resume.
- **Rollback & sequencing:** trivially revertible; no migration, no persisted state, no deploy ordering.

## Test plan

- [x] `make verify-linux` from the worktree (daemon `go test`, `flutter pub get`, `flutter test`, Linux build) — passed (exit 0, "✅ Linux build verification passed").
- [x] Added `connection_status_banner_test.dart` widget test (asserts the reconnecting message + spinner render), mirroring `circuit_breaker_banner_test.dart`.
- [ ] `flutter analyze` — not run locally (no Flutter toolchain on host; `verify-linux` covers `flutter test` but not analyze); CI runs it.
- [ ] Reviewer: manually confirm the banner appears on an isolated `/events` drop and clears on reconnect, while the global indicator behaves as before.

Closes #572
EOF

### DIFF
--- a/flutter_app/lib/features/server/widgets/connection_status_banner.dart
+++ b/flutter_app/lib/features/server/widgets/connection_status_banner.dart
@@ -1,0 +1,42 @@
+import 'package:flutter/material.dart';
+
+/// Thin in-tab banner shown when the Events tab's own SSE stream has dropped.
+///
+/// The tab owns a separate [SseClient] from the shared stream that drives the
+/// global connection indicator, so a drop isolated to this connection would
+/// otherwise leave the tab silently stalled while the global indicator still
+/// reads "connected" (#572). The client auto-reconnects, hence the
+/// "reconnecting" wording rather than a hard "offline".
+class ConnectionStatusBanner extends StatelessWidget {
+  const ConnectionStatusBanner({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    const amber = Color(0xFFFFB347);
+    return Container(
+      width: double.infinity,
+      padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
+      color: amber.withValues(alpha: 0.15),
+      child: Row(
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          const SizedBox(
+            width: 12,
+            height: 12,
+            child: CircularProgressIndicator(strokeWidth: 2, color: amber),
+          ),
+          const SizedBox(width: 8),
+          Flexible(
+            child: Text(
+              'Live stream disconnected — reconnecting…',
+              style: TextStyle(
+                fontSize: 12,
+                color: Theme.of(context).colorScheme.onSurface,
+              ),
+            ),
+          ),
+        ],
+      ),
+    );
+  }
+}
--- a/flutter_app/lib/features/server/widgets/connection_status_banner.dart
+++ b/flutter_app/lib/features/server/widgets/connection_status_banner.dart
@@ -13,29 +13,34 @@ class ConnectionStatusBanner extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     const amber = Color(0xFFFFB347);
-    return Container(
-      width: double.infinity,
-      padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
-      color: amber.withValues(alpha: 0.15),
-      child: Row(
-        mainAxisSize: MainAxisSize.min,
-        children: [
-          const SizedBox(
-            width: 12,
-            height: 12,
-            child: CircularProgressIndicator(strokeWidth: 2, color: amber),
-          ),
-          const SizedBox(width: 8),
-          Flexible(
-            child: Text(
-              'Live stream disconnected — reconnecting…',
-              style: TextStyle(
-                fontSize: 12,
-                color: Theme.of(context).colorScheme.onSurface,
+    // liveRegion so assistive tech announces the drop/recovery as it toggles.
+    return Semantics(
+      liveRegion: true,
+      label: 'Live stream disconnected, reconnecting',
+      child: Container(
+        width: double.infinity,
+        padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
+        color: amber.withValues(alpha: 0.15),
+        child: Row(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            const SizedBox(
+              width: 12,
+              height: 12,
+              child: CircularProgressIndicator(strokeWidth: 2, color: amber),
+            ),
+            const SizedBox(width: 8),
+            Flexible(
+              child: Text(
+                'Live stream disconnected — reconnecting…',
+                style: TextStyle(
+                  fontSize: 12,
+                  color: Theme.of(context).colorScheme.onSurface,
+                ),
               ),
             ),
-          ),
-        ],
+          ],
+        ),
       ),
     );
   }
--- a/flutter_app/lib/features/server/widgets/events_tab.dart
+++ b/flutter_app/lib/features/server/widgets/events_tab.dart
@@ -7,6 +7,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import '../../../core/api/sse_client.dart';
 import '../../../core/platform/platform_services_provider.dart';
 import '../event_summary.dart';
+import 'connection_status_banner.dart';
 import 'event_row.dart';
 
 /// Server > Events tab — operational dashboard of live SSE events.
@@ -38,6 +39,11 @@ class _EventsTabState extends ConsumerState<EventsTab> {
   final _expanded = <int>{};
   int _nextRowId = 0;
   bool _autoScroll = true;
+  // This tab's own SSE connection health. The shared stream that drives the
+  // global indicator is separate, so a drop isolated to this connection needs
+  // its own in-tab signal (#572). Optimistic on connect — no events arrive for
+  // up to a polling cycle (60 s) under normal operation.
+  bool _connected = true;
   final Set<String> _enabledGroups = {'pr', 'issue', 'polling', 'state', 'circuit_breaker'};
   String _searchQuery = '';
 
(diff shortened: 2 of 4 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Clean, well-scoped Flutter change that adds an in-tab SSE connection banner for the Events tab, addressing the gap where a drop isolated to the tab's own SSE client left it silently stalled (#572). The widget is well-documented and the state logic (flip flag on error/done, restore on event) is correct and idempotent via _setConnected's guard. Main gaps are around test coverage of the actual feature behavior and the onDone handling assumption.

### Issues

⚠️ **flutter_app/test/features/server/widgets/connection_status_banner_test.dart:0** — Tests only cover the static rendering of ConnectionStatusBanner (message + spinner present). The actual feature logic introduced in events_tab.dart — banner appearing on onError/onDone and clearing when _onEvent fires again — is untested, which is the core behavior of #572. Impact: regressions in the connect/disconnect state transitions (e.g. banner not clearing, or stuck visible) would not be caught. Recommended fix: add a widget/unit test for EventsTab driving a controllable SSE stream (emit error → assert banner shown; emit event → assert banner cleared).

### Suggestions

- onDone always calls _setConnected(false) implying 'reconnecting…', but if SseClient.connect() can emit onDone on a permanent close (rather than reconnecting internally), the banner would remain shown indefinitely with no recovery path since no further _onEvent would fire. Verify SseClient's reconnect semantics; if done means terminal, consider distinct wording/handling for that case.
- The amber color (0xFFFFB347) and its alpha are hardcoded in connection_status_banner.dart while the text color uses Theme.of(context).colorScheme. Consider sourcing the accent from the theme (or a shared color constant) for consistency and dark-mode correctness.
- Consider wrapping the banner in a Semantics widget (or using a liveRegion) so screen readers announce the disconnected/reconnecting state.
- In _onEvent, _connected = true is assigned directly before setState while _setConnected centralizes the guarded transition; routing both paths through a single helper would keep the state-mutation pattern consistent.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, focused PR that adds an in-tab reconnection banner for the Events tab's own SSE stream. One real bug: `_connected = true` in `_onEvent` bypasses the `_setConnected` helper and its `mounted` guard, risking a `setState() called after dispose()` crash. Test coverage is minimal.

### Issues

⚠️ **flutter_app/lib/features/server/widgets/events_tab.dart:97** — Bug: `_connected = true` is assigned directly instead of going through `_setConnected(true)`. The `_setConnected` helper was specifically written with a `mounted` guard to prevent `setState() called after dispose()`. If the stream delivers a final event after `dispose()` cancels the subscription (async race), this direct assignment followed by the unguarded `setState` on the next line will throw. Use `_setConnected(true)` and move it inside the `setState` closure, or add a `if (!mounted) return` early exit to `_onEvent`.
🟡 **flutter_app/lib/features/server/widgets/events_tab.dart:97** — Style: State mutation (`_connected = true`) is outside the `setState` closure. Flutter convention is to place all state mutations inside `setState(() { ... })` so that the framework and future readers can see what changed. Since `setState` is called immediately after, it works, but it's inconsistent with the `_setConnected` pattern used for the error/done paths.
🟡 **flutter_app/test/features/server/widgets/connection_status_banner_test.dart:1** — Test gap: Only the banner widget is tested in isolation. There are no tests verifying that EventsTab shows the banner when an SSE error occurs or hides it when a new event arrives. These are the core behaviors introduced by this PR.

### Suggestions

- Replace the direct `_connected = true` assignment in `_onEvent` with `_setConnected(true)` and add an early `if (!mounted) return;` guard at the top of `_onEvent` to protect the existing `setState` call as well.
- Add a widget test for EventsTab that simulates a stream error (banner appears) followed by a new event (banner disappears) to cover the integration between the state flag and the banner widget.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR de Flutter que añade un banner in-tab para indicar caída del stream SSE propio del Events tab (la conexión es independiente de la que alimenta el indicador global), con su widget, integración en events_tab y un test del widget. Cambio acotado, bien documentado y consistente con lo que dice el título. Sin issues bloqueantes.

### Issues

🟡 **flutter_app/lib/features/server/widgets/events_tab.dart:97** — En `_onEvent` se muta `_connected = true` fuera de `setState`. Funciona porque hay un `setState` inmediatamente después (mismo frame), pero acopla la corrección del flag a ese setState; si en el futuro se reordena o se hace early-return antes del setState, el banner no se limpiaría. Es un nit, no un bug actual.
🟡 **flutter_app/test/features/server/widgets/connection_status_banner_test.dart:1** — El test solo cubre el widget banner aislado. La lógica no trivial real (transiciones connected/disconnected vía onError/onDone/_onEvent y la condición `if (!_connected)` en el build de events_tab) no se ejercita en ningún test. No es bloqueante, pero la parte con riesgo de regresión queda sin cobertura.

### Suggestions

- Añadir un widget test sobre EventsTab que simule un onError/onDone del stream y verifique que aparece el ConnectionStatusBanner, y que un evento posterior lo oculta (cubre la lógica de estado, no solo el widget visual).
- Considerar mover `_connected = true` de `_onEvent` dentro del propio `setState` (o usar `_setConnected(true)`) para que la actualización del flag no dependa del orden con el setState que añade la fila.

---

vbuenog:
Thanks @sergiotejon and @Muriano — addressed in e224071. Item-by-item:

**✅ Implemented**

- **`_connected = true` bypassing the `mounted` guard in `_onEvent`** (both reviewers; @sergiotejon flagged as a bug). Added `if (!mounted) return;` at the top of `_onEvent` and moved `_connected = true` inside the `setState` closure, so a late event after `dispose()` can no longer trigger `setState()`-after-dispose, and the mutation now lives where Flutter convention expects.
- **Test gap — no coverage of the actual show-on-error / clear-on-event behaviour** (both reviewers). Added an injectable `SseClient` seam (`@visibleForTesting`) to `EventsTab` and `events_tab_test.dart`, which drives a controllable transport: error → banner appears, reconnect + event → banner clears. This is the core behaviour of #572.
- **Accessibility** (@Muriano). Wrapped the banner in `Semantics(liveRegion: true, …)` so assistive tech announces the disconnect/recovery as it toggles.

**↩️ Pushed back (no change)**

- **`onDone` could leave the banner stuck on a permanent close** (@Muriano). Verified against `sse_client.dart`: the listener's `onDone` fires *only* when `_controller.close()` runs, which happens exclusively in `disconnect()` (i.e. `dispose()`/`onCancel`). The internal stream's `onDone` schedules a reconnect and never closes the broadcast controller. So there's no path where the tab stays alive with a stuck banner — by the time `onDone` reaches the listener, the widget is being disposed. Kept the `onDone → _setConnected(false)` handler for parity with `logs_screen.dart`.

**🙅 Declined (with reasoning)**

- **Source the amber from the theme instead of hardcoding `0xFFFFB347`** (@Muriano). That hex is the repo's established WARN colour — `logs_screen.dart` uses the exact same value, and `circuit_breaker_banner.dart` hardcodes `Colors.red.shade50`. The banner's *text* already uses `Theme…onSurface`, so it adapts to dark mode while the accent stays consistent with the rest of the UI. Honest reason: convention consistency. Happy to revisit if you'd prefer a shared colour constant across both widgets as a separate cleanup.

`make verify-linux` (daemon tests + `flutter test` + Linux build) passes green with the new test.

PR #577, opened 2026-06-24
fix(github): deterministic same-second timeline order so re-requests win re-review (refs ai-platform-workspace#602)

vbuenog:
## Summary

Heimdallm intermittently failed to re-review a PR after the author resolved a *request changes* review and explicitly re-requested a review — exactly the symptom in freepik-company/ai-platform-workspace#602. Root cause: a non-deterministic ordering of timeline events when a dismiss and a re-request share the same whole-second timestamp, which made the re-review gate skip a legitimately requested re-review about half the time.

## How it works

The re-review gate `pipeline.shouldBypassSHASkipForReReview` (`daemon/internal/pipeline/pipeline.go`) proceeds with a re-review **only** when the single most-recent `review_requested`/`review_dismissed` timeline event for the bot is a `review_requested` newer than the previous review. It relies on `GetPRTimelineEventsForReviewer` (`daemon/internal/github/client.go`) returning those events sorted ascending and reads the **last** element.

Both GitHub's `created_at` and our stored review anchor are RFC3339 **second-granularity** (`store.go`: `sqliteTimeFormat = time.RFC3339`). When an author dismisses the blocking review and re-requests back-to-back (or automation does both), the two events collapse to the same second. The previous `sort.Slice` is **not stable**, so the dismiss landed last ≈50% of the time → the gate read "most recent event = dismiss" → silently skipped the re-review. Intermittent by construction.

Fix: `sort.SliceStable` with an explicit same-second tiebreak that orders `review_dismissed` **before** `review_requested`, so the explicit re-request ends up last and wins the tie. When the true sub-second order is unrecoverable, we honour the operator's intent to be re-reviewed.

Key files:
- `daemon/internal/github/client.go` — the ordering fix in `GetPRTimelineEventsForReviewer`.
- `daemon/internal/github/timeline_test.go` — new regression test (both raw input orderings).

## Scope

In scope:
- Deterministic ordering of bot timeline events on a same-second `review_dismissed`/`review_requested` tie.

Out of scope (deliberately unchanged — these are by-design fail-closed guards, not the reported bug):
- Fail-closed on timeline API error / pagination cap (`pipeline.go`, `client.go`).
- Tier-2 "ghost" skip when the Pulls API lags the Search index (`cmd/heimdallm/main.go`).
- Circuit-breaker per-PR/per-repo rate caps.
- Distinct-timestamp `dismiss-after-request` still correctly keeps the skip.

- **Runtime behavior change**: only when the bot's timeline has a `review_dismissed` and `review_requested` at the same whole-second — that tie now deterministically resolves to "proceed with re-review" instead of a coin-flip. Same single timeline fetch, no new outbound calls, no new resource use (`SliceStable` over a tiny per-bot filtered slice), no contract/response changes.
- **Rollout failure modes**: none. Pure in-process logic; no secret/config/permission/migration. Mixed-version pods during a rolling deploy are both safe.
- **Blast radius**: narrow — only PRs where the bot already reviewed *and* a dismiss+request collide in the same second. Outcome is one Claude re-review the operator explicitly asked for. The fail-closed posture for API errors, missing deps, and distinct-timestamp dismiss-after-request is unchanged, so it does **not** widen the surface for non-requested reviews.
- **What to watch**: the skip log `pipeline: skipping re-review, no explicit re-request` (`reason=sha_unchanged` / `no_rereview_request`) should **drop** for genuinely re-requested PRs (success signal), with a small matching rise in `explicit re-request detected — proceeding with review`. Watch circuit-breaker per-PR-24h counters for any uptick on hyperactive PRs (realistically negligible). Claude spend stays effectively flat.
- **Rollback & sequencing**: trivially revertible — single stateless function, no migration, no flag.

## Evidence

The new test reproduces the bug, then proves the fix (and order-independence — a non-stable sort passes one ordering and fails the other):

<details>
<summary>Before the fix (non-stable sort.Slice)</summary>

```
=== RUN   TestGetPRTimelineEventsForReviewer_SameSecondReRequestWins/request_then_dismiss_in_payload
    timeline_test.go:63: most-recent event = "review_dismissed", want review_requested to win the same-second tie
--- FAIL: .../request_then_dismiss_in_payload
--- PASS: .../dismiss_then_request_in_payload
```
</details>

<details>
<summary>After the fix (sort.SliceStable + tiebreak)</summary>

```
=== RUN   TestGetPRTimelineEventsForReviewer_SameSecondReRequestWins
--- PASS: .../request_then_dismiss_in_payload
--- PASS: .../dismiss_then_request_in_payload
ok  	github.com/heimdallm/daemon/internal/github
ok  	github.com/heimdallm/daemon/internal/pipeline
```
</details>

## Test plan

- [x] New regression test fails on the old code and passes on the new code (both raw orderings).
- [x] `go test ./internal/github/ ./internal/pipeline/` green (incl. existing `TestPipeline_Run_DismissAfterReRequestKeepsSkip` — distinct-timestamp dismiss-after-request still keeps the skip).
- [x] `go build ./...` and `go vet ./...` clean.
- [x] Pre-existing unrelated failures confirmed on base `main` too: `TestDetect_Fallback` (host has a real `codex` binary) and a flaky `cmd/heimdallm` timing test (passes on re-run).
- [ ] Reviewer: confirm the same-second tiebreak semantics (re-request wins) match the desired fail-closed/honour-intent policy.

Refs freepik-company/ai-platform-workspace#602

### DIFF
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -1019,10 +1019,13 @@ func (c *Client) GetPRTimelineEventsForReviewer(repo string, number int, login s
 			// about half the time. On a tie we order review_dismissed BEFORE
 			// review_requested — the explicit re-request wins and ends up last
 			// — because when we cannot recover the true sub-second order we
-			// honour the operator's intent to be re-reviewed.
+			// honour the operator's intent to be re-reviewed. The tiebreak is
+			// kept total (a strict weak ordering): two same-second events of
+			// the same type compare as equal so the stable sort preserves
+			// their input order, rather than reporting both i<j and j<i.
 			sort.SliceStable(out, func(i, j int) bool {
 				if out[i].CreatedAt.Equal(out[j].CreatedAt) {
-					return out[i].Event == "review_dismissed"
+					return out[i].Event == "review_dismissed" && out[j].Event == "review_requested"
 				}
 				return out[i].CreatedAt.Before(out[j].CreatedAt)
 			})
--- a/daemon/internal/github/client.go
+++ b/daemon/internal/github/client.go
@@ -1009,7 +1009,21 @@ func (c *Client) GetPRTimelineEventsForReviewer(repo string, number int, login s
 			// A single sort here is cheap (events is filtered down to just
 			// review_requested / review_dismissed for one login) and immunises
 			// the caller against an undocumented re-ordering on GitHub's side.
-			sort.Slice(out, func(i, j int) bool {
+			//
+			// Stable sort + an explicit same-second tiebreak (#602): both
+			// GitHub's created_at and our stored review anchor are RFC3339
+			// second-granularity, so a dismiss and a re-request done back to
+			// back (or by automation) collapse to the same timestamp. The
+			// caller decides on the single most-recent relevant event, so a
+			// non-deterministic order silently dropped legitimate re-reviews
+			// about half the time. On a tie we order review_dismissed BEFORE
+			// review_requested — the explicit re-request wins and ends up last
+			// — because when we cannot recover the true sub-second order we
+			// honour the operator's intent to be re-reviewed.
+			sort.SliceStable(out, func(i, j int) bool {
+				if out[i].CreatedAt.Equal(out[j].CreatedAt) {
+					return out[i].Event == "review_dismissed"
+				}
 				return out[i].CreatedAt.Before(out[j].CreatedAt)
 			})
 			return out, nil
--- a/daemon/internal/github/timeline_test.go
+++ b/daemon/internal/github/timeline_test.go
@@ -17,6 +17,7 @@ func timelineServer(t *testing.T, raw string) *httptest.Server {
 		if r.URL.Path != "/repos/org/repo/issues/7/timeline" {
 			t.Errorf("path = %q", r.URL.Path)
 		}
+		w.Header().Set("Content-Type", "application/json")
 		_, _ = w.Write([]byte(raw))
 	}))
 }
@@ -34,10 +35,20 @@ func timelineServer(t *testing.T, raw string) *httptest.Server {
 // Both raw input orderings are exercised to prove the result is
 // order-independent (a non-stable sort would pass one and fail the
 // other).
+// reqEventAt / disEventAt build raw timeline JSON for a review_requested /
+// review_dismissed event targeting heimdallm-bot at the given created_at.
+func reqEventAt(ts string) string {
+	return fmt.Sprintf(`{"event":"review_requested","created_at":%q,"actor":{"login":"alice"},"requested_reviewer":{"login":"heimdallm-bot"}}`, ts)
+}
+
+func disEventAt(ts string) string {
+	return fmt.Sprintf(`{"event":"review_dismissed","created_at":%q,"actor":{"login":"alice"},"dismissed_review":{"user":{"login":"heimdallm-bot"}}}`, ts)
+}
+
 func TestGetPRTimelineEventsForReviewer_SameSecondReRequestWins(t *testing.T) {
 	const ts = "2026-06-24T11:00:00Z"
-	reqEvent := fmt.Sprintf(`{"event":"review_requested","created_at":%q,"actor":{"login":"alice"},"requested_reviewer":{"login":"heimdallm-bot"}}`, ts)
-	disEvent := fmt.Sprintf(`{"event":"review_dismissed","created_at":%q,"actor":{"login":"alice"},"dismissed_review":{"user":{"login":"heimdallm-bot"}}}`, ts)
+	reqEvent := reqEventAt(ts)
+	disEvent := disEventAt(ts)
 
 	cases := []struct {
 		name string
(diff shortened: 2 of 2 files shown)


ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

El PR cambia sort.Slice por sort.SliceStable con un tiebreak explícito en GetPRTimelineEventsForReviewer para que, ante created_at de igual segundo, un review_requested gane al review_dismissed (quede último) y el caller no descarte re-reviews legítimos. Cambio correcto, bien documentado y con un test de regresión que ejercita ambos órdenes de entrada. LGTM.

### Issues

🟡 **daemon/internal/github/client.go:1020** — El comparador del tiebreak no es estrictamente consistente para dos eventos del mismo tipo con idéntico created_at: para dos 'review_dismissed' devuelve true en ambos sentidos (less(i,j) y less(j,i)), lo que viola la antisimetría que sort espera. Funcionalmente es benigno (ambos son dismiss y el caller solo mira el último evento relevante), pero contradice ligeramente el objetivo de orden determinista del PR. El patrón estrictamente correcto sería comparar ambos lados: `return out[i].Event == "review_dismissed" && out[j].Event == "review_requested"`.

### Suggestions

- Considera endurecer el comparador a `out[i].Event == "review_dismissed" && out[j].Event == "review_requested"` para que sea un less estrictamente válido también ante dos eventos del mismo tipo en el mismo segundo.
- Opcional: añadir un caso de test con dos review_dismissed del mismo segundo para documentar que ese orden no afecta al resultado del caller.

---

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Good fix for a real non-determinism bug, but the same-second tiebreaker comparator violates Go's sort contract (strict weak ordering) when two review_dismissed events share a timestamp, which can cause undefined behavior including panics.

### Issues

🔴 **daemon/internal/github/client.go:1021** — Sort comparator violates strict weak ordering. `return out[i].Event == "review_dismissed"` returns true for BOTH less(a,b) and less(b,a) when both events are review_dismissed with the same timestamp. Go's sort package requires a strict weak ordering (if less(a,b) then !less(b,a)). Violating this can cause undefined behavior — incorrect results, index-out-of-bounds panics, or infinite loops in pathological cases. Fix: `return out[i].Event == "review_dismissed" && out[j].Event != "review_dismissed"` — this returns false when both events are the same type (preserving stable-sort original order) and only returns true when dismissed precedes non-dismissed.
⚠️ **daemon/internal/github/timeline_test.go:39** — Test does not cover the case that exposes the comparator bug: two review_dismissed events (or two review_requested events) with the same timestamp. Adding a subtest with [dismissed, dismissed, requested] at the same second would catch the strict-weak-ordering violation and verify no panic.
🟡 **daemon/internal/github/timeline_test.go:18** — The test server does not set Content-Type: application/json on the response. While this works because the client presumably doesn't check the header, it's a minor fidelity gap that could mask issues if content-type validation is ever added.

### Suggestions

- Fix the tiebreaker to: `return out[i].Event == "review_dismissed" && out[j].Event != "review_dismissed"` — this preserves the desired dismissed-before-requested ordering while satisfying the sort contract when both events share the same type.
- Add a test case with duplicate event types at the same timestamp (e.g., [dismissed, dismissed, requested]) to guard against the sort contract violation and verify the slice doesn't panic.
- Consider extracting the event-type priority into a small helper (e.g., eventOrder(string) int) if more event types are added to the filter in the future — but this is not necessary for the current two-type filter.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Small, well-scoped fix that makes same-second timeline ordering deterministic by switching to sort.SliceStable with an explicit dismissed-before-requested tiebreak, so an explicit re-request reliably wins the tie and is treated as the most-recent event. The intent is correct, the comment explains the rationale clearly, and the new test is deterministic and exercises both input orderings (the key property that would distinguish a stable from a non-stable sort). One minor comparator-totality concern, otherwise clean.

### Issues

🟡 **daemon/internal/github/client.go:1019** — The same-second tiebreak comparator is not a strict weak ordering for the dismissed/dismissed case. When two review_dismissed events share the same CreatedAt, less(i,j)=='review_dismissed'==true AND less(j,i)=='review_dismissed'==true, i.e. the comparator reports both i<j and j<i for the same pair, violating sort's Less contract. Practical impact is currently nil (if both equal-time events are dismissals the last element is a dismissal regardless of order, so the caller's outcome is unchanged), but it is the kind of contract violation that can yield undefined ordering under future sort implementations and is ironic in a PR whose purpose is determinism. Fix: make the tiebreak total, e.g. only treat dismissed<requested when the two events differ (return out[i].Event=='review_dismissed' && out[j].Event=='review_requested'), so equal events compare as equal.

### Suggestions

- Consider adding a regression case where the same-second pair is review_dismissed/review_dismissed (and request/request) to lock in the comparator-totality behavior alongside the existing dismiss/request cases.
- The tiebreak reasoning lives only in the code comment; a brief note in the function doc or a named constant/helper (e.g. dismissBeforeRequestOnTie) could make the intent discoverable without reading the inline block.

---

vbuenog:
Thanks @sergiotejon, @Muriano, @ivanmunozruiz — all three reviews converged on the same real defect in my tiebreak, and you're right. Addressed in `e58360ae0794b920228fda1871f5ba2484e81880`.

**Implemented**
- **🔴 Strict-weak-ordering violation (client.go):** the comparator returned `out[i].Event == "review_dismissed"`, which reports both `less(i,j)` and `less(j,i)` as true for two same-second `review_dismissed` events — a contract violation (ironic in a determinism PR). Made it total: `return out[i].Event == "review_dismissed" && out[j].Event == "review_requested"`. Same-type same-second events now compare as equal, so the stable sort preserves input order; the dismissed-before-requested policy (re-request wins) is unchanged.
- **⚠️ Missing coverage:** added `TestGetPRTimelineEventsForReviewer_SameSecondDuplicateTypes` with `[dismissed, dismissed, requested]`, `[requested, requested]`, and `[dismissed, requested, dismissed]` at the same second — guards the comparator totality and that no ordering/panic regression slips in.
- **🟡 Content-Type:** the test server now sets `Content-Type: application/json`.

**Declined (with reasoning)**
- **`eventOrder(string) int` helper:** skipping — as @sergiotejon noted, it's unnecessary for the current two-type filter; I'd rather not add indirection for a hypothetical third type (YAGNI). Easy to introduce if/when the filter grows.
- **Move tiebreak rationale to function doc / named constant (`dismissBeforeRequestOnTie`):** keeping the rationale as the inline comment right at the comparator — it's where a reader hits the logic, and a named-constant indirection adds a hop without making intent clearer. Open to changing if you feel strongly.

`go test ./internal/github/ ./internal/pipeline/` and `go vet` are green.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All three previous findings (strict-weak-ordering violation, missing duplicate-type coverage, Content-Type header) have been addressed in the updated diff. The comparator is now correct, tests cover the edge cases that would have caught the original bug, and the declined suggestions (eventOrder helper, named constant) are reasonable YAGNI calls. No new issues found.

---

PR #585, opened 2026-07-09
refactor(review): drop PR-review suggestions; rename issue-triage suggestions → next_steps

Muriano:
Limpieza del concepto **suggestions** en Heimdallm, en dos mitades.

## 1. PR review: se elimina "suggestions"

Un review o levanta una issue o no lo hace; la lista libre de "suggestions" añadía ruido sin valor accionable. Eliminado end-to-end del flujo de review de PRs:

- **executor**: fuera `suggestions` del schema JSON del prompt (plantilla por defecto y la de instrucciones custom) y del struct `ReviewResult`.
- **pipeline**: deja de marshalar/publicar suggestions; la sección `### Suggestions` desaparece de ambos bodies de GitHub (review único y multi-summary). La columna `reviews.suggestions` del store se mantiene por compatibilidad con filas históricas y se escribe siempre `"[]"`.
- **flutter**: fuera del modelo `Review` + su JSON generado, la normalización cliente en `_parseReviewMap`, y la sección "Suggestions" del PR detail.

## 2. Issue triage: "suggestions" → `next_steps` (lenguaje ubicuo)

La lista del triage de issues **ya se renderizaba como "Next steps"** en el comentario de GitHub y en la GUI, pero el campo / clave JSON / columna DB seguían llamándose `suggestions`. Se alinea el identificador con el concepto:

- **store**: renombra la columna `issue_reviews.suggestions` → `next_steps` (migración idempotente `ALTER TABLE ... RENAME COLUMN`; no-op en DBs nuevas donde el schema ya crea `next_steps`) y el campo `IssueReview.Suggestions` → `NextSteps` (`json:"next_steps"`).
- **issues pipeline**: `IssueReviewResult.NextSteps` + `"next_steps"` en el schema del prompt de triage; wiring de `triage_context`/`refinement`.
- **server**: `issueReviewResponse` expone `next_steps`.
- **flutter**: `TrackedIssueReview.nextSteps` (`JsonKey next_steps`), clave de normalización cliente y label del issue detail ("Next steps").

Rename puro, sin cambio de comportamiento.

## Verificación

- `go build ./...` ✅ · `go vet ./...` ✅
- `go test ./...` daemon ✅ (el flaky `TestExecuteRawAddsDetectedWorkDirFlags` — sonda `claude --help` con timeout 2s bajo carga del suite — pasa aislado; ajeno a estos cambios)
- `flutter analyze` limpio ✅ · `flutter test` 202/202 ✅

[...]

PR #588, opened 2026-07-10
fix(executor): keep coreutils on PATH in CLI-exec tests (Linux release gate)

Muriano:
## Problema

El release de **v0.7.3** se cortó pero el build de binarios falló: el job `Tests` de `release.yml` (que corre en **ubuntu-22.04**) falló en el paquete `executor`, saltándose los 4 jobs de build de assets. El release quedó sin binarios.

Causa raíz:
```
executor_test.go:68: execute: run claude: exit status 127 (output: .../testdata/bin/claude: 3: cat: not found)
```
`TestExecute` y `TestExecuteRawCodexUsesExecAndReadsPromptFromStdin` **reemplazaban** `$PATH` por solo el directorio del CLI falso. Los scripts falsos (`claude`/`codex`) invocan `cat` (heredoc / captura de stdin), que **no es builtin** del `/bin/sh` basado en dash de los runners Linux → exit 127 `cat: not found`. En runners macOS (donde corre `tests.yml`) no se notaba.

Es fragilidad **pre-existente** (el código de test es byte-idéntico a v0.7.2); la salió a la luz la imagen actual del runner. No es regresión de #585.

## Fix

Anteponer el bin dir falso al `$PATH` existente en vez de reemplazarlo: el CLI falso sigue ganando (va primero) y los coreutils del sistema (`cat`) siguen resolviéndose. Los tests de aislamiento de detección (`TestDetect*`) no se tocan.

## Verificación

- Reproducido el bug con el script real bajo PATH restringido (`cat: command not found`) y confirmado que el fix lo resuelve.
- `go test ./internal/executor/` ✅ · `go vet ./...` ✅ · build ✅.

### DIFF
--- a/daemon/internal/executor/executor_test.go
+++ b/daemon/internal/executor/executor_test.go
@@ -60,7 +60,12 @@ func TestDetect_NoneAvailable(t *testing.T) {
 func TestExecute(t *testing.T) {
 	_, file, _, _ := runtime.Caller(0)
 	binDir := filepath.Join(filepath.Dir(file), "testdata", "bin")
-	t.Setenv("PATH", binDir)
+	// Prepend (not replace) so the fake `claude` is resolved first while the
+	// system coreutils the fake script relies on (`cat`) stay on PATH. On a
+	// dash-based /bin/sh (Linux CI runners) `cat` is not a shell builtin, so a
+	// replaced PATH makes the fake CLI exit 127 with "cat: not found". The fake
+	// binary wins because binDir comes first.
+	t.Setenv("PATH", binDir+string(os.PathListSeparator)+os.Getenv("PATH"))
 
 	e := executor.New()
 	result, err := e.Execute("claude", "Review this diff", executor.ExecOptions{})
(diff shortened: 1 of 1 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

Correct, well-scoped test fix: prepending binDir to PATH (instead of replacing it) keeps system coreutils resolvable for the fake CLI scripts that shell out to `cat`, while the fake binaries still win resolution because binDir comes first. The root cause is accurately diagnosed in the comments. Only minor consistency/robustness nits remain.

### Issues

🟡 **daemon/internal/executor/executor_test.go:99** — Other tests in the same file (TestDetect, TestDetect_Fallback, TestExecuteRawAddsDetectedWorkDirFlags, TestExecuteRawFallsBackToCWDWhenWorkDirFlagUnsupported) still fully replace PATH with binDir. They pass today only because their fake scripts happen to use shell builtins (printf), but any future edit that adds an external coreutil to those scripts will reintroduce the same exit-127 failure this PR fixes. Inconsistent treatment leaves the trap in place.
🟡 **daemon/internal/executor/executor_test.go:65** — If the inherited PATH is empty, binDir+separator+"" yields a trailing empty PATH entry, which POSIX interprets as the current directory — a (very unlikely) way for stray executables in the CWD to be picked up during the test. Guarding the concat when os.Getenv("PATH") is empty would avoid it.

### Suggestions

- Extract a small helper, e.g. prependToPath(t *testing.T, dir string), and use it in all five tests that manipulate PATH so the prepend-not-replace fix (and its explanatory comment) lives in one place.
- Apply the same prepend treatment to the remaining t.Setenv("PATH", binDir) call sites now, rather than waiting for a fake script to grow a coreutils dependency and fail the release gate again.
- In the helper, skip appending the separator when the existing PATH is empty to avoid introducing an implicit current-directory entry.

---

PR #592, opened 2026-07-10
fix(app): folder pickers + explicit Org save (no lost edits)

Muriano:
Tres problemas reportados en Settings y la vista Organizations:

## 1. No se puede seleccionar carpeta (Global + Organization)
"Clone directory" (Global) y "Local directory" (Organization) solo se podían escribir a mano. Ahora tienen un **botón de carpeta (Browse)**:
- `OverrideTextField` (compartido) gana un flag `isDirectory` cableado a `FilePicker.getDirectoryPath`.
- El campo global de clone-dir usa el mismo botón.
- (Repositories ya tenía picker nativo.)

## 2. Organization: no había botón de guardar + se perdían cambios
El auto-guardado con debounce **se cancelaba en `dispose()` sin flush**: editar un campo y salir de la pantalla antes de 800ms **descartaba el cambio en silencio**. Además el baseline era el valor pre-edición, así que editar varios campos rápido solo persistía el último.

Ahora:
- **Botón "Save" explícito** en el AppBar (activo mientras hay cambios sin guardar).
- Diffs calculados contra el último valor **persistido** (`_saved`), no el pre-edición.
- `PopScope` que **hace flush del guardado antes de cerrar** la pantalla.
- El auto-save se mantiene como conveniencia.

## Retrocompatibilidad
No hay cambios de formato de config ni del daemon: `config.toml` y la tabla `configs` de SQLite ya **persisten** entre reinstalaciones de la app (verificado en disco). La pérdida que veías era este descarte client-side de la vista de org, ahora corregido.

## Verificación
- `flutter analyze` sin issues ✅ · `flutter test` 202/202 ✅

### DIFF
--- a/flutter_app/lib/features/config/config_screen.dart
+++ b/flutter_app/lib/features/config/config_screen.dart
@@ -1,3 +1,4 @@
+import 'package:file_picker/file_picker.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
@@ -87,6 +88,7 @@ class ConfigScreen extends ConsumerStatefulWidget {
 class _ConfigScreenState extends ConsumerState<ConfigScreen> {
   final _tokenController = TextEditingController();
   final _pollController = TextEditingController();
+  final _cloneDirController = TextEditingController();
   final _pollFieldKey = GlobalKey<FormFieldState<String>>();
   bool _obscureToken = true;
   bool _tokenFromGh = false; // true = auto-detected from gh CLI
@@ -133,6 +135,7 @@ class _ConfigScreenState extends ConsumerState<ConfigScreen> {
   void dispose() {
     _tokenController.dispose();
     _pollController.dispose();
+    _cloneDirController.dispose();
     _devMaxTurnsController.dispose();
     _devTimeoutController.dispose();
     _claimLeaseController.dispose();
@@ -610,6 +613,7 @@ class _ConfigScreenState extends ConsumerState<ConfigScreen> {
     _globalNeverApproveWithIssues = config.globalNeverApproveWithIssues;
     _globalTriageOwner = config.globalTriageOwner;
     _globalCloneDir = config.globalCloneDir;
+    _cloneDirController.text = config.globalCloneDir;
     _globalAutoPromoteTriage = config.globalAutoPromoteTriage ?? false;
     _globalAutoPromoteRefinement = config.globalAutoPromoteRefinement ?? false;
     _globalGeneratePRDescription = config.globalGeneratePRDescription;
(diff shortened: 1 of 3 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid UX improvement: folder pickers reuse the established FilePicker.getDirectoryPath pattern already on main, and the explicit Save + baseline-diff (_saved) design in org_detail_screen correctly fixes the lost-edit race where rapid edits diffed against a stale 'previous' snapshot. However, the save-on-leave guard has a hole: the AppBar's custom back button calls go_router's context.pop(), which invokes NavigatorState.pop directly and does not consult PopScope.canPop — PopScope only intercepts system back (maybePop/popRoute). Since dispose() cancels the pending debounce timer, edits made within the 800ms debounce window are still silently dropped when leaving via the visible back arrow (the primary navigation path on desktop), which is exactly the bug the PR claims to close. Secondary smaller issues around save-failure-on-pop and overlapping saves.

### Issues

⚠️ **flutter_app/lib/features/organizations/org_detail_screen.dart:221** — The PopScope unsaved-edits guard is bypassed by the screen's own back button. The custom leading IconButton calls context.pop(), and go_router's pop() calls NavigatorState.pop directly (not maybePop), so PopScope.canPop/onPopInvokedWithResult never run. dispose() cancels the pending _debounce timer, so edits made in the ~800ms before clicking the back arrow are still silently lost — the guard only protects system back gestures. Use Navigator.maybePop (or an explicit 'if (_dirty) await _save()' in the button handler, or the default BackButton) so the flush actually runs on the primary navigation path.
🟡 **flutter_app/lib/features/organizations/org_detail_screen.dart:213** — In onPopInvokedWithResult, navigation proceeds even when _save() fails: the catch shows a toast but the code still calls context.pop(), destroying the screen (and the toast context) with the edits unpersisted. For a PR whose goal is 'no lost edits', a failed flush should keep the user on the screen (or prompt to discard) rather than pop.
🟡 **flutter_app/lib/features/organizations/org_detail_screen.dart:58** — _save() is not guarded against concurrent invocations: the debounce Timer can fire while a manual/pop-triggered save is still in flight (the Save button is disabled during _saving, but the timer isn't). Two overlapping PATCH requests can complete out of order, causing updateFromServer to apply a stale server snapshot and _saved to regress to an older baseline, re-marking the screen dirty or re-sending fields. Gate on _saving (re-scheduling or awaiting the in-flight request) to serialize saves.
🟡 **flutter_app/lib/features/config/config_screen.dart:677** — The folder-picker callback assigns _globalCloneDir = dir without .trim(), while the field's onChanged stores v.trim(); also, navigation via PopScope-less routes means clone-dir edits on this screen still rely on its existing save flow. Minor inconsistency only.

### Suggestions

- Make the AppBar back button honor the guard: replace the custom leading with the default BackButton (which calls Navigator.maybePop and therefore respects PopScope), or have its onPressed do 'if (_dirty) await _save();' before context.pop().
- On save failure during pop, stay on the screen (or show a discard-changes confirmation dialog) instead of popping, so the error toast is visible and edits aren't dropped.
- Serialize saves: track the in-flight save Future and have _save() await/coalesce with it before computing the diff, preventing out-of-order PATCH responses from regressing _saved and the shared config state.
- Trim the picker result in config_screen (_globalCloneDir = dir.trim()) to match the text-input path, and consider extracting the duplicated folder-picker suffixIcon (config_screen, override_field, repo_detail_screen) into a shared widget.
- Consider also flushing pending edits when navigation happens via routes that don't pop (e.g. sidebar context.go), since PopScope only guards pops — e.g. flush in deactivate()/dispose() with a fire-and-forget PATCH.

---

PR #594, opened 2026-07-10
fix: restore macOS entitlements in local package + stop false "Server unavailable"

Muriano:
Dos defectos sobre la app de escritorio en marcha:

## 1. Los botones Browse (carpeta) no hacían nada en builds locales
`make package-macos` re-firmaba con `codesign --sign -` **sin `--entitlements`**, lo que **borra** los entitlements que `flutter build` había aplicado — quitando `com.apple.security.files.user-selected.read-write`, así que el diálogo nativo de `file_picker` no-op silenciosamente.

El job `build-macos` de `release.yml` ya firmaba con `--entitlements`; el Makefile ahora hace lo mismo. (El código del picker y `Release.entitlements` ya eran correctos; solo la re-firma local era destructiva. El DMG distribuido en las releases nunca tuvo este problema.)

Verificado: la app reinstalada ahora expone `files.user-selected.read-write` y el picker abre.

## 2. "Server unavailable" intermitente con el daemon sano
El watchdog de conexión saltaba directo a `offline` ante **cualquier** error del stream SSE, aunque una conexión SSE de larga duración se resetea por muchas razones benignas (idle, daemon ocupado un instante, blip de red). Eso es lo que hacía parpadear el "Server unavailable" falso.

Ahora un error de SSE muestra "Reconnecting…" y pasa por la verificación de salud existente: reconecta si el daemon responde y **solo** reporta offline si un health check falla de verdad. Añade backoff 1→16s antes de reconectar para que un stream que falla al instante no haga hot-loop.

## Verificación
- `flutter analyze` sin issues ✅ · `flutter test` 202/202 ✅
- Daemon local: log sin errores de servidor; estaba sano mientras se veía "unavailable" → confirma que era falso positivo del cliente.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -404,7 +404,13 @@ restart: _check-docker
 
 package-macos: _check-macos build-daemon build-app
 	cp $(DAEMON_BIN) "$(APP_BUNDLE)/Contents/MacOS/heimdalld"
-	codesign --force --deep --sign - "$(APP_BUNDLE)"
+	# Re-sign WITH the entitlements — a bare `codesign --sign -` strips the
+	# entitlements `flutter build` applied, dropping the
+	# files.user-selected.read-write capability so file_picker's native
+	# directory picker silently no-ops. Mirrors the release.yml build-macos job.
+	codesign --force --deep --sign - \
+	  --entitlements flutter_app/macos/Runner/Release.entitlements \
+	  "$(APP_BUNDLE)"
 	mkdir -p dist
 	hdiutil create \
 	  -volname "Heimdallm" \
--- a/flutter_app/lib/features/dashboard/dashboard_providers.dart
+++ b/flutter_app/lib/features/dashboard/dashboard_providers.dart
@@ -89,11 +89,22 @@ class DaemonConnectionNotifier extends Notifier<DaemonConnectionStatus> {
         state = const DaemonConnectionStatus.connecting();
       }
       if (next.hasError) {
-        state = DaemonConnectionStatus(
-          phase: DaemonConnectionPhase.offline,
-          lastEventAt: state.lastEventAt,
-          message: next.error.toString(),
-        );
+        // A dropped/errored SSE stream does NOT mean the daemon is down: the
+        // long-lived connection resets for many benign reasons (idle timeout,
+        // the daemon momentarily busy, a network blip). Going straight to
+        // "offline" here is what flashes a bogus "Server unavailable" while the
+        // daemon is perfectly healthy. Instead show "reconnecting" and verify
+        // health first — _verifyAndReconnect reconnects when the daemon answers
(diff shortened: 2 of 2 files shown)


vbuenog (approved):
## 🤖 Heimdallm AI Review

Solid, well-reasoned fix. The Makefile change correctly re-applies Release.entitlements (which exists and declares files.user-selected.read-write) after the ad-hoc re-sign, mirroring the release workflow. The Dart change replaces the immediate offline transition on SSE stream error with a health-verified reconnect plus capped exponential backoff; checkHealth() catches all exceptions and returns bool, so the offline fallback path is preserved when the daemon is genuinely down, and the _checkingHealth flag correctly collapses overlapping error emissions across the backoff delay. Only minor issues found.

### Issues

🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:155** — The backoff comment claims delays of '1s, 2s, 4s… capped', but _reconnectAttempts is incremented before the shift is computed, so the actual sequence is 2s, 4s, 8s, 16s (capped at 16s). The first retry never waits 1s. Behavior is fine; the comment is inaccurate.
🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:91** — If the SSE endpoint is persistently broken while /health keeps answering 200, the UI shows 'Reconnecting…' forever — there is no attempt cap or eventual stale/degraded signal, so a real (partial) daemon fault is masked as a transient reconnect. Likely an acceptable tradeoff given the PR's goal, but worth a max-attempts escape hatch.
🟡 **flutter_app/lib/features/dashboard/dashboard_providers.dart:96** — The original stream error (next.error) is no longer surfaced anywhere: the connecting state shows the generic 'Reconnecting…' and the offline fallback shows 'Health check failed'. Diagnostic detail (e.g. connection refused vs TLS error) is lost to the user and logs.
🟡 **Makefile:411** — codesign --deep combined with --entitlements applies the app's entitlements to all nested code, including the copied heimdalld binary. Harmless here because app-sandbox is false (entitlements are just capability declarations), but --deep is deprecated by Apple and signing nested binaries with the app's entitlements is discouraged; signing heimdalld separately then the bundle without --deep would be more future-proof.

### Suggestions

- Fix the backoff comment to match the actual 2s/4s/8s/16s sequence, or compute the shift from the pre-increment attempt count so the first retry really waits 1s.
- Consider surfacing a degraded state (e.g. stale with a message) after N consecutive healthy-but-stream-failing reconnect attempts instead of showing 'Reconnecting…' indefinitely.
- Preserve the underlying stream error in the DaemonConnectionStatus message (or a debug log) so connection failures remain diagnosable.
- In the Makefile, sign the copied heimdalld first and then sign the bundle without --deep, matching Apple's current codesign guidance.

---

Issue #437, opened 2026-05-10
feat: split release pipeline — per-component releases + unified installer


Muriano:
## Context

Currently the release pipeline produces a single release with all artifacts bundled. For #436 (auto-update per component) to work, each component must be independently versioned and downloadable. Additionally, a unified release that bundles the three binaries must continue to exist for first-time installs and full upgrades.

### Per-component releases

Each component gets its own release lifecycle with independent semver:

| Component | Tag format | Artifacts |
|-----------|-----------|-----------|
| Backend (daemon) | `daemon/vX.Y.Z` | `heimdalld_{version}_{os}_{arch}.tar.gz` |
| GUI (Flutter desktop) | `gui/vX.Y.Z` | `Heimdallm-{version}.dmg` (macOS), `.deb`, `.rpm`, `.AppImage` (Linux) |
| TUI (CLI) | `cli/vX.Y.Z` | `heimdallm-cli_{version}_{os}_{arch}.tar.gz` |

Each release includes its own `checksums.txt` for integrity verification by the auto-updater.

### Unified release (full installer)

Continues the current `vX.Y.Z` tag format and bundles all three components:

| Platform | Artifact | Contents |
|----------|----------|----------|
| macOS | `Heimdallm-vX.Y.Z.dmg` | Flutter app + embedded daemon + CLI |
| Linux | `.deb` / `.rpm` / `.AppImage` | Flutter app + daemon + CLI |
| Docker | `ghcr.io/theburrowhub/heimdallm:X.Y.Z` | Daemon + web UI (unchanged) |

The unified release references which component versions it includes (e.g., in release notes or a `versions.json` manifest).

### 1. Monorepo release-please configuration

Switch from single-project to multi-project release-please config. Each component has its own path and changelog:

```json
{
  "packages": {
    "daemon": { "release-type": "go", "component": "daemon" },
    "flutter_app": { "release-type": "dart", "component": "gui" },
    "cli": { "release-type": "go", "component": "cli" }
  }
}
```

### 2. Component-specific CI workflows

- `release-daemon.yml` — triggers on `daemon/v*` tags, builds Go daemon for all platforms
- `release-gui.yml` — triggers on `gui/v*` tags, builds Flutter desktop for macOS + Linux
- `release-cli.yml` — triggers on `cli/v*` tags, builds CLI via goreleaser for all platforms
- `release-unified.yml` — triggers on `v*` tags (no prefix), assembles all components into platform installers

### 3. Version manifest

[...]

Issue #436, opened 2026-05-10
feat: auto-update mechanism for backend, GUI, and TUI


Muriano:
## Context

Currently, updating Heimdallm requires manual intervention: downloading a new DMG/package, stopping the daemon, replacing binaries, and relaunching. This proposal adds self-update capabilities to each component independently.

Each component ships and updates on its own lifecycle — they must be updatable separately.

### 1. Backend (daemon)

- Check for new releases on GitHub (`gh release` or GitHub API)
- Download the new daemon binary
- Gracefully stop the running daemon (finish in-flight work or timeout)
- Replace the binary on disk
- Restart the daemon (via LaunchAgent on macOS, systemd on Linux, or direct spawn)
- Rollback if the new binary fails health check within a timeout

### 2. GUI (Flutter desktop app)

- Check for new releases on startup and/or on a configurable interval
- Notify the user that an update is available (non-intrusive — toast/banner, not modal)
- Download the new app bundle in the background
- Apply update on next launch, or offer "restart now"
- Platform-specific:
  - **macOS**: download DMG, extract `.app`, replace in `/Applications/`, re-apply `xattr -cr`
  - **Linux**: download `.deb`/`.rpm`/`.AppImage` and install appropriately

### 3. TUI (CLI binary)

- `heimdallm update` command to check and apply updates
- `heimdallm version --check` to check without applying
- Download the new CLI binary for the current OS/arch
- Replace in-place (or via temp + rename for atomicity)
- Print changelog summary after update

## Shared Infrastructure

- **Release channel**: GitHub Releases (already used for distribution)
- **Version comparison**: semver from `.release-please-manifest.json` or binary `--version`
- **Update check interval**: configurable globally (default: once per day)
- **Config fields**:
  - `auto_update.enabled` — bool, default `true`
  - `auto_update.check_interval` — duration, default `24h`
  - `auto_update.channel` — `stable` (tags only) or `latest` (includes pre-releases)

## Safety

- Never auto-restart the daemon during an active review/implementation
- Always verify the downloaded binary's checksum against `checksums.txt` from the release
- Keep the previous version as a backup for rollback (e.g., `heimdalld.bak`)
- Docker deployments are excluded — they update via `make up-build` or image pull

PR #635, opened 2026-07-29
feat(makefile): add macOS install and uninstall targets

### LINKED ISSUE #604: feat(makefile): add install-macos / uninstall-macos targets
## Problem

There is no way to install Heimdallm on macOS from the repo. The `Makefile` has
`install-linux` / `uninstall-linux` (user-local, no sudo, `Makefile:586-754`),
but on macOS the only targets are `package-macos` (build a DMG) and
`release-local` (build + sign + notarize + publish). Neither installs anything.

The result is that the documented macOS install path is a manual sequence of
`curl` + `hdiutil attach` + `cp -R` + `xattr -cr` (`README.md:38-48`,
`LLM-HOW-TO-INSTALL.md` steps 1-3), and day-to-day usage falls back to
`make dev`, which builds the daemon and runs the Flutter app in **debug** mode
— a development flow, not an installed app.

## Proposal

Add `install-macos` and `uninstall-macos` to the root `Makefile`, mirroring the
Linux targets.

```
make install-macos                  # download the latest release DMG → /Applications
make install-macos RELEASE=v0.7.5   # pin a specific release
make uninstall-macos                # remove the app; config and data preserved
make uninstall-macos PURGE=1        # also wipe config, data and logs
```

### `install-macos`

1. `_check-macos` guard (already exists, `Makefile:233`).
2. Resolve the tag: `RELEASE` if provided, otherwise `tag_name` from
   `https://api.github.com/repos/theburrowhub/heimdallm/releases/latest`.
   Parsed with `sed` — `jq` is not a dependency of this repo, and the repo is
   public so no authentication is needed.
3. Download `Heimdallm-<tag>.dmg` into a `mktemp -d`, with a `trap` that
   detaches the volume and removes the temp dir on any exit path.
4. `hdiutil attach -nobrowse -quiet`, then locate `Heimdallm.app` in the volume.
5. **Validate the bundle before touching `/Applications`** — port the two guards
   from `LLM-HOW-TO-INSTALL.md`: both `Contents/MacOS/Heimdallm` and
   `Contents/MacOS/heimdalld` must exist, and `cmp -s` must show they differ.
   Identical binaries mean the case-collision fork-bomb state that
   `install-linux` already guards against (`Makefile:605-608`).
6. Stop running instances with `pkill -f /Applications/Heimdallm.app` (and the
   matching `heimdalld`). Scoped to the install path on purpose, so an
   in-flight `make dev` is not killed — same reasoning as the comment block in
   `uninstall-linux` (`Makefile:691-705`).
7. `rm -rf` + `cp -R` into `/Applications/Heimdallm.app`.
8. `xattr -cr` — required, see "Known limitation" below.
9. Print the installed version, `open -a Heimdallm`, and a note that config and
   data are untouched.

### `uninstall-macos`

Stop the app and daemon processes; if the LaunchAgent is installed, run
`heimdalld uninstall` (the daemon exposes that subcommand,
`daemon/cmd/heimdallm/main.go:84`) **before** deleting the bundle; then
`rm -rf /Applications/Heimdallm.app`. With `PURGE=1`, also remove
`~/.config/heimdallm`, `~/.local/share/heimdallm` and `~/Library/Logs/heimdallm`.

### Naming note

The target cannot reuse the existing `VERSION` variable: `Makefile:146` defines
it as the *next* auto-incremented semver for `release-local`, so
`make install-macos` would try to download a tag that does not exist yet. Hence
`RELEASE`.

## Known limitation: no integrity verification

The DMG is downloaded over HTTPS and not verified beyond that. Two reasons this
cannot be improved within this change:

- **No published checksum.** `checksums.txt` in the release covers only the
  GoReleaser CLI binaries; `heimdallm_<v>_checksums.txt` covers only the `.deb`
  and `.rpm`. The DMG (and the AppImage) appear in neither.
- **Ad-hoc signature.** `release.yml:487-505` signs the bundle with
  `codesign --sign -`, so the DMG is not notarized and `spctl --assess` cannot
  be used as a validation step either. This is also why `xattr -cr` is
  mandatory.

Publishing a checksum for the DMG is a separate concern in the release
pipeline; deliberately out of scope here.

## Documentation

Update `README.md` and `LLM-HOW-TO-INSTALL.md` to lead with the target, keeping
the manual steps as the fallback.

## Verification

The `Makefile` has no automated test coverage and this change does not add a
framework for it. Manual verification, to be recorded in the PR:
`make -n install-macos` to inspect the expansion, then install → `open` →
confirm the version in the UI → `make uninstall-macos`.

## Related

- #436 (auto-update for backend/GUI/TUI) — overlaps in intent: its GUI section
  describes the app updating itself by downloading a DMG and replacing
  `/Applications`. This issue is the manual, available-today counterpart, with
  no dependency on that work.
- #437 (split release pipeline + unified installer) — future restructuring of
  the release artifacts. This target consumes the artifacts as they exist now.

ivanmunozruiz:
## Summary

- add `make install-macos`, `make uninstall-macos`, and `make test-install-macos`
- install release DMGs through a POSIX helper with validated staging, same-filesystem swap, bounded process shutdown, selective `sudo`, and rollback
- migrate existing LaunchAgents to the daemon inside `/Applications/Heimdallm.app`, including loaded-disabled and unloaded states
- remove the LaunchAgent during normal uninstall while preserving user state
- make exact `PURGE=1` delete only canonical config, review-history, and log directories while preserving Keychain and custom override paths
- document the preferred checkout workflow and add a Linux CI non-regression gate

## Why

macOS currently has packaging and release targets but no repository-driven install or uninstall flow. The manual DMG procedure does not safely migrate a service installed from a checkout and can leave `launchd` retrying a missing or stale daemon.

The new flow keeps executable service state consistent with the installed bundle, limits privilege escalation to `/Applications`, and fails closed around launchd, process, port, mount, database, and rollback inspection.

## Security-sensitive files

- [installer at ded023f](https://github.com/theburrowhub/heimdallm/blob/ded023fc4751ef11f939ad39b18f459562fc53b0/scripts/macos-install.sh) ([raw](https://raw.githubusercontent.com/theburrowhub/heimdallm/ded023fc4751ef11f939ad39b18f459562fc53b0/scripts/macos-install.sh))
- [test harness at ded023f](https://github.com/theburrowhub/heimdallm/blob/ded023fc4751ef11f939ad39b18f459562fc53b0/scripts/test-macos-install.sh) ([raw](https://raw.githubusercontent.com/theburrowhub/heimdallm/ded023fc4751ef11f939ad39b18f459562fc53b0/scripts/test-macos-install.sh))
- [concise ADR](https://github.com/theburrowhub/heimdallm/blob/ded023fc4751ef11f939ad39b18f459562fc53b0/docs/superpowers/specs/2026-07-29-install-macos-target-design.md)

GitHub's per-file API exposes both script patches completely (2,184 installer lines and 288 harness lines), even if an aggregated reviewer context truncates the PR diff.

## Validation

- `make test-install-macos` — 96/96
- `/bin/dash scripts/test-macos-install.sh` — 96/96
- Alpine Linux read-only smoke — syntax, pure tests, Make guards, direct helper guards, and unchanged Linux dry-runs passed
- `make test-docker` — daemon vet and full Go suite passed in the pinned container
- `make test-cli` — passed
- `make -n install-linux` and `make -n uninstall-linux` — existing recipes still parse unchanged
- workflow YAML parse and `git diff --check` — passed

## Manual verification

The five real macOS integration scenarios in the ADR have not been run against `/Applications` or live user data from this checkout. They remain a merge gate/checklist for clean install, running-app reinstall, LaunchAgent migration, state-preserving uninstall, and guarded purge.

Closes #604
Related to #436
Related to #437

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,15 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Syntax and pure-logic tests
-        run: make test-install-macos
+        run: |
+          set -eu
+          for script in scripts/macos-install.sh scripts/test-macos-install.sh; do
+            if [ ! -x "$script" ]; then
+              echo "::error::$script must be executable"
+              exit 1
+            fi
+          done
+          make test-install-macos
 
       - name: Existing Linux install targets still parse
         run: |
@@ -38,19 +46,27 @@ jobs:
               exit 1
             fi
             printf '%s\n' "$output"
-            printf '%s\n' "$output" | grep -F "This target requires macOS." >/dev/null
+            if ! printf '%s\n' "$output" |
+              grep -Fq "This target requires macOS."; then
+              echo "::error::$target did not stop at the expected macOS guard"
+              exit 1
+            fi
           done
 
       - name: Direct helper invocation is also harmless on Linux
         run: |
           set -eu
           for operation in install uninstall; do
-            if output="$(sh scripts/macos-install.sh "$operation" 2>&1)"; then
+            if output="$(./scripts/macos-install.sh "$operation" 2>&1)"; then
               echo "::error::direct $operation unexpectedly succeeded on Linux"
               exit 1
             fi
             printf '%s\n' "$output"
-            printf '%s\n' "$output" | grep -F "This installer requires macOS." >/dev/null
+            if ! printf '%s\n' "$output" |
+              grep -Fq "This installer requires macOS."; then
+              echo "::error::direct $operation did not stop at the expected Linux guard"
+              exit 1
+            fi
           done
 
   daemon:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,6 +15,44 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  installer-shell:
+    name: macOS installer shell (Linux)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Syntax and pure-logic tests
+        run: make test-install-macos
+
+      - name: Existing Linux install targets still parse
+        run: |
+          make -n install-linux >/dev/null
+          make -n uninstall-linux >/dev/null
+
+      - name: macOS targets stop at the platform guard
+        run: |
+          set -eu
+          for target in install-macos uninstall-macos; do
+            if output="$(make "$target" 2>&1)"; then
+              echo "::error::$target unexpectedly succeeded on Linux"
+              exit 1
+            fi
+            printf '%s\n' "$output"
+            printf '%s\n' "$output" | grep -F "This target requires macOS." >/dev/null
+          done
+
+      - name: Direct helper invocation is also harmless on Linux
+        run: |
+          set -eu
+          for operation in install uninstall; do
+            if output="$(sh scripts/macos-install.sh "$operation" 2>&1)"; then
+              echo "::error::direct $operation unexpectedly succeeded on Linux"
+              exit 1
+            fi
+            printf '%s\n' "$output"
+            printf '%s\n' "$output" | grep -F "This installer requires macOS." >/dev/null
+          done
+
   daemon:
     name: Daemon (Go)
     runs-on: macos-14
--- a/LLM-HOW-TO-INSTALL.md
+++ b/LLM-HOW-TO-INSTALL.md
@@ -2,10 +2,58 @@
 
 Step-by-step guide for Claude Code or any AI agent to install Heimdallm on macOS or Linux.
 
-- **macOS**: steps 1–5 below (DMG installer)
+- **macOS with a repository checkout**: use the preferred Make target below
+- **macOS without a checkout**: use steps 1–5 (manual DMG fallback)
 - **Linux**: jump to [Linux installation](#linux-installation)
 
 ---
+
+## macOS from a repository checkout (preferred)
+
+Run from the repository root as the normal user:
+
+```bash
+make install-macos                  # latest release
+make install-macos RELEASE=v0.7.5   # specific release
+```
+
+Never run `sudo make install-macos` or `sudo make uninstall-macos`. If writing
+to `/Applications` requires elevation, the helper obtains one sudo ticket and
+elevates only the app-bundle operations. It keeps LaunchAgent and home-directory
+work in the invoking user's domain.
+
+If a Heimdallm LaunchAgent already exists, install migrates it to the daemon
+inside `/Applications/Heimdallm.app`. Uninstall removes the app and LaunchAgent
+even without purge:
+
+```bash
+make uninstall-macos                # preserve config, history, logs, and Keychain
+make uninstall-macos PURGE=1        # irreversibly delete canonical user state
+```
+
+`PURGE=1` deletes exactly `~/.config/heimdallm`,
+`~/.local/share/heimdallm` (including the review database), and
+`~/Library/Logs/heimdallm`. It does not follow custom config/data path
+overrides, and it preserves the Keychain item. To remove the credential too,
+run:
+
+```bash
+security delete-generic-password -s heimdallm -a github-token
+```
+
+After a successful install, skip the manual download steps and continue at
+[Launch](#4-launch).
+
+---
+
+## macOS manual DMG fallback (no checkout)
+
+Use this path only when the repository and its Make target are unavailable.
+Unlike the preferred target, these commands do not migrate a pre-existing
+LaunchAgent. If
+`~/Library/LaunchAgents/com.heimdallm.daemon.plist` exists, stop here and use
+the repository target; the manual fallback can otherwise leave an old
+`KeepAlive` service running against the replacement app.
 
 ## 1. Get the latest version
 
@@ -49,7 +97,10 @@ fi
 echo "✓ Bundle looks good ($(ls -lh "$DAEMON_BIN" | awk '{print $5}') daemon, $(ls -lh "$FLUTTER_BIN" | awk '{print $5}') launcher)"
 
 # Install
-pkill -9 -f Heimdallm 2>/dev/null; sleep 1
+APP_PATTERN='^/Applications/Heimdallm[.]app/Contents/MacOS/(Heimdallm|heimdalld)([[:space:]]|$)'
+pkill -TERM -U "$(id -u)" -f "$APP_PATTERN" 2>/dev/null || true
+sleep 2
+pkill -KILL -U "$(id -u)" -f "$APP_PATTERN" 2>/dev/null || true
 rm -rf /Applications/Heimdallm.app
 cp -R "$APP_SRC" /Applications/Heimdallm.app
 echo "✓ Installed to /Applications"
@@ -216,4 +267,7 @@ If it exits, there's no config yet — the app creates it on first launch.
 
 ## Updating
 
-Repeat from step 1. Config (`~/.config/heimdallm/`) and history (`~/.local/share/heimdallm/`) are preserved.
+From a repository checkout, rerun `make install-macos` (optionally with
+`RELEASE=vX.Y.Z`). It preserves config (`~/.config/heimdallm/`), history
+(`~/.local/share/heimdallm/`), logs, custom override paths, and the Keychain
+credential. Without a checkout, repeat the manual macOS steps from step 1.
--- a/LLM-HOW-TO-INSTALL.md
+++ b/LLM-HOW-TO-INSTALL.md
@@ -23,8 +23,11 @@ elevates only the app-bundle operations. It keeps LaunchAgent and home-directory
 work in the invoking user's domain.
 
 If a Heimdallm LaunchAgent already exists, install migrates it to the daemon
-inside `/Applications/Heimdallm.app`. Uninstall removes the app and LaunchAgent
-even without purge:
+inside `/Applications/Heimdallm.app`. A loaded, enabled LaunchAgent is
+regenerated from the bundled daemon, which discards manual plist edits. For an
+unloaded or disabled LaunchAgent, only the daemon executable path changes and
+the remaining plist keys are preserved. Uninstall removes the app and
+LaunchAgent even without purge:
 
 ```bash
 make uninstall-macos                # preserve config, history, logs, and Keychain
@@ -40,6 +43,10 @@ run:
 ```bash
 security delete-generic-password -s heimdallm -a github-token
 ```
+
+Leave `PURGE` empty for a normal uninstall. Any non-empty value other than
+exact `PURGE=1` is an error and aborts before changing the LaunchAgent, app, or
+user data.
 
 After a successful install, skip the manual download steps and continue at
 [Launch](#4-launch).
@@ -99,7 +106,7 @@ echo "✓ Bundle looks good ($(ls -lh "$DAEMON_BIN" | awk '{print $5}') daemon, 
 # Install
 APP_PATTERN='^/Applications/Heimdallm[.]app/Contents/MacOS/(Heimdallm|heimdalld)([[:space:]]|$)'
 pkill -TERM -U "$(id -u)" -f "$APP_PATTERN" 2>/dev/null || true
-sleep 2
+sleep 6
 pkill -KILL -U "$(id -u)" -f "$APP_PATTERN" 2>/dev/null || true
 rm -rf /Applications/Heimdallm.app
 cp -R "$APP_SRC" /Applications/Heimdallm.app
--- a/Makefile
+++ b/Makefile
@@ -238,7 +238,6 @@ release-local: _check-macos _check-signing _check-gh build-daemon
 _check-macos:
 	@if [ "$$(uname -s)" != "Darwin" ]; then \
 	  echo "❌  This target requires macOS."; \
-	  echo "    For a Linux install, use 'make install-linux'."; \
 	  exit 1; \
 	fi
 
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,10 @@ else
 endif
 
 .PHONY: build-daemon build-app build-web build-cli test test-cli lint-cli dev-cli test-docker dev dev-daemon dev-stop \
-        release-local package-macos install-service verify-linux run-linux \
-        install-linux uninstall-linux \
+        release-local package-macos install-service verify-linux run-linux test-install-macos \
+        install-macos uninstall-macos install-linux uninstall-linux \
         setup up up-build up-daemon up-build-daemon down logs logs-daemon \
-        ps restart clean clean-clones _check-docker _check-env _check-linux _post-up-hints
+        ps restart clean clean-clones _check-docker _check-env _check-macos _check-linux _post-up-hints
 
 # ── Build ─────────────────────────────────────────────────────────────────────
 
@@ -50,6 +50,11 @@ test-cli:
 
 lint-cli:
 	$(MAKE) -C cli lint
+
+test-install-macos:
+	@sh -n scripts/macos-install.sh
+	@sh -n scripts/test-macos-install.sh
+	@sh scripts/test-macos-install.sh
 
 # ── Sandboxed Go tests (EDR-safe) ─────────────────────────────────────────────
 #
@@ -233,6 +238,7 @@ release-local: _check-macos _check-signing _check-gh build-daemon
 _check-macos:
 	@if [ "$$(uname -s)" != "Darwin" ]; then \
 	  echo "❌  This target requires macOS."; \
+	  echo "    For a Linux install, use 'make install-linux'."; \
 	  exit 1; \
 	fi
 
@@ -417,6 +423,27 @@ package-macos: _check-macos build-daemon build-app
 	  -srcfolder "$(APP_BUNDLE)" \
 	  -ov -format UDZO \
 	  "dist/Heimdallm.dmg"
+
+# ── Native macOS install / uninstall (release DMG) ────────────────────────────
+#
+# Downloads a release DMG and installs the validated app bundle transactionally
+# into /Applications. The helper owns launchd state, rollback, selective sudo,
+# and cleanup; keep these recipes thin so Make never evaluates macOS commands at
+# parse time.
+#
+# Usage:
+#   make install-macos
+#   make install-macos RELEASE=v0.7.5
+#   make uninstall-macos
+#   make uninstall-macos PURGE=1
+
+install-macos: export RELEASE := $(RELEASE)
+install-macos: _check-macos
+	@./scripts/macos-install.sh install
+
+uninstall-macos: export PURGE := $(PURGE)
+uninstall-macos: _check-macos
+	@./scripts/macos-install.sh uninstall
 
 # ── Docker-based Linux build verification ─────────────────────────────────────
 #
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ endif
         release-local package-macos install-service verify-linux run-linux test-install-macos \
         install-macos uninstall-macos install-linux uninstall-linux \
         setup up up-build up-daemon up-build-daemon down logs logs-daemon \
-        ps restart clean clean-clones _check-docker _check-env _check-macos _check-linux _post-up-hints
+        ps restart clean clean-clones _check-docker _check-env _check-macos _check-macos-user _check-linux _post-up-hints
 
 # ── Build ─────────────────────────────────────────────────────────────────────
 
@@ -54,7 +54,7 @@ lint-cli:
 test-install-macos:
 	@sh -n scripts/macos-install.sh
 	@sh -n scripts/test-macos-install.sh
-	@sh scripts/test-macos-install.sh
+	@./scripts/test-macos-install.sh
 
 # ── Sandboxed Go tests (EDR-safe) ─────────────────────────────────────────────
 #
@@ -239,6 +239,13 @@ _check-macos:
 	@if [ "$$(uname -s)" != "Darwin" ]; then \
 	  echo "❌  This target requires macOS."; \
 	  echo "    For a Linux install, use 'make install-linux'."; \
+	  exit 1; \
+	fi
+
+_check-macos-user: _check-macos
+	@if [ "$$(id -u)" -eq 0 ]; then \
+	  echo "❌  Do not install or uninstall Heimdallm as root."; \
+	  echo "    Run the macOS target as your normal user, without 'sudo make'."; \
 	  exit 1; \
 	fi
 
@@ -436,13 +443,15 @@ package-macos: _check-macos build-daemon build-app
 #   make install-macos RELEASE=v0.7.5
 #   make uninstall-macos
 #   make uninstall-macos PURGE=1
+#
+# RELEASE/PURGE supplied on the make command line or inherited from the
+# environment already reach the recipe environment; do not export them
+# globally, where they could leak into unrelated targets.
 
-install-macos: export RELEASE := $(RELEASE)
-install-macos: _check-macos
+install-macos: _check-macos-user
 	@./scripts/macos-install.sh install
 
-uninstall-macos: export PURGE := $(PURGE)
-uninstall-macos: _check-macos
+uninstall-macos: _check-macos-user
 	@./scripts/macos-install.sh uninstall
 
 # ── Docker-based Linux build verification ─────────────────────────────────────
--- a/README.md
+++ b/README.md
@@ -37,7 +37,32 @@ A Flutter Web dashboard (`:3000`) with Dashboard, PR list, Issue list, prompt/ag
 
 ## Installation
 
-### macOS (DMG)
+### macOS
+
+From a checkout of this repository, use the Make targets (preferred):
+
+```bash
+make install-macos                  # install the latest release
+make install-macos RELEASE=v0.7.5   # install a specific release
+make uninstall-macos                # remove app + service; keep user state
+make uninstall-macos PURGE=1        # also delete config, history, and logs
+```
+
+Run these targets as your normal user, never with `sudo make`. If
+`/Applications` needs elevated permissions, the installer requests a sudo
+ticket once and limits it to the app-bundle operations. A pre-existing
+LaunchAgent is migrated to the installed bundle; uninstall always removes the
+LaunchAgent so `launchd` cannot keep retrying a missing executable.
+
+Normal uninstall preserves config, review history, logs, and the macOS
+Keychain credential. `PURGE=1` irreversibly removes only the canonical
+`~/.config/heimdallm`, `~/.local/share/heimdallm`, and
+`~/Library/Logs/heimdallm` directories. It does not follow custom config/data
+path overrides, and it still preserves the Keychain credential. Remove that
+separately, if desired, with
+`security delete-generic-password -s heimdallm -a github-token`.
+
+Without a checkout, use the manual DMG fallback:
 
 1. Download `Heimdallm-vX.Y.Z.dmg` from [Releases](https://github.com/theburrowhub/heimdallm/releases)
 2. Open the DMG and drag **Heimdallm** to **Applications**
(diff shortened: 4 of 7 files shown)


Muriano (commented):
## 🤖 Heimdallm AI Review

PR adds `make install-macos` / `uninstall-macos` thin Make targets delegating to a new POSIX helper (`scripts/macos-install.sh`), a portable logic-test harness, a Linux CI guard job, a detailed design doc, and README/LLM-HOW-TO-INSTALL rewrites. The visible portions are well-structured: recipes stay thin so Make never evaluates macOS commands at parse time, `_check-macos` now points Linux users at `install-linux`, the manual DMG snippet's dangerous `pkill -9 -f Heimdallm` is replaced by a UID-scoped, argv[0]-anchored TERM→KILL sequence (a real safety improvement — the old pattern could kill an unrelated `make dev`), and the docs consistently state the sudo boundary, LaunchAgent migration, and the exact PURGE contract. Important caveat: the diff supplied is truncated before `scripts/macos-install.sh` and `scripts/test-macos-install.sh`, so the actual installer state machine (rollback, selective sudo, launchd handling, purge guards) could not be reviewed — the findings below concern only what is visible. No blocking defect found in the reviewed portion.

### Issues

⚠️ **.github/workflows/tests.yml:18** — CI invokes the helper as `sh scripts/macos-install.sh`, while the Makefile invokes it as `./scripts/macos-install.sh`. The two paths differ in what they require: `./` needs the committed file mode to include the execute bit and a valid shebang, and CI never exercises that. If the script is committed 644 (or with a bad shebang), `make install-macos` fails on a user's Mac with a bare 'Permission denied' while the whole Linux job stays green. Fix: either add a mode/exec assertion to the job (e.g. `test -x scripts/macos-install.sh`) or make the Makefile call `sh scripts/macos-install.sh` so recipe and CI use the same invocation.
⚠️ **docs/superpowers/specs/2026-07-29-install-macos-target-design.md:300** — The automated test surface is deliberately limited to `sh -n` plus pure-logic units (tag validation, PURGE parsing, path construction, listener classification), while the highest-risk logic — the staging/swap transaction, armed-vs-disarmed rollback, signal-handler teardown, and the purge abort guards — is only covered by manually recorded macOS scenarios. Those are exactly the paths where a regression destroys a user's `/Applications` bundle or review database, and manual records do not run again on the next PR. The reasoning against stubbing `hdiutil`/`launchctl`/`sudo` is sound, but the state machine around them (flag transitions, idempotency of cleanup/rollback, ordering) is testable without stubbing the tools if it is factored into pure functions. Recommend extending the harness to assert rollback/cleanup state transitions so the destructive paths have non-manual coverage.
⚠️ **docs/superpowers/specs/2026-07-29-install-macos-target-design.md:268** — The default install path downloads and installs an executable app bundle into `/Applications` (with sudo when elevation is required) with no integrity verification of the artifact: the design records that no published checksum covers the DMG and that the signature is ad-hoc (`codesign --sign -`), so `hdiutil verify` and `codesign --verify` detect corruption but authenticate nothing. Provenance rests entirely on GitHub HTTPS, and a compromised or swapped release asset would be installed silently. The limitation is documented and tracked in #604, but it is not surfaced to the person running the target. Minimum fix: print an explicit one-line notice at install time that the bundle is ad-hoc signed and unverified against a published checksum; the durable fix is publishing a DMG checksum (or Developer ID/notarization) and verifying it here.
🟡 **README.md:47** — The purge flag is contractually exact-match (`PURGE=1`) per the design, so plausible variants like `PURGE=true`, `PURGE=yes`, or `PURGE=on` are silently treated as 'no purge'. Silent no-op on a destructive flag is the wrong failure direction for user expectations: the user believes state was wiped and it was not. Fix: reject any non-empty `PURGE` value other than `1` with a clear error naming the accepted value, rather than ignoring it.
🟡 **Makefile:437** — `install-macos: export RELEASE := $(RELEASE)` / `uninstall-macos: export PURGE := $(PURGE)` are redundant self-assignments: variables set on the make command line are already exported to recipe subshells, and the target-specific `export` also leaks onto the prerequisite `_check-macos`. Harmless today, but it reads as if it were load-bearing and invites copy-paste into cases where the self-assignment shadows an inherited environment value. Recommend a plain `export RELEASE PURGE` (or dropping the lines and letting the recipes reference `$(RELEASE)`/`$(PURGE)` directly) with a one-line comment on why the helper reads them from the environment.

---

---

ivanmunozruiz (commented):
## Review

Reviewed against the approved spec (`docs/superpowers/specs/2026-07-29-install-macos-target-design.md`) and the daemon/Flutter code the installer interacts with. Verified locally on macOS: `make test-install-macos` → **73/73**.

No blocking defects: no data-corruption path, no privilege-escalation hole, no destructive command reachable with an unvalidated path. One real behavioral inconsistency and a few smaller items below.

### Evaluation

| Area | Verdict | Notes |
|---|---|---|
| Correctness | :warning: | Purge guards abort at two different phases, leaving different end states |
| Spec conformance | :white_check_mark: | 4×4 matrix, LaunchAgent migration, ownership, `sudo -v`/`sudo -n` split all match |
| Safety of destructive paths | :white_check_mark: | Every `rm -rf`/`mv` under `sudo` is gated by `safe_application_work_path`/`safe_temp_path` |
| Process/port handling | :white_check_mark: | Exact-executable pattern, UID scoping, and `lsof` classification verified against real output |
| Testing | :white_check_mark: | Pure-logic only, as agreed; no tautological stubs |
| Linux non-regression | :white_check_mark: | CI job passes in 4s and asserts both guard layers |
| Docs | :white_check_mark: | README + LLM-HOW-TO-INSTALL lead with the target, fallback kept and hardened |

### 1. Purge guards abort at inconsistent phases

`uninstall_macos` runs in this order: `remove_launchagent` → `stop_bundle_processes` → `handle_ui_pid_file` → `remove_installed_app` → `purge_user_data`.

The two guards that can cancel a purge sit on opposite sides of `remove_installed_app`:

- **Live development UI** (`handle_ui_pid_file`, `scripts/macos-install.sh:1760`) dies *before* the app is removed → LaunchAgent gone, processes stopped, **app still installed**.
- **Open database** (`verify_database_closed`, called from `purge_user_data:1860`) dies *after* → LaunchAgent gone, **app removed**, purge cancelled.

Both are "purge refused", but they leave the machine in different states, and the first one is a silent partial uninstall: the user asked to remove the app and it is still there.

This is the same late-check pattern that was moved to preflight on the install path (step 3 of the spec). The fix is symmetrical: when `PURGE=1`, evaluate the `ui.pid` and database guards at the top of `uninstall_macos`, before `remove_launchagent`, and abort while the operation is still read-only. Scenario 5 of the spec's manual verification (`aborts before deleting any user state`) is satisfied by both paths today, but only by accident of what "state" means.

### 2. Migrated service gets ~6–12s to come up

`wait_for_loaded_service_ready` (`scripts/macos-install.sh:1493`) loops six times before failing, and failure means `die` → full rollback of an otherwise-correct install.

The daemon normally boots in well under a second, but it also runs store migrations on `store.Open`. On a large `heimdallm.db` with a pending migration, a correct installation could roll back purely on timing. Worth either raising the bound for this specific wait or treating "plist migrated correctly but service slow to listen" as a warning rather than a rollback trigger.

### 3. `sudo -v` ticket can expire mid-transaction

`configure_application_privilege` takes one interactive ticket, then everything transactional uses `sudo -n`. The default `timestamp_timeout` is 5 minutes. `ditto` of the bundle, the swap, the LaunchAgent waits, and `wait_for_loaded_service_ready` can plausibly exceed that on a slow disk — and a `sudo -n` failure mid-swap sends us into rollback, which also needs `sudo -n`.

The failure mode is safe (backup preserved, manual recovery printed), so this is not a correctness bug. A cheap mitigation is a second `sudo -v` immediately before `swap_staged_bundle`, where the remaining privileged work is seconds rather than minutes.

### 4. The 4×4 matrix does not actually drive the flow

`decide_install_policy` returns tokens like `stop-bundle+migrate-disabled`, but `install_macos` only branches on `abort-inconsistent`, `abort-foreign`, and `*warn*`. The rest of the flow is decided by `LAUNCHAGENT_STATE` inside `migrate_launchagent`, and `stop_bundle_processes` runs unconditionally.

Behavior is correct today — stopping bundle processes is always safe, and the state machine agrees with the tokens. But the matrix is effectively a test oracle rather than the controller, so the two can drift without any test failing. Either drive the flow from the token, or add a comment stating the matrix is documentation + test surface and the flow derives from `LAUNCHAGENT_STATE`.

`absent:service` and `present-unloaded:service` are unreachable by construction (`classify_listener` only returns `service` when a non-empty captured PID matches, which requires a loaded job). Fine as defensive code; worth a comment so a future reader does not hunt for the scenario.

### 5. `heimdalld install` replaces a hand-edited plist

The `loaded-enabled` branch regenerates the canonical plist via `launchagent.Install` (`daemon/launchagent/plist.go:44`). That was the right call — it is the source of truth and avoids reimplementing plist surgery — but it silently discards manual customization (added `EnvironmentVariables`, changed `StandardOutPath`, etc.). The `unloaded/disabled` branch preserves everything because it uses `plutil -replace`.

Not a defect, but the asymmetry deserves one line in README/LLM-HOW-TO-INSTALL so nobody is surprised.

### Nits

- `process_executable` (`:264`) is correct — I verified on a real process that the first `ftxt`/`n` pair from `lsof -d txt` is the executable and matches `ps -p PID -o comm=`. `ps -o comm=` would give the same answer with no `awk`, if you ever want to simplify.
- Temp-file creation is inconsistent: `create_private_temp` validates `TMPDIR` against a safe-path allowlist, `verify_database_closed:1800` uses `${TMPDIR:-/tmp}` unvalidated, and `inspect_port:537` hardcodes `/tmp`. No security impact (`mktemp` is 0600), but three styles for the same need.
- `inspect_port`'s scratch file and `verify_database_closed`'s are created before the `EXIT` trap covers them in some paths, so a signal at exactly the wrong moment can orphan a 0-byte file in `/tmp`. Cosmetic.
- The irreversible-purge warning prints inside `purge_user_data`, i.e. after the app and LaunchAgent are already gone. It is still before any data deletion (which is what the spec requires), but as a "last chance to Ctrl-C" it arrives late.
- `chown -R` in `stage_bundle` follows symlinks inside the bundle (no `-h`). All Flutter framework symlinks are bundle-internal, so this is safe today; `-h` would make it robust by construction.
- Ownership verification only stats the bundle root, not the tree.

### Required Changes

- [ ] Move the `PURGE=1` guards (`handle_ui_pid_file` live-UI case and `verify_database_closed`) into a preflight at the start of `uninstall_macos`, so a refused purge never leaves the app half-removed
- [ ] Decide on the migrated-service wait: raise the bound or downgrade a slow-but-correct start from rollback to warning
- [ ] Document that the `loaded-enabled` migration regenerates the plist from scratch and discards manual plist edits

### Non-blocking follow-ups

- [ ] Refresh the sudo ticket immediately before `swap_staged_bundle`
- [ ] Comment (or wire up) the relationship between `decide_install_policy` and the actual control flow, including the unreachable `*:service` cells
- [ ] Unify temp-file creation on the validated helper

Once the three required items are addressed, this is good to merge from my side — but the five real-macOS scenarios in the spec are still unrun per the PR description, and `build-test`/`Flutter` checks were still pending at review time. Those gate the merge, not this comment.

sergiotejon (changes requested):
## 🤖 Heimdallm AI Review

Well-structured PR that adds macOS install/uninstall Make targets with thorough CI coverage and documentation. The Makefile changes are clean thin wrappers, the CI Linux non-regression testing is solid, and documentation is comprehensive. However, the core implementation scripts (scripts/macos-install.sh, scripts/test-macos-install.sh) are not present in the diff — these contain all the security-sensitive logic (sudo handling, process signaling, filesystem mutations, network downloads) and cannot be reviewed. The visible changes are high quality but the review is incomplete without the scripts.

### Issues

🔴 **scripts/macos-install.sh:0** — The core installer script is referenced by the Makefile and CI workflow but is not included in the diff (truncated). This file handles sudo privilege escalation, process signaling (TERM/KILL), filesystem transactions (staging/swap in /Applications), network downloads, hdiutil mount/unmount, and LaunchAgent migration. It is the most security-critical component of this PR and must be reviewed before merge.
⚠️ **scripts/test-macos-install.sh:0** — The test harness script is also missing from the diff. Without seeing it, there's no way to verify that the pure-logic tests (release-tag validation, PURGE parsing, path construction, listener classification) adequately cover the security-sensitive code paths in the installer.
🟡 **Makefile:238** — The _check-macos guard added a helpful 'For a Linux install' message but the guard itself only checks uname -s. The design spec requires rejecting effective UID 0 ('run without sudo') at the guard level, but this check appears to be deferred entirely to the script. If the script is somehow bypassed or replaced, the Makefile provides no root-rejection safety net. Consider adding an EUID != 0 check in the Makefile target prerequisites as defense-in-depth.
🟡 **docs/superpowers/specs/2026-07-29-install-macos-target-design.md:1** — The 382-line design spec duplicates and extends what the code and commit messages should convey. It will drift from the implementation over time. Consider converting it to an ADR (Architecture Decision Record) format focused on the 'why' decisions (privilege boundary, no notarization, PURGE semantics) rather than step-by-step implementation that mirrors the script. Remove procedural details that are better expressed as code comments or tests.
🟡 **LLM-HOW-TO-INSTALL.md:100** — The manual DMG fallback section uses 'sleep 2' between TERM and KILL signals, while the design spec for the automated script specifies 'up to six seconds'. Two seconds may be insufficient for a Flutter app performing SQLite WAL checkpoint or other cleanup on shutdown. Consider increasing to match the script's timeout or adding a note that this is intentionally shorter for the manual path.
🟡 **.github/workflows/tests.yml:30** — The CI step 'macOS targets stop at the platform guard' correctly validates failure, but the grep pattern checks for 'This target requires macOS.' (from the Makefile). If the _check-macos message ever changes, this CI step silently becomes a no-op (grep fails, caught by set -eu, but the error message won't be helpful). Consider using grep -q with an explicit || { echo 'expected message not found'; exit 1; } for clearer failure diagnostics.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Well-engineered and unusually thoroughly documented addition of native macOS install/uninstall Make targets backed by a POSIX shell helper, with an ADR, pure-logic tests, and Linux CI guard checks. The visible code (platform/root guards, release-tag and PURGE validation, listener/policy classification matrix, rollback predicates) is careful and internally consistent with the ADR. Remaining concerns are structural rather than outright bugs: the installer downloads an unauthenticated ad-hoc-signed DMG and strips quarantine, the ~2,200-line shell script's real install/rollback paths are only verifiable manually on macOS (CI exercises only syntax, pure functions, and Linux guard failures), and the sheer size of the shell installer is a significant maintenance surface. Note: the diff was truncated in the provided input and PR discussion could not be fetched, so the middle of macos-install.sh and the test script were not directly reviewed.

### Issues

⚠️ **scripts/macos-install.sh:1** — The installer downloads a release DMG that has no published checksum and only an ad-hoc code signature, then clears the quarantine attribute (xattr -cr) before installing to /Applications. hdiutil verify and codesign --verify detect corruption but do not authenticate the publisher, so integrity rests entirely on GitHub HTTPS; stripping quarantine also bypasses Gatekeeper for the unauthenticated bundle. This is honestly disclosed in the ADR and runtime output, but publishing a SHA-256 checksum (and eventually Developer ID/notarization) should be tracked as a prerequisite for recommending this as the preferred install path.
⚠️ **.github/workflows/tests.yml:18** — CI only validates shell syntax, dependency-free pure functions, and that targets fail at the platform guards on Linux. None of the actual install/uninstall behavior (hdiutil mount, transactional swap, sudo escalation, LaunchAgent migration, rollback, purge) is exercised by any automated job, even though macOS runners are already used elsewhere in this workflow (daemon job runs on macos-14). The five 'required macOS scenarios' in the ADR are manual. A smoke job on the existing macOS runner (e.g. uninstall idempotency on a clean machine, or install from a fixture DMG) would materially reduce regression risk for a 2,000+ line installer.
🟡 **scripts/macos-install.sh:1** — The helper is ~2,184 lines of hand-rolled POSIX sh implementing transactional swaps, signal-safe rollback, launchd state machines, and lsof-based process classification. This is a large, hard-to-review maintenance surface in the least type-safe language available; a fair portion of the risk it defends against (e.g. selective sudo, rollback bookkeeping) stems from its own ambition. Worth considering whether some responsibilities (e.g. service management) could be delegated to the bundled `heimdalld install` to shrink the script.
🟡 **LLM-HOW-TO-INSTALL.md:107** — The manual-fallback kill sequence always sleeps 6 seconds between TERM and KILL even when pkill matched no processes; checking pkill's exit status (0 = matched) before sleeping would avoid the fixed delay on clean systems. Minor UX nit.
⚠️ **scripts/macos-install.sh:1** — The diff supplied for review is truncated mid-file (approximately line 500 of macos-install.sh) and omits scripts/test-macos-install.sh entirely, and the PR discussion could not be retrieved from this environment; the untruncated majority of the installer (download, mount, staging, sudo, swap, rollback, migration, purge implementations) was not directly reviewed and this assessment cannot vouch for it.

---

Muriano (commented):
## 🤖 Heimdallm AI Review

Re-review: los cinco hallazgos previos están resueltos en este diff. (1) CI ahora afirma explícitamente el bit de ejecución de ambos scripts y además invoca `./scripts/macos-install.sh install|uninstall` directamente, con lo que el modo committeado y el shebang quedan ejercitados en el job de Linux; el Makefile mantiene `./scripts/...` de forma coherente. (2) El diseño y el helper factorizan la máquina de estados destructiva en funciones puras (`rollback_required`, `rollback_old_move_detected`, `rollback_new_move_detected`, `classify_purge_holder`, `decide_install_policy`) y `make test-install-macos` cubre las 16 celdas de la matriz, decisiones de rollback y limpieza armada/desarmada/idempotente, más los guards de purge — la cobertura ya no es solo manual. (3) La falta de checksum publicado y la firma ad-hoc se anuncian en tiempo de ejecución antes de la descarga, y queda documentado en README/ADR que la procedencia depende de HTTPS de GitHub. (4) `validate_purge_value` rechaza cualquier valor no vacío distinto de `1` y aborta antes de tocar LaunchAgent, app o datos, documentado en README y en LLM-HOW-TO-INSTALL. (5) Los self-assignments `export RELEASE`/`export PURGE` desaparecieron: las recetas son finas y un comentario explica correctamente que las variables de línea de comandos o de entorno ya llegan al entorno de la receta. El diseño (guard UID 0, patrón anclado de procesos, TERM→espera→KILL, swap transaccional con backup en el mismo filesystem, purge limitado a rutas canónicas sin seguir overrides y preservando Keychain) es sólido y consistente entre script, Makefile y documentación. Solo queda un detalle cosmético introducido en este diff. Nota: el diff recibido está truncado en `scripts/macos-install.sh` (2184 líneas) y no incluye `scripts/test-macos-install.sh`, por lo que la verificación de las secciones no mostradas se limita a lo descrito en el ADR.

### Issues

🟡 **Makefile:241** — La nueva línea `echo "    For a Linux install, use 'make install-linux'."` se añadió dentro de `_check-macos`, que es prerequisito compartido por `release-local` y `package-macos`, no solo por los nuevos targets de instalación. Impacto: quien ejecute `make package-macos` o `make release-local` en Linux recibe un consejo sin sentido para su intención (empaquetar/publicar un DMG, no instalar), lo que degrada un diagnóstico que antes era neutro y correcto para todos los consumidores del guard. Fix recomendado: mover la sugerencia al guard exclusivo de instalación (`_check-macos-user`), o dejar `_check-macos` con el mensaje genérico y emitir la pista de `install-linux` desde las recetas `install-macos`/`uninstall-macos`. Severidad: NIT.

---

---

ivanmunozruiz:
Implemented the three review rounds in `ded023f` and requested re-review from @Muriano and @sergiotejon.

### Muriano — [review](https://github.com/theburrowhub/heimdallm/pull/635#pullrequestreview-4806250980)

- CI now asserts both scripts are executable and invokes them directly, exercising the shebang.
- The pure harness grew from 73 to 96 checks: strict PURGE parsing, rollback-required states, real idempotent temp cleanup, known/foreign purge-holder decisions, and the full 4×4 policy matrix.
- Install prints the checksum/ad-hoc-signature limitation before download.
- Any non-empty value except exact `PURGE=1` aborts before mutation.
- The redundant target-specific self-exports are gone.

### Implementation review — [review](https://github.com/theburrowhub/heimdallm/pull/635#pullrequestreview-4806302987)

- `PURGE=1` now performs a read-only preflight before launchd/process/app changes. It allows only the known LaunchAgent or exact installed bundle holders, rejects foreign DB/UI holders, then retains a strict late `lsof` guard. The data directory is removed immediately after that final guard.
- Migrated-service readiness is bounded at 60 seconds and still rolls back on failure.
- README and install guide document the loaded/enabled plist regeneration asymmetry.
- The matrix is explicitly documented as preflight policy/test oracle; defensive unreachable cells are identified.
- Uninstall now obtains any needed `sudo -v` ticket before bootout/pkill, so cancelled authentication cannot leave a predictable partial uninstall. Transactional commands remain `sudo -n`; there is no late interactive prompt.

### sergiotejon — [requested changes](https://github.com/theburrowhub/heimdallm/pull/635#pullrequestreview-4806696408)

- The complete security-sensitive files are directly reviewable at this exact SHA:
  - [installer (2,184 lines)](https://github.com/theburrowhub/heimdallm/blob/ded023fc4751ef11f939ad39b18f459562fc53b0/scripts/macos-install.sh) · [raw](https://raw.githubusercontent.com/theburrowhub/heimdallm/ded023fc4751ef11f939ad39b18f459562fc53b0/scripts/macos-install.sh)
  - [harness (288 lines)](https://github.com/theburrowhub/heimdallm/blob/ded023fc4751ef11f939ad39b18f459562fc53b0/scripts/test-macos-install.sh) · [raw](https://raw.githubusercontent.com/theburrowhub/heimdallm/ded023fc4751ef11f939ad39b18f459562fc53b0/scripts/test-macos-install.sh)
- Verified after push that GitHub's per-file API returns both patches completely (65,177 and 11,856 characters); the earlier omission was aggregate-context truncation.
- Make now has a dedicated `_check-macos-user` prerequisite for install/uninstall that rejects UID 0 without changing package/release targets.
- The 382-line procedural spec is now a concise 223-line ADR.
- The manual fallback matches the six-second TERM→KILL grace period.
- Both CI message checks use `grep -Fq` with explicit `::error::` diagnostics.

The final audit also caught and fixed a pre-existing edge in this PR: Flutter writes `ui.pid` without a newline, while POSIX `read` returns non-zero at EOF. The helper now preserves that assigned PID, with a regression test, so a live development UI cannot be silently ignored during purge.

Validation at `ded023f`:

- `make test-install-macos`: 96/96
- `/bin/dash scripts/test-macos-install.sh`: 96/96
- Linux read-only container smoke: syntax, direct/Make platform guards, pure tests, and Linux dry-runs passed
- `make test-docker`: daemon vet + full suite passed in the pinned container
- `make test-cli`: passed
- workflow YAML and `git diff --check`: passed

The five destructive/live-state macOS scenarios remain explicitly unrun and remain a merge checklist; I am not treating this update as approval. @Muriano @sergiotejon, please re-review the new SHA.

Muriano (commented):
## 🤖 Heimdallm AI Review

Re-review at ded023f: los tres rounds previos están efectivamente atendidos en el diff — CI ahora exige ejecutables e invoca los scripts directamente (ejercitando el shebang), el aviso de ausencia de checksum/firma ad-hoc se imprime antes de la descarga, `validate_purge_value` rechaza cualquier valor no vacío distinto de `1` antes de mutar estado, existe `_check-macos-user` dedicado que rechaza UID 0 sin tocar package/release, la spec se condensó en un ADR de 223 líneas, el fallback manual del DMG replica la gracia TERM→KILL de 6s con patrón anclado y `-U $(id -u)`, y las comprobaciones de CI usan `grep -Fq` con diagnósticos `::error::`. La matriz 4×4, rollback y clasificación de holders están documentadas y cubiertas por el harness puro (96 checks). No hay concerns de revisores sin resolver ni blockers pendientes. Solo persiste el NIT anterior, no atendido. Nota no bloqueante: los 5 escenarios destructivos/live-state en macOS siguen explícitamente sin ejecutar y son checklist de merge declarada por el autor.

### Issues

🟡 **Makefile:241** — Persiste sin cambios el NIT del review previo: `echo "    For a Linux install, use 'make install-linux'."` sigue dentro de `_check-macos`, guard compartido por `release-local` y `package-macos`, no solo por los targets de instalación. Impacto: quien ejecute `make package-macos` o `make release-local` en Linux recibe un consejo irrelevante para su intención (empaquetar/publicar un DMG, no instalar), degradando un diagnóstico que antes era neutro y correcto para todos los consumidores del guard. Fix: mover la pista al guard exclusivo de instalación (`_check-macos-user`) o emitirla desde las recetas `install-macos`/`uninstall-macos`, dejando `_check-macos` con el mensaje genérico. Severidad: NIT — no bloquea merge.

---

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Re-review: The author addressed three of five previous findings (Makefile root guard, sleep timing, ADR format). The 2184-line installer script is partially visible in the diff (~350 lines of pure helpers and validation). The visible code is well-structured with proper input validation, full-path system binary references, and namespace-prefixed variables. The core security-critical sections (download/curl, sudo escalation, filesystem swap, rollback/cleanup, trap handling) remain in the truncated portion and are not reviewable from the diff alone. The test harness is also not in the diff. These are the same concerns as the previous review, now at reduced severity since the visible architecture inspires confidence and the design spec is thorough.

### Issues

⚠️ **scripts/macos-install.sh:0** — 2184-line POSIX shell installer with only ~350 lines visible in the diff. The visible portion (pure helpers, validation, process inspection) is well-written. However, the security-critical sections — curl download with URL construction, hdiutil mount/verify, sudo -v escalation and sudo -n usage, ditto staging, mv-based atomic swap, trap/rollback/cleanup, LaunchAgent migration via plutil, and the uninstall/purge teardown — are all in the truncated portion. A reviewer must read the full file before approving. If you've already reviewed those sections outside this diff, this is not blocking.
🟡 **scripts/test-macos-install.sh:0** — The test harness is referenced by CI and the Makefile but its content is not in the diff. The design spec claims it covers all 16 matrix cells, rollback decisions, purge-holder classification, and cleanup idempotency. Without seeing the actual test code, coverage claims cannot be verified.
🟡 **README.md:488** — The README Make-targets table lists 'make install-macos' and 'make uninstall-macos' with brief descriptions, but the new macOS installation section above (lines ~39-72 in the diff) is ~35 lines of dense prose about LaunchAgent migration semantics, PURGE behavior, and plist regeneration. This level of detail belongs in the LLM doc or a dedicated install guide — the README should stay concise with a link to detailed docs. The Linux section is 4 lines for comparison.

---

ivanmunozruiz:
Resolved Muriano’s remaining nit in `c9ec671`: `_check-macos` is generic again, so shared consumers such as `package-macos` and `release-local` no longer receive an irrelevant install hint. The install/uninstall helper’s direct Linux guard remains scoped to those operations. `make test-install-macos` still passes 96/96 and `git diff --check` is clean. @Muriano, please re-review this head. @sergiotejon, your requested-changes re-review is still requested as well.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review del head c9ec671. El único hallazgo previo (NIT: pista de 'make install-linux' dentro del guard compartido _check-macos) está resuelto: el diff actual muestra _check-macos con el mensaje genérico y la pista específica de instalación queda en el guard Linux del helper, exclusivo de install/uninstall. Los puntos de las rondas anteriores también aparecen aplicados en el diff: CI verifica el bit ejecutable e invoca ambos scripts directamente (ejercitando el shebang), la validación estricta de PURGE aborta antes de mutar LaunchAgent/app/datos, install imprime la limitación de checksum/firma ad-hoc antes de la descarga, la readiness del servicio migrado está acotada a 60s con rollback, el preflight de purge es read-only y la política del matriz 4x4 está documentada como oráculo de test. La ADR sustituye la spec procedimental y README/LLM-HOW-TO-INSTALL documentan la asimetría de regeneración del plist cargado+habilitado y el grace period TERM→KILL de seis segundos, coherente con BUNDLE_PROCESS_PATTERN y el pkill acotado por UID del fallback manual. No detecto issues confirmados nuevos en el diff. Queda como recordatorio operativo (no bloqueante y ya asumido explícitamente por el autor) que los cinco escenarios destructivos/live-state en macOS siguen sin ejecutar y son checklist de merge; los tests automatizados solo cubren lógica pura y guards de plataforma en Linux, no hdiutil/launchd/sudo/swaps reales.

---

PR #636, opened 2026-07-29
fix: launch installed macOS bundle by exact path

ivanmunozruiz:
## Summary

- print an exact `/Applications/Heimdallm.app` launch command after `make install-macos`
- avoid `open -a Heimdallm`, which can resolve a registered Debug build with the same bundle name and identifier
- add a regression test for the exact-path launch hint

## Reproduction

LaunchServices had both the checkout Debug bundle and v0.7.8 under `com.theburrowhub.heimdallm`. The installer hint opened the Debug bundle, which does not contain the packaged `heimdalld`, while the installed bundle did.

## Verification

- `make test-install-macos` (97 tests)
- launched `/Applications/Heimdallm.app` by exact path and verified both UI and daemon run from the installed bundle

Linux targets and runtime code are unchanged.


vbuenog (approved):
## 🤖 Heimdallm AI Review

Small, safe change: replaces the ambiguous 'open -a Heimdallm' launch hint with an exact-path '/usr/bin/open "/Applications/Heimdallm.app"' command, extracted into a helper so it can be unit-tested. Verified against the repo: APP_PATH is a top-level constant set before the helper runs, the test harness sources the script with HEIMDALLM_MACOS_INSTALL_SOURCE_ONLY=1 so the new assert_output test resolves APP_PATH correctly, and the trailing newline emitted by printf is stripped by command substitution in both finish_install and assert_output. No correctness, security, or portability problems found.

### Issues

🟡 **scripts/macos-install.sh:62** — installed_app_launch_command is placed in the 'Pure helpers' section but reads the APP_PATH global rather than taking it as a parameter, unlike the other helpers around it (e.g. validate_release_tag). Passing the path as an argument would keep the section's convention and make the test independent of sourcing order. Cosmetic/consistency only.

---

Muriano (commented):
## 🤖 Heimdallm AI Review

Small, well-scoped change: the post-install hint now prints an exact-path launch command (`/usr/bin/open "$APP_PATH"`) instead of the ambiguous `open -a Heimdallm`, extracted into a pure helper (`installed_app_launch_command`) and covered by a deterministic assertion in the existing pure-logic test harness. It matches the project's established pattern of factoring shell logic into testable functions, is backward-compatible (informational output only), and introduces no security, performance or observability concerns. Note the change only fixes the printed hint; if any other code path still launches the bundle via `open -a Heimdallm`, the underlying LaunchServices ambiguity would remain there (not verifiable from this diff). Findings are cosmetic/robustness-level only.

### Issues

🟡 **scripts/macos-install.sh:62** — The helper builds the command string with `printf '/usr/bin/open "%s"\n' "$APP_PATH"`, which only survives spaces. If `$APP_PATH` ever contains a double quote, `$`, backtick or backslash (e.g. because the install prefix becomes overridable), the printed hint is a command the user copies and pastes that is mis-quoted or performs unintended expansion. Impact is limited to a broken/misleading copy-paste hint today, but it is a latent shell-quoting hazard. Recommended fix: emit single-quoted form with embedded quotes escaped (e.g. a small `shell_quote()` helper, or `printf "/usr/bin/open '%s'\n" "${APP_PATH//\'/\'\\\'\'}"` in bash), and keep the assertion for that case in the test file.
🟡 **scripts/test-macos-install.sh:114** — The new assertion hardcodes the expected output `/usr/bin/open "/Applications/Heimdallm.app"` while relying on `APP_PATH` being implicitly set to its default at source time — it is placed before the `set_user_paths` block, so the test silently depends on definition order and on the default never moving. Impact: a later refactor that derives `APP_PATH` inside `set_user_paths` (or an env override) would make this test pass/fail for reasons unrelated to the helper, and the non-default path case is untested. Recommended fix: set `APP_PATH` explicitly in the assertion's scope (e.g. run the helper in a subshell with `APP_PATH=/Applications/Heimdallm.app`) and add a second case with a path containing a space to lock the quoting contract.
🟡 **scripts/macos-install.sh:1634** — The install hint now uses the absolute `/usr/bin/open`, while `LLM-HOW-TO-INSTALL.md:74` still documents `open /Applications/Heimdallm.app`. Impact: cosmetic inconsistency between installer output and the documented launch instruction, which can confuse users following the docs alongside the installer. Recommended fix: align the documentation with the `/usr/bin/open "/Applications/Heimdallm.app"` form emitted by the installer.

---

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-scoped fix that replaces an ambiguous `open -a Heimdallm` hint with an exact-path launch command, preventing LaunchServices from picking a wrong bundle. Includes a matching test. No issues found.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #636: all three findings from the previous round are addressed. The helper now emits a single-quoted command with embedded single quotes escaped via sed (`s/'/'\''/g` after shell unescaping, which is the correct POSIX single-quote escaping form), the test no longer relies on implicit `APP_PATH` state (the path is passed as an explicit argument) and adds a second case locking the quoting contract for a path containing `$`, backticks, double quotes, backslash and an apostrophe, and LLM-HOW-TO-INSTALL.md's launch step was aligned to the `/usr/bin/open "/Applications/Heimdallm.app"` form. Verified the quoting by hand: `printf "...\"s/'/'\\\\''/g\""` reaches sed as `s/'/'\''/g`, whose replacement yields `'\''` — a correctly terminated/reopened single-quoted string, so the emitted hint is safe to copy-paste for any path. Absolute `/usr/bin/open` and `/usr/bin/sed` match the defensive absolute-binary style already used elsewhere in the installer, and the helper's arity guard follows the same pattern as `validate_purge_value`/`validate_home_path`. Scope stays minimal (hint text + test + doc), no behavioural change to installation itself, and the change is a real root-cause fix for the LaunchServices ambiguity between a dev bundle and the installed one. No new issues found.

---

PR #638, opened 2026-07-29
fix(test): isolate Compose integration runs from production volumes

### LINKED ISSUE #606: fix(test): isolate Compose integration runs from production volumes
## Contexto

Los runners E2E usan los mismos archivos Compose y, por defecto, el mismo nombre de proyecto que un despliegue normal iniciado desde el repositorio. Ambos registran un `trap` que ejecuta `down -v --remove-orphans`.

## Evidencia

- `docker/scripts/test-local.sh`: `COMPOSE_TEST="docker compose -f docker/docker-compose.yml -f docker/docker-compose.test.yml"` y cleanup con `down -v`.
- `docker/scripts/test-web.sh`: mismo patrón.
- No se fija un nombre de proyecto exclusivo para tests antes de eliminar volúmenes.

## Riesgo

Ejecutar un smoke/E2E mientras existe el stack local puede detener servicios y borrar `heimdallm-data` y `heimdallm-config`, incluyendo base de datos, token y configuración persistente.

## Criterios de aceptación

- Cada ejecución usa un proyecto Compose inequívocamente de test y nunca puede coincidir con el proyecto de desarrollo/producción.
- El cleanup solo opera sobre recursos creados por esa ejecución.
- Las ejecuciones paralelas no comparten contenedores, redes ni volúmenes.
- Se elimina o parametriza cualquier `container_name` que impida el aislamiento por proyecto.
- El script falla cerrado si no puede demostrar que el proyecto es de test.
- La guía E2E explica el aislamiento y la recuperación esperada.

## Pruebas de no regresión

- Un stack normal permanece activo y conserva sus volúmenes después de ejecutar y limpiar el test.
- Dos runs de test concurrentes no colisionan.
- Una salida temprana o señal limpia únicamente los recursos del run afectado.
- Test del wrapper Compose con un binario fake que compruebe todos los argumentos destructivos sin levantar el daemon.

## Relacionado

Hallazgo de auditoría 2026-07-29; no se encontró issue abierto equivalente.

ivanmunozruiz:
## Qué cambia

- Aísla cada ejecución de pruebas Compose con un nombre de proyecto, contenedores, red, volúmenes y puertos efímeros propios.
- Centraliza el ciclo de vida en un helper con marcador fail-closed y cleanup protegido.
- Evita que los scripts de smoke reutilicen o eliminen los recursos del stack normal.
- Añade pruebas de aislamiento, wiring en CI y documentación de recuperación.
- Conserva el flujo macOS introducido en #604.

## Causa raíz e impacto

Los scripts compartían los nombres, puertos y volúmenes por defecto de Compose. Una prueba podía colisionar con una instalación activa o ejecutar un `down -v` sobre sus datos. Ahora el ámbito destructivo queda limitado al proyecto efímero creado por esa ejecución.

## Validación

- `make test-compose-isolation` — 8/8 casos.
- `make test-install-macos` — 98/98 casos.
- Comprobaciones de sintaxis POSIX/Bash y `git diff --check`.
- Dos ejecuciones reales del smoke Docker con cleanup verificado: 16/17 checks funcionales; el único fallo es la instalación preexistente de OpenCode en arm64/musl, registrada en #631 y relacionada con #621.
- Revisión independiente sin hallazgos bloqueantes, altos ni medios.

## Riesgo residual

`SIGKILL` no permite ejecutar traps; el proyecto efímero queda identificado para limpieza manual. La concurrencia se cubre con el harness determinista y la configuración Compose, además de dos ejecuciones Docker reales secuenciales.

Closes #606

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -109,6 +109,15 @@ jobs:
         working-directory: cli
         run: go vet ./...
 
+  compose-isolation:
+    name: Docker Compose test isolation
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Test destructive-command guards without starting services
+        run: make test-compose-isolation
+
   flutter:
     name: Flutter
     runs-on: macos-14
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,8 @@ else
   APP_BUNDLE       := $(FLUTTER_BUILD)/bundle
 endif
 
-.PHONY: build-daemon build-app build-web build-cli test test-cli lint-cli dev-cli test-docker dev dev-daemon dev-stop \
+.PHONY: build-daemon build-app build-web build-cli test test-cli lint-cli test-compose-isolation \
+        test-smoke test-github test-e2e test-web dev-cli test-docker dev dev-daemon dev-stop \
         release-local package-macos install-service verify-linux run-linux test-install-macos \
         install-macos uninstall-macos install-linux uninstall-linux \
         setup up up-build up-daemon up-build-daemon down logs logs-daemon \
@@ -55,6 +56,23 @@ test-install-macos:
 	@sh -n scripts/macos-install.sh
 	@sh -n scripts/test-macos-install.sh
 	@./scripts/test-macos-install.sh
+
+# Static/fake-Docker regression suite for the destructive Compose test wrappers.
+# It does not build images, start containers, or touch daemon data.
+test-compose-isolation:
+	./docker/scripts/tests/test-compose-isolation.sh
+
+test-smoke:
+	./docker/scripts/test-local.sh smoke
+
+test-github:
+	./docker/scripts/test-local.sh github
+
+test-e2e:
+	./docker/scripts/test-local.sh e2e
+
+test-web:
+	./docker/scripts/test-web.sh
 
 # ── Sandboxed Go tests (EDR-safe) ─────────────────────────────────────────────
 #
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -1,21 +1,22 @@
-# Override for local testing.
-# Usage: docker compose -f docker/docker-compose.yml -f docker/docker-compose.test.yml up
+# Override for isolated local testing.
+#
+# Do not invoke this file directly. docker/scripts/test-local.sh and
+# docker/scripts/test-web.sh generate a unique Compose project and required
+# container names for every run.
 services:
   heimdallm:
-    container_name: heimdallm-test
+    container_name: "${HEIMDALLM_TEST_DAEMON_CONTAINER_NAME:?use an isolated Docker test runner}"
     restart: "no"
-    ports:
-      - "${HEIMDALLM_PORT:-7842}:${HEIMDALLM_PORT:-7842}"
     environment:
       - HEIMDALLM_POLL_INTERVAL=${HEIMDALLM_POLL_INTERVAL:-1m}
     volumes:
       - heimdallm-test-data:/data
-      - ./config:/config
+      - heimdallm-test-config:/config
       # Gemini OAuth tokens from host (uncomment for Level 3 E2E test):
       # - ~/.gemini:/home/heimdallm/.gemini:ro
 
   web:
-    container_name: heimdallm-web-test
+    container_name: "${HEIMDALLM_TEST_WEB_CONTAINER_NAME:?use an isolated Docker test runner}"
     restart: "no"
     # Set HEIMDALLM_API_TOKEN to empty (the entrypoint's `[ -n "$VAR" ]`
     # check treats empty and unset the same, falling through to the file
@@ -33,4 +34,6 @@ services:
 
 volumes:
   heimdallm-test-data:
+    driver: local
+  heimdallm-test-config:
     driver: local
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -6,7 +6,12 @@ services:
     container_name: heimdallm
     restart: unless-stopped
     ports:
-      - "${HEIMDALLM_PORT:-7842}:${HEIMDALLM_PORT:-7842}"
+      # Compose-only host binding knobs let the isolated test runners request
+      # a loopback ephemeral port without adding a second merged port entry.
+      # When HOST_IP is unset the mapping remains exactly the original
+      # host-port:container-port form, preserving Docker's default bind
+      # semantics. Tests set it to 127.0.0.1 and the host port to 0.
+      - "${HEIMDALLM_COMPOSE_DAEMON_HOST_IP:+${HEIMDALLM_COMPOSE_DAEMON_HOST_IP}:}${HEIMDALLM_COMPOSE_DAEMON_HOST_PORT:-${HEIMDALLM_PORT:-7842}}:${HEIMDALLM_PORT:-7842}"
     environment:
       # ── Heimdallm config ───────────────────────────────────────────────
       - GITHUB_TOKEN=${GITHUB_TOKEN}
@@ -134,7 +139,7 @@ services:
     container_name: heimdallm-web
     restart: unless-stopped
     ports:
-      - "${HEIMDALLM_WEB_PORT:-3000}:3000"
+      - "${HEIMDALLM_COMPOSE_WEB_HOST_IP:+${HEIMDALLM_COMPOSE_WEB_HOST_IP}:}${HEIMDALLM_COMPOSE_WEB_HOST_PORT:-${HEIMDALLM_WEB_PORT:-3000}}:3000"
     environment:
       # Upstream daemon URL inside the compose network. Host + port line
       # up with the heimdallm service above.
--- a/docker/scripts/lib/compose-test.sh
+++ b/docker/scripts/lib/compose-test.sh
@@ -1,0 +1,280 @@
+#!/bin/sh
+#
+# Shared Docker Compose isolation helpers for integration-test runners.
+#
+# This file is sourced by both test-local.sh (bash) and test-web.sh (POSIX sh),
+# so keep it portable and do not enable shell options here.
+
+compose_test_error() {
+    printf 'compose-test: %s\n' "$*" >&2
+}
+
+compose_test_init() {
+    if [ "${COMPOSE_TEST_INITIALIZED:-0}" = "1" ]; then
+        compose_test_error "isolation is already initialized"
+        return 1
+    fi
+
+    COMPOSE_TEST_KIND="${1:-}"
+    COMPOSE_TEST_REPO_ROOT="${2:-}"
+
+    case "$COMPOSE_TEST_KIND" in
+        ''|*[!a-z0-9-]*)
+            compose_test_error "invalid test kind '$COMPOSE_TEST_KIND'"
+            return 1
+            ;;
+    esac
+
+    if [ ! -d "$COMPOSE_TEST_REPO_ROOT" ]; then
+        compose_test_error "repository root does not exist: $COMPOSE_TEST_REPO_ROOT"
+        return 1
+    fi
+
+    COMPOSE_TEST_BASE_FILE="$COMPOSE_TEST_REPO_ROOT/docker/docker-compose.yml"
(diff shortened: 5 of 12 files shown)


Muriano (commented):
## 🤖 Heimdallm AI Review

Solid, well-targeted fix: per-run Compose project names, run-scoped container names, test-only named volumes (no more ./config bind), loopback ephemeral ports, fail-closed guards before every destructive `down -v`, plus a real fake-Docker regression suite wired into CI. The base compose port expression preserves production semantics when the new knobs are unset, and 8/8 of the new suite verifies that with a real `docker compose config`. Remaining problems are documentation drift and two degradations in the failure/diagnostic paths of test-local.sh — none of them undo the isolation goal.

### Issues

⚠️ **docker/docker-compose.test.yml:8** — Making `container_name` a `${VAR:?...}` required variable turns every documented direct invocation of the override file into a hard error, and the docs were not updated. `docs/e2e-test-guide.md` still instructs `docker compose -f docker/docker-compose.yml -f docker/docker-compose.test.yml up --build` (line 77), the same form for reading the API token (line 137) and for `down -v` (line 156). Impact: anyone following the E2E guide now gets `required variable HEIMDALLM_TEST_DAEMON_CONTAINER_NAME is missing a value: use an isolated Docker test runner` instead of a running stack. Fix: update docs/e2e-test-guide.md to use the new `make test-e2e` / `make test-web` / `./docker/scripts/test-local.sh` entrypoints (and note that the host port is now ephemeral, so `BASE` comes from the runner output, not `HEIMDALLM_PORT`).
⚠️ **docker/scripts/test-local.sh:60** — `resolve_daemon_url` is called immediately after `compose_test up -d` and, under `set -euo pipefail`, its failure aborts the script before the existing diagnostic path runs. If the daemon container starts and then exits (the most common local failure), `compose_test port` yields no binding, `compose_test_published_port` returns 1, and the script dies with only `compose-test: could not resolve published port for heimdallm:7842` — the `fail "Container starts and /health responds"` + `compose_test logs --tail=30` block in test_smoke/test_github/test_e2e becomes unreachable. Impact: loss of the container logs that used to be printed on startup failure. Fix: tolerate the failure (`if ! resolve_daemon_url; then ... compose_test logs --tail=30; return 1; fi`) so the log dump still happens.
🟡 **docker/scripts/test-local.sh:216** — The API token read was moved before the endpoint checks and `api_header` is built unconditionally as `X-Heimdallm-Token: ${api_token}`. When the token cannot be read (exec fails, /data/api_token not yet written), `api_token` is empty; curl drops a header with an empty value, so GET /config, /prs, /stats, /agents and /events all report as hard failures instead of being skipped, masking the real cause. The original code kept an explicit `skip "..." "no API token available"` path for exactly this case. Fix: after reading the token, skip (not fail) the authenticated checks when `api_token` is empty, or `fail` once with a clear "could not read API token" message before running them.
🟡 **docker/scripts/lib/compose-test.sh:232** — On cleanup failure `compose_test_cleanup` deliberately preserves `$COMPOSE_TEST_STATE_DIR` (marker + cleanup.log) but nothing ever reclaims it: the recovery command printed by `compose_test_display` only covers Docker resources, and there is no documented step to remove the leftover `${TMPDIR}/heimdallm-compose-test-*` directory or a startup sweep for stale ones. Impact: repeated failing runs accumulate orphan temp dirs and orphan Compose projects with no operator-facing runbook. Fix: print the exact `rm -rf <state dir>` recovery line alongside the marker path (and/or document the recovery procedure next to the new Makefile targets).

---

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Well-constructed test-isolation PR: unique per-run Compose project names, named volumes, ephemeral loopback ports, and a fake-docker regression suite correctly prevent test runs from destroying production Compose volumes. The guard logic in lib/compose-test.sh is defensive and portable, and the auth-token additions fix the smoke checks against authenticated endpoints. Only minor edge-case issues found: the new parameterized port-mapping interpolation requires Docker Compose v2 (breaks EOL docker-compose v1 users of the production file), and trap-cleanup ordering in test-web.sh can skip Docker cleanup in rare early-failure paths.

### Issues

🟡 **docker/docker-compose.yml:14** — The port mapping now uses ${VAR:+...} alternate-value interpolation with nested defaults. This is only supported by Docker Compose v2 (compose-go); legacy docker-compose v1 users of the production file will get an interpolation error even with the test env vars unset. v1 is EOL so impact is limited, but this silently raises the minimum Compose version for the production stack, not just tests.
🟡 **docker/scripts/test-web.sh:30** — cleanup() runs `rm -f "${INDEX_HTML:-}" ...` before compose_test_cleanup under `set -eu`. If the script dies between compose_test_init and the mktemp calls (signal or mktemp failure), the vars expand to empty strings and GNU `rm -f ""` exits non-zero, aborting the trap before compose_test_cleanup runs and before `exit "$exit_code"` — leaking the isolation state dir/marker and losing the intended exit code. Guarding the rm with `|| true` (as test-local.sh does) or moving it after compose_test_cleanup would close this window.
🟡 **docker/scripts/test-local.sh:58** — In cleanup(), `rm -f "$HTTP_BODY_FILE"` must succeed for compose_test_release's `rmdir "$COMPOSE_TEST_STATE_DIR"` to work; any leftover per-run file makes the rmdir fail silently (`2>/dev/null || true`), leaking the temp directory. compose_test_release could remove known per-run files or use a recursive-but-path-validated removal instead of relying on callers to empty the directory first.
🟡 **docker/scripts/lib/compose-test.sh:236** — compose_test_published_port picks the binding via `tail -n 1` of `docker compose port` output. With the forced 127.0.0.1 host IP this is fine today, but if the IP knobs are ever changed to a dual-stack bind, multiple lines are returned and the selected line is arbitrary. A comment or an explicit IPv4 filter would make the assumption durable.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-engineered isolation layer for Docker Compose integration tests. Unique project names, ephemeral loopback ports, marker-file identity assertions, and a comprehensive fake-fixture test suite. The production docker-compose.yml port syntax change is backward-compatible. Signal handling and cleanup are correct. No security or correctness issues found.

### Issues

🟡 **docker/docker-compose.yml:12** — The nested variable expansion '${HEIMDALLM_COMPOSE_DAEMON_HOST_IP:+${HEIMDALLM_COMPOSE_DAEMON_HOST_IP}:}${HEIMDALLM_COMPOSE_DAEMON_HOST_PORT:-${HEIMDALLM_PORT:-7842}}:${HEIMDALLM_PORT:-7842}' is correct and backward-compatible, but its complexity makes it hard to audit at a glance. A YAML anchor or x-extension field with a comment showing the expanded forms for production vs test would aid future maintainers.
🟡 **docker/docker-compose.test.yml:14** — Config mount changed from bind mount './config:/config' to named volume 'heimdallm-test-config:/config'. The container will see an empty /config directory. If any test scenario or future config-file feature depends on files being present in /config, it will silently see nothing. Intentional for isolation, but the rationale should be documented in a comment.
🟡 **docker/scripts/lib/compose-test.sh:168** — The compose_test() wrapper only applies the 'assert_safe' guard (with the 'refusing destructive Compose command' message) for the 'down' subcommand. Other destructive subcommands like 'rm' or 'kill' go through the functionally-equivalent assert_identity but without the explicit destructive-intent framing. Consider guarding 'rm|kill|stop' alongside 'down' for defense-in-depth.
🟡 **docker/scripts/tests/fixtures/curl:46** — The catch-all '-*) shift ;;' consumes any unrecognized flag as argument-less. If a caller passes a flag that takes a value (e.g., '-u user:pass' or '--header val') not listed in the explicit cases, the value would be misinterpreted as the URL positional. Acceptable for current test scope but fragile if test scenarios expand.
🟡 **docker/scripts/tests/fixtures/docker:16** — Pattern matching uses '$*' (space-joined) to detect subcommands like '" port web 3000"', but if any argument contains a space (e.g., a label filter), the pattern could false-match. Not a problem with current usage, but worth noting for robustness.

---

ivanmunozruiz:
He subido las correcciones del review en `78b0563`:

- Las rutas de fallo al resolver el puerto conservan el diagnóstico y los logs en smoke, GitHub, E2E y web.
- Un token vacío produce un único fallo claro y omite las seis comprobaciones autenticadas, sin cascada de 401 opacos.
- El cleanup fallido imprime comandos exactos y limitados al proyecto/run; el estado privado se recupera con `rm -f` de archivos conocidos + `rmdir`, nunca borrado recursivo.
- La guía usa los runners aislados, URL/puerto impresos y cabecera de autenticación para GET/SSE y mutaciones.
- Se añadió protección contra variables de `.env` que intenten desviar el fichero temporal fuera del directorio privado.

Validado con sintaxis Bash/POSIX/dash, `make test-compose-isolation` (8/8 y subcasos), Compose config y `git diff --check`. Queda pendiente una nueva revisión explícita de Muriano sobre este HEAD.

Muriano (commented):
## 🤖 Heimdallm AI Review

Re-review of HEAD (78b0563). The four previous findings are addressed in the diff: resolve_daemon_url failures now degrade gracefully with fail + show_container_logs in smoke/github/e2e; the API token is read once and the six authenticated checks are skipped (not failed) via api_token_ready when the token is missing or <32 chars; cleanup failure now prints exact, project-scoped recovery commands plus a non-recursive state-dir reclamation line (compose_test_display_state_cleanup); and the helper hardens the per-run file deletion against docker/.env clobbering internal variables. The port-isolation design (HOST_IP:+ prefix + host port 0, preserving the original mapping when unset) is sound and keeps the production default binding unchanged, and the new required container_name variables are what make direct invocation of the override file fail loudly. Only one minor residual defect found; no blockers. Note: the docs update to docs/e2e-test-guide.md falls in the truncated portion of the diff and could not be verified in this pass — it is not re-flagged, but confirm the guide no longer uses the direct 'docker compose -f docker/docker-compose.yml -f docker/docker-compose.test.yml' form and that BASE comes from the runner output rather than HEIMDALLM_PORT.

### Issues

🟡 **docker/scripts/test-local.sh:20** — compose_test_init creates the private state directory (mktemp -d under $TMPDIR) but 'trap cleanup EXIT' is installed roughly 40 lines later, after HTTP_BODY_FILE=$(compose_test_run_path http-body) and the color/helper definitions. Any failure or signal in that window (e.g. compose_test_run_path returning non-zero under 'set -euo pipefail', or Ctrl-C during sourcing) exits the script with no cleanup handler registered, leaking a ${TMPDIR}/heimdallm-compose-test-local.XXXXXX directory containing the marker file. Impact: silent accumulation of orphan run directories exactly in the failure paths the new isolation logic is meant to make recoverable. Fix: install the EXIT/HUP/INT/TERM traps immediately after compose_test_init succeeds (before computing HTTP_BODY_FILE), or have compose_test_init itself register a minimal release handler that the fuller cleanup() later replaces.

---

---

ivanmunozruiz:
Corregido el único punto del re-review en `755b425`: el runner instala un handler de cleanup de inicialización inmediatamente después de `compose_test_init`, antes de calcular `HTTP_BODY_FILE`; cuando ya existen todos los helpers se sustituye por el cleanup completo. Añadí una aserción de regresión sobre el orden de instalación del trap. Pasan sintaxis Bash/POSIX, `make test-compose-isolation` 8/8 y `git diff --check`.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of the Compose test-isolation PR at current HEAD (755b425). The single LOW finding from the previous pass is resolved: test-local.sh now installs an initialization-only cleanup handler (cleanup_initialization + HUP/INT/TERM traps) immediately after compose_test_init succeeds and before HTTP_BODY_FILE=$(compose_test_run_path http-body), so the previously unguarded window can no longer leak the private mktemp state directory; the full Docker-aware cleanup handler replaces it once all runtime state and helpers exist, and the exit code is preserved through both handlers. The rest of the changes remain sound: no direct invocation of docker-compose.test.yml (required container-name vars use :? guards), production volumes are no longer bind-mounted into test runs (heimdallm-test-config replaces ./config), destructive `down` is gated behind compose_test_assert_safe with marker+PID+prefix+charset validation, recovery paths use rm -f of known files plus rmdir instead of rm -rf, deletion targets are derived from validated identity rather than caller/.env-influenced variables, and the port knobs preserve the original host-port:container-port form when the IP var is unset. The new make targets and the compose-isolation CI job give the guards a regression suite that runs without starting containers. No unresolved reviewer concerns remain.

---

PR #602, opened 2026-07-21
feat(review): clearer never-approve downgrade note + min-severity threshold (closes #597)

vbuenog:
## Summary

Fixes the two actionable problems from #597. The `never_approve_with_issues` downgrade note said "issues were found", which a PR author misread as "no GitHub issue is linked to this PR" — the note now quotes the number of review findings and tells the author how to get an approval. And the downgrade fired on ANY finding (even a single 🟡 LOW nit), which silently removed auto-approval for repos with required approvals; a new companion setting `never_approve_min_severity` lets repos keep approvals flowing past low-severity nits while still withholding on medium+.

## How it works

- `daemon/internal/pipeline/pipeline.go` — `downgradeNote` const becomes `downgradeNoteFor(findingCount)`: *"Not approving: this review raised N finding(s) above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the findings and re-request a review to get an approval."* All three `AnnotateBodyForEvent` call sites (initial publish, `PublishPending` retry, NATS publish-worker) pass the stored finding count.
- `ReviewEvent` now takes the max finding severity (new helper `MaxIssueSeverity`, "" = no findings) plus the threshold. The downgrade only fires when at least one finding ranks ≥ `never_approve_min_severity`. Empty/unset threshold ranks as "low" (any finding downgrades) — exactly the pre-#597 behavior, so existing configs are unaffected.
- `daemon/internal/config/config.go` — `never_approve_min_severity` (`"low"|"medium"|"high"`) mirrors the base flag's plumbing: global `[ai]`, `[ai.orgs.*]`, `[ai.repos.*]` overrides with empty = inherit, resolved in `AIForRepo`, validated at load (`Validate()` rejects anything outside the enum at every scope).
- `daemon/cmd/heimdallm/main.go` — threaded into `pipeline.RunOptions`, exposed in `GET /config` (global map + org/repo override maps), and classified as a dynamic field in `configReloadRestartSnapshot` (no poller restart on change). `PATCH /config` needs no changes (generic TOML deep-merge).

## Scope

**In scope:** Problem 1 (note wording) and Problem 2 (severity threshold) from #597.
**Out of scope:**
- Problem 3 (per-user config divergence visible on the PR) — needs its own design discussion.
[...]

PR #658, opened 2026-08-03
feat!: expose never-approve severity threshold in the UI, default it to medium

Muriano:
## Problem

`never_approve_min_severity` landed in #602 (v0.7.8) but never got a UI control. The `never_approve_with_issues` toggle is exposed in three places — AI defaults, org detail, repo detail — and the threshold that qualifies it is exposed in none of them, so the only way to move it was hand-editing `config.toml`. Anyone looking for the setting in the app concluded it didn't exist.

Worse, the unset default was the strictest possible value. `severityRank("")` falls through to `default: return 1` — the rank of `low` — so **any** finding withheld the approval, including a `🟡 LOW` nit.

That combination is what makes review loops fail to converge with the current generation of models. Real example on a feature PR: a re-review pass reported *"the six previous findings are all fixed, CI green"* and still did not approve, because two `low` findings remained — where a godoc comment sat relative to its function, and a latent trap that needs an unrelated future change to become reachable. A strong model finds something like that on essentially every diff, and each fix commit creates fresh lines for the next pass to find more in. The approval was unreachable.

## Changes

**UI (the actual gap).** The threshold now sits next to the toggle it qualifies, at all three scopes:

- **AI defaults** — `Never approve — minimum severity` dropdown (`low` / `medium` / `high`). Disabled while the toggle is off, since the threshold has no effect then and a live-looking control that changes nothing reads as a bug.
- **Org detail / Repo detail** — `OverrideDropdown` with the same inherit / override / reset semantics as `Review mode`, including the repo→org→global inheritance label.

`AppConfig`, `OrgConfig` and `RepoConfig` gained the field with the existing sentinel-`copyWith` contract; the diff builders emit `''` to clear a string override (same asymmetry as `review_mode`, documented in `repo_diff.dart`).

**Default.** Unset now resolves to `medium` via `pipeline.DefaultNeverApproveMinSeverity` instead of falling through `severityRank` to `low`. All-low reviews keep their approval; the findings are still listed in the review body, they just no longer withhold it.

Resolution happens at the use site rather than by materialising a value into the operator's `config.toml`, so `""` keeps meaning "inherit" at org/repo scope and nothing rewrites a user's file.

No daemon API changes: `GET /config` already served the key at all three scopes (`main.go:1850`, `addCommonAIOverrideFields`), `PATCH /config` already accepted it, and the validator already bounded it to `low|medium|high`.

## Behaviour

| Findings | Before (unset) | After (unset) |
|---|---|---|
| none | APPROVE | APPROVE |
| max `low` | **COMMENT** | **APPROVE** |
| max `medium` | COMMENT | COMMENT |
| final `high` | REQUEST_CHANGES | REQUEST_CHANGES |

⚠️ **Breaking**, hence `feat!`: repos with `never_approve_with_issues` on and no explicit threshold change behaviour. `never_approve_min_severity = "low"` restores the old strictness. Retitle to `feat:` if you'd rather this went out as a patch.

## Testing

- `go test ./... -timeout 60s` — 17 packages, all pass. `go vet ./...` clean.
- `TestReviewEvent` updated for the new default, plus new cases: `medium`/`high` findings above the default threshold, and a whitespace-only threshold resolving to the default rather than to `low`.
- `flutter analyze` — no issues. `flutter test` — 234 pass.
- 5 new model tests: round-trip at all three scopes, unset-serves-the-default, empty-scoped-stays-null (inherit), `hasOverride`/`hasAiOverride` detection, diff-clears-with-`''`.
- 2 new widget tests on `ConfigScreen`: the dropdown renders the configured threshold and is enabled with the toggle on; it is disabled with the toggle off.
- `org_detail_screen_test.dart` extended to assert the override renders the org's stored value instead of the global.
- `make build-daemon` produces a working binary.

Not verified by running a second daemon against live config on purpose — spawning one would have it poll GitHub with the operator's token and write to the production SQLite. The decision function is covered directly by `TestReviewEvent` and the payload path by the existing `main_test.go` override-map tests.

### DIFF
--- a/daemon/internal/config/config.go
+++ b/daemon/internal/config/config.go
@@ -480,10 +480,10 @@ type AIConfig struct {
 
 	// NeverApproveMinSeverity is the minimum finding severity that triggers
 	// the NeverApproveWithIssues downgrade: "low", "medium" or "high".
-	// Empty defaults to "low" (any finding downgrades — backwards compat).
-	// With "medium", reviews whose findings are all low-severity nits still
-	// approve. Only meaningful when NeverApproveWithIssues is on.
-	// Overridable per-org and per-repo.
+	// Empty resolves to pipeline.DefaultNeverApproveMinSeverity ("medium"),
+	// so reviews whose findings are all low-severity nits still approve; set
+	// "low" explicitly to downgrade on any finding at all. Only meaningful
+	// when NeverApproveWithIssues is on. Overridable per-org and per-repo.
 	NeverApproveMinSeverity string `toml:"never_approve_min_severity"`
 
 	// ReviewResponse configures phase 2 of the PR review-state vigilance
@@ -1710,7 +1710,8 @@ func deleteLegacyAgentFieldAliases(m map[string]any, canonical string) {
 
 // validateNeverApproveMinSeverity bounds never_approve_min_severity to the
 // canonical severities at every scope. Empty means "inherit" (repo/org) or
-// "low" (global), so it is always accepted.
+// "use the default" (global — pipeline.DefaultNeverApproveMinSeverity), so it
+// is always accepted.
 func (c *Config) validateNeverApproveMinSeverity() error {
 	check := func(scope, v string) error {
 		switch v {
--- a/daemon/internal/pipeline/pipeline.go
+++ b/daemon/internal/pipeline/pipeline.go
@@ -1219,6 +1219,32 @@ func MaxIssueSeverity(issues []executor.Issue) string {
 	return rankToSeverity(maxRank)
 }
 
+// DefaultNeverApproveMinSeverity is the threshold applied when the operator
+// leaves never_approve_min_severity unset at every scope.
+//
+// It is "medium", not "low": a strong review model reports low-severity nits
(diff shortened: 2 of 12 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured PR that changes the default never_approve_min_severity from 'low' to 'medium' and exposes it in the Flutter UI. The daemon logic, config model plumbing, UI wiring, and tests are all consistent. One stale test comment is the only concrete issue.

### Issues

🟡 **daemon/internal/pipeline/pipeline_test.go:1705** — Stale comment: '// flag ON, explicit "low" threshold — same as default' is now wrong — the default changed to "medium", so "low" is no longer the same as default. Should read something like '// flag ON, explicit "low" threshold — any finding downgrades'.
🟡 **flutter_app/lib/features/config/config_screen.dart:777** — Uses `initialValue:` on DropdownButtonFormField. In standard Flutter the constructor parameter is `value:`, not `initialValue:` (the latter is inherited from FormField's interface but not forwarded as a named constructor parameter in most Flutter versions). If this compiles in the project's Flutter version it works, but it's fragile across upgrades — prefer `value:` for clarity and forward-compatibility.

---

PR #648, opened 2026-07-31
fix(build): stamp daemon version into binaries (closes #645)

vbuenog:
## Summary

The daemon's `main.version` defaults to `"dev"` and no build path ever passed `-ldflags "-X main.version=..."`, so the desktop app's Server → Status tab shows **"Heimdallm dev"** everywhere — local `make install-linux` builds *and* official goreleaser artifacts (goreleaser's explicit `ldflags: [-s -w]` replaces its default, which is what normally injects the version). This PR stamps the version on every build path. Fixes #645.

## How it works

- `daemon/Makefile` — accepts `VERSION ?= dev` and passes `-X main.version=$(VERSION)`; a bare `make build` still produces a `dev` binary.
- `daemon/.goreleaser.yml` / `daemon/.goreleaser-docker.yml` — ldflags now include `-X main.version={{.Version}}`, matching what `cli/.goreleaser.yml` already did.
- `Dockerfile.linux-verify` — new `ARG VERSION=dev` forwarded to `make build`.
- Root `Makefile` — new `GIT_VERSION := git describe --tags --always --dirty` (the *current* checkout; deliberately distinct from the existing `VERSION` var, which computes the **next** release tag for `release-local`). Passed to `build-daemon` (so `make dev` and the macOS packaging paths get stamped too) and to `verify-linux` as a `--build-arg`.

## Scope

**In scope:** every daemon build path — local Linux install (`verify-linux` → `install-linux`), local dev builds (`build-daemon`), goreleaser `.deb`/`.rpm` packages, goreleaser GHCR image, macOS `release-local`, CI release jobs, and (after review round 2) the self-hosted compose image (`docker/Dockerfile`, absorbing #649).

**Out of scope:** hardening the three pre-existing inline `${{ inputs.tag }}` interpolations elsewhere in `release.yml` (this PR's new steps use the safe `env:` form).

## Production impact & what to watch

**Runtime behavior change:** none functional — the only observable difference is the daemon health endpoint reporting a real version string instead of `"dev"`, which the Flutter Status tab renders. No new calls, config, or startup requirements; ldflags `-X` only sets an existing string variable.

**Rollout failure modes:** goreleaser template errors would fail the release build (not production) — `{{.Version}}` is the standard template already used by `cli/.goreleaser.yml` in this repo. `GIT_VERSION` falls back to `dev` outside a git checkout. `ARG VERSION` defaults to `dev`, so direct `docker build` invocations without the build-arg keep working.

[...]

PR #660, opened 2026-08-04
tui: Report the daemon's version in the Server tab

vbuenog:
The TUI's Server tab showed the CLI binary's own build version instead of the daemon's, and `cli/Makefile` never stamped `main.version` — so the row read `dev` no matter which daemon was running. Neither defect was addressed by the daemon version stamping in #648, which is why that fix appeared not to work.

**Changes:**
- `api`: add `Health` + `GetHealth()`; a missing `version` field is tolerated so daemons predating stamping don't error the whole fetch
- `tui`: render the daemon's `/health` version as `Version`, with the CLI's own version on a separate `CLI` row so a mismatch stays visible. `/health` is fetched first and best-effort since it needs no token, so the version still shows when the authenticated endpoints return 401
- `build`: stamp `main.version` in `cli/Makefile`, driven by the root Makefile's existing `GIT_VERSION` exactly as `build-daemon` drives `daemon/Makefile` (release builds already stamped via `cli/.goreleaser.yml`)

Before / after against a daemon reporting `0.8.0`:

```
  Version    dev                 Version    0.8.0
                                 CLI        0.8.0
```

Follow-up for the desktop app's equivalent staleness: #661

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ build-app:
 	cd flutter_app && flutter build $(FLUTTER_DEVICE) --release
 
 build-cli:
-	$(MAKE) -C cli build
+	$(MAKE) -C cli build VERSION=$(GIT_VERSION)
 
 # Flutter Web bundle, consumed by docker/Dockerfile.web (served via Nginx).
 # --base-href=/ matches the Nginx server block that expects assets at the root.
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -1,9 +1,14 @@
 BINARY = bin/heimdallm-cli
 CMD    = ./cmd/heimdallm-cli
 
-# Stamped into main.version. Overridden by the root Makefile with GIT_VERSION,
-# matching how daemon/Makefile is driven, so both binaries report the same
-# format (e.g. "0.8.0"). Left as "dev" for a bare `make -C cli build`.
+# Stamped into main.version. Left as "dev" for a bare `make -C cli build`.
+#
+# Invariant: the root Makefile passes GIT_VERSION here and to daemon/Makefile
+# from the same variable (build-cli and build-daemon), so the CLI and the daemon
+# always report the same format (e.g. "0.8.0"). If that variable is renamed or
+# its format changes, both call sites must move together — a CLI stamped from a
+# different source than the daemon makes the Server tab's Daemon/CLI comparison
+# meaningless, which is the whole point of showing them side by side.
 VERSION ?= dev
 
 .PHONY: build test test-race lint clean dev
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -1,11 +1,16 @@
 BINARY = bin/heimdallm-cli
 CMD    = ./cmd/heimdallm-cli
 
+# Stamped into main.version. Overridden by the root Makefile with GIT_VERSION,
+# matching how daemon/Makefile is driven, so both binaries report the same
+# format (e.g. "0.8.0"). Left as "dev" for a bare `make -C cli build`.
+VERSION ?= dev
+
 .PHONY: build test test-race lint clean dev
 
 build:
 	@mkdir -p $(dir $(BINARY))
-	go build -o $(BINARY) $(CMD)
+	go build -ldflags "-X main.version=$(VERSION)" -o $(BINARY) $(CMD)
 
 test:
 	go test -timeout 60s ./...
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -1,14 +1,9 @@
 BINARY = bin/heimdallm-cli
 CMD    = ./cmd/heimdallm-cli
 
-# Stamped into main.version. Left as "dev" for a bare `make -C cli build`.
-#
-# Invariant: the root Makefile passes GIT_VERSION here and to daemon/Makefile
-# from the same variable (build-cli and build-daemon), so the CLI and the daemon
-# always report the same format (e.g. "0.8.0"). If that variable is renamed or
-# its format changes, both call sites must move together — a CLI stamped from a
-# different source than the daemon makes the Server tab's Daemon/CLI comparison
-# meaningless, which is the whole point of showing them side by side.
+# Stamped into main.version; "dev" for a bare `make -C cli build`. The root
+# Makefile must keep passing the same GIT_VERSION to build-cli and build-daemon,
+# or the Server tab's Daemon/CLI rows stop being comparable.
 VERSION ?= dev
 
 .PHONY: build test test-race lint clean dev
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -4,7 +4,15 @@ CMD    = ./cmd/heimdallm-cli
 # Stamped into main.version; "dev" for a bare `make -C cli build`. The root
 # Makefile must keep passing the same GIT_VERSION to build-cli and build-daemon,
 # or the Server tab's Daemon/CLI rows stop being comparable.
+#
+# ?= is not enough on its own: a command-line VERSION= counts as defined even
+# when empty, so it would stamp an empty version — and cobra then drops the
+# --version flag entirely, rather than reporting "dev". `override` is required
+# because a plain assignment here would lose to the command line.
 VERSION ?= dev
+ifeq ($(strip $(VERSION)),)
+override VERSION := dev
+endif
 
 .PHONY: build test test-race lint clean dev
 
--- a/cli/internal/api/client.go
+++ b/cli/internal/api/client.go
@@ -88,19 +88,21 @@ const (
 	versionDisplayMax = 48
 )
 
-// displayField reduces a daemon-supplied string to something safe to print: both
-// Status and Version reach a TUI row and a terminal line verbatim, where a
-// control byte would corrupt the layout and an unbounded string would overrun
-// it. The cap counts runes, not bytes, so multibyte input is not truncated far
-// short of it.
+// DisplayText reduces a daemon-supplied string to something safe to print. Every
+// such string reaching a TUI row or a terminal line goes through here: a control
+// byte would corrupt the layout and an unbounded string would overrun it. The cap
+// counts runes, not bytes, so multibyte input is not truncated far short of it.
+//
+// unicode.IsPrint already excludes tab, newline and carriage return, so they need
+// no separate test.
 //
 // Note what this does NOT do: dropping ESC is what defuses an ANSI sequence, so
 // the inert remainder ("[31m") is deliberately kept rather than parsed out.
-func displayField(s string, maxRunes int) string {
+func DisplayText(s string, maxRunes int) string {
 	var b strings.Builder
 	kept := 0
 	for _, r := range s {
-		if r == '\t' || r == '\n' || r == '\r' || !unicode.IsPrint(r) {
+		if !unicode.IsPrint(r) {
 			continue
 		}
 		if kept >= maxRunes {
@@ -118,7 +120,7 @@ func (h *Health) DisplayStatus() string {
 	if h == nil {
 		return ""
 	}
-	return displayField(h.Status, statusDisplayMax)
+	return DisplayText(h.Status, statusDisplayMax)
 }
 
 // DisplayVersion is the same treatment for Version, which travels the same path
@@ -127,7 +129,7 @@ func (h *Health) DisplayVersion() string {
 	if h == nil {
 		return ""
 	}
-	return displayField(h.Version, versionDisplayMax)
+	return DisplayText(h.Version, versionDisplayMax)
 }
 
 // GetHealth returns the daemon's health payload, including the version it was
--- a/cli/internal/api/client.go
+++ b/cli/internal/api/client.go
@@ -70,6 +70,29 @@ func (c *Client) Health() error {
 	return err
 }
 
+// Health is the subset of /health the dashboard renders. Version is empty for
+// daemons built before version stamping (they omit the field), so callers must
+// treat "" as unknown rather than an error.
+type Health struct {
+	Status    string    `json:"status"`
+	Version   string    `json:"version"`
+	StartedAt time.Time `json:"started_at"`
+}
+
+// GetHealth returns the daemon's health payload, including the version it was
+// built from — the authoritative server version, distinct from this CLI's.
+func (c *Client) GetHealth() (*Health, error) {
+	data, err := c.do("GET", "/health")
+	if err != nil {
+		return nil, err
+	}
+	var h Health
+	if err := json.Unmarshal(data, &h); err != nil {
+		return nil, fmt.Errorf("decoding health: %w", err)
+	}
+	return &h, nil
+}
+
 // PR types matching daemon API responses.
 
 type Review struct {
--- a/cli/internal/api/client.go
+++ b/cli/internal/api/client.go
@@ -64,9 +64,10 @@ func (c *Client) do(method, path string) ([]byte, error) {
 	return data, nil
 }
 
-// Health checks daemon connectivity.
+// Health checks daemon connectivity. Shares GetHealth's transport, so a
+// degraded-but-reachable daemon is not reported as a connectivity failure.
 func (c *Client) Health() error {
-	_, err := c.do("GET", "/health")
+	_, err := c.GetHealth()
 	return err
 }
 
@@ -81,10 +82,29 @@ type Health struct {
 
 // GetHealth returns the daemon's health payload, including the version it was
 // built from — the authoritative server version, distinct from this CLI's.
+//
+// Deliberately does not go through do(), which errors on any status >= 400: the
+// daemon answers 503 with status "degraded" whenever a nats/sqlite/last_poll
+// check fails, and that response still carries the version — precisely the state
+// in which knowing the running build matters most. 200 and 503 both decode; the
+// caller reads Status to tell them apart. Everything else (401, a proxy's HTML
+// 502, an undecodable body) stays an error.
 func (c *Client) GetHealth() (*Health, error) {
-	data, err := c.do("GET", "/health")
+	req, err := c.newRequest("GET", "/health", nil)
 	if err != nil {
 		return nil, err
+	}
+	resp, err := c.HTTPClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("request failed: %w", err)
+	}
+	defer resp.Body.Close()
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("reading response: %w", err)
+	}
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusServiceUnavailable {
+		return nil, fmt.Errorf("HTTP %d: %s", resp.StatusCode, strings.TrimSpace(string(data)))
 	}
 	var h Health
 	if err := json.Unmarshal(data, &h); err != nil {
--- a/cli/internal/api/client.go
+++ b/cli/internal/api/client.go
@@ -81,29 +81,53 @@ type Health struct {
 	StartedAt time.Time `json:"started_at"`
 }
 
-// statusDisplayMax bounds the rendered status so one field cannot overrun a TUI
-// row. Comfortably above the daemon's "ok"/"degraded".
-const statusDisplayMax = 32
+// Caps for the rendered fields, in runes. Both sit comfortably above what the
+// daemon emits — "ok"/"degraded", and versions like "0.8.0-3-g86f49fa-dirty".
+const (
+	statusDisplayMax  = 32
+	versionDisplayMax = 48
+)
 
-// DisplayStatus returns Status reduced to something safe to print: the daemon's
-// value reaches a TUI row and a terminal line verbatim, where a control byte
-// would corrupt the layout and an unbounded string would overrun it. Today the
+// displayField reduces a daemon-supplied string to something safe to print: both
+// Status and Version reach a TUI row and a terminal line verbatim, where a
+// control byte would corrupt the layout and an unbounded string would overrun
+// it. The cap counts runes, not bytes, so multibyte input is not truncated far
+// short of it.
+//
+// Note what this does NOT do: dropping ESC is what defuses an ANSI sequence, so
+// the inert remainder ("[31m") is deliberately kept rather than parsed out.
+func displayField(s string, maxRunes int) string {
+	var b strings.Builder
+	kept := 0
+	for _, r := range s {
+		if r == '\t' || r == '\n' || r == '\r' || !unicode.IsPrint(r) {
+			continue
+		}
+		if kept >= maxRunes {
+			break
+		}
+		b.WriteRune(r)
+		kept++
+	}
+	return b.String()
+}
+
+// DisplayStatus returns Status bounded and stripped of control bytes. Today the
 // daemon only emits "ok" and "degraded", so this is about not depending on that.
 func (h *Health) DisplayStatus() string {
 	if h == nil {
 		return ""
 	}
-	var b strings.Builder
-	for _, r := range h.Status {
-		if r == '\t' || r == '\n' || r == '\r' || !unicode.IsPrint(r) {
-			continue
-		}
-		if b.Len() >= statusDisplayMax {
-			break
-		}
-		b.WriteRune(r)
+	return displayField(h.Status, statusDisplayMax)
+}
+
+// DisplayVersion is the same treatment for Version, which travels the same path
+// to the same rows and was previously printed raw and unbounded.
+func (h *Health) DisplayVersion() string {
+	if h == nil {
+		return ""
 	}
-	return b.String()
+	return displayField(h.Version, versionDisplayMax)
 }
 
 // GetHealth returns the daemon's health payload, including the version it was
--- a/cli/internal/api/client.go
+++ b/cli/internal/api/client.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"strings"
 	"time"
+	"unicode"
 )
 
 const DefaultHost = "http://localhost:7842"
@@ -78,6 +79,31 @@ type Health struct {
 	Status    string    `json:"status"`
 	Version   string    `json:"version"`
 	StartedAt time.Time `json:"started_at"`
+}
+
+// statusDisplayMax bounds the rendered status so one field cannot overrun a TUI
+// row. Comfortably above the daemon's "ok"/"degraded".
+const statusDisplayMax = 32
+
+// DisplayStatus returns Status reduced to something safe to print: the daemon's
+// value reaches a TUI row and a terminal line verbatim, where a control byte
+// would corrupt the layout and an unbounded string would overrun it. Today the
+// daemon only emits "ok" and "degraded", so this is about not depending on that.
+func (h *Health) DisplayStatus() string {
+	if h == nil {
+		return ""
+	}
+	var b strings.Builder
+	for _, r := range h.Status {
+		if r == '\t' || r == '\n' || r == '\r' || !unicode.IsPrint(r) {
+			continue
+		}
+		if b.Len() >= statusDisplayMax {
+			break
+		}
+		b.WriteRune(r)
+	}
+	return b.String()
 }
 
 // GetHealth returns the daemon's health payload, including the version it was
--- a/cli/internal/api/client_test.go
+++ b/cli/internal/api/client_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 )
@@ -149,5 +150,36 @@ func TestGetHealthErrorsOnUnexpectedStatusAndGarbage(t *testing.T) {
 				t.Errorf("GetHealth on %s = nil error, want error", tc.name)
 			}
 		})
+	}
+}
+
+// Status is rendered into a TUI row and a terminal line, so control bytes would
+// corrupt the layout and an unbounded string would overrun it. The daemon only
+// emits "ok"/"degraded" today; this keeps a future value from breaking callers.
+func TestHealthDisplayStatus(t *testing.T) {
+	cases := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{"plain", "degraded", "degraded"},
+		{"empty", "", ""},
+		// Dropping ESC is what defuses the sequence; the remaining "[31m" is
+		// inert literal text, so it is deliberately kept rather than parsed out.
+		{"defuses ANSI by dropping ESC", "deg\x1b[31mraded\n", "deg[31mraded"},
+		{"strips tabs and CR", "de\tgra\rded", "degraded"},
+		{"caps length", strings.Repeat("x", 80), strings.Repeat("x", 32)},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			h := &Health{Status: tc.in}
+			if got := h.DisplayStatus(); got != tc.want {
+				t.Errorf("DisplayStatus() = %q, want %q", got, tc.want)
+			}
+		})
+	}
+	var nilH *Health
+	if got := nilH.DisplayStatus(); got != "" {
+		t.Errorf("nil receiver DisplayStatus() = %q, want empty", got)
 	}
 }
--- a/cli/internal/api/client_test.go
+++ b/cli/internal/api/client_test.go
@@ -47,3 +47,42 @@ func TestStreamEventsReturnsWhenContextCanceledWithBlockedSend(t *testing.T) {
 		t.Fatal("StreamEvents did not return after context cancellation")
 	}
 }
+
+func TestGetHealthParsesDaemonVersion(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != "/health" {
+			t.Errorf("unexpected path %q", r.URL.Path)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		fmt.Fprint(w, `{"status":"ok","version":"0.8.0","started_at":"2026-08-04T09:20:26Z"}`)
+	}))
+	t.Cleanup(srv.Close)
+
+	h, err := New(srv.URL, "").GetHealth()
+	if err != nil {
+		t.Fatalf("GetHealth: %v", err)
+	}
+	if h.Version != "0.8.0" {
+		t.Errorf("Version = %q, want 0.8.0", h.Version)
+	}
+	if h.Status != "ok" {
+		t.Errorf("Status = %q, want ok", h.Status)
+	}
+}
+
+// A daemon built before version stamping omits the field entirely; GetHealth
+// must succeed with an empty Version rather than failing the whole fetch.
+func TestGetHealthToleratesMissingVersion(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprint(w, `{"status":"ok"}`)
+	}))
+	t.Cleanup(srv.Close)
+
+	h, err := New(srv.URL, "").GetHealth()
+	if err != nil {
+		t.Fatalf("GetHealth: %v", err)
+	}
+	if h.Version != "" {
+		t.Errorf("Version = %q, want empty", h.Version)
+	}
+}
--- a/cli/internal/api/client_test.go
+++ b/cli/internal/api/client_test.go
@@ -183,3 +183,47 @@ func TestHealthDisplayStatus(t *testing.T) {
 		t.Errorf("nil receiver DisplayStatus() = %q, want empty", got)
 	}
 }
+
+// The cap is documented in runes, so it must count runes: a byte-based check
+// truncated multibyte input at ~11 characters and could still emit 32+ bytes.
+func TestDisplayFieldsCapInRunes(t *testing.T) {
+	multibyte := strings.Repeat("é", 80) // 2 bytes per rune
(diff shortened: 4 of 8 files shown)


Muriano (commented):
## 🤖 Heimdallm AI Review

Well-scoped change: the Server tab now reports the daemon's version from /health and shows the CLI's own build version in a separate row, with the cli/Makefile stamping main.version (the var exists in cli/cmd/heimdallm-cli/main.go, and the root Makefile's GIT_VERSION is defined, so the wiring is correct and mirrors daemon/Makefile). Tests cover the happy path, the missing-field back-compat path, the render rows and the Update() plumbing. One real defect: the daemon returns HTTP 503 from /health when it is degraded (daemon/internal/server/handlers.go:394-418) and Client.do treats >=400 as an error, so GetHealth fails exactly in the degraded case even though the body carries the version. Non-blocking observations: (1) api.Health.StartedAt is parsed but never used — the Uptime row still derives from d.startTime (the CLI's own start), so 'Uptime' in the Server tab remains the CLI's uptime, not the server's, which now reads slightly inconsistently next to a genuinely server-side Version row; using health.StartedAt there would be a natural follow-through; (2) once fetched, d.daemonVersion is never cleared, so after the daemon stops the tab shows '● stopped' alongside a stale version — arguably 'last known' and acceptable, but worth a deliberate decision; (3) the GetHealth error is swallowed with no logging, so a persistent '(unknown)' gives no diagnostic trail.

### Issues

⚠️ **cli/internal/tui/dashboard.go:151** — GetHealth goes through Client.do, which returns an error for any status >= 400 (cli/internal/api/client.go:61). The daemon's /health handler responds 503 with status "degraded" whenever the nats, sqlite or last_poll check fails (daemon/internal/server/handlers.go:401-406), while still including the version field in the body. Failure scenario: polling stalls so last_poll goes stale, the daemon answers 503 {"status":"degraded","version":"0.8.0",...}; GetHealth returns an error, msg.health stays nil and the Server tab renders Version as "(unknown)" for a daemon that is up and reporting its version — precisely the state in which an operator needs to know which build is running. Fix: fetch /health with a path that decodes the body on 503 as well (e.g. a raw request in GetHealth, or a do variant that tolerates 503 for /health) and only error on transport failures or undecodable payloads.

---

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured PR that adds daemon version reporting to the TUI. Good test coverage, proper handling of missing version from older daemons, and correct best-effort semantics for the unauthenticated /health call. Only minor issues.

### Issues

🟡 **cli/internal/tui/dashboard.go:313** — When the daemon disconnects (health fetch fails), daemonVersion retains its last-known value because msg.health is nil and the field is not cleared. The Server tab will show the stale version alongside a 'disconnected' status badge, and if the daemon restarts with a different version, the old version persists until a successful health fetch. Consider clearing daemonVersion when connected becomes false, or documenting this as intentional last-known-value behavior.
🟡 **cli/internal/api/client.go:70** — The existing method Health() (returns error) and the new GetHealth() (returns *Health, error) both hit the same /health endpoint but with different response handling. The naming asymmetry is confusing — Health() discards the body while GetHealth() parses it. Consider whether Health() callers (the SSE watchdog) could use GetHealth() instead, eliminating the duplicate method and extra HTTP round-trip if both ever run close together.
🟡 **cli/internal/tui/dashboard.go:1108** — The label asymmetry ('Version' for daemon, 'CLI' for CLI binary) is slightly inconsistent — 'Version' reads as the overall version while 'CLI' reads as a component name. Consider 'Server' and 'CLI', or 'Daemon' and 'CLI' for clearer parallel labeling.

---

Muriano (commented):
## 🤖 Heimdallm AI Review

El PR sigue siendo correcto en lo esencial: separa la versión del daemon (/health) de la del CLI, tolera daemons antiguos que omiten el campo, sella la versión del CLI vía ldflags desde el Makefile raíz, y añade tests razonables (parseo, campo ausente, render del tab Server, aplicación del dataMsg antes de la rama de error). El fetch best-effort antes de los endpoints autenticados es un buen detalle: la versión sobrevive a un 401.

Sin embargo, el único hallazgo de la revisión previa NO se ha abordado: el diff de `GetHealth` es idéntico al de la pasada anterior (sigue usando `c.do`, que en cli/internal/api/client.go:59-61 devuelve error para cualquier status >= 400), y `daemon/internal/server/handlers.go:401-418` responde 503 con `status: "degraded"` incluyendo el campo `version`. No hay commit nuevo ni respuesta del autor descartando el punto, así que se re-flaggea.

Nits (no bloquean): el comentario de `cli/Makefile` menciona que el formato coincide con daemon/Makefile — conviene que ambos sigan usando la misma variable si el root Makefile cambia; y `TestServerSectionShowsDaemonVersionNotCLIVersion` localiza la fila haciendo `strings.Contains(line, "Version")` sobre la primera coincidencia, lo cual funciona hoy solo porque la fila del CLI se etiqueta "CLI" — es frágil si se renombra la etiqueta a algo como "CLI Version".

### Issues

⚠️ **cli/internal/api/client.go:85** — GetHealth sigue delegando en Client.do, que devuelve error para cualquier status >= 400 (client.go:59-61). El handler /health del daemon responde 503 con status "degraded" cuando falla cualquiera de los checks nats/sqlite/last_poll (daemon/internal/server/handlers.go:401-406), pero el cuerpo SÍ incluye el campo version (handlers.go:411-413). Escenario de fallo concreto: el polling se atasca y last_poll queda obsoleto; el daemon responde 503 {"status":"degraded","version":"0.8.0",...}; GetHealth devuelve error, msg.health queda nil en fetchData (dashboard.go:157-159) y el tab Server pinta Version como "(unknown)" para un daemon que está arriba y reportando su build — justo el estado en el que el operador necesita saber qué versión corre. Fix: obtener /health por una vía que decodifique el cuerpo también en 503 (una petición cruda dentro de GetHealth, o una variante de do que tolere 503 para /health) y fallar solo en errores de transporte o payload indescifrable; el status "degraded" ya viaja en Health.Status para que la UI lo refleje.

---

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Re-review: the diff is unchanged from the prior review. All three previous LOW findings (stale daemonVersion on disconnect, Health/GetHealth duplication, Version/CLI label asymmetry) remain as-is, but none are correctness bugs — they are minor polish items the author may have intentionally deferred. No new issues introduced.

---

vbuenog:
All findings addressed in 7b8be9b.

**@Muriano — MEDIUM, /health 503 (re-flagged): accepted, fixed.** Verified against `daemon/internal/server/handlers.go:401-418`: 503 + `status:"degraded"` with `version` still in the body, and `do()` errors on `>= 400`. `GetHealth` no longer goes through `do()` — it issues its own request and decodes 200 and 503 alike, leaving 401, a proxy's HTML 502 and undecodable bodies as errors. Test `TestGetHealthDecodesDegraded503` fails on the old code with exactly the reported symptom.

Your non-blocking observations, all fixed rather than deferred:
- **(1) `StartedAt` parsed but unused / Uptime inconsistency** — the Uptime row now derives from `/health` `started_at`, so the Server section reports the *daemon's* uptime. The CLI's own uptime stays in the footer status bar. Verified live: 5h46m for a daemon restarted 5h46m earlier, against a CLI started seconds before.
- **(2) `daemonVersion` never cleared** — decided deliberately, and against keeping it: a failed fetch now clears version and start time, so a last-known value cannot sit next to a stopped badge or outlive a restart onto a different version.
- **(3) error swallowed with no diagnostic trail** — the failure is stored and rendered on the Daemon row: `Daemon (unknown) — request failed: dial tcp …`.

Nits: the `cli/Makefile` comment now states the invariant that `build-cli` and `build-daemon` must keep stamping from the same variable. And you were right about `TestServerSection…` being fragile — the row lookup is now anchored on the label via a `serverRow` helper, so renaming a row cannot make another row's assertion vacuous.

**@sergiotejon — all three LOW findings fixed** (they were not deliberately deferred):
- stale `daemonVersion` on disconnect → cleared, per above
- `Health()` / `GetHealth()` duplication → `Health()` now delegates to `GetHealth()`, so there is one `/health` path. Note the behaviour change this implies: a degraded-but-reachable daemon is no longer a watchdog connectivity failure, which I think is correct — it is up.
- label asymmetry → rows are now `Daemon` and `CLI`, your suggested parallel pair.

Rendered against a live daemon:

```
  Status     ● running          Status     ● stopped
  Daemon     0.8.0              Daemon     (unknown) — request failed: dial tcp …
  CLI        0.8.0-dirty        CLI        0.8.0-dirty
  Uptime     5h46m58s           Uptime     (unknown)
```

`make test-cli` and `make lint-cli` pass locally; 9 new tests in this round.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previous findings have been addressed thoroughly. The revision is clean: GetHealth correctly handles 200/503 while rejecting other statuses, Health() delegates to GetHealth() eliminating duplication, version clearing on disconnect prevents stale state, and the Server section now shows distinct Daemon/CLI rows with daemon uptime. Test coverage is comprehensive. No new issues found.

---

Muriano (commented):
## 🤖 Heimdallm AI Review

El hallazgo MEDIUM anterior está correctamente resuelto: GetHealth ya no pasa por do(), emite su propia petición y decodifica 200 y 503 (TestGetHealthDecodesDegraded503), manteniendo 401/502-HTML/cuerpos indescifrables como error (TestGetHealthErrorsOnUnexpectedStatusAndGarbage). También verifiqué las tres observaciones no bloqueantes: Uptime deriva ahora de /health started_at con fallback (unknown), un fetch fallido limpia daemonVersion/daemonStartedAt (TestUnreachableDaemonClearsVersion), y el error se persiste y se pinta truncado en la fila Daemon. El helper serverRow ancla el lookup en la etiqueta, eliminando la fragilidad que señalé. Confirmé además que el stamping de versión es coherente en los dos caminos de build: cli/.goreleaser.yml ya usa -X main.version={{.Version}} y cli/cmd/heimdallm-cli/main.go:9 declara var version = "dev", así que el nuevo -ldflags del cli/Makefile stampea el símbolo real (nada silenciosamente inerte) y comparte GIT_VERSION con build-daemon (Makefile:39,42).

Queda un efecto colateral del cambio de Health() que sí es observable fuera del dashboard: el comando `status` (cli/internal/cli/status.go:16) usaba Health() como puerta y ahora un daemon degradado la atraviesa, imprimiendo "Status: online" sin matiz. Lo reporto abajo.

Nits/observaciones no bloqueantes: (1) en el dashboard, un daemon degradado (503) sigue pintando el badge "● running" porque d.connected solo depende de los endpoints autenticados — el Status "degraded" que ahora viaja en api.Health no se usa en la UI; podría alimentar un badge intermedio. (2) time.Since(d.daemonStartedAt) puede dar una duración negativa si el reloj del host del daemon va adelantado respecto al del CLI (escenario remoto); cosmético. (3) fetchData añade una petición /health por tick de refresco; despreciable pero real. (4) El comentario-invariante del cli/Makefile es largo para lo que documenta, aunque el contenido es correcto y útil.

### Issues

⚠️ **cli/internal/api/client.go:68** — Health() ahora delega en GetHealth(), que trata 503 como éxito. Eso es lo correcto para el dashboard, pero cambia el comportamiento de `heimdallm-cli status`, que usa Health() como única puerta y luego imprime literalmente "Status: online" (cli/internal/cli/status.go:16,32). Escenario concreto: el polling se atasca o cae NATS; el daemon responde 503 {"status":"degraded",...}; antes de este PR `status` fallaba (con el mensaje impreciso "daemon unreachable"), ahora pasa la puerta y el operador ve "Heimdallm Daemon Status / Status: online" sin ninguna indicación de degradación — output sin ambigüedad pero incorrecto justo en el estado que importa. Fix: en status.go llamar a GetHealth() y derivar la línea del payload (p. ej. imprimir h.Status, "online"/"degraded", y opcionalmente h.Version), reservando el error solo para el fallo de transporte.

---

---

vbuenog:
Round 2 addressed in 86f49fa.

**@Muriano — MEDIUM, `status` regression: accepted, and a good catch.** You are right that this was a knock-on of my own change: `Health()` tolerating 503 fixed the dashboard and broke `heimdallm-cli status`, which used it as its only gate and then printed a literal `online` (`status.go:16,32`). A daemon with stalled polling or NATS down reported healthy in the one command an operator runs to check. It now calls `GetHealth()` and derives the line from the payload, printing the daemon's own word, plus the version while it is there. `statusLine` is a pure function so the mapping is tested rather than eyeballed.

Against the live daemon:
```
  Status:       online
  Version:      0.8.0
```
and a 503 payload yields `Status: degraded`.

**Nit (1), degraded badge — fixed, and it was the same blind spot.** `api.Health.Status` was parsed and unused, and a degraded daemon answers every endpoint, so `d.connected` alone rendered `● running`. Verified against a stub serving 503 on `/health` and 401 elsewhere:
```
  Status     ● degraded
  Daemon     0.8.0
  CLI        0.8.0-dirty
  Uptime     5h58m43s
```
That is the exact state your original MEDIUM was about, now legible end to end.

**Nit (2), negative uptime — fixed.** Reproduced first (`Uptime -1h59m59s` with a started_at two hours ahead), then clamped at zero, with a test.

**Nit (4), comment length — fixed**, trimmed to two lines.

**Nit (3), extra `/health` per tick — deliberately not changed.** One unauthenticated request against localhost every 10s, and it is precisely what lets the version and the degraded state survive a 401 on the authenticated endpoints. Happy to revisit if you disagree.

Also confirming your independent check on the stamping paths matches mine: `cli/.goreleaser.yml` already stamped `-X main.version`, so only the local `make` path was inert.

`make test-cli` and `make lint-cli` pass; 8 further tests this round.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review pass 3. Both previously reported findings are addressed in the visible diff and I verified the fixes rather than taking the comment at face value.

(1) MEDIUM `status` regression — resolved. `cli/internal/cli/status.go` no longer uses `Health()` as a pass/fail gate: it calls `GetHealth()` and derives the printed line from the payload via the pure `statusLine()`, so a 503 `{"status":"degraded"}` daemon now prints `Status: degraded` instead of a flat `online`, and the transport error path still reports `daemon unreachable`. The mapping table in `status_test.go` covers ok/degraded/unknown-word/empty/nil, including the nil-receiver branch, so the exact scenario I reported is now covered by a failing-on-old-code test.

(2) Degraded badge — resolved. `serverStatusBadge()` gives the daemon's own word precedence over mere connectedness, which closes the loop on the original MEDIUM: the state is now legible in both the TUI and the CLI. `TestServerStatusBadgeReflectsDegraded` asserts both the presence of `degraded` and the absence of `running`, so it cannot pass vacuously.

(3) Negative uptime — resolved, clamped at zero with `TestServerSectionUptimeNeverNegative`.

On the deliberately-declined nit (extra unauthenticated `/health` per 10s tick): your reasoning holds. That request is what keeps the version and the degraded state visible when the authenticated endpoints 401, and one localhost GET per tick is not a cost worth trading for that. No objection.

Non-blocking observations, none of which clear the reporting bar:
- `statusLine()` and `serverStatusBadge()` both render `h.Status` verbatim. Today the daemon only emits `ok`/`degraded`, so this is fine; it becomes a rendering concern only if the field ever carries free-form text or control bytes, which would need an unrelated daemon change first. Latent, not an issue.
- `cli/Makefile` uses `VERSION ?= dev`, but a command-line `VERSION=` from the root Makefile counts as defined-even-if-empty, so an empty `GIT_VERSION` stamps an empty version and the CLI row falls back to `(unknown)` rather than `dev`. Cosmetic, and the invariant comment you added already points at the coupling.
- The Daemon row truncates the health error at 60 runes via `truncateRunes`, which is the right call for layout; long `dial tcp` errors lose the tail but the leading cause survives, which is the useful part.

Good convergence: the diff now separates three genuinely independent facts (daemon build, CLI build, daemon-reported health) that were previously conflated into one `Version` row and one `connected` boolean, and each of them has a test that fails on the pre-PR behaviour. Nothing in this pass warrants withholding approval.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previously flagged issues have been addressed. The 503 degraded handling, status regression, negative uptime, stale version clearing, label asymmetry, and badge fixes are all resolved with thorough tests. The extra /health per tick was deliberately kept with sound justification. The code is clean and well-tested.

---

ivanmunozruiz (commented):
## 🤖 Heimdallm AI Review

Version reporting is well implemented, but the Server status can still hide a known degraded state.

### Issues

⚠️ **cli/internal/tui/dashboard.go:1108** — serverStatusBadge checks !connected before daemonStatus. fetchData can successfully receive a 503 health payload with status "degraded" and then fail on ListPRs/config/stats; Update preserves daemonStatus but sets connected=false, causing the badge to report "stopped" instead of the authoritative "degraded" state. Check the current health status first or track health reachability separately.

---

---

vbuenog:
Round 3 in 44a51fd.

**@ivanmunozruiz — MEDIUM, badge ordering: accepted, and you are right.** Reproduced before fixing: `dataMsg{health: degraded, err: …}` leaves `daemonStatus="degraded"` with `connected=false`, and the badge rendered `● stopped` — hiding a known-degraded daemon behind an outright wrong state, the opposite of what surfacing it was for.

Fixed by your first suggestion, since `/health` answering at all *is* the reachability signal: health is now consulted before `!connected`, and the disconnected badge only applies when `/health` itself failed. `connected` only ever tracked the authenticated endpoints, so it was never the right authority for up/down once a health payload exists. Two tests: `TestServerStatusBadgeDegradedOutranksDisconnected` (the exact scenario you described) and `TestServerStatusBadgeStoppedWhenHealthUnreachable` (so the fix cannot swing too far and call an unreachable daemon healthy).

**@Muriano — thanks for the pass-3 verification.** Both remaining observations fixed:
- **`h.Status` rendered verbatim** — `api.Health.DisplayStatus()` strips control bytes and caps at 32 runes, shared by the badge and `statusLine` so the two cannot drift. Worth noting what it does *not* do: dropping ESC is what defuses an ANSI sequence, so `deg\x1b[31mraded` becomes `deg[31mraded` — inert literal text, deliberately kept rather than parsed out. Asserted that way.
- **`VERSION ?= dev` vs an empty command-line `VERSION=`** — real, and the symptom is worse than reported: cobra drops the `--version` flag *entirely* rather than showing `(unknown)`. My first fix did not work, because a plain assignment inside the `ifeq` still loses to the command line; it needs `override`. Verified across all three paths — `VERSION=` → `dev`, `make build-cli` → `0.8.0-3-g86f49fa-dirty`, bare `make -C cli build` → `dev`.
- **60-rune truncation of the health error** — left as-is, per your assessment.

`make test-cli` and `make lint-cli` pass.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All prior review findings have been addressed across three rounds. The badge ordering fix correctly prioritizes /health's word over the authenticated-endpoint connectedness flag. DisplayStatus sanitizes untrusted status strings. The Makefile's `override` handles the empty VERSION= edge case. Test coverage is thorough, including the degraded-outranks-disconnected scenario, negative uptime clamping, unreachable-clears-version, and DisplayStatus sanitization. No new defects found.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Round 3 converges. Verified in the current diff: (1) the badge-ordering defect @ivanmunozruiz raised is fixed — serverStatusBadge() now consults the health payload (daemonHealthErr == nil && daemonStatus != "") before the !connected branch, so a 503 "degraded" payload followed by a failure on /prs renders "● degraded" instead of "● stopped", and TestServerStatusBadgeDegradedOutranksDisconnected / TestServerStatusBadgeStoppedWhenHealthUnreachable pin both directions so the fix cannot swing into calling an unreachable daemon healthy; the pre-first-fetch state (empty daemonStatus, nil err) still falls through to the connected/stopped logic, which is correct. (2) Status is no longer rendered verbatim — api.Health.DisplayStatus() drops control bytes and bounds the length, and it is shared by both the badge and cli statusLine so the two cannot drift; nil-receiver and ANSI cases are asserted. (3) The Makefile VERSION handling is now actually correct: `VERSION ?= dev` plus `ifeq ($(strip $(VERSION)),)` / `override VERSION := dev` is the only form that beats an empty command-line `VERSION=` (a plain assignment loses to the command line), and the root Makefile passes GIT_VERSION through to build-cli, so a stray empty value stamps "dev" rather than an empty string that makes cobra drop --version. No previously reported finding remains unaddressed, and nothing in the new lines clears the issue bar.

Non-blocking observations, deliberately not raised as issues:

- statusDisplayMax is applied against b.Len(), which is bytes, not runes, while the comment and TestHealthDisplayStatus ("caps length") describe it as a 32-rune cap. For ASCII statuses these coincide; for multibyte input the effective cap is ~11 runes, and because the check runs before WriteRune the output can reach 32+3 bytes. The safety property the function exists for (bounded output) holds either way, so this is a doc/precision nit rather than a defect — but if the intent is a rune cap, count runes explicitly.
- Asymmetry worth a thought: Status is now sanitised precisely because it reaches a TUI row and a terminal line, but health.Version travels the same path unsanitised — printed raw in `status` (`Version: %s`) and in the Server tab's Daemon row, with no length bound. It is stamped via -ldflags on the operator's own daemon, so there is no reachable failure today and this does not clear the bar; it is the same latent argument the DisplayStatus comment makes about Status ("this is about not depending on that"). Reusing a shared display-sanitiser for Version would close the gap for one line of code.
- The extra unauthenticated /health per refresh tick and the 60-rune truncation of the health error were both discussed and consciously kept; I agree with both calls — the /health request is exactly what lets the version and degraded state survive a 401 on the authenticated endpoints.

Test coverage for this diff is unusually good: the 503-decodes path, the missing-version tolerance, the 401/HTML/garbage rejections, the clock-skew clamp, the daemon-vs-CLI uptime distinction, and the badge precedence are each pinned by a test that fails on the pre-fix code.

---

vbuenog:
Thanks — both observations were inconsistencies in my own change, so fixed rather than carried: 3692b4a.

**Byte-vs-rune cap.** Correct, and the mismatch was between the code and my own comment/test. `statusDisplayMax` was checked against `strings.Builder.Len()`, so multibyte input capped at ~11 runes while the output could still reach 32+ bytes. It now counts runes, with `TestDisplayFieldsCapInRunes` pinning the rune count for both fields.

**`Version` unsanitised while `Status` is.** You are right that this was the same latent argument applied inconsistently — `Version` reached the same TUI row and the same terminal line raw and unbounded. Both now go through a shared `displayField`, so they cannot drift, and `DisplayVersion` gets its own 48-rune cap (above the longest `git describe` output, e.g. `0.8.0-3-g86f49fa-dirty`). Verified against the live daemon: `Version: 0.8.0` unchanged.

**Deliberately kept, per your agreement:** the extra unauthenticated `/health` per tick, and the 60-rune truncation of the health error.

`make test-cli` and `make lint-cli` pass; 2 further tests.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review pass 4 (3692b4a). Both prior findings are genuinely fixed. Verified: (1) the cap now counts runes — `displayField` increments `kept` per kept rune and breaks at `maxRunes`, so the 32/48 caps match their documented units, and `TestDisplayFieldsCapInRunes` pins both with 2-byte input; (2) `Version` and `Status` now share the same `displayField` path, so they cannot drift, with `DisplayVersion` capped at 48 runes (comfortably above `git describe` output like `0.8.0-3-g86f49fa-dirty`). Nil receivers on both Display* methods are covered. The rest of the diff is unchanged from the previously converged state: `GetHealth` decodes 200 and 503 while still erroring on 401/HTML/undecodable bodies; the badge consults the health payload before `!connected`; uptime derives from `started_at` and is clamped at zero; a failed health fetch clears version/start/status and surfaces the error on the Daemon row; the Makefile `override VERSION := dev` correctly defeats an empty command-line `VERSION=`. Nothing in this pass clears the reporting bar. Minor observations, none blocking and none worth a change on their own: `statusLine` compares the raw `h.Status` against "ok" rather than the sanitised value, so a hypothetical `"ok\n"` from the daemon would take the pass-through branch and print a sanitised "ok" instead of "online" — cosmetic, and the daemon does not emit that; the explicit `'\t' || '\n' || '\r'` checks in `displayField` are redundant with `!unicode.IsPrint(r)`, kept presumably for readability; and `d.daemonStatus` stores the raw string while every render path sanitises it, which is consistent but relies on that discipline holding for future call sites. Test coverage for this round is proportionate to the change. Converged — approve.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previously raised findings have been addressed across multiple review rounds. The current diff is clean: displayField now counts runes correctly, Version goes through the same sanitization as Status, the badge ordering reflects daemon health before connectedness, the Makefile handles empty VERSION via override, uptime is clamped at zero, and Health() delegates to GetHealth(). Test coverage is thorough and matches all the behavioral claims. No new issues found.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Re-review round 4. The previous MEDIUM (serverStatusBadge checking !connected before daemonStatus) is properly fixed: /health results are now consulted before the connected flag, with TestServerStatusBadgeDegradedOutranksDisconnected covering the exact reported scenario and TestServerStatusBadgeStoppedWhenHealthUnreachable guarding against over-correction. Muriano's byte-vs-rune cap and unsanitized-Version findings are also fixed via the shared displayField helper, and the Makefile 'override VERSION' correctly handles an empty command-line VERSION= (verified GIT_VERSION is defined in the root Makefile). The two deliberately-kept items (extra unauthenticated /health per tick, 60-rune error truncation) had explicit reviewer agreement. One minor new inconsistency remains on the error path; it does not block merge.

### Issues

🟡 **cli/internal/tui/dashboard.go:1148** — The health-fetch error rendered on the Daemon row goes through truncateRunes (length only) but not displayField-style sanitization. GetHealth embeds the raw response body in the error (fmt.Errorf("HTTP %d: %s", ..., strings.TrimSpace(string(data))) — TrimSpace strips only leading/trailing whitespace), so a proxy's HTML 502 body containing interior newlines or ESC bytes within the first 60 runes reaches the TUI row raw and can break the Server-section layout. This is the same argument the PR itself used to sanitize Status and Version, applied inconsistently to the third daemon-supplied string on the same row; prior discussion covered only the truncation length, not control bytes. Low impact: error path only, typically localhost.

---

vbuenog:
Fixed in f6be8f5.

**@ivanmunozruiz — 🟡 unsanitised health error: accepted, and you identified the inconsistency exactly.** It was the PR's own argument applied to two of three daemon-supplied strings on that row. `displayField` is now the exported `DisplayText` and the error goes through it too, so all three share one path.

Worth flagging what your finding exposed in my *test*: my first regression test asserted on the Daemon **row**, and it passed against the unsanitised code. A leaked newline splits the row, so the control bytes land on lines a row-scoped assertion never inspects — the very mechanism you described. The test now asserts over the whole section plus a line-count comparison, and I mutation-tested it: without the fix it fails with `leaks "\x1b"` and `has 16 lines, want 14`. Printable remnants (`<body>`) are deliberately kept, same as `[31m` once its ESC is gone.

**@Muriano — all three minor observations fixed**, since two of them were latent-by-discipline rather than latent-by-impossibility:
- `statusLine` compared the raw `h.Status` against `"ok"`, so a hypothetical `"ok\n"` took the pass-through branch; it compares the sanitised value now.
- `d.daemonStatus` is stored sanitised rather than sanitised at each render site — you were right that the previous shape relied on that discipline holding for future call sites.
- Dropped the redundant `'\t'/'\n'/'\r'` tests; verified `unicode.IsPrint` returns false for all three, so they were pure noise.

`make test-cli` and `make lint-cli` pass.

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Re-revisión tras 3692b4a. Los hallazgos de la ronda anterior sobre el tope byte-vs-runa y la sanitización de Version están correctamente resueltos: displayField compartido cuenta runas, DisplayVersion tiene su propio tope de 48 runas, y ambos campos quedan cubiertos por tests (TestDisplayFieldsCapInRunes, TestHealthDisplayVersion). La lógica del Makefile con override, la ordenación del badge (salud antes que connected) y statusLine son sólidas y están testeadas. Queda un único hallazgo LOW sin cambios: el error de /health renderizado en la fila Daemon pasa por truncateRunes (solo longitud) pero no por sanitización de bytes de control; la discusión del autor solo acordó dejar la longitud del truncado como está, no cubrió los bytes de control. No hay preocupaciones bloqueantes pendientes de otros revisores.

### Issues

🟡 **cli/internal/tui/dashboard.go:1148** — Sin cambios desde la revisión anterior: el error de la petición a /health se renderiza en la fila Daemon vía truncateRunes(d.daemonHealthErr.Error(), 60), que acota la longitud pero no elimina bytes de control. GetHealth incrusta el cuerpo crudo de la respuesta en el error (fmt.Errorf("HTTP %d: %s", ..., strings.TrimSpace(string(data))) — TrimSpace solo recorta espacios en los extremos), así que un 502 HTML de un proxy con saltos de línea interiores o bytes ESC dentro de las primeras 60 runas llega crudo a la fila del TUI y puede romper el layout de la sección Server. Es el mismo argumento que el propio PR usa para sanear Status y Version con displayField, aplicado de forma inconsistente a la tercera cadena suministrada por el daemon en la misma fila. La discusión previa solo acordó mantener la longitud del truncado (60 runas), no cubrió los bytes de control. Impacto bajo: solo ruta de error, típicamente localhost.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Pass 5 re-review of #660. All previously reported findings are resolved in the current diff: (a) GetHealth no longer routes through do(), so a 503 "degraded" payload decodes and still yields the version, while 401 / HTML 502 / undecodable bodies remain errors — covered by TestGetHealthDecodesDegraded503 and the status/garbage table; (b) the status command no longer prints a flat "online" — statusLine derives from the payload and now compares the *sanitised* status, with the nil/empty-status case reported honestly; (c) the badge consults the health payload before !connected, so a degraded daemon that answered /health is never reported as stopped (TestServerStatusBadgeDegradedOutranksDisconnected) while an unreachable /health still renders stopped; (d) daemonVersion/startedAt/status are cleared on a health-fetch failure so a last-known version cannot sit next to a stopped badge; (e) Uptime is the daemon's, clamped at zero against clock skew; (f) all three daemon-supplied strings on that row (Status, Version, health error) now share the single exported DisplayText path, capped in runes rather than bytes (TestDisplayFieldsCapInRunes pins that); (g) the cli/Makefile stamping uses `override` inside the ifeq, which is genuinely required — a plain assignment would lose to the command line, and the empty `VERSION=` path now yields "dev" instead of dropping cobra's --version flag. Nits/observations, none blocking: the extra unauthenticated /health per 10s tick and the 60-rune truncation of the health error were both consciously kept and I agree with both calls; sseStale is still evaluated before the health payload, so a degraded daemon with a stalled SSE stream renders "● unresponsive" rather than "● degraded" — defensible precedence (a dead event stream is the more actionable fact) but worth being aware of if the two ever need to compose; DisplayText's documented behaviour of keeping the inert remainder of a defused ANSI sequence ("[31m") is intentional and asserted, which is the right trade-off over parsing sequences. The comment density in this diff is unusually high for the codebase, but it is explanatory rather than noise. Note the supplied diff was truncated in the dashboard_test.go tail; I reviewed what was visible and did not speculate about the unseen hunks. Test coverage is thorough for a change of this size, including mutation-verified regression tests. Converged — nothing clears the issue bar.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All prior review findings have been addressed across multiple rounds. The code is well-structured with consistent sanitization, proper 503 handling, correct badge priority ordering, and thorough test coverage. No new issues found.

---

vbuenog:
**@ivanmunozruiz — refuting this one, with evidence.** The finding states *"Sin cambios desde la revisión anterior: el error … se renderiza … vía `truncateRunes(d.daemonHealthErr.Error(), 60)`"*, but your review was submitted against `f6be8f5`, and that commit already contains the fix — it landed in `f6be8f5` itself, in response to the identical finding on `3692b4a`.

At the reviewed commit:

```
$ git show f6be8f5:cli/internal/tui/dashboard.go | grep -n daemonHealthErr
  75:    daemonHealthErr error
 340:            d.daemonHealthErr = nil
 345:            d.daemonHealthErr = msg.healthErr
1113:    if d.daemonHealthErr == nil && d.daemonStatus != "" {
1147:        if d.daemonHealthErr != nil {
1149:                " — " + api.DisplayText(d.daemonHealthErr.Error(), 60))
```

`truncateRunes` no longer appears on that path anywhere in the file, and `git ls-remote` confirms `f6be8f5` is the branch head, so this is not a stale checkout on my side either.

The substance of the finding was right and is already implemented: `DisplayText` (the exported `displayField`) now handles all three daemon-supplied strings on that row, and `TestDaemonRowSanitisesHealthError` pins it — asserting over the whole Server section rather than the Daemon row, precisely because a leaked newline splits the row and hides the following bytes from a row-scoped assertion. Without the fix that test reports `leaks "\x1b"` and `has 16 lines, want 14`.

So: no code change for this one. If you are seeing `truncateRunes` on that line in your view, it is worth checking which commit was fetched — happy to be shown otherwise if I have misread it.

Nothing else outstanding on this PR: `make test-cli` and `make lint-cli` pass on `f6be8f5`.

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Re-revisión tras la refutación del autor: verificada y aceptada. En el diff actual, el error de /health se renderiza vía api.DisplayText(d.daemonHealthErr.Error(), 60), de modo que las tres cadenas suministradas por el daemon en la fila del Server (Status, Version y el error) comparten la misma ruta de saneado, con TestDaemonRowSanitisesHealthError cubriendo la sección completa. Mi hallazgo anterior queda resuelto — la refutación era correcta y el fix está en el commit revisado. El resto de rondas (503 degradado, regresión de `status`, orden del badge, tope en runas, Makefile con `override`) siguen correctamente aplicadas. Solo queda una micro-inconsistencia nueva en statusLine, sin impacto práctico.

### Issues

🟡 **cli/internal/cli/status.go:96** — statusLine decide el caso 'unreported' con el Status crudo (h.Status == "") pero imprime el valor saneado: un status no vacío compuesto solo de bytes no imprimibles (p. ej. "\n") pasa el guard, DisplayStatus() lo reduce a "" y el comando imprime una línea 'Status:' en blanco en vez de 'online (status unreported)'. El dashboard no tiene esta asimetría porque almacena d.daemonStatus ya saneado y su guard evalúa el valor saneado. Comprobar el valor saneado (o reutilizarlo para el guard) cierra el hueco. Impacto puramente teórico: el daemon solo emite 'ok'/'degraded'.

---

vbuenog:
Thanks for verifying the refutation, @ivanmunozruiz — and the new finding is right, same class as the previous one and again in my own code. Fixed in 1b90f55.

`statusLine` decided the "unreported" case from the **raw** `h.Status` but printed the **sanitised** value, so a status of only non-printable bytes was non-empty, passed the guard, sanitised to `""`, and printed a blank `Status:` line. Reproduced before fixing — `statusLine() = "", want "online (status unreported)"`.

Every branch now reads the sanitised value, and since `DisplayStatus` is nil-safe the nil-payload case folds into the same guard rather than needing its own. Two cases added: `"\n"` → `online (status unreported)`, and `"\nok\r"` → `online`. Your diagnosis of *why* the dashboard was immune (it stores `daemonStatus` already sanitised) is exactly the shape the CLI now matches.

`make test-cli` and `make lint-cli` pass; live output unchanged (`Status: online`, `Version: 0.8.0`).

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-revisión sobre 1b90f55. La última observación (guard de `statusLine` evaluando el valor crudo en lugar del saneado) está corregida: las tres ramas leen `h.DisplayStatus()`, que es nil-safe, y hay casos de test para `"\n"` → "online (status unreported)" y `"\nok\r"` → "online". Verificado también que los hallazgos anteriores siguen aplicados: `GetHealth` decodifica 200 y 503 y mantiene error en 401/HTML/cuerpos indescifrables; `Health()` delega en `GetHealth` con una sola ruta a /health; los tres strings suministrados por el daemon (status, version, error) pasan por `api.DisplayText`, con cap contado en runas; el badge consulta la salud antes de `!connected`, de modo que un daemon degradado no se reporta como "stopped"; uptime se deriva de `started_at` con clamp a cero ante skew de reloj; version y started_at se limpian cuando el fetch falla; y el `override VERSION` en cli/Makefile cubre el `VERSION=` vacío de línea de comandos. Cobertura de tests amplia y determinista (httptest, funciones puras, helper `serverRow` anclado en la etiqueta). Observaciones no bloqueantes, ya discutidas y aceptadas conscientemente por el autor: una petición extra no autenticada a /health por tick (10s, localhost) y el truncado a 60 runas del error de salud, que puede recortar mensajes de dial largos. Nit menor: la densidad de comentarios explicativos en `cli/Makefile` y en los comentarios de `dashboard.go`/`client.go` es alta respecto al resto del repo — útil como registro de la discusión, algo verboso como documentación permanente. Nada de esto alcanza el umbral de issue. El PR ha convergido.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previous review findings have been thoroughly addressed across multiple rounds. The code is well-structured: GetHealth correctly decodes both 200 and 503, DisplayText sanitizes all daemon-supplied strings consistently, statusLine operates on sanitized values, the dashboard clears stale state on disconnect, the status badge correctly prioritizes health over connectedness, and the Makefile handles empty VERSION. Test coverage is comprehensive with 20+ new tests covering normal paths, edge cases (nil receiver, multibyte runes, control bytes, clock skew), and error conditions. No new issues found.

---

PR #662, opened 2026-08-04
install: Stop the outgoing daemon on install-linux

### LINKED ISSUE #661: Upgrades silently keep the previous daemon running (stale binary after install-linux)
## Problem

`make install-linux` stages new binaries but never stops the daemon that is already running, and the desktop app adopts whatever is already healthy on the port. The result: after an upgrade the **previous** daemon keeps serving requests indefinitely, while the new binary sits unused on disk.

This is how it surfaced: the Server → Status tab reported `Heimdallm dev` on a machine where `~/.local/opt/heimdallm/heimdalld --version` reported `0.8.0`. The version display was correct all along — it was faithfully reporting a daemon process from a build predating the version stamping in #648. The stale process had 21h uptime and had survived that day's install.

The version string is only the visible symptom. The real cost is running the previous release's review logic, bugfixes and migrations without any indication that you are.

## Root cause

- `Makefile:671-696` (`install-linux`) — `rm -rf` + `cp` into `~/.local/opt/heimdallm/`, no daemon stop, no notice that a running daemon is now stale.
- `flutter_app/lib/core/daemon/daemon_lifecycle.dart:21` — `ensureRunning()` is `if (await isRunning()) return;`. Any healthy daemon on the port is adopted, whatever binary it came from. It is also only called at app startup, so nothing re-checks later.

Reproduce:
```
curl -s localhost:7842/health | jq .version   # e.g. "dev"
~/.local/opt/heimdallm/heimdalld --version    # 0.8.0
```
Restarting the daemon resolves it — which is precisely the step an upgrade never prompts for.

## Proposed fix

1. **`install-linux` stops the running daemon** after staging, and says so, so the next app launch spawns the new binary. Where the defect actually is; small and contained.
2. **App surfaces the mismatch** — compare `/health`'s version against the bundled `heimdalld --version` at startup and show a "restart server to finish updating" banner in the Status tab. Covers daemons started by something other than the installer.

Related: #648 (daemon version stamping), #660 (the TUI had a separate, unrelated defect in the same area).

vbuenog:
Closes #661.

`install-linux` replaced the binaries but left the running daemon alone, and `DaemonLifecycle.ensureRunning()` adopts any healthy daemon on the port rather than spawning the one just installed. Every upgrade kept serving the previous release indefinitely — stale review logic, stale version in the Status tab, no indication anything was wrong.

**Changes:**
- `install-linux`: SIGTERM the previous daemon after staging, wait up to 10s, and point at Server → Status → Start server. Runs last, so a failure earlier in the target never takes down a working daemon
- `dev-stop`, `uninstall-linux`: add a missing `-x` (see below)
- `scripts/test-linux-install.sh`: TAP regression guard, wired to a new `test-install-linux` target. Starts no daemon and touches no installed files

**Second defect found while fixing this** — worth a look, it is not cosmetic. `pkill -f <path>` inside a make recipe also matches the recipe's own `sh -c` command line, which embeds the pattern. pkill SIGTERMs the shell running it and the target dies mid-recipe, whether or not anything matched; `|| true` does not help, because the signal hits the shell rather than pkill. Consequences on `main` today:

- `dev-stop` aborts at its first line, taking `make dev` and `make dev-daemon` with it
- `uninstall-linux` aborts before removing the desktop entry, icons and PATH symlink

Reproduced with the `dev-stop` line verbatim and nothing running — `make` reports `*** [t] Terminated` and `REACHED-END` never prints:

```make
DAEMON_BIN := $(shell pwd)/daemon/bin/heimdallm
t:
	@pkill -f "$(DAEMON_BIN)" 2>/dev/null && echo "parado" || true
	@echo REACHED-END
```

`-x` requires the whole cmdline to equal the pattern: the daemon and app are spawned with no arguments so they still match, while the longer shell cmdline cannot.

The guard is static (every `pkill`/`pgrep` in the Makefile passes `-x`) plus behavioural (`-f` alone kills its recipe; `-x -f` does not, and still signals a process whose cmdline is exactly the path). The behavioural half matters because the failure leaves no trace in normal output — the target just stops after printing its progress lines.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,17 +56,21 @@ jobs:
           # and only the stderr detail survives.
           log="$RUNNER_TEMP/pkill-guard.log"
           make test-install-linux 2>&1 | tee "$log"
-          output="$(cat "$log")"
+          # greps read $log directly rather than piping a variable into grep -q:
+          # under pipefail, grep -q exiting on first match can SIGPIPE the writer
+          # and invert the condition. Unreachable at this output size, but the
+          # guard is built to be un-skippable, so the hole goes for free.
+          #
           # A silent skip here would defeat the purpose: the guard exits 0 when it
           # cannot run, which is right locally on macOS but must never pass in CI.
-          if printf '%s\n' "$output" | grep -q '^1\.\.0'; then
+          if grep -q '^1\.\.0' "$log"; then
             echo "::error::the pkill guard skipped entirely — it must run on ubuntu-latest"
             exit 1
           fi
           # Per-test skips are also coverage loss, not success: on this runner
           # every assertion is meant to execute, so any "# SKIP" means the
           # environment stopped reproducing what the guard exists to prove.
-          if printf '%s\n' "$output" | grep -q '# SKIP'; then
+          if grep -q '# SKIP' "$log"; then
             echo "::error::a pkill guard assertion skipped — coverage is incomplete"
             exit 1
           fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,6 +37,27 @@ jobs:
           make -n install-linux >/dev/null
           make -n uninstall-linux >/dev/null
 
+      # The parse checks above cannot catch a dropped -x: `make -n` expands the
+      # recipe without ever running pkill. This job is the only place the pkill
+      # semantics are executed, and ubuntu-latest has the /proc the guard needs.
+      - name: pkill semantics in the Linux targets
+        run: |
+          set -eu
+          for script in scripts/pkill-escape.sh scripts/test-linux-install.sh; do
+            if [ ! -x "$script" ]; then
+              echo "::error::$script must be executable"
+              exit 1
+            fi
+          done
+          output="$(make test-install-linux)"
+          printf '%s\n' "$output"
+          # A silent skip here would defeat the purpose: the guard exits 0 when it
+          # cannot run, which is right locally on macOS but must never pass in CI.
+          if printf '%s\n' "$output" | grep -q '^1\.\.0'; then
+            echo "::error::the pkill guard skipped entirely — it must run on ubuntu-latest"
+            exit 1
+          fi
+
       - name: macOS targets stop at the platform guard
         run: |
           set -eu
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -57,6 +57,13 @@ jobs:
             echo "::error::the pkill guard skipped entirely — it must run on ubuntu-latest"
             exit 1
           fi
+          # Per-test skips are also coverage loss, not success: on this runner
+          # every assertion is meant to execute, so any "# SKIP" means the
+          # environment stopped reproducing what the guard exists to prove.
+          if printf '%s\n' "$output" | grep -q '# SKIP'; then
+            echo "::error::a pkill guard assertion skipped — coverage is incomplete"
+            exit 1
+          fi
 
       - name: macOS targets stop at the platform guard
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,15 +42,21 @@ jobs:
       # semantics are executed, and ubuntu-latest has the /proc the guard needs.
       - name: pkill semantics in the Linux targets
         run: |
-          set -eu
+          # pipefail so a guard failure still fails the step through the tee.
+          set -euo pipefail
           for script in scripts/pkill-escape.sh scripts/test-linux-install.sh; do
             if [ ! -x "$script" ]; then
               echo "::error::$script must be executable"
               exit 1
             fi
           done
-          output="$(make test-install-linux)"
-          printf '%s\n' "$output"
+          # Streamed through tee rather than captured into a variable: with
+          # `output="$(make …)"` under set -e, a failing guard aborts the step
+          # before the echo runs, so the ok/not ok TAP lines never reach the log
+          # and only the stderr detail survives.
+          log="$RUNNER_TEMP/pkill-guard.log"
+          make test-install-linux 2>&1 | tee "$log"
+          output="$(cat "$log")"
           # A silent skip here would defeat the purpose: the guard exits 0 when it
           # cannot run, which is right locally on macOS but must never pass in CI.
           if printf '%s\n' "$output" | grep -q '^1\.\.0'; then
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,8 @@ dev-stop:
 	@# aborts — taking `make dev` / `make dev-daemon` with it, whether or not a
 	@# daemon was running. The dev daemon is spawned with no arguments, so its
 	@# cmdline equals the pattern exactly. Guarded by scripts/test-linux-install.sh.
-	@DAEMON_RE=$$($(PKILL_ESCAPE) "$(DAEMON_BIN)"); \
+	@DAEMON_RE=$$($(PKILL_ESCAPE) "$(DAEMON_BIN)") || exit 1; \
+	[ -n "$$DAEMON_RE" ] || { echo "❌  pkill-escape.sh produced an empty pattern"; exit 1; }; \
 	pkill -x -f "$$DAEMON_RE" 2>/dev/null && echo "↓  Daemon parado" || true
 	@UI_PID_FILE="$$HOME/.local/share/heimdallm/ui.pid"; \
 	 if [ -f "$$UI_PID_FILE" ]; then \
@@ -780,7 +781,9 @@ install-linux: _check-linux verify-linux
 	@# cmdline equal the pattern, which the daemon satisfies (spawned with no
 	@# arguments) and the longer shell cmdline does not. Guarded by
 	@# scripts/test-linux-install.sh.
-	@DAEMON_RE=$$($(PKILL_ESCAPE) "$$HOME/.local/opt/heimdallm/heimdalld"); \
+	@DAEMON_RE=$$($(PKILL_ESCAPE) "$$HOME/.local/opt/heimdallm/heimdalld") || exit 1; \
+	[ -n "$$DAEMON_RE" ] || { echo "❌  pkill-escape.sh produced an empty pattern — the"; \
+	  echo "   previous daemon is still running; stop it before launching."; exit 1; }; \
 	if pkill -x -f "$$DAEMON_RE" 2>/dev/null; then \
 	  echo ""; \
 	  echo "↓  Stopped the previously running daemon (it was serving the old binary)."; \
@@ -854,10 +857,12 @@ uninstall-linux: _check-linux
 	@# its cmdline is longer than the pattern. Same rationale as above — the
 	@# rm -rf is safe against a running process — and plain -f never actually
 	@# reached that case anyway, because it killed this recipe's shell first.
-	@APP_RE=$$($(PKILL_ESCAPE) "$$HOME/.local/opt/heimdallm/heimdallm"); \
-	 pkill -x -f "$$APP_RE" 2>/dev/null || true
-	@DAEMON_RE=$$($(PKILL_ESCAPE) "$$HOME/.local/opt/heimdallm/heimdalld"); \
-	 pkill -x -f "$$DAEMON_RE" 2>/dev/null || true
+	@APP_RE=$$($(PKILL_ESCAPE) "$$HOME/.local/opt/heimdallm/heimdallm" 2>/dev/null); \
+	 if [ -n "$$APP_RE" ]; then pkill -x -f "$$APP_RE" 2>/dev/null || true; \
+	 else echo "⚠  could not build the app pattern; leaving any running app alone"; fi
+	@DAEMON_RE=$$($(PKILL_ESCAPE) "$$HOME/.local/opt/heimdallm/heimdalld" 2>/dev/null); \
+	 if [ -n "$$DAEMON_RE" ]; then pkill -x -f "$$DAEMON_RE" 2>/dev/null || true; \
+	 else echo "⚠  could not build the daemon pattern; leaving any running daemon alone"; fi
 	@rm -f "$$HOME/.local/share/heimdallm/ui.pid"
 	rm -f "$$HOME/.local/share/applications/com.theburrowhub.heimdallm.desktop"
 	@for SIZE in 48 128 256 512; do \
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ endif
 .PHONY: build-daemon build-app build-web build-cli test test-cli lint-cli test-web-tooling test-compose-isolation \
         test-smoke test-github test-e2e test-web dev-cli test-docker dev dev-daemon dev-stop \
         release-local package-macos install-service verify-linux run-linux test-install-macos \
+        test-install-linux \
         install-macos uninstall-macos install-linux uninstall-linux \
         setup up up-build up-daemon up-build-daemon down logs logs-daemon \
         ps restart clean clean-clones _check-docker _check-buildkit _check-env \
@@ -69,6 +70,12 @@ test-install-macos:
 	@sh -n scripts/macos-install.sh
 	@sh -n scripts/test-macos-install.sh
 	@./scripts/test-macos-install.sh
+
+# Guards the pkill invocations in dev-stop / install-linux / uninstall-linux.
+# Static-plus-behavioural; starts no daemon and touches no installed files.
+test-install-linux:
+	@sh -n scripts/test-linux-install.sh
+	@./scripts/test-linux-install.sh
 
 test-web-tooling:
 	@sh docker/scripts/tests/test-web-tooling.sh
@@ -149,7 +156,12 @@ dev-cli:
 	$(MAKE) -C cli dev
 
 dev-stop:
-	@pkill -f "$(DAEMON_BIN)" 2>/dev/null && echo "↓  Daemon parado" || true
+	@# -x is REQUIRED: without it the pattern also matches this recipe's own
+	@# `sh -c` command line, so pkill SIGTERMs the shell running it and dev-stop
+	@# aborts — taking `make dev` / `make dev-daemon` with it, whether or not a
+	@# daemon was running. The dev daemon is spawned with no arguments, so its
+	@# cmdline equals the pattern exactly. Guarded by scripts/test-linux-install.sh.
+	@pkill -x -f "$(DAEMON_BIN)" 2>/dev/null && echo "↓  Daemon parado" || true
 	@UI_PID_FILE="$$HOME/.local/share/heimdallm/ui.pid"; \
 	 if [ -f "$$UI_PID_FILE" ]; then \
 	   UI_PID=$$(cat "$$UI_PID_FILE"); \
@@ -742,6 +754,39 @@ install-linux: _check-linux verify-linux
 	  update-desktop-database "$$HOME/.local/share/applications/" 2>/dev/null || true
 	@command -v gtk-update-icon-cache >/dev/null 2>&1 && \
 	  gtk-update-icon-cache -q -t "$$HOME/.local/share/icons/hicolor/" 2>/dev/null || true
+	@# Stop the outgoing daemon — see #661. Replacing the binary above does not
+	@# touch the process already serving it (Linux unlink-while-running), and
+	@# DaemonLifecycle.ensureRunning() adopts any healthy daemon on the port
+	@# instead of spawning the one just installed. Left alone, the previous
+	@# release keeps running indefinitely: stale review logic, stale version in
+	@# the Status tab, no indication anything is wrong.
+	@#
+	@# Runs last, so a failure earlier in this target never takes down a working
+	@# daemon. SIGTERM (pkill's default) is the daemon's own graceful path — it
+	@# drains HTTP and sweeps in-flight agents.
+	@#
+	@# -x is REQUIRED, not a refinement: with -f alone the pattern also matches
+	@# the recipe's own `sh -c` command line, which contains the path — pkill
+	@# SIGTERMs the shell running it and the target dies mid-recipe (`|| true`
+	@# does not help; the signal hits the shell, not pkill). -x demands the whole
+	@# cmdline equal the pattern, which the daemon satisfies (spawned with no
+	@# arguments) and the longer shell cmdline does not. Guarded by
+	@# scripts/test-linux-install.sh.
+	@DAEMON="$$HOME/.local/opt/heimdallm/heimdalld"; \
+	if pkill -x -f "$$DAEMON" 2>/dev/null; then \
+	  echo ""; \
+	  echo "↓  Stopped the previously running daemon (it was serving the old binary)."; \
+	  for _ in 1 2 3 4 5 6 7 8 9 10; do \
+	    pgrep -x -f "$$DAEMON" >/dev/null 2>&1 || break; \
+	    sleep 1; \
+	  done; \
+	  if pgrep -x -f "$$DAEMON" >/dev/null 2>&1; then \
+	    echo "⚠  It has not exited after 10s — stop it manually, or it will keep"; \
+	    echo "   serving the old binary."; \
+	  else \
+	    echo "   Start the new one from the app: Server → Status → Start server."; \
+	  fi; \
+	fi
 	@echo ""
 	@echo "✅  Heimdallm installed:"
 	@echo "    Bundle:  $$HOME/.local/opt/heimdallm/"
@@ -774,6 +819,13 @@ uninstall-linux: _check-linux
 	@echo "▶  Uninstalling Heimdallm from $$HOME/.local/..."
 	@# Stop running instances (best-effort — ignored if nothing is running).
 	@#
+	@# -x is REQUIRED: with -f alone the pattern also matches the recipe's own
+	@# `sh -c` command line, so pkill SIGTERMs the shell running it and this
+	@# target dies here — before the desktop entry, icons and symlink below are
+	@# removed (`|| true` does not help; the signal hits the shell, not pkill).
+	@# -x demands the whole cmdline equal the pattern, which the shell's longer
+	@# cmdline never does. Guarded by scripts/test-linux-install.sh.
+	@#
 	@# Coverage caveat: pkill -f matches against /proc/<pid>/cmdline. This
 	@# catches:
 	@#   - launcher-launched Flutter apps (desktop entry Exec= is absolute,
@@ -788,8 +840,8 @@ uninstall-linux: _check-linux
 	@# running process, and the user can close the window whenever. We
 	@# intentionally avoid a broader `-f 'heimdallm'` match to prevent
 	@# hitting unrelated dev processes (e.g. `flutter run` of this repo).
-	@pkill -f "$$HOME/.local/opt/heimdallm/heimdallm" 2>/dev/null || true
-	@pkill -f "$$HOME/.local/opt/heimdallm/heimdalld" 2>/dev/null || true
+	@pkill -x -f "$$HOME/.local/opt/heimdallm/heimdallm" 2>/dev/null || true
+	@pkill -x -f "$$HOME/.local/opt/heimdallm/heimdalld" 2>/dev/null || true
 	@rm -f "$$HOME/.local/share/heimdallm/ui.pid"
 	rm -f "$$HOME/.local/share/applications/com.theburrowhub.heimdallm.desktop"
 	@for SIZE in 48 128 256 512; do \
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,13 @@ OS := $(shell uname -s)
 
 DAEMON_BIN  := $(shell pwd)/daemon/bin/heimdallm
 
+# pkill/pgrep -f treat their pattern as an ERE, so a path containing regex
+# metacharacters silently fails to match — a $HOME or checkout directory with
+# `+`, `(` or `[` in it would leave the daemon running with no error. The script
+# is the single implementation, shared with scripts/test-linux-install.sh, which
+# asserts it against a metacharacter-laden path.
+PKILL_ESCAPE = ./scripts/pkill-escape.sh
+
 ifeq ($(OS),Darwin)
   FLUTTER_DEVICE   := macos
   FLUTTER_BUILD    := flutter_app/build/macos/Build/Products
@@ -161,7 +168,8 @@ dev-stop:
 	@# aborts — taking `make dev` / `make dev-daemon` with it, whether or not a
 	@# daemon was running. The dev daemon is spawned with no arguments, so its
 	@# cmdline equals the pattern exactly. Guarded by scripts/test-linux-install.sh.
-	@pkill -x -f "$(DAEMON_BIN)" 2>/dev/null && echo "↓  Daemon parado" || true
+	@DAEMON_RE=$$($(PKILL_ESCAPE) "$(DAEMON_BIN)"); \
+	pkill -x -f "$$DAEMON_RE" 2>/dev/null && echo "↓  Daemon parado" || true
 	@UI_PID_FILE="$$HOME/.local/share/heimdallm/ui.pid"; \
 	 if [ -f "$$UI_PID_FILE" ]; then \
 	   UI_PID=$$(cat "$$UI_PID_FILE"); \
@@ -772,15 +780,15 @@ install-linux: _check-linux verify-linux
 	@# cmdline equal the pattern, which the daemon satisfies (spawned with no
 	@# arguments) and the longer shell cmdline does not. Guarded by
 	@# scripts/test-linux-install.sh.
-	@DAEMON="$$HOME/.local/opt/heimdallm/heimdalld"; \
-	if pkill -x -f "$$DAEMON" 2>/dev/null; then \
+	@DAEMON_RE=$$($(PKILL_ESCAPE) "$$HOME/.local/opt/heimdallm/heimdalld"); \
+	if pkill -x -f "$$DAEMON_RE" 2>/dev/null; then \
 	  echo ""; \
 	  echo "↓  Stopped the previously running daemon (it was serving the old binary)."; \
 	  for _ in 1 2 3 4 5 6 7 8 9 10; do \
-	    pgrep -x -f "$$DAEMON" >/dev/null 2>&1 || break; \
+	    pgrep -x -f "$$DAEMON_RE" >/dev/null 2>&1 || break; \
 	    sleep 1; \
 	  done; \
-	  if pgrep -x -f "$$DAEMON" >/dev/null 2>&1; then \
+	  if pgrep -x -f "$$DAEMON_RE" >/dev/null 2>&1; then \
 	    echo "⚠  It has not exited after 10s — stop it manually, or it will keep"; \
 	    echo "   serving the old binary."; \
 	  else \
@@ -840,8 +848,16 @@ uninstall-linux: _check-linux
 	@# running process, and the user can close the window whenever. We
 	@# intentionally avoid a broader `-f 'heimdallm'` match to prevent
 	@# hitting unrelated dev processes (e.g. `flutter run` of this repo).
-	@pkill -x -f "$$HOME/.local/opt/heimdallm/heimdallm" 2>/dev/null || true
-	@pkill -x -f "$$HOME/.local/opt/heimdallm/heimdalld" 2>/dev/null || true
+	@#
+	@# -x narrows this further: an app launched WITH arguments (e.g.
+	@# `~/.local/opt/heimdallm/heimdallm --some-flag`) no longer matches, since
+	@# its cmdline is longer than the pattern. Same rationale as above — the
+	@# rm -rf is safe against a running process — and plain -f never actually
+	@# reached that case anyway, because it killed this recipe's shell first.
+	@APP_RE=$$($(PKILL_ESCAPE) "$$HOME/.local/opt/heimdallm/heimdallm"); \
+	 pkill -x -f "$$APP_RE" 2>/dev/null || true
+	@DAEMON_RE=$$($(PKILL_ESCAPE) "$$HOME/.local/opt/heimdallm/heimdalld"); \
+	 pkill -x -f "$$DAEMON_RE" 2>/dev/null || true
 	@rm -f "$$HOME/.local/share/heimdallm/ui.pid"
 	rm -f "$$HOME/.local/share/applications/com.theburrowhub.heimdallm.desktop"
 	@for SIZE in 48 128 256 512; do \
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,23 @@
 # ── Platform detection ─────────────────────────────────────────────────────────
 OS := $(shell uname -s)
 
-DAEMON_BIN  := $(shell pwd)/daemon/bin/heimdallm
+# Anchored to this Makefile's directory for the same reason as PKILL_ESCAPE
+# below: with $(shell pwd) a `make -f /path/to/Makefile` from elsewhere pointed
+# dev-stop at a daemon path that does not exist. Identical from the repo root.
+DAEMON_BIN  := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))/daemon/bin/heimdallm
 
 # pkill/pgrep -f treat their pattern as an ERE, so a path containing regex
 # metacharacters silently fails to match — a $HOME or checkout directory with
 # `+`, `(` or `[` in it would leave the daemon running with no error. The script
 # is the single implementation, shared with scripts/test-linux-install.sh, which
 # asserts it against a metacharacter-laden path.
-PKILL_ESCAPE = ./scripts/pkill-escape.sh
+#
+# Anchored to this Makefile's own directory rather than the cwd: with a relative
+# path, `make -f /path/to/Makefile` from elsewhere cannot find the script, and
+# since the callers now abort on an empty pattern that would turn an unusual
+# invocation into a hard failure of dev-stop and install-linux.
+REPO_ROOT_DIR := $(dir $(lastword $(MAKEFILE_LIST)))
+PKILL_ESCAPE = $(REPO_ROOT_DIR)scripts/pkill-escape.sh
 
 ifeq ($(OS),Darwin)
   FLUTTER_DEVICE   := macos
--- a/scripts/pkill-escape.sh
+++ b/scripts/pkill-escape.sh
@@ -25,4 +25,16 @@ if [ "$#" -ne 1 ]; then
   exit 2
 fi
 
+# sed is line-oriented, so a path containing an embedded newline would be
+# escaped per line and the resulting pattern could never match the target's
+# cmdline — pkill would silently no-op, which is the failure this script exists to
+# prevent. Refuse it loudly instead.
+# Counted rather than glob-matched: "$(printf '\n')" collapses to an empty
+# string in a command substitution, so `case $1 in *""*)` matches everything.
+# printf '%s' adds no trailing newline, so wc -l counts only embedded ones.
+if [ "$(printf '%s' "$1" | wc -l)" -gt 0 ]; then
+  echo "$(basename "$0"): refusing a path containing a newline" >&2
+  exit 2
+fi
+
 printf '%s' "$1" | sed -e 's/\\/\\\\/g' -e 's/[]^$.*+?(){}|[]/\\&/g'
--- a/scripts/pkill-escape.sh
+++ b/scripts/pkill-escape.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+#
+# Escapes a literal path so it can be used as a pkill/pgrep -f pattern.
+#
+# `pkill -f` and `pgrep -f` treat their argument as an ERE, so a path containing
+# regex metacharacters does not match itself: a $HOME or checkout directory with
+# `+`, `(` or `[` in it makes the pattern silently miss, and the Makefile's
+# daemon-stop leaves the old daemon running with no error at all.
+#
+# Escaping runs in two passes because a single bracket expression cannot hold
+# both `\` and the rest of the metacharacters, and because `[.` inside a bracket
+# expression is read as the start of a POSIX collating symbol — hence `[` sits
+# last in the class. Backslashes go first so the escapes added by the second pass
+# are not escaped again.
+#
+# Used by the Makefile's dev-stop / install-linux / uninstall-linux and asserted
+# by scripts/test-linux-install.sh, so both share one implementation.
+#
+# Usage: pkill-escape.sh <literal-path>
+
+set -eu
+
+if [ "$#" -ne 1 ]; then
+  echo "usage: $(basename "$0") <literal-path>" >&2
+  exit 2
+fi
+
+printf '%s' "$1" | sed -e 's/\\/\\\\/g' -e 's/[]^$.*+?(){}|[]/\\&/g'
--- a/scripts/test-linux-install.sh
+++ b/scripts/test-linux-install.sh
@@ -147,6 +147,11 @@ fi
 #
 # No target process exists here on purpose: the self-match is what ends the
 # recipe, so the bug reproduces against nothing at all.
+#
+# PAT goes through the ENVIRONMENT, not make's command line: a command-line
+# `PAT=…` lands in make's own /proc cmdline, so `pkill -f` would match the make
+# process too and the test would only prove "something in the chain self-matched"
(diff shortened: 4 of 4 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-crafted fix for a real and subtle bug where `pkill -f` self-matches the `sh -c` recipe shell, silently aborting Makefile targets. The `-x` fix is correct, thoroughly documented, and guarded by both static and behavioral tests. No correctness issues found; minor observations only.

### Issues

🟡 **scripts/test-linux-install.sh:63** — The static grep pattern `(-[a-zA-Z]*x[a-zA-Z]*[[:space:]])` requires `-x` immediately after `pkill/pgrep`. If someone writes flags in a different order with `-x` as a separate arg (e.g., `pkill -f -x "pattern"`), the check would incorrectly flag it. Not a current problem since all Makefile uses put `-x` first, but the test would give a false positive if someone rearranges the flags.
🟡 **scripts/test-linux-install.sh:86** — The behavioral test asserts that `pkill -f` *must* self-match the recipe shell, and fails (not skips) if the environment's pkill doesn't reproduce this. If CI runs on a system where pkill doesn't exhibit this behavior (e.g., BusyBox pkill, non-procps-ng), the test will hard-fail rather than SKIP. Consider using the same `1..0 # SKIP` pattern already used for missing pkill/pgrep on line 47.
🟡 **scripts/test-linux-install.sh:113** — The `sleep 1` after spawning the fixture process is a time-based race-condition guard. On heavily loaded CI runners this could be insufficient, causing `kill -0 $TARGET` to fail and the test to report the fixture didn't start. A retry loop (similar to the 10-iteration wait in the Makefile) would be more robust, though in practice 1s is usually plenty for `/bin/cat`.

---

Muriano (commented):
## 🤖 Heimdallm AI Review

Solid, well-reasoned fix. I verified the core claim end-to-end: the daemon is spawned as `Process.start(daemonBinaryPath, [])` (flutter_app/lib/core/daemon/daemon_lifecycle.dart:28), `DAEMON_BIN` is absolute (`$(shell pwd)/daemon/bin/heimdallm`, Makefile:4), `dev-daemon` runs it with only an env prefix (Makefile:146, which does not appear in /proc/<pid>/cmdline), and the desktop entry uses a bare absolute `Exec=` with no field codes (Makefile:708). So every intended target's cmdline equals the pattern exactly and `-x` still matches them; the added strictness in uninstall-linux narrows nothing that was previously being caught (launcher/symlink cases were already out of reach with plain `-f`). Running the stop block last in install-linux, and using SIGTERM to hit the daemon's graceful drain, are both the right calls. Two problems with the new guard script itself are below. Non-blocking notes: (a) the static check's exclusion regex requires the x-flag to be the *first* token after pkill, so a future `pkill -f -x ...` or `--exact` would be reported as an offender even though it is correct; (b) `$DAEMON`/`$FAKE` are used as EREs, so a `$HOME` containing regex metacharacters (`+`, `[`, `(`) would silently fail to match and leave the old daemon running — latent, not reachable today; (c) test 3 depends on the 30s `sleep` writer outliving the assertion, so on a heavily loaded machine the fixture could exit on its own and the test would pass vacuously — pinning it with a longer sleep or asserting the process is still alive right before `make with-x` would harden it; (d) the 140-line script has no shellcheck/`sh -n`-only entry beyond the Makefile target, which is consistent with test-macos-install.sh.

### Issues

⚠️ **scripts/test-linux-install.sh:36** — The only skip guard is `command -v pkill/pgrep`, but both exist on macOS/BSD while `/proc` does not. On macOS — the platform this repo is primarily developed on — `make test-install-linux` reaches test 3, `tr '\0' ' ' < /proc/$TARGET/cmdline` yields the empty string via `|| echo ''`, and the script reports `not ok - fixture cmdline is not exactly the binary path` and exits 1. That is a false failure on a correct tree, which trains developers to ignore the target. Guard the whole script (or at least test 3) on the platform it is written for, e.g. `[ "$(uname -s)" = Linux ] && [ -d /proc ] || { printf '1..0 # SKIP requires Linux /proc\n'; exit 0; }`.
⚠️ **Makefile:74** — The new regression guard is never executed automatically: `.github/workflows/tests.yml` runs `make test-install-macos` plus `make -n install-linux/uninstall-linux` parse checks, and this PR does not add `make test-install-linux` to it. A future edit that drops `-x` from any of the three pkill sites passes CI green (the `-n` parse checks cannot detect it) and silently reintroduces exactly the failure this PR fixes — `make dev` aborting in dev-stop and `make uninstall-linux` dying before removing the desktop entry, icons and symlink. The `installer-shell` job already runs on ubuntu-latest, where /proc exists and all three tests are meaningful, so wiring it in is a one-line addition to that job.

---

---
_Not approving: this review raised 2 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

Muriano (commented):
## 🤖 Heimdallm AI Review

Re-review of PR #662. The branch is unchanged since my previous pass — a single commit (1e47ea9), same diff, no fix commits. The core change is sound: the `-x` addition to the three `pkill` sites is a correct root-cause fix for the recipe-shell self-match, the ordering of the daemon stop at the end of `install-linux` is deliberate and right (a failure earlier in the target can no longer take down a working daemon), SIGTERM is the daemon's graceful path, and the 10s bounded wait with an explicit warning on timeout is good operator UX. `scripts/test-linux-install.sh` is a genuinely well-constructed guard: test 2 proves the bug reproduces against no target process at all, and test 3 uses `/bin/cat` on a fifo precisely so argv[] equals the path (a shell script would have made the test pass vacuously) — that reasoning is documented in the header. Nits, not issues: the static check in test 1 is regex-based and would miss a `pkill` invocation split across a line continuation or one written as `@ pkill` with the flag after the pattern; and `uninstall-linux` narrowing from `-f` to `-x -f` does slightly reduce coverage for an app launched with arguments (e.g. `heimdallm --some-flag` from a terminal), which the existing coverage-caveat comment does not yet mention — worth a line there. Both of my previous findings remain unaddressed in the code: the script still has no Linux/`/proc` guard, and `make test-install-linux` is still not wired into `.github/workflows/tests.yml` (verified on the branch: the `installer-shell` job still runs only `test-install-macos` plus the two `make -n` parse checks).

### Issues

⚠️ **scripts/test-linux-install.sh:43** — STILL UNCHANGED. The only skip guard is `command -v pkill/pgrep`, but both binaries exist on macOS/BSD while `/proc` does not. Running `make test-install-linux` on macOS — the platform this repo is primarily developed on — reaches test 3, where `tr '\0' ' ' < /proc/$TARGET/cmdline 2>/dev/null || echo ''` yields the empty string, so `[ "$actual" != "$FAKE" ]` is true and the script reports `not ok - fixture cmdline is not exactly the binary path` and exits 1 on a perfectly correct tree. A test target that fails deterministically on the developer's own machine trains people to stop running it. Fix: guard the script (or at minimum test 3) on the platform it is written for, e.g. `[ "$(uname -s)" = Linux ] && [ -d /proc ] || { printf '1..0 # SKIP requires Linux /proc\n'; exit 0; }`.
⚠️ **Makefile:74** — STILL UNCHANGED. The new regression guard is never executed automatically. Verified on the PR branch: `.github/workflows/tests.yml`'s `installer-shell` job runs `make test-install-macos` and `make -n install-linux` / `make -n uninstall-linux`, and this PR adds nothing to it. A future edit that drops `-x` from any of the three `pkill` sites merges green — `make -n` only expands the recipe, it never runs pkill, so the parse checks are structurally incapable of catching it — and silently reintroduces exactly the failure this PR fixes: `make dev` aborting inside `dev-stop`, `make uninstall-linux` dying before removing the desktop entry, icons and PATH symlink, and `install-linux` leaving the outgoing daemon alive. The `installer-shell` job already runs on `ubuntu-latest`, where `/proc` exists and all three tests are meaningful, so this is a one-line addition to that job.

---

---
_Not approving: this review raised 2 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured fix for a real issue (#661). The `-x` flag addition to all pkill/pgrep calls is correct and well-documented, the daemon stop logic in install-linux is properly sequenced (runs last), and the new test file provides both static and behavioral coverage. My three previous LOW findings (grep pattern brittleness, hard-fail vs skip on non-procps environments, sleep-based race) appear unaddressed in the current diff but none represent correctness bugs — they are robustness nits for edge-case CI environments. No new issues found.

---

vbuenog:
Both blocking findings addressed in 9ac2ee6, plus every nit.

**@Muriano — MEDIUM, no Linux//proc guard (re-flagged): accepted, fixed.** The script now exits `1..0 # SKIP` unless `uname -s` is Linux and `/proc` exists, so `make test-install-linux` no longer fails deterministically on macOS.

**@Muriano — MEDIUM, guard not wired into CI (re-flagged): accepted, fixed.** Added a `pkill semantics in the Linux targets` step to the `installer-shell` job, exactly as you suggested — it already runs on `ubuntu-latest`. Your point that `make -n` is structurally incapable of catching a dropped `-x` is the whole reason this needed to exist, so the step also **fails if the guard reports a wholesale skip** (`1..0`): a guard that silently skips in CI is no guard. It also asserts both scripts are executable, matching the existing macOS step.

**Nit (b), the ERE metacharacter hazard — not latent, and worth more than a nit.** I reproduced it: with a path containing `+`, `(` and `[`, `pgrep -x -f "$PATH"` does not match its own process, so the daemon-stop silently no-ops. Fixed via `scripts/pkill-escape.sh`, shared by all three Makefile sites and asserted by tests 5-7 (unescaped misses → escaped matches → escaped does not match an unrelated path). Escaping needs two passes: one bracket expression cannot hold both `\` and the rest, and `[.` inside one is read as a POSIX collating symbol, so `[` sits last in the class. Verified end-to-end — `install-linux`'s stop block now works with `HOME=…/we+ird(x)`, which silently failed before.

**Nit (c), test 3's writer could outlive the assertion** — writer now sleeps an hour, fixture startup polls `/proc` instead of sleeping, and the process is re-checked alive immediately before the assertion.

**Nit, `uninstall-linux` coverage caveat** — the comment now records that `-x` also excludes an app launched *with arguments*, and why that is acceptable (the `rm -rf` is safe against a running process, and plain `-f` never reached that case anyway — it killed the recipe's shell first).

**@sergiotejon — all three LOW findings fixed:**
- static check is now flag-order independent: `-f -x`, bundled `-xf` and `--exact` all pass. Mutation-tested both ways — reordered flags are not flagged; a removed `-x` is, with the line number.
- test 2 now `# SKIP`s instead of hard-failing where pkill does not self-match (BusyBox / non-procps), since the `-x` fix stays correct there.
- the `sleep 1` race is gone: startup polls `/proc`, and exit waits poll too.

Guard is now 7 tests, all passing on Linux; skips cleanly off it. `make test-install-macos` still passes (98 tests). Both `make -n` parse checks still pass.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previous findings have been addressed. The Linux/proc guard, CI wiring, ERE metacharacter escaping via pkill-escape.sh, and test robustness improvements are all correctly implemented. The two-pass sed escaping in pkill-escape.sh is correct (backslashes first, then metacharacters with ] first and [ last in the bracket expression to avoid POSIX collating-symbol ambiguity). The static check is now flag-order independent. No new issues introduced.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #662 after 9ac2ee6. Both previously blocking findings are genuinely fixed, and I verified them rather than taking the reply at face value.

(1) Platform guard: scripts/test-linux-install.sh now exits `1..0 # SKIP` unless `uname -s` is Linux AND `/proc` exists, before any fixture runs, plus a second skip for missing pkill/pgrep. `make test-install-linux` no longer fails deterministically on macOS, and the rationale comment states why guarding on the binaries alone was insufficient (pkill/pgrep exist on BSD, /proc does not). Resolved.

(2) CI wiring: the `installer-shell` job gains a `pkill semantics in the Linux targets` step on ubuntu-latest that runs `make test-install-linux`, asserts both scripts are executable, and — the part that actually closes the hole — fails the build if the output contains `^1..0`, so a wholesale skip cannot masquerade as a pass. A future edit dropping `-x` is now caught behaviourally, which `make -n` structurally could not do. Resolved.

I independently verified the load-bearing assumption behind `-x`: the daemon is spawned with an empty argv (`Process.start(daemonBinaryPath, [])` at flutter_app/lib/core/daemon/daemon_lifecycle.dart:28 and `Process.start(binaryPath, [], mode: detached)` at flutter_app/lib/core/platform/platform_services_desktop.dart:265), and the desktop entry is `Exec=$HOME/.local/opt/heimdallm/heimdallm` with no arguments or `%U` (Makefile:708). So `pkill -x -f` matches both the daemon and the launcher-launched app; the narrowing that `-x` introduces is real but does not lose the cases this PR targets, and the uninstall comment now records that trade-off honestly.

The ERE-escaping fix is correct as written. `sed -e 's/\\/\\\\/g' -e 's/[]^$.*+?(){}|[]/\\&/g'` covers the full ERE metacharacter set (`. [ ] \ ( ) * + ? { } | ^ $`), backslashes are escaped first so the second pass cannot double-escape them, `]` sits first and `[` last in the class as required, and `-` is correctly absent (not a metacharacter outside a bracket expression). Test 4's fixture path `we+ird(dir)[x]` genuinely fails to match unescaped and matches once escaped, and the negative assertion against an unrelated path shows the escaping is not too loose. Single implementation shared by all three call sites, so the Makefile and the test cannot drift.

Install-order reasoning also checks out: the stop block runs last in install-linux, so an earlier failure never SIGTERMs a working daemon, and the 10×1s pgrep poll with an explicit "stop it manually" warning means a daemon that ignores SIGTERM is surfaced rather than assumed dead.

Nothing in this pass clears the reporting bar. Non-blocking observations, all for the summary only:

- Empty-pattern fallback (latent, not reachable as committed): `DAEMON_RE=$$($(PKILL_ESCAPE) …)` silently yields an empty string if the script cannot run — a lost exec bit, or `make -f /abs/path/Makefile` from elsewhere, since `PKILL_ESCAPE` is the relative `./scripts/pkill-escape.sh`. The compound recipe line swallows the non-zero status, and the result is `pkill -x -f ""`. On procps that matches nothing (tasks with a NULL cmdline are skipped under `-f`), so the worst case degrades to the pre-PR silent no-op rather than a broad kill — but a `[ -n "$$DAEMON_RE" ] || { echo …; exit 1; }` would turn it into a loud failure. The relative path itself matches the existing convention (`test-install-macos` already uses `./scripts/…`), so that part is consistent.
- Static check test 1 uses a greedy `.*` before `\(pkill\|pgrep\)`, so on a hypothetical line with two invocations only the last one's flags are inspected; no such line exists today. Also note `\|` is a GNU-sed BRE extension — harmless precisely because the script skips off Linux.
- The CI `1..0` assertion catches a wholesale skip but not per-test `# SKIP`s (test 2 on non-procps pkill, test 5 on a pgrep that matches metacharacters literally). Asserting the expected plan line (`1..7`) instead of just "not 1..0" would pin coverage exactly.
- `mktemp -u` for the fifo name is a theoretical TOCTOU inside a private `mktemp -d`; not worth changing.
- Comment density in the Makefile hunks is high relative to the rest of the file, but every block explains a non-obvious silent failure mode, which is the right trade here.

No unresolved reviewer concerns remain: @sergiotejon's three LOW findings (flag-order independence, test 2 skipping instead of hard-failing, the sleep race) are all visibly addressed in the diff. Recommend approve.

---

vbuenog:
Thanks both for the approvals. I fixed the non-blocking observations rather than carrying them — d158e1a.

**@Muriano — empty-pattern fallback.** Confirmed reachable enough to matter, so fixed: a stubbed escaper printing nothing produced `pkill -x -f ""` and a silent no-op, i.e. exactly the pre-PR behaviour the PR exists to remove. `dev-stop` and `install-linux` now abort with a message naming the consequence; `uninstall-linux` warns and continues instead, because aborting there would resurrect the half-finished uninstall this PR fixes and the `rm -rf` is already safe against a running process. Verified both paths.

**@Muriano — greedy `.*` in test 1.** Fixed: every invocation on a line is inspected via `grep -oE`, not just the last. Mutation-tested with `pkill -f "$DAEMON_RE"; pkill -x -f "$DAEMON_RE"` on one line — previously the correct neighbour would have masked it, now it is reported with its line number.

**@Muriano — CI catches `1..0` but not per-test `# SKIP`.** Fixed. I went with failing on any `# SKIP` rather than asserting a literal `1..7`: same coverage guarantee, and it does not need editing every time a test is added.

**Left as-is, agreeing with your own read:** the `mktemp -u` fifo name (theoretical TOCTOU inside a private `mktemp -d`) and the Makefile comment density.

Also noted your independent verification of the escaping character set and of the empty-argv assumption behind `-x` (including `platform_services_desktop.dart:265`, which I had not cited) — that matches what I found.

**@sergiotejon** — your three LOW findings from the first pass remain fixed; nothing in this round touched them.

Guard is 7/7 on Linux, both `make -n` parse checks pass, `make test-install-macos` unaffected.

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previous findings have been addressed. The empty-pattern guards, per-invocation static check, and CI skip detection are correctly implemented. The ERE escaping in pkill-escape.sh is sound (two-pass to avoid double-escaping backslashes, bracket expression ordered correctly for literal ] and [). No new issues found.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #662 after d158e1a. All previously reported findings are addressed in the current diff and I could not reproduce a new defect that clears the reporting bar.

Verified this pass:
1. **Empty-pattern fallback** — `dev-stop` and `install-linux` now abort with an explicit message when `pkill-escape.sh` yields an empty pattern (`|| exit 1` plus `[ -n "$DAEMON_RE" ]`), so a broken escaper can no longer degrade into `pkill -x -f ""` and a silent no-op. `uninstall-linux` warns and continues instead, which is the right asymmetry: aborting there would reintroduce the half-finished uninstall (desktop entry / icons / symlink left behind) that this PR fixes, and the subsequent `rm -rf` is safe against a running process. Note the assignment status in `uninstall-linux` is intentionally discarded (`2>/dev/null`, no `exit`), and the `[ -n ... ]` guard covers the missing/non-executable-script case, so that path is sound.
2. **Greedy match in static test 1** — extraction now iterates every `grep -oE '(pkill|pgrep)[^"]*'` occurrence per line rather than inspecting only the last, so a bare `pkill -f` can no longer hide behind a correct neighbour on the same line. Comment exclusion (`^[0-9]+:[[:space:]]*@?#`) still matches the recipes' `\t@#` lines and the top-of-file `#` block, so the deliberate `pkill -f` prose mentions are not flagged, and `PKILL_ESCAPE` / `pkill-escape.sh` do not false-positive (uppercase, and `pkill-` is not `pkill<space>`).
3. **Per-test skip detection in CI** — the new step fails on `1..0` *and* on any `# SKIP`, which is the stronger guarantee and avoids pinning a literal test count. On `ubuntu-latest` (procps + `/proc`, fractional `sleep` available) every assertion is expected to execute, so this is enforceable rather than aspirational.
4. **Escaping** — `sed -e 's/\\/\\\\/g' -e 's/[]^$.*+?(){}|[]/\\&/g'` covers the full ERE metacharacter set, backslash-first ordering is correct, and `]` leading / `[` trailing inside the bracket expression is the right POSIX construction. Tests 5-7 assert unescaped-misses → escaped-matches → escaped-does-not-match-an-unrelated-path, which is the meaningful triple.
5. **Fixture shape** — `/bin/cat` on a fifo gives argv exactly `[path]`, which is the only shape that makes the `-x` assertions non-vacuous; startup polls `/proc`, the writer outlives the assertion by an hour, and liveness is re-checked immediately before acting. No fixed-sleep races remain.

Non-blocking observations, deliberately not filed as issues:
- The CI step captures the guard's stdout into `$output` under `set -eu`, so on a guard failure the `printf` never runs and the `ok`/`not ok` stdout lines are lost — only the stderr `not ok` line and the failure detail reach the log. Enough to debug, but streaming (e.g. `tee`) would preserve the full TAP output.
- In test 2 the `without-x` pattern also matches the inner `make -C "$WORK"` cmdline, so that process may be signalled alongside the recipe shell. Harmless — the pattern is confined to the private `mktemp -d`, and either outcome yields the same absence of `RECIPE-COMPLETED` — but it means the test proves "something in the chain self-matched", slightly broader than the comment's claim.
- `-x` intentionally narrows `uninstall-linux` so an app launched *with* arguments is no longer matched; this is documented in the recipe with the rationale, and plain `-f` never reached that case anyway. Recording it as a known behaviour change rather than a defect.
- Makefile comment density in the three recipes is high; the author consciously kept it, and given that both failure modes are invisible in normal output I agree with that call.

Converged. No blockers, no unresolved reviewer concerns.

---

vbuenog:
Both observations fixed in 624053c — thanks, the second one was a real gap between what test 2 asserts and what its comment claimed.

**Test 2 matched more than the recipe shell.** Confirmed and fixed. A command-line `PAT=` lands in make's own `/proc` cmdline, so `pkill -f` signalled the make process too. Evidence, same fixture in both modes:

```
command line:   /bin/sh -c sleep 3; echo done …/opt/heimdalld
                make -C …/patcheck slow PAT=…/opt/heimdalld      ← also matched
environment:    /bin/sh -c sleep 3; echo done …/opt/heimdalld     ← only this
```

The pattern now travels through the environment, so the test pins the self-match on the recipe shell, which is what the comment says. The comment records why.

**CI lost the TAP stdout on failure.** Correct — `output="$(make …)"` under `set -e` aborts the step before the echo. Now streamed through `tee`, with `set -o pipefail` so a failing guard still fails the step rather than being masked by tee's exit status. Verified by removing an `-x`: **0** `ok` lines survived before, **6** after, step exit 2 in both cases.

**Left as-is, agreeing with your read:** `-x` narrowing `uninstall-linux` for an app launched with arguments (documented in the recipe with the rationale), and the comment density in the three hunks.

Guard still 7/7 on Linux; both `make -n` parse checks pass.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review pass. The two observations from the previous round are both genuinely fixed and I could not find anything new that clears the reporting bar.

Verified in this diff:
- **Test 2 self-match attribution.** `PAT` now reaches the fixture Makefile through the environment (`PAT=… make -C "$WORK" without-x`), so the pattern no longer appears in make's own `/proc/<pid>/cmdline`. The `without-x` recipe's `sh -c` line is now the only cmdline containing it, which is exactly what the assertion claims to prove. The `with-x` control target still completes, so the test discriminates rather than passing vacuously. The added comment records the reasoning, so a future edit back to a command-line assignment is unlikely to slip through unnoticed.
- **CI keeps the TAP stdout on failure.** `make test-install-linux 2>&1 | tee "$log"` with `set -o pipefail` preserves both properties that matter: a failing guard still fails the step (tee's zero exit cannot mask it), and the `ok`/`not ok` lines reach the Actions log before the abort. Reading the captured `$log` back into `$output` keeps the two grep guards (`^1\.\.0` wholesale skip, and any per-test `# SKIP`) operating on the full transcript. Failing on any `# SKIP` rather than asserting a literal `1..7` is the right call — same coverage guarantee, no edit needed per new test.

Also re-checked the pieces the fix commit touched indirectly: the escape class `[]^$.*+?(){}|[]` still covers every ERE metacharacter with `\` handled in the first pass and `[` last in the class (so `[.` is never read as a collating symbol); `-x` still guards every `pkill`/`pgrep` that targets a process in the Makefile, and the static check in test 1 still inspects every invocation on a line rather than only the last.

Non-blocking observations, none of which I would hold an approval over:
- `printf '%s\n' "$output" | grep -q …` under `pipefail` is theoretically invertible by SIGPIPE on the `printf` side if the guard's output ever exceeded the 64 KiB pipe buffer (grep -q exits on first match, printf takes 141, pipefail propagates it, the `if` reads false). The TAP output is a handful of lines, so this is unreachable in practice — noting it only because the guard is deliberately built to be un-skippable, and `grep -q "…" "$log"` reading the file directly would remove the theoretical hole for free.
- Comment density in the three Makefile hunks remains high — I agree with your earlier read that the rationale is worth keeping inline, since the whole class of bug is invisible in normal output.
- Latent, not an issue: the `-x` contract depends on the installed daemon being spawned with empty argv. If a future change ever adds a flag to that spawn, the stop becomes a silent no-op again. The guard asserts the shape, which is the right place for it to be caught.

Guard reported 7/7 on Linux with both `make -n` parse checks passing. Converged — no findings.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previously raised findings have been addressed. The PAT-via-environment fix for test 2, the tee+pipefail fix for CI TAP output, the pkill-escape.sh implementation, the empty-pattern guards, and the per-invocation -x static check are all sound. No new issues found in the current diff.

---

vbuenog:
Fixed in 8014f80.

**@Muriano — the SIGPIPE hole in `grep -q`.** You are right that it is unreachable at this output size, and equally right that it is not worth leaving in a step whose entire purpose is to be un-skippable. Both guards now read `"$log"` directly instead of piping a variable into `grep -q`, which removes the inversion path for free.

**Left as-is, agreeing with your read:** the Makefile comment density, and the `-x` contract depending on the daemon being spawned with empty argv — the guard asserts that shape, which is the right place for a future change to be caught.

Guard still 7/7 on Linux; both `make -n` parse checks pass.

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

Solid, carefully-engineered PR that fixes a real silent-failure class (pkill -f self-matching the recipe shell and ERE metacharacter paths) in dev-stop/install-linux/uninstall-linux, with a shared escaping script and a thorough behavioural TAP guard wired into CI. All three rounds of reviewer feedback (Muriano, sergiotejon) were addressed and explicitly closed out; the remaining left-as-is items were mutually agreed. Only minor residual risks remain.

### Issues

🟡 **Makefile:9** — PKILL_ESCAPE is a cwd-relative path (./scripts/pkill-escape.sh). It works for `make` and `make -C` from the repo root, but invoking the Makefile via `make -f /path/to/Makefile <target>` from another directory makes the escape call fail, which now aborts dev-stop and install-linux at the empty-pattern guard. Consistent with the existing `$(shell pwd)`-based conventions in this Makefile, but worth anchoring to the Makefile's own directory (e.g. $(dir $(lastword $(MAKEFILE_LIST)))).
🟡 **Makefile:784** — The -x fix couples correctness to the invariant that the daemon/app is always exec'd with argv exactly equal to the binary path (verified today at platform_services_desktop.dart:265). If a future change spawns heimdalld with any argument, pkill -x -f silently stops matching and install-linux again leaves the old daemon running with no error — the same silent failure mode this PR removes. The invariant is documented in comments but has no automated cross-check tying the Dart spawn site to the Makefile assumption.
🟡 **scripts/pkill-escape.sh:28** — sed processes its input line by line, so a path containing an embedded newline is escaped per-line and the resulting pattern cannot match the target cmdline (pkill would then silently no-op, mirroring the failure the script exists to prevent). Pathological input, but a guard (reject arguments containing newlines with exit 2) would make it explicit.
🟡 **scripts/test-linux-install.sh:118** — The static check extracts invocations with grep -oE '(pkill|pgrep)[^"]*', which truncates at the first double quote. A legal GNU-getopt-permuted form like `pkill -f "$PAT" -x` would be reported as a false-positive offender, and conversely the `*-x*` glob matches any '-x' substring anywhere in the extracted text (e.g. inside an unquoted argument), not specifically the flag. Acceptable for the current Makefile contents, but brittle if invocation styles change.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review pass on 8014f80. The only change since the last round is the CI guard step: both assertions now grep the tee'd log file (`$RUNNER_TEMP/pkill-guard.log`) instead of piping a captured variable into `grep -q`, which removes the SIGPIPE-inversion path and, together with the earlier `tee` + `set -o pipefail` change, keeps the TAP `ok`/`not ok` lines in the step log even when the guard exits non-zero. I traced the new step end to end: `set -euo pipefail` means an unset `RUNNER_TEMP` fails loudly rather than silently writing to `/pkill-guard.log`; a non-zero `make test-install-linux` still fails the step through the pipe; `grep -q` returning 1 inside an `if` does not trip `set -e`; and both the wholesale-skip (`^1\.\.0`) and per-test (`# SKIP`) detectors are anchored on strings only the guard's own `printf`s emit, so there is no false-positive source in the merged 2>&1 stream. Everything raised in prior rounds remains addressed and I found nothing new that clears the bar. Spot-checks on the unchanged parts held up: `VAR=$(cmd) || exit 1` does propagate the command-substitution status for a simple assignment, so the `|| exit 1` guards in `dev-stop`/`install-linux` are real; `PKILL_ESCAPE`'s relative `./scripts/...` is safe because make chdir's under `-C`; the two-pass sed class `[]^$.*+?(){}|[]` covers the ERE metacharacter set with `]` first and `[` last as documented; and the fifo fixture genuinely produces argv == [path] (a shell script would have matched `-x` vacuously). Non-blocking observations, all previously acknowledged and none worth holding the PR for: the comment-to-code ratio in the three Makefile hunks is very high — accurate, but it makes the recipes harder to scan, and a future reader editing the stop block has ~25 lines of prose to get through; the `-x` contract still rests on the daemon being spawned with empty argv, which test 3 pins, so a future change that adds a flag will fail the guard rather than silently regress; `-x` in `uninstall-linux` narrows the app match so an app launched with arguments is no longer signalled, which the recipe documents along with why the `rm -rf` is safe against a running process; and the `mktemp -u` fifo name is a theoretical TOCTOU inside an already-private `mktemp -d`. Converged.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previously raised findings have been addressed. The SIGPIPE hole in grep -q is fixed (guards now read from $log directly), the PAT variable travels through the environment, CI catches per-test SKIPs, the static check inspects every invocation on a line, and the empty-pattern fallback is handled. The pkill-escape.sh sed expression correctly orders ] first and [ last in the bracket expression and escapes backslashes in a separate first pass. No new issues found.

---

ivanmunozruiz (approved):
## 🤖 Heimdallm AI Review

PR sólido y muy trabajado: corrige el auto-match de `pkill -f` contra el shell de la propia receta (añadiendo `-x`), el fallo silencioso con rutas que contienen metacaracteres ERE (vía scripts/pkill-escape.sh compartido), y añade un guard conductual (scripts/test-linux-install.sh) cableado en CI con protecciones contra skips silenciosos. Todas las observaciones de los revisores (Muriano, sergiotejon) a lo largo de cuatro rondas fueron aceptadas y corregidas o explícitamente acordadas como aceptables; no quedan preocupaciones sin resolver. Solo restan riesgos residuales menores, documentados en el propio código.

### Issues

🟡 **Makefile:786** — El contrato de `-x` depende de que el daemon se lance siempre con argv vacío (cmdline == ruta exacta). Si un cambio futuro en el spawner (p. ej. platform_services_desktop.dart) añadiera argumentos, el stop de install-linux/dev-stop dejaría de emparejar silenciosamente — el mismo modo de fallo que este PR elimina. El guard asegura la forma del fixture pero no puede detectar un cambio en el código Dart real. Riesgo residual documentado y ya discutido con los revisores.
🟡 **scripts/test-linux-install.sh:106** — El chequeo estático extrae invocaciones con `grep -oE '(pkill|pgrep)[^"]*'`, que corta en la primera comilla doble: una invocación con `-x` situado después de un argumento entrecomillado (p. ej. `pkill -f "$PAT" -x`) se marcaría como ofensora falsamente, y patrones sin comillas podrían producir falsos negativos. Es una heurística suficiente para el estilo actual del Makefile, pero frágil ante reordenaciones de flags poco convencionales.
🟡 **scripts/pkill-escape.sh:28** — El escapado vía sed es orientado a líneas: una ruta que contuviera un salto de línea no se escaparía correctamente. Caso extremo casi imposible en $HOME o en un checkout real, y pkill -f tampoco lo manejaría bien de todos modos; solo informativo.

---

vbuenog:
All four accepted and fixed in 7f816c9 — thanks, these were the most substantive of the round.

**🟡 `PKILL_ESCAPE` cwd-relative.** Correct, and worse than before my last commit: since the callers now abort on an empty pattern, `make -f /abs/Makefile` from another directory turned into a *hard failure* of `dev-stop` and `install-linux`. Anchored to `$(dir $(lastword $(MAKEFILE_LIST)))` as you suggested. I also anchored `DAEMON_BIN`, which had the same cwd-dependence via `$(shell pwd)` — pre-existing, but running from `/tmp` pointed it at `/tmp/daemon/bin/heimdallm`, so leaving one relative and one absolute would have been the worse half of a fix. Verified from `/tmp` with an absolute `-f`.

**🟡 No automated cross-check for the empty-argv invariant.** This was the best catch of the round, so it now has one: a new static test asserts every `Process.start` at the two known spawn sites passes `[]`. Mutation-tested — changing `Process.start(daemonBinaryPath, [])` to `['--foo']` fails the guard with *"a daemon/app spawn passes arguments, which breaks the pkill -x contract"*. It also fails loudly if the spawn sites move, rather than passing vacuously.

That change required restructuring the platform gate: the static tests are portable and are exactly the ones that catch this, so they now run everywhere and **only the behavioural half skips** off Linux. On macOS you get 2 real tests plus 6 skips instead of `1..0`.

**🟡 sed is line-oriented in `pkill-escape.sh`.** Fixed with your suggested `exit 2`. Worth noting the trap I hit implementing it: `case $1 in *"$(printf '\n')"*)` matches *every* path, because command substitution strips the trailing newline and leaves an empty pattern. It counts newlines instead.

**🟡 Brittle extraction in the static check.** Fixed both halves you identified: quoted arguments are blanked before extraction, so `pkill -f "$PAT" -x` is no longer a false positive; and flags are matched as whole tokens, so an `-x` inside an argument is no longer accepted. Mutation-tested all four forms (flag after the pattern → clean; `-x` inside an argument → flagged; `--exact` → clean; no `-x` → flagged).

**Two defects in my own work this round**, found by verifying it rather than by review: the early exit on the skip path ignored `FAIL_COUNT`, so a failing static test would have reported success on every non-Linux machine; and test 1 passed vacuously when the Makefile was unreadable. Both fixed and pinned.

Guard is 8/8 on Linux; both `make -n` parse checks pass.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-revisión de #662 tras 7f816c9. Verificado que las cuatro observaciones del pase anterior están resueltas: (1) `PKILL_ESCAPE` y `DAEMON_BIN` quedan ancladas a `$(dir $(lastword $(MAKEFILE_LIST)))`/`$(abspath ...)`, de modo que `make -f /abs/Makefile` desde otro directorio ya no convierte el aborto por patrón vacío en un fallo duro de `dev-stop`/`install-linux`; (2) existe ahora una comprobación estática automatizada del contrato argv-vacío sobre los dos sitios de `Process.start`, que además falla ruidosamente (no vacuamente) si no encuentra ninguna llamada; (3) `pkill-escape.sh` rechaza con `exit 2` una ruta con newline, contando líneas con `printf '%s' | wc -l` en lugar del glob `*"$(printf '\n')"*` que —correctamente señalado por el autor— habría coincidido con todo; (4) la extracción estática blanquea los argumentos entrecomillados antes de extraer y compara flags como tokens completos, con lo que `pkill -f "$PAT" -x` deja de ser falso positivo y un `-x` embebido en un argumento deja de aceptarse. También verificado el reordenamiento del gate de plataforma: los tests estáticos corren en todas partes y la ruta de skip ahora honra `FAIL_COUNT`, cerrando el agujero por el que un estático fallido reportaba éxito en macOS. La clase de caracteres del sed (`[]^$.*+?(){}|[]`) sigue siendo correcta y completa para ERE, con `]` primero y `[` último por el símbolo de colación POSIX. Nada en este pase alcanza el umbral de issue.

Observaciones no bloqueantes, para el registro: (a) el matcher de spawn `*"Process.start("*", []"*` es por línea, así que una llamada `Process.start(` partida en varias líneas o escrita con `const []` se reportaría como offender aunque sea correcta; requiere un cambio futuro para volverse alcanzable, y el modo de fallo es ruidoso, no silencioso; (b) `spawn_found` es un contador global sobre los dos ficheros, de modo que si el spawn del daemon se mueve a un fichero nuevo mientras el otro sitio conserva el suyo, la guarda pasa — el `file not found` sí lo cubre para los paths listados, pero no para un tercer sitio nuevo; (c) la comprobación de `1..0` en CI quedó efectivamente muerta al pasar los estáticos a correr siempre (el plan mínimo es ahora `1..8`), aunque es inofensiva y el chequeo de `# SKIP` cubre el caso real; (d) `$(abspath)` no resuelve symlinks, a diferencia del `$(shell pwd)` lógico anterior — irrelevante para la comparación de cmdline que hace `pkill -x`; (e) la densidad de comentarios en los tres hunks del Makefile sigue siendo alta, ya discutida y aceptada conscientemente. Guarda 8/8 en Linux, 2 reales + 6 skips en macOS, ambos `make -n` pasan.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All four previously flagged findings have been addressed. The PKILL_ESCAPE path is now anchored to the Makefile's directory, the empty-argv invariant has an automated cross-check, the sed newline edge case is handled, and the static extraction logic is robust against flag ordering and quoted arguments. The author also self-caught two additional defects (vacuous pass on unreadable Makefile, FAIL_COUNT ignored on skip path) and fixed them. No new issues found in this revision.

---

vbuenog:
Thanks both. @Muriano — two of your five observations were actionable, so they are fixed in 7d5d926 rather than carried.

**(a) Per-line spawn matcher.** Right, and it would have been a false *offender* on correct code — the worst kind of guard failure, since it teaches people to ignore it. Each file is now joined onto one logical line before the call is extracted, so the check judges the argument list rather than the formatting. Mutation-tested all four: added argument → fails; call wrapped across lines → passes; `const []` → passes; spawn removed → fails.

**(b) Global `spawn_found`.** Also right: the guard would still have passed if the daemon spawn moved to a new file while the other listed site kept its own. Counted per file now — each listed file must contribute at least one call, and losing its spawn fails the guard (case D above).

**Left as-is, with reasons:**
- **(c) the `1..0` assertion is dead** — correct, the minimum plan is `1..8` now. Keeping it: it costs nothing and would catch a future restructure that reintroduced a wholesale skip, which is precisely the failure mode it was added for. The `# SKIP` check covers the reachable case, as you say.
- **(d) `$(abspath)` does not resolve symlinks** — agreed it is irrelevant here: `pkill -x` compares the cmdline string the spawner used, not a canonicalised path, so resolving symlinks could actually *break* the match rather than help it.
- **(e) comment density** — already discussed and consciously kept.

Guard 8/8 on Linux, 2 real + 6 skips off it; both `make -n` parse checks pass.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review de 7d5d926 sobre #662. Verificado lo que cambió respecto al pase anterior: (a) el matcher de spawn ahora une cada fichero en una línea lógica (`tr '\n' ' '`) antes de extraer la llamada, por lo que una `Process.start` partida en varias líneas o escrita como `const []` ya no se reporta como offender; y (b) el conteo de `Process.start(` es por fichero (`calls=$(grep -c … "$f")`, con `continue` y offender si es 0), de modo que si el spawn del daemon se moviera a otro fichero el guard falla en vez de pasar apoyado en el otro site. Comprobado contra el código real del repo: `flutter_app/lib/core/daemon/daemon_lifecycle.dart:28` (`Process.start(daemonBinaryPath, [])`) y `flutter_app/lib/core/platform/platform_services_desktop.dart:265` (`Process.start(binaryPath, [], mode: ProcessStartMode.detached)`) son las únicas llamadas en esos ficheros y ambas satisfacen el patrón `*", []"*`, así que el nuevo test 2 pasa de forma no vacua y el contrato de argv vacío que sostiene `pkill -x` queda anclado. El resto del PR (escapado ERE compartido en scripts/pkill-escape.sh con rechazo de newline, anclaje de PKILL_ESCAPE y DAEMON_BIN al directorio del Makefile, aborto con mensaje ante patrón vacío en dev-stop/install-linux y warn+continue en uninstall-linux, paso de CI con tee+pipefail y detección de `1..0` y `# SKIP`) sigue igual que en el pase que ya validé. No encuentro nada en las líneas nuevas que alcance el umbral de issue; devuelvo `issues: []` como resultado de convergencia.

Observaciones que quedan fuera del umbral, para el autor:

- **Trampa latente en la extracción del spawn (no reportable hoy).** `grep -oE 'Process\.start\([^)]*'` corta en el primer `)`. Con el código actual no hay problema porque ninguna de las dos llamadas anida un paréntesis antes del `, []`. Pero si alguien escribiera `Process.start(_resolveBin(path), [])`, la extracción se quedaría en `Process.start(_resolveBin(path` y el guard reportaría un falso offender sobre código correcto — exactamente la clase de fallo que ya corregiste en (a). Un `[^;]*` acotado, o exigir el `, []` dentro de la misma llamada balanceada, cerraría el hueco; requiere un cambio futuro para volverse alcanzable, así que va aquí y no en issues.
- **`, []` se compara con un espacio exacto.** `Process.start(bin,[])` o `Process.start(bin, [ ])` serían offenders. `dart format` produce siempre la forma con un espacio, así que en la práctica no muerde; normalizar espacios antes del `case` lo haría inmune al formateo.
- **El join con `tr '\n' ' '` es sobre el fichero entero**, así que una `Process.start` comentada o dentro de un string literal contribuiría al conteo y podría generar un offender espurio. Hoy no ocurre en ninguno de los dos ficheros.
- **Etiqueta desalineada:** el encabezado dice "Behavioural half (tests 3-7)" pero el bloque de skips lista 6 asserts (tests 3-8, con el plan mínimo ahora en `1..8`). Puramente cosmético.
- De acuerdo con tus decisiones de mantener el chequeo de `1..0` (coste cero, cubre una reestructuración futura), la no resolución de symlinks en `$(abspath)` (irrelevante: `pkill -x` compara la cadena de cmdline del spawner) y la densidad de comentarios en el Makefile: son load-bearing porque documentan por qué `-x` no es un refinamiento sino un requisito.

---

sergiotejon (approved):
## 🤖 Heimdallm AI Review

All previously identified issues have been addressed across multiple revision rounds. The pkill-escape implementation, static and behavioural guards, CI integration, and error-handling paths are thorough and well-tested. No new defects found in the current diff.

---

ivanmunozruiz (changes requested):
## 🤖 Heimdallm AI Review

El diff suministrado corrige el problema principal, pero requiere cambios: las rutas se usan como expresiones regulares sin escapar y el nuevo guard no es portable ni se ejecuta en CI.

### Issues

⚠️ **Makefile:776** — pkill/pgrep -f interpreta DAEMON como una ERE, no como una ruta literal. Un HOME válido con metacaracteres como +, ( o [ impide el match o invalida el patrón, dejando silenciosamente el daemon anterior en ejecución; afecta también a dev-stop y uninstall-linux.
⚠️ **scripts/test-linux-install.sh:43** — El guard solo comprueba que existan pkill y pgrep. macOS dispone de ambos pero carece de /proc, por lo que la lectura posterior de /proc/$TARGET/cmdline produce un fallo determinista en un árbol correcto.
⚠️ **Makefile:76** — test-install-linux no se invoca desde make test ni desde el job installer-shell de CI; los dry-runs existentes no ejercitan pkill. Eliminar accidentalmente -x seguiría dejando CI en verde y reintroduciría la regresión.

---

ivanmunozruiz (commented):
## 🤖 Heimdallm AI Review

La lógica de parada y escape de pkill es correcta, pero el flujo de recuperación anunciado no funciona con la app abierta y el nuevo guard conserva una carrera pre-exec.

### Issues

⚠️ **Makefile:795** — Tras el SIGTERM externo, la app abierta conserva daemonHealthProvider=true; Status sigue mostrando “Stop server” y tanto Stop como Restart fallan al llamar /shutdown antes de invalidar el estado o lanzar el nuevo daemon. La instrucción impresa no puede seguirse y el servicio queda apagado hasta cerrar completamente y relanzar la app.
🟡 **scripts/test-linux-install.sh:96** — /proc/$pid/cmdline es legible inmediatamente después de fork, antes de que el hijo ejecute la copia de cat, por lo que este bucle puede no esperar y cmdline_of puede leer el argv heredado del shell. Esto introduce fallos intermitentes en CI; debe esperarse hasta que el cmdline sea exactamente _sf_path.

---

---
_Not approving: this review raised 2 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

ivanmunozruiz (commented):
## 🤖 Heimdallm AI Review

El PR corrige el daemon obsoleto en instalaciones normales, pero el match exacto falla con rutas canonicalizadas mediante symlinks y el fixture mantiene una carrera que puede volver CI flaky.

### Issues

⚠️ **Makefile:784** — El patrón usa la ruta textual $HOME/.local/opt/heimdallm/heimdalld, pero la aplicación deriva el daemon de Platform.resolvedExecutable, que resuelve symlinks. Si $HOME o un directorio intermedio es un symlink, argv[0] contiene la ruta canonicalizada; pkill -x -f no encuentra el proceso y la instalación termina con éxito dejando el daemon antiguo activo. Debe canonicalizarse la ruta antes de escaparla.
🟡 **scripts/test-linux-install.sh:95** — Esperar solo a que /proc/$pid/cmdline sea legible no sincroniza con exec: ese archivo ya existe tras fork y todavía puede contener el cmdline heredado del shell. La comprobación siguiente puede fallar espuriamente bajo carga; debe esperarse hasta que cmdline sea exactamente $_sf_path o el proceso termine.

---

---
_Not approving: this review raised 2 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

ivanmunozruiz (commented):
## 🤖 Heimdallm AI Review

El PR aborda correctamente el auto-match de pkill, pero aún puede dejar activo el daemon anterior cuando la ruta instalada contiene symlinks. Además, el guard de CI conserva una carrera de arranque.

### Issues

⚠️ **Makefile:784** — DAEMON_RE usa la ruta léxica $HOME/.local/opt/heimdallm/heimdalld, mientras la aplicación deriva la ruta del daemon de Platform.resolvedExecutable, que resuelve symlinks. Si $HOME, .local u otro ancestro es un symlink, argv[0] contiene la ruta canónica y pkill -x -f no encuentra el proceso; install-linux deja el daemon anterior ejecutándose silenciosamente. Canonicaliza la ruta antes de escaparla y añade un fixture con un ancestro enlazado.
🟡 **scripts/test-linux-install.sh:97** — El bucle solo espera que /proc/$pid/cmdline sea legible, pero ese archivo existe inmediatamente después de fork y antes de que el hijo ejecute el binario fixture. En un runner cargado, cmdline_of puede observar todavía el argv del shell y producir un fallo intermitente. Espera hasta que argv[0] sea exactamente $_sf_path y detecta si el proceso termina.

---

---
_Not approving: this review raised 2 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

ivanmunozruiz (commented):
## 🤖 Heimdallm AI Review

El PR implementa la parada del daemon anterior y añade cobertura Linux, pero aún puede completar la instalación dejando activa la versión antigua. Requiere cambios.

### Issues

⚠️ **Makefile:784** — El patrón usa la ruta léxica de $HOME, pero la aplicación construye la ruta del daemon desde Platform.resolvedExecutable, que resuelve symlinks. Con $HOME o .local enlazado, el argv del daemon contiene la ruta canónica y pkill -x no lo encuentra, reproduciendo el fallo que corrige el PR.
⚠️ **Makefile:787** — Cualquier salida no cero de pkill se interpreta como «no había daemon». Si pkill no está instalado o falla operativamente, stderr se descarta y install-linux anuncia éxito mientras el daemon antiguo continúa ejecutándose; pkill/pgrep no se validan como prerrequisitos.
⚠️ **scripts/test-linux-install.sh:153** — Las pruebas de comportamiento ejecutan un Makefile sintético, no el bloque real de install-linux. Eliminar completamente la nueva parada, usar una ruta incorrecta o evitar el escaper seguiría dejando los siete tests en verde, por lo que el comportamiento central de #661 no queda protegido.

---

---
_Not approving: this review raised 3 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

PR #666, opened 2026-08-05
fix(pipeline): re-check HEAD SHA before publishing a review

### LINKED ISSUE #664: Review publish doesn't recheck HEAD SHA, so a slow run can publish a stale verdict over a newer commit
## Symptom

On `freepik-company/ai-bumblebee-proxy#1802` (8 force-pushes in ~20 minutes), Heimdallm approved commit `306bfcc7` at 07:39:51, then 103 seconds later published `REQUEST_CHANGES` **on that same already-superseded commit** at 07:41:34 — after the PR had already moved on:

```
07:39:51  review published  pr=1802  event=APPROVE            head_sha=306bfcc7...  (github_review_id=4862145756)
07:40:34  review-worker: stale message (HEAD SHA changed), skipping   msg_sha=0ef4748a...  current_sha=1b284050...
07:41:31  review published  pr=1802  event=APPROVE            head_sha=eebf6a82...  (github_review_id=4862164226)
07:41:35  review published  pr=1802  event=REQUEST_CHANGES    head_sha=306bfcc7...  (github_review_id=4862165122)  <- stale
07:43:18  review published  pr=1802  event=APPROVE            head_sha=0ef4748a...  (github_review_id=4862181389)
```

The `REQUEST_CHANGES` at 07:41:35 targets `306bfcc7`, a commit that was already approved and already superseded by two newer pushes by the time it got published. Net effect: a human watching the PR sees an approval, then a change-request that references stale code, then another approval — confusing and noisy, and in the `high`-severity case it would incorrectly block merge on outdated code.

## Root cause

`daemon/internal/pipeline/pipeline.go`, step 8 (`Run`, around line 826-850): once a review run has fetched the diff and gotten a result from Claude, it stores the row and calls `p.gh.SubmitReview` unconditionally — there is no check that the PR's current HEAD SHA still matches the SHA this run started against:

```go
ghReviewID, ghReviewState, publishErr := p.gh.SubmitReview(
    pr.Repo, pr.Number,
    AnnotateBodyForEvent(reviewBody, reviewEvent, len(result.Issues)),
    reviewEvent,
)
```

The codebase already has a "stale message" guard at intake (`review-worker: stale message (HEAD SHA changed), skipping`, seen in the same log window at 07:40:34), which correctly discards *queued* work for an outdated SHA. But once a run has already started, nothing re-validates freshness before the final publish — a long-running Claude invocation (the `review_timing` stats show a `bucket_very_slow` tail up to ~3525s) can finish and publish well after the PR has moved on.

This gap predates all of the later hardening work (#243/#245-247, #322, #325, #359) — it's present in the original `SubmitReview` call from the initial pipeline implementation (`6f874cef`) and was never addressed by any of the subsequent dedup/circuit-breaker/orphan-marking passes, which all guard *before* a run starts or *after* a publish fails, not immediately before a successful publish.

## Suggested fix

Immediately before `p.gh.SubmitReview`, re-fetch (or reuse an already-fresh) current HEAD SHA for the PR and compare against `pr.Head.SHA` used for this run. If they differ, skip publishing and mark the review row the same way the existing `SupersededReviewID` sentinel is used elsewhere (`pipeline.go` already has this concept for a different case — reuse it here), rather than posting a verdict on code that's no longer at HEAD.

## Repro data
- Repo: `freepik-company/ai-bumblebee-proxy` PR #1802
- Author force-pushed 8 times in ~20 minutes
- Stale publish: `github_review_id=4862165122`, event=REQUEST_CHANGES, severity=high, head_sha=306bfcc7d5efa0227b8779432330f8a07b2308b5 (already 2 commits behind HEAD `1b2840500787c1f393ad0f08d431c227f691209a` at publish time)

sergiotejon:
Fixes #664.

## Problem

`pipeline.Run` step 8 published unconditionally. Nothing revalidated that the commit the run analysed was still at HEAD, and the run holds no lock on the PR while the CLI executes — a window whose `bucket_very_slow` tail reaches past 3000s.

On `freepik-company/ai-bumblebee-proxy#1802` (8 force-pushes in ~20 min) that produced `REQUEST_CHANGES` at `severity=high` on `306bfcc7`, a commit already approved and superseded by two further pushes:

```
07:39:51  APPROVE           head_sha=306bfcc7   (4862145756)
07:41:31  APPROVE           head_sha=eebf6a82   (4862164226)
07:41:35  REQUEST_CHANGES   head_sha=306bfcc7   (4862165122)  <- stale
07:43:18  APPROVE           head_sha=0ef4748a   (4862181389)
```

## Fix

Re-resolve the live HEAD at the top of step 8 and retire the stored row with the existing `SupersededReviewID` sentinel when it no longer matches.

**This guard already existed — on the other publish path.** The deferred publish-worker (`main.go:1183-1216`) does `GetPRSnapshot` + `pendingReviewInvalidReason` immediately before `SubmitReview`, with the comment *"A deferred review is valid only for the commit it analysed."* Only the hot path was missing it. This PR is the hot-path half; the two sites now agree and reuse the same sentinel.

## Three decisions worth reviewing

**1. The guard precedes the multi-mode comment loop, not just `SubmitReview`.**
In `multi` mode the per-issue `PostComment` loop runs first and is already a write to the PR. Checking only before `SubmitReview` would leave N inline comments about superseded code with *no* summary to explain them — worse than the bug. Pinned by `TestPipeline_Run_SkipsPerIssueCommentsWhenHeadMoved`.

**2. Retiring the row is load-bearing, not bookkeeping.**
Step 7 stores the row with `github_review_id == 0`, which is exactly what `ListUnpublishedReviews` selects on. Skipping the publish without retiring would hand the stale verdict to the publish-worker to post later — the bug would survive the fix. Pinned by the sentinel assertion in `TestPipeline_Run_RetiresReviewWhenHeadMovedDuringRun`.

**3. Resolution failure fails open (publishes).**
Deferring on a transient error would delay every publish whenever GitHub blips and would need a lifecycle state that doesn't exist. Publishing stale additionally requires HEAD to have moved *and* the call to fail in the same run. The deferred path stays hardened either way, since it revalidates under its own claim. Happy to switch to fail-closed if you'd rather — it's a one-line change plus a new skip reason.

## API cost

One extra `GetPRHeadSHA` per *published* review — reuses the `HeadSHAResolver` already on the pipeline's `gh` interface, so no interface change. Bounded by the breaker (3/PR/24h, 20/repo/hour). Runs are unaffected that skip before the publish block.

`TestPipeline_Run_HydratesHeadSHAWhenMissing` asserted exactly one `GetPRHeadSHA` call, so it needed updating. I kept its intent rather than loosening it: the two calls are documented as non-interchangeable (hydration resolves an unknown SHA *before* work; the re-check detects drift *after*), and `TestPipeline_Run_NoRecheckWhenRunSkipsEarly` is new coverage that the re-check never fires outside the publish block. Flagging it explicitly since relaxing a call-count assertion deserves a second opinion.

Worth coordinating with `feat/gh-api-efficiency-v2` if that branch changes how PR snapshots are fetched.

## Tests

New `daemon/internal/pipeline/pipeline_head_moved_test.go`:

| Test | Asserts |
|---|---|
| `RetiresReviewWhenHeadMovedDuringRun` | no `SubmitReview`; row marked `SupersededReviewID`; analysed SHA preserved; `Run` returns nil |
| `SkipsPerIssueCommentsWhenHeadMoved` | no `PostComment` in multi mode — pins guard placement |
| `PublishesWhenHeadUnchanged` | control; exactly 1 re-check call |
| `PublishesWhenHeadRecheckFails` | documents fail-open |
| `NoRecheckWhenRunSkipsEarly` | API budget: no re-check when the run short-circuits |

`make test-docker` green across all 17 daemon packages (vet included).

## Not covered here

`SkipReasonHeadChanged` is emitted for the retired review. The Flutter activity tile has a catch-all (`_ => reason.replaceAll('_', ' ')`) so it renders without a UI change; adding a friendly label is a trivial follow-up if you want one.

### DIFF
--- a/daemon/internal/pipeline/pipeline.go
+++ b/daemon/internal/pipeline/pipeline.go
@@ -46,6 +46,31 @@ type HeadSHAResolver interface {
 	GetPRHeadSHA(repo string, number int) (string, error)
 }
 
+// PRSnapshotFetcher returns the live state of a PR, not just its HEAD.
+//
+// Optional capability, discovered by type assertion on the pipeline's gh
+// dependency rather than added to the required set. *github.Client implements
+// it; the package's many test doubles do not, and widening the mandatory
+// interface would force every one of them to grow a method that most tests do
+// not exercise. When absent, the publish guard degrades to a HEAD-only check via
+// HeadSHAResolver.
+type PRSnapshotFetcher interface {
+	GetPRSnapshot(repo string, number int) (*github.PRSnapshot, error)
+}
+
+// CommitAnchoredReviewer submits a review pinned to an explicit commit.
+//
+// Optional for the same reason as PRSnapshotFetcher, and for the reason stated
+// on github.Client.SubmitReviewForCommit itself: keeping it off the
+// GitHubReviewer contract preserves the test doubles. Anchoring matters because
+// the plain SubmitReview lets GitHub attach the review to whatever is at HEAD
+// when the request lands, so a push racing the submit would misattribute
+// findings from the analysed commit to newer code. With the anchor, the worst
+// case degrades from "misattributed" to "shows as outdated".
+type CommitAnchoredReviewer interface {
+	SubmitReviewForCommit(repo string, number int, body, event, commitID string) (int64, string, error)
+}
+
 // CLIExecutor detects and runs an AI CLI tool.
 type CLIExecutor interface {
 	Detect(primary, fallback string) (string, error)
@@ -211,74 +236,134 @@ func (p *Pipeline) stopIfRepoBecameIneligible(
 	return true, nil
 }
 
-// retireIfHeadMoved re-resolves the PR's live HEAD immediately before anything
-// is written to GitHub and retires the freshly stored review when the commit it
-// analysed is no longer at HEAD.
+// retireIfPRMoved re-resolves the PR's live state immediately before anything is
+// written to GitHub and retires the freshly stored review when the PR has moved
+// on from what was analysed — either the commit is no longer at HEAD, or the PR
+// is no longer open.
 //
-// Why a re-check is needed at all: every dedup layer upstream decides whether
-// to *start* a review, and none of them can speak for what happens while the
-// CLI runs. That window is long — the review_timing stats carry a
-// bucket_very_slow tail out past 3000s — so an author who force-pushes inside
-// it gets a verdict describing code that is no longer there. Observed on
+// Why a re-check is needed at all: every dedup layer upstream decides whether to
+// *start* a review, and none of them can speak for what happens while the CLI
+// runs. That window is long — the review_timing stats carry a bucket_very_slow
+// tail out past 3000s — so an author who force-pushes, merges or closes inside it
+// gets a verdict describing code that is no longer there. Observed on
 // freepik-company/ai-bumblebee-proxy#1802: a REQUEST_CHANGES landed on a commit
 // that had already been approved and then superseded by two further pushes
 // (theburrowhub/heimdallm#664). At `high` severity that also blocks merge on
 // outdated code.
 //
-// The deferred publish path in cmd/heimdallm already does exactly this
-// (GetPRSnapshot + pendingReviewInvalidReason immediately before SubmitReview);
-// only the hot path was missing it, so the two publish sites disagreed. This is
-// the hot-path half, and it reuses the same SupersededReviewID sentinel so both
-// retire a stale row identically.
+// This mirrors pendingReviewInvalidReason on the deferred publish path, which
+// checks both conditions for one API call. Checking only the SHA would have left
+// a PR merged mid-run still receiving its verdict.
 //
 // Marking the row is load-bearing, not bookkeeping: step 7 stored it with
 // GitHubReviewID == 0, which is exactly what ListUnpublishedReviews selects on.
 // Skipping the publish without retiring the row would simply hand the stale
-// verdict to the publish-worker to post later.
+// verdict to the publish-worker to post later. The sentinel differs by reason,
+// matching the deferred path: SupersededReviewID for a moved HEAD (which Run
+// then treats as "no previous review", so the new commit can be reviewed), and
+// orphanedReviewID for a closed PR, which is terminal.
 //
 // Returns (true, …) when the caller must stop. Run then returns (nil, nil),
 // matching the other skip paths so main.go's `rev == nil` filter keeps the
 // review_completed SSE, the "review done" log line and the activity_log row
 // suppressed for a review that was never published (#322 Bug 4).
-func (p *Pipeline) retireIfHeadMoved(rev *store.Review, pr *github.PullRequest) (bool, error) {
-	if pr.Head.SHA == "" {
-		// Nothing to compare against. The empty-SHA/legacy-row handling
-		// upstream owns this case; re-deciding it here would duplicate it.
-		return false, nil
-	}
-	liveSHA, err := p.gh.GetPRHeadSHA(pr.Repo, pr.Number)
+func (p *Pipeline) retireIfPRMoved(rev *store.Review, pr *github.PullRequest) (bool, error) {
+	live, err := p.resolveLivePR(pr)
 	if err != nil {
-		// Deliberately fail-open. Deferring to the publish-worker instead
-		// would be safer in isolation but would delay every publish whenever
-		// the API has a blip, and needs a lifecycle state that does not exist
-		// yet. Publishing stale requires HEAD to have moved AND this call to
-		// fail in the same run; leaving the guard to the confirmable case
-		// keeps this change to a single behavioural difference. The deferred
-		// path stays hardened regardless, because it re-validates the SHA
-		// under its own claim before submitting.
-		slog.Warn("pipeline: could not re-resolve HEAD before publishing, publishing anyway",
+		// Deliberately fail-open. Deferring to the publish-worker instead would
+		// be safer in isolation but would delay every publish whenever the API
+		// has a blip, and needs a lifecycle state that does not exist yet.
+		// Publishing stale requires the PR to have moved AND this lookup to fail
+		// in the same run. The deferred path stays hardened regardless, because
+		// it re-validates under its own claim before submitting.
+		slog.Warn("pipeline: could not re-resolve PR before publishing, publishing anyway",
 			"repo", pr.Repo, "pr", pr.Number, "review_id", rev.ID,
 			"head_sha", pr.Head.SHA, "err", err)
 		return false, nil
 	}
-	if liveSHA == "" || liveSHA == pr.Head.SHA {
+
+	reason := SkipReasonNone
+	terminalReviewID := SupersededReviewID
+	switch {
+	case live.stateKnown && live.state != "open":
+		reason, terminalReviewID = SkipReasonNotOpen, orphanedReviewID
+	case live.headSHA != "" && live.headSHA != pr.Head.SHA:
+		reason, terminalReviewID = SkipReasonHeadChanged, SupersededReviewID
+	}
+	if reason == SkipReasonNone {
 		return false, nil
 	}
-	if err := p.store.MarkReviewPublished(rev.ID, SupersededReviewID, "", time.Now().UTC()); err != nil {
+
+	if err := p.store.MarkReviewPublished(rev.ID, terminalReviewID, "", time.Now().UTC()); err != nil {
 		// Report rather than swallow, but still stop: an unretired row keeps
-		// GitHubReviewID == 0, so the publish-worker picks it up and applies
-		// its own SHA re-validation. The stale verdict cannot reach GitHub
-		// either way — only the bookkeeping is left inconsistent.
+		// GitHubReviewID == 0, so the publish-worker picks it up and applies its
+		// own re-validation. The stale verdict cannot reach GitHub either way —
+		// only the bookkeeping is left inconsistent.
 		return true, fmt.Errorf("pipeline: retire superseded review %d: %w", rev.ID, err)
 	}
-	rev.GitHubReviewID = SupersededReviewID
-	slog.Info("pipeline: HEAD moved during review — retiring instead of publishing",
-		"repo", pr.Repo, "pr", pr.Number, "review_id", rev.ID,
-		"reviewed_head_sha", pr.Head.SHA, "current_head_sha", liveSHA,
-		"event", rev.Event, "severity", rev.Severity)
-	p.publishSkipped(pr, SkipReasonHeadChanged)
+	slog.Info("pipeline: PR moved during review — retiring instead of publishing",
+		"repo", pr.Repo, "pr", pr.Number, "review_id", rev.ID, "reason", string(reason),
+		"reviewed_head_sha", pr.Head.SHA, "current_head_sha", live.headSHA,
+		"current_state", live.state, "event", rev.Event, "severity", rev.Severity)
+	p.publishSkipped(pr, reason)
 	return true, nil
 }
+
+// livePR is the freshness view the publish guard needs.
+type livePR struct {
+	state   string
+	headSHA string
+	// stateKnown distinguishes "the PR is open" from "state was never fetched".
+	// Without it the HEAD-only fallback would report an empty state and the
+	// caller would read that as a closed PR, retiring every review.
+	stateKnown bool
+}
+
+// resolveLivePR fetches the PR's current state, preferring the full snapshot and
+// falling back to a HEAD-only lookup when the gh dependency does not provide one.
+//
+// Both paths retry once after a short pause, matching the hydration lookup at the
+// top of Run: #243's failure mode was rate-limit 429s, and a back-to-back retry
+// is useless against those because the window is still active. A 429 during a
+// long review is the likeliest way this call fails, so the retry meaningfully
+// shrinks the fail-open window.
+func (p *Pipeline) resolveLivePR(pr *github.PullRequest) (livePR, error) {
+	if snapshotter, ok := p.gh.(PRSnapshotFetcher); ok {
+		snap, err := snapshotter.GetPRSnapshot(pr.Repo, pr.Number)
+		if err != nil {
+			time.Sleep(headResolveRetryDelay)
+			snap, err = snapshotter.GetPRSnapshot(pr.Repo, pr.Number)
+		}
+		if err != nil {
+			return livePR{}, err
+		}
+		if snap == nil {
+			return livePR{}, fmt.Errorf("pipeline: nil PR snapshot")
+		}
+		return livePR{state: snap.State, headSHA: snap.HeadSHA, stateKnown: true}, nil
+	}
+
+	sha, err := p.gh.GetPRHeadSHA(pr.Repo, pr.Number)
+	if err != nil {
+		time.Sleep(headResolveRetryDelay)
+		sha, err = p.gh.GetPRHeadSHA(pr.Repo, pr.Number)
+	}
+	if err != nil {
+		return livePR{}, err
+	}
+	if sha == "" {
+		// Empty-but-nil-error is a lookup failure, not "unchanged" — the same
+		// reading the hydration path at the top of Run applies to this value.
+		// Folding it into the error branch keeps the log honest about which case
+		// occurred instead of silently claiming the HEAD was confirmed.
+		return livePR{}, fmt.Errorf("pipeline: resolve HEAD SHA: empty SHA returned")
+	}
+	return livePR{headSHA: sha}, nil
+}
+
+// headResolveRetryDelay is the pause before the single retry of a HEAD/snapshot
+// lookup. Same value and rationale as the hydration lookup in Run.
+const headResolveRetryDelay = 500 * time.Millisecond
 
 // publish emits an SSE lifecycle event with the given payload. No-op
 // when no publisher is wired. A marshal failure on a map[string]any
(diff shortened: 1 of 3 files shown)


Muriano (approved):
## 🤖 Heimdallm AI Review

Solid, well-scoped fix. `retireIfHeadMoved` is correctly placed *before* the multi-mode per-issue comment loop (not just before `SubmitReview`), so a superseded run posts nothing at all — pinned by `TestPipeline_Run_SkipsPerIssueCommentsWhenHeadMoved`. Verified against the repo: (a) the retired row is written with the existing `SupersededReviewID` (-2) sentinel, which `ListUnpublishedReviews` no longer selects (`github_review_id != 0`), so the publish path cannot post the stale verdict later; (b) `SkipReasonHeadChanged` already exists in guards.go and is the same reason the deferred worker emits, so the two publish sites now agree; (c) the fail-open branches (`liveSHA == ""`, API error) are safe against every existing fake in the package — `fakeGH`, `fakeGHCounter`, `fakeGHCommentsError` and `fakeGHTransientSubmit` all return `""` from `GetPRHeadSHA`, and `fakeGHReloop`/`fakeGHBreaker`/`fakeGHLockedSubmit` return the same SHA that hydration installed, so only `TestPipeline_Run_HydratesHeadSHAWhenMissing` needed its call-count updated (correctly done, 1 → 2, with both calls justified). Test coverage is good for a fix this size: retire, multi-mode, unchanged-HEAD control, re-check failure (fail-open), and a no-extra-call test for early-skip runs. Two non-blocking observations for the record, neither of which meets the reporting bar. First, the retirement writes the sentinel that `Run` deliberately treats as "no previous review" (pipeline.go:538), so after a mid-review force-push the next automatic cycle will re-review the new HEAD *without* an explicit re-request — a behavioural change relative to today (publish stale, then require a re-request). That is almost certainly the intent, and repeated force-pushes remain bounded by the per-repo circuit breaker at pipeline.go:719, but it is worth stating explicitly in the PR body since it widens the automatic-review surface, not just the skip surface. Second, the doc comment says an unretired row "is picked up by the publish-worker which applies its own SHA re-validation" — true for the production path (the NATS worker in cmd/heimdallm/main.go:1183-1218 re-fetches the snapshot and calls `pendingReviewInvalidReason`), but the in-package `Pipeline.PublishPending` (pipeline.go:944) has no such re-validation. It is not wired in production (`tier2Adapter.publishPending` only enqueues to NATS), so this is a latent trap rather than a defect; narrowing the comment to name the NATS worker would keep it accurate if that legacy loop is ever re-activated. Nits: the doc comment on `retireIfHeadMoved` is unusually long even for this codebase's style, and the empty-`liveSHA` case is treated fail-open here while the hydration path at pipeline.go:518 treats the same value fail-closed — the asymmetry is deliberate and explained, but a one-line cross-reference between the two sites would help the next reader.

---

vbuenog (commented):
## 🤖 Heimdallm AI Review

Correct, well-tested fix for the core problem: the hot publish path now re-resolves HEAD before writing anything to GitHub, and the guard is deliberately placed ahead of the multi-mode comment loop (the tests pin that placement). I verified the interfaces compile (p.gh already embeds HeadSHAResolver), the retire sentinel is handled downstream (SupersededReviewID makes the next Run treat the row as absent, so the new HEAD does get re-reviewed while the circuit breaker still bounds cost), and that no existing pipeline test fake breaks on the extra GetPRHeadSHA call — only the one test the PR updates. Two gaps remain against the PR's stated goal of making the two publish sites agree: the hot path still submits unpinned (leaving a real, if narrow, TOCTOU window that an existing helper already closes), and it only checks the SHA, not the PR state.

### Issues

⚠️ **daemon/internal/pipeline/pipeline.go:902** — The re-check narrows the race but does not close it: after retireIfHeadMoved returns, the multi-mode comment loop and SubmitReview still run, and SubmitReview is the unpinned variant, so GitHub attaches the review to whatever HEAD is current at submit time. github.Client already has SubmitReviewForCommit (client.go:604) written for exactly this failure — 'GitHub otherwise defaults to the pull request's current HEAD, which lets a concurrent push misattribute findings from an older commit' — and the deferred path uses it (main.go:1229). So the two publish sites still disagree on the load-bearing half: the deferred one is race-free by construction, the hot one is check-then-act. Anchoring commit_id here (via an optional interface assertion on p.gh, keeping the GitHubReviewer contract intact for the test doubles) would make the guard's failure mode 'review shows as outdated' instead of 'review misattributed to new code'.
⚠️ **daemon/internal/pipeline/pipeline.go:246** — Parity gap with the path this change cites as its model. pendingReviewInvalidReason (main.go:5484) retires a pending review on TWO conditions: snapshot.State != "open" (SkipReasonNotOpen) and a HEAD mismatch. retireIfHeadMoved only implements the second, so a PR merged or closed during the CLI run — a window the PR itself documents as reaching past 3000s — still gets its REQUEST_CHANGES/APPROVE posted. Calling GetPRSnapshot instead of GetPRHeadSHA covers both for the same one API call and the same one-extra-call cost the control test pins.
🟡 **daemon/internal/pipeline/pipeline.go:250** — The fail-open error path has no retry, and treats an empty-but-nil SHA as 'unchanged'. Every other SHA resolution in the codebase is harder: the hydration lookup 400 lines up retries once after a 500ms sleep specifically because #243's failure mode was 429s, and treats an empty SHA as a lookup failure; the deferred path hard-errors on `rev.HeadSHA != "" && snapshot.HeadSHA == ""` rather than proceeding. Since a 429 during a review is the most likely way this call fails, reusing the same single short retry would shrink the fail-open window at near-zero cost, and folding `liveSHA == ""` into the failure branch (log + publish) rather than the 'confirmed unchanged' branch would keep the log honest about which case occurred.
🟡 **daemon/internal/pipeline/pipeline.go:241** — Two lines in the new function cannot affect behaviour, and their comments imply otherwise. (a) `pr.Head.SHA == ""` is unreachable at step 8: Run either receives a non-empty SHA or hydrates it, failing closed on both error and empty — so the 'upstream owns this case' comment is describing why the branch is dead, not why it's needed. (b) `rev.GitHubReviewID = SupersededReviewID` mirrors state onto a struct the caller never receives, since this path returns (nil, err). Harmless, but both read as load-bearing.
🟡 **daemon/internal/pipeline/pipeline.go:262** — Cost consequence worth stating in the commit message, since the PR presents the change as a single behavioural difference. Retiring with SupersededReviewID makes the next Run set prevReview = nil (pipeline.go:538), which bypasses the #509 re-request gate — correct, the outstanding request deserves an answer — but it means each force-push during a review now costs one discarded CLI run plus one full re-review. Because PerPR24h counts per HEAD SHA (circuitbreaker.go:73), every new commit gets a fresh allowance, so only PerRepoHr bounds a push storm. The tradeoff is the right one versus publishing stale verdicts; it just isn't free, and the pre-existing amplification is now easier to hit.

---

---
_Not approving: this review raised 5 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

sergiotejon:
@vbuenog is right and the PR was overclaiming: it said the two publish sites now agree, and on the load-bearing half they did not. Fixed in 8178df8 — thanks, this was the most useful finding across the three PRs.

**Unanchored submit.** `SubmitReviewForCommit` (client.go:609) exists for exactly this failure and the deferred path already used it (main.go:1229); the hot path did not. The guard was check-then-act, so a push landing between the re-check and the submit still had GitHub attach the review to the new HEAD. Now anchored, so the worst case degrades from "findings misattributed to code they were not written about" to "review shows as outdated".

**State parity gap.** Also correct — `pendingReviewInvalidReason` retires on two conditions and I implemented one, so a PR merged or closed inside a window I documented as reaching past 3000s still got its verdict posted. Now covered by preferring `GetPRSnapshot`, at the same one-API-call cost the control test pins. A closed PR takes the terminal orphan sentinel rather than `Superseded`, since `Superseded` means "re-review the new commit", which would be wrong there.

Both are optional interfaces discovered by type assertion, per your suggestion — `*github.Client` implements them, the test doubles don't, and widening the mandatory `gh` set would force a method onto every double for a path most tests never reach.

**Retry + empty SHA.** Both applied: one retry after 500ms matching the hydration lookup (a 429 mid-review being the likeliest failure), and `liveSHA == ""` folded into the failure branch rather than the "confirmed unchanged" branch, so the log no longer claims a confirmation it didn't get.

**Dead code.** Both correct. Verified the hydration block fails closed on error *and* on empty and assigns `pr.Head.SHA`, so the `== ""` branch was unreachable at step 8 and its comment explained why it was dead rather than why it was needed. Removed, along with the `rev.GitHubReviewID` mirror onto a struct the caller never receives on that path.

**Cost consequence.** Now in the commit message, as both you and @Muriano asked: retiring with `SupersededReviewID` bypasses the #509 re-request gate, so a force-push mid-review costs one discarded run plus one full re-review, and since `PerPR24h` counts per HEAD SHA only `PerRepoHr` bounds a push storm.

One new safeguard the extension needed: a `stateKnown` flag separating "PR is open" from "state was never fetched". Without it the HEAD-only fallback reports an empty state, the guard reads that as closed, and *every* review on a dependency lacking `GetPRSnapshot` gets retired. Pinned by a test.

@Muriano — on your note about `Pipeline.PublishPending` (pipeline.go:944) having no re-validation: correct, and I've left it alone since it isn't wired in production, but it does mean my doc comment was too broad. Worth a follow-up issue rather than widening this PR.

Four new tests: closed-PR retirement, commit anchoring, unanchored fallback, and the unknown-state case. `make test-docker` green across all 17 daemon packages.

Since this was already approved and the change is materially larger than what you approved, I'd rather it get another look than merge on the old approval.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of the 8178df8 follow-up. All four previously reported findings are addressed and verified against the surrounding code, not just the diff:

1. **Commit anchoring** — the hot path now prefers `SubmitReviewForCommit`, matching the deferred publish path (main.go:1229). Verified `github.Client.SubmitReviewForCommit` (client.go:609) rejects an empty commit_id, and the pipeline guards `pr.Head.SHA != ""` before taking that branch, so the fallback is reached rather than an error. `TestPipeline_Run_AnchorsPublishToAnalysedCommit` pins the commit_id and `TestPipeline_Run_FallsBackToUnanchoredSubmit` pins the degraded path.
2. **State parity** — `retireIfPRMoved` now checks both conditions, mirroring `pendingReviewInvalidReason` (main.go:5483). Checked the string comparison specifically: `GetPRSnapshot` reads the REST Pulls API with `application/vnd.github+json` and copies `pr.State` verbatim, which is lowercase `open`/`closed` (merged PRs normalise to `closed`, per the doc comment at client.go:957), so `live.state != "open"` cannot misfire the way an uppercase GraphQL enum would. The sentinel split (orphan for closed, Superseded for a moved HEAD) is correct and matches the deferred path's own split.
3. **Optional-capability wiring actually binds in production** — `pipeline.New(s, ghClient, …)` at main.go:283 passes `gh.NewClient(token)` (`*github.Client`) directly, with no adapter in between, so both type assertions succeed in the daemon and the snapshot/anchored paths are the real production paths, not dead code exercised only by tests.
4. **Retry + empty SHA + dead code** — one retry after 500ms on both resolution paths; the empty-SHA case folded into the failure branch; the unreachable `== ""` branch and the `rev.GitHubReviewID` mirror removed.

Guard placement ahead of the multi-mode comment loop is right and is pinned by `TestPipeline_Run_SkipsPerIssueCommentsWhenHeadMoved`. `stateKnown` is a genuinely necessary addition and is tested. The API-cost story is honest: the control test pins exactly one extra call per published review, and the snapshot test proves the state condition is free.

Non-blocking observations, none of which clear the issue bar:

- **Snapshot path does not log an unresolved HEAD.** When `GetPRSnapshot` returns 200 with an empty `HeadSHA`, `resolveLivePR` returns `stateKnown: true, headSHA: ""` and the SHA comparison is silently skipped. The *outcome* is identical to the HEAD-only path (both fail open and publish), so this is not a behavioural gap — but the fallback emits a warn line in that case and the snapshot path emits nothing, so an operator debugging a stale publish would see no trace. A one-line `slog.Warn` when `snap.HeadSHA == ""` would close the asymmetry. Note the deferred path chose the opposite direction here (main.go:1219 errors so the row retries rather than being orphaned) — worth a sentence in the doc comment explaining why the hot path deliberately diverges, since the comment currently claims to mirror it.
- **Anchored submit adds a new 422 shape.** In the residual check-then-act window, a force-push landing between the re-check and the submit can leave `commit_id` no longer part of the PR, which GitHub answers with 422. Whether `classifyPermanentSubmit422` treats that as permanent (row orphaned terminally, no re-review of the old verdict) or transient is worth a glance. The window is microseconds-wide and the deferred path already carries the same shape, so this is a latent note, not a defect in this diff.
- **Test wall-clock.** `TestPipeline_Run_PublishesWhenHeadRecheckFails` pays a real 500ms `time.Sleep` for the retry. One test is fine; if more fail-open cases get added, the delay is worth making injectable rather than a package const.
- `Pipeline.PublishPending` (pipeline.go:944) still has no re-validation. Agreed it is out of scope while unwired — the follow-up issue should be filed rather than left in the discussion thread, and the `retireIfPRMoved` doc comment's "the deferred path stays hardened regardless" is accurate only for the main.go worker, not for `PublishPending`.

The change is materially larger than the version that was approved, and it holds up. Nothing here blocks merge.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Both previous MEDIUM findings are genuinely fixed: the hot path now prefers GetPRSnapshot (state + HEAD in the same one call, closed PRs take the terminal -1 sentinel, moved HEAD takes -2) and anchors the submit via SubmitReviewForCommit, matching the deferred path at main.go:1183-1235. The dead branches are gone, the retry and empty-SHA handling were applied to the fallback path, and the stateKnown flag correctly prevents the HEAD-only fallback from reading 'never fetched' as 'closed' — a real trap, well caught and pinned by a test. I verified the production wiring (main.go:283 passes *github.Client, which does implement both new interfaces with matching signatures) and that the anchored-submit failure mode is safe: an unclassified 422 leaves github_review_id == 0, and production's PublishPending is the tier2 adapter that enqueues to the hardened NATS worker, not the unvalidated Pipeline.PublishPending. Remaining items are all low: the empty-SHA hardening was only applied to one of the two branches, the optional interfaces have no compile-time assertion, the retry rationale doesn't hold against the client's own rate-limit cooldown, and the closed-PR retirement is unrecoverable if the PR is reopened.

### Issues

🟡 **daemon/internal/pipeline/pipeline.go:343** — The empty-SHA hardening was applied to only one of the two branches. The HEAD-only fallback now folds `sha == ""` into the error branch (correct, and matches the hydration lookup), but the snapshot branch returns `livePR{headSHA: snap.HeadSHA, ...}` unconditionally — so a snapshot that comes back with an empty HeadSHA passes the `live.headSHA != ""` short-circuit in retireIfPRMoved and publishes as if freshness were confirmed, with no log recording that the SHA was never resolved. The deferred path this change models explicitly refuses that case (`if rev.HeadSHA != "" && snapshot.HeadSHA == "" { return fmt.Errorf(...) }`, main.go:1219). Since *github.Client is the only production implementer and it takes the snapshot branch, the branch that got hardened is the one production never uses. Treating an empty snapshot HeadSHA the same way as the fallback (error -> documented fail-open with an honest log) would make the two branches agree.
🟡 **daemon/internal/pipeline/pipeline.go:58** — Both new capabilities are discovered by type assertion and degrade silently when the assertion misses, but nothing in the package binds *github.Client to them. I confirmed the signatures currently match and that main.go:283 passes the concrete client, so the feature works today — but every new test drives a hand-written fake, so if GetPRSnapshot or SubmitReviewForCommit ever gains a ctx parameter, or the pipeline is wired through a wrapper the way tier 2/3 wrap ghClient elsewhere in main.go, the closed-PR guard reverts to HEAD-only and the anchoring silently disappears with no test failing and no log line. A one-line `var _ PRSnapshotFetcher = (*github.Client)(nil)` / `var _ CommitAnchoredReviewer = (*github.Client)(nil)` in the package turns that from a silent runtime downgrade into a compile error.
🟡 **daemon/internal/pipeline/pipeline.go:326** — The retry's stated rationale doesn't hold for the failure mode it names. The comment says a 429 mid-review is the likeliest way this call fails and that the 500ms retry 'meaningfully shrinks the fail-open window' — but on a 429 the client arms its own cooldown (rlPausedUntil, client.go:381-396) and every subsequent GET returns RateLimitError without touching the network until the window expires, which is orders of magnitude longer than 500ms. So against 429s the retry adds 500ms of latency and cannot succeed; it only helps transient 5xx/network blips. The retry is still worth keeping for those, but the comment overstates what it buys (the hydration lookup it mirrors has the same property, so this is pre-existing reasoning being propagated rather than newly introduced).
🟡 **daemon/internal/pipeline/pipeline.go:289** — Closed-PR retirement is unrecoverable for the close-then-reopen case. Retiring with orphanedReviewID (-1) leaves a stored row at the analysed HeadSHA, and the dedup gate (pipeline.go:607) skips any later run where prevReview is non-nil and the SHA matches, unless a review_requested bypass fires — and there is no reopen handling anywhere in the daemon. So a PR closed mid-run and reopened at the same commit gets no review at all: the verdict the CLI already paid for is discarded and no replacement is ever produced. For a merged PR the terminal sentinel is clearly right, and choosing -1 over Superseded for 'closed' is well argued; the gap is only that 'closed' conflates permanent and reversible. Worth either narrowing the sentinel to merged-only, or accepting it explicitly with a note (and ideally a test) that reopen requires a manual re-request.

---

PR #45, opened 2026-04-17
feat(issues): auto_implement pipeline — branch, commit, PR

ivanmunozruiz:
## Summary

Fourth step of the fase-2 issue-tracking pipeline (#27). Implements the \`auto_implement\` mode end-to-end: for an issue classified as \`develop\` **with** a \`local_dir\` configured, the agent edits the working tree in place, the daemon commits + pushes the branch, and opens a PR linked back to the issue via \`Closes #N\`. Integration with the poll cycle and HTTP endpoints is deferred to #28 — **main.go is untouched here**.

### GitHub client (\`internal/github/repos.go\`)

- \`GetDefaultBranch(repo)\` via \`GET /repos/{repo}\` — we base the work branch on whatever trunk the repo actually uses, not a guessed \`main\`.
- \`CreatePR(repo, title, body, head, base)\` via \`POST /repos/{repo}/pulls\` returning the PR number so the \`IssueReview\` row can record \`pr_created\`.
- Both validate inputs and surface wrapped errors with status codes and scrubbed bodies.
- Skipped the \`CreateBranch\` listed in the original issue body: \`git checkout -B\` + \`git push\` materialises the branch on the remote without a second round-trip; a method with no caller would be dead code.

### Git ops (\`internal/issues/gitops.go\`)

New \`GitOps\` interface + a default \`GitExec\` that shells out to the \`git\` binary (3-minute per-command timeout). Four operations:

- \`CheckoutNewBranch(dir, branch, base)\` — \`git fetch origin <base>\` + \`git checkout -B <branch> origin/<base>\`. \`-B\` is intentional: on a re-run for the same issue we want the branch reset to the current base, not picking up stale state.
- \`HasChanges(dir)\` — \`git status --porcelain\`, tracks both modified and untracked files because the agent may create new files.
- \`CommitAll(dir, msg)\` — \`git add -A\` + commit with the Heimdallm identity set per-command via \`-c user.name=…\`. The repo-level or global git config is never mutated.
- \`Push(dir, repo, branch, token)\` — HTTPS push with the token in the URL userinfo. The token is never persisted to git config. Any stderr echo from git is scrubbed (\`ReplaceAll(token, \"***\")\`) before the error bubbles up.

### Pipeline (\`internal/issues/pipeline.go\`)

[...]

PR #667, opened 2026-08-05
fix(daemon): kill git's process group on timeout and reap orphans in PID 1

### LINKED ISSUE #665: git subprocess timeout only kills the parent, leaking orphaned ssh children as zombies
## Symptom

The running container (`ghcr.io/theburrowhub/heimdallm:0.8.0`, up 23h) accumulates zombie processes under PID 1:

```
docker exec heimdallm ps aux
    1 heimdall  2:09 /usr/local/bin/heimdallm
 5183 heimdall  0:00 [git]     <- zombie
 5213 heimdall  0:00 [git]     <- zombie
 9536 heimdall  0:00 [ssh]     <- zombie
 9647 heimdall  0:00 [ssh]     <- zombie
 ... (13 ssh + 2 git zombies total)
```

15 zombies accumulated over ~23 hours of uptime. Not an immediate resource problem (a zombie is just a process-table entry), but it's an unbounded leak that will eventually hit the container's PID limit on a long-lived instance, and it means the daemon is losing track of subprocess lifecycles it spawns for repo access.

## Root cause

`daemon/internal/repoctx/manager.go`, `execGit.Run` (introduced in `302ea3f1`, 2026-05-07):

```go
func (execGit) Run(ctx context.Context, dir string, env []string, args ...string) error {
	runCtx, cancel := context.WithTimeout(ctx, gitTimeout)
	defer cancel()
	cmd := exec.CommandContext(runCtx, "git", args...)
	cmd.Dir = dir
	cmd.Env = env  // or os.Environ()
	...
	if err := cmd.Run(); err != nil { ... }
```

`exec.CommandContext` only sends the kill signal to the `git` process itself when the context expires — it does not touch anything `git` may have forked (notably `ssh`, when a repo is accessed over SSH rather than HTTPS). If a `git fetch`/`clone` times out while it has an `ssh` child running:

1. `git` gets killed.
2. Its `ssh` child is orphaned and reparented to PID 1 (the heimdallm binary itself — there's no `tini`/init process in this image).
3. heimdallm never calls `wait()` on processes it didn't directly spawn via its own `exec.Cmd`, so when that orphaned `ssh` eventually exits, it becomes a permanent zombie.

The `git`-only zombies (2 of them, vs. 13 `ssh`) suggest a smaller number of cases where `git` itself was killed before `cmd.Run()`'s internal `Wait()` completed cleanup — worth double-checking exit-code handling here too, but the dominant pattern is the orphaned-ssh-child case.

## Suggested fix

Put the `git` invocation in its own process group and kill the whole group on cancellation, not just the single PID:

```go
cmd := exec.CommandContext(runCtx, "git", args...)
cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
...
context.AfterFunc(runCtx, func() {
    if cmd.Process != nil {
        syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL) // negative pid = process group
    }
})
```

(`exec.CommandContext`'s default `Cancel`/`WaitDelay` hooks in newer Go versions can also be used instead of a manual `AfterFunc`, whichever fits the current Go version in go.mod.)

## Repro data
- Container: `heimdallm` (`ghcr.io/theburrowhub/heimdallm:0.8.0`), uptime 23h at time of observation
- Zombie count: 13 `[ssh]` + 2 `[git]`, all direct children of PID 1
- Introducing commit: `302ea3f1` (2026-05-07), `daemon/internal/repoctx/manager.go` `execGit.Run`

sergiotejon:
Fixes #665.

## Two causes, both required

The issue diagnosed cause (1). Investigating it turned up a second one, and **the suggested fix alone would not have removed the reported symptom**:

1. `exec.CommandContext` signals only the process it started. `git` forks `ssh` for SSH remotes, so a `gitTimeout` killed `git` and left `ssh` running, reparented onto PID 1.
2. **Nothing reaps it there.** The kernel does not auto-reap on PID 1's behalf, and `ENTRYPOINT` was the daemon binary with no init. Every orphan that exited stayed a zombie forever.

Killing the process group makes the orphan *die*; it is still reparented onto PID 1 and still needs a reaper. So (1) without (2) leaves the zombies. (2) without (1) hides the leak while the daemon keeps losing track of subprocesses. Both are here.

### Empirical confirmation

PID 1 = a process that does not reap adopted children (like the Go daemon):

```
# no init                          # with init
PID  PPID STAT COMMAND             PID  PPID STAT COMMAND
  1     0 S    node                  1     0 S    tini
 15     1 Z    sleep   <- zombie      7     1 S    node
```

Same shape as the report (`[ssh]`/`[git]` with `PPID 1`). Also verified with the exact `ENTRYPOINT ["/sbin/tini", "--", …]` form used here, and that tini forwards SIGTERM so graceful shutdown still works (`docker stop` → handler runs → exit 0).

## Scope: a third call site the issue missed

The issue named only `repoctx/manager.go` (`302ea3f1`). `issues/gitops.go` `captureGit` has the identical pattern and is **20 days older** (`1186af4e`, #45). It is the shared helper behind `CheckoutNewBranch`, `Push`, `DeleteRemoteBranch`, `HasChanges`, `CommitAll`, `Diff` — the push/fetch ones fork `ssh` just the same. Fixing only `repoctx` would leave the auto_implement pipeline leaking. Both are routed through the new helper.

## Design notes for review

**`internal/procgroup` wraps `Start`+`Wait` instead of offering a "configure this cmd" helper.** The escalation goroutine must be told when the child was reaped; a caller who forgot would eventually signal a pgid the kernel had recycled onto an unrelated group. Owning the lifecycle makes that unforgettable.

**One precondition:** the cmd must come from `exec.CommandContext`. `os/exec` refuses to start a command with a `Cancel` hook and no context, so misuse fails immediately with an explicit error. Semantically correct too — with no context there is nothing to cancel. Pinned by `TestRun_RejectsCommandWithoutContext`.

**`internal/executor` was left alone.** It already has a richer variant of this logic (group tracking for `TerminateAll`, tolerating `ErrWaitDelay` when the payload is complete) added in #656 for #614. Folding those needs into a shared helper would turn this PR into a refactor of the most delicate path in the daemon. Flagging the duplication deliberately — converging them is follow-up for #614, which this PR should probably be linked to since it is the same bug class.

**`ErrWaitDelay` is not special-cased for git.** The executor returns collected output when the child exited 0 but a descendant held the pipes. git output is parsed (`status --porcelain`, `diff --name-only -z`), so silently returning a truncated read could produce wrong decisions. It propagates as an error instead.

## Tests

New `daemon/internal/procgroup/procgroup_test.go`, 7 tests. The regression test was **verified to fail without the fix** (sabotaged `Setpgid` + parent-only kill → fails in 10.3s; with the fix, passes in 0.30s).

One thing worth knowing: the first version of that test used `kill(pid, 0)` for liveness and was **wrong in both directions** — that call also succeeds for a zombie, which is exactly the state under test. It now uses an inherited pipe and waits for EOF, which distinguishes "running" from "exited but not reaped" and is portable to the macOS CI runners (no `/proc`).

`make test-docker` green across all 18 daemon packages (vet included); `make test-compose-isolation` green.

## Not covered here

The `--init`/tini change only affects the container images. Host installs (`install-linux`) run under systemd, which already reaps. The process-group fix helps both.

### DIFF
--- a/daemon/internal/issues/gitops.go
+++ b/daemon/internal/issues/gitops.go
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/heimdallm/daemon/internal/procgroup"
 )
 
 // gitTimeout caps each `git` invocation so a hung network or huge fetch
@@ -382,7 +384,12 @@ func captureGit(ctx context.Context, dir string, env []string, args ...string) (
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
-	if err := cmd.Run(); err != nil {
+	// procgroup.Run rather than cmd.Run: this helper backs every git call in
+	// GitExec, including CheckoutNewBranch / Push / DeleteRemoteBranch, which
+	// fork `ssh` for SSH remotes. exec.CommandContext's cancellation reaches
+	// only git, leaving that ssh child orphaned onto PID 1 as a zombie
+	// (theburrowhub/heimdallm#665).
+	if err := procgroup.Run(cmd); err != nil {
 		// Cap stderr to protect against pathological output (e.g. huge
 		// merge-conflict reports, repeated progress lines, etc).
 		errText := stderr.String()
--- a/daemon/internal/procgroup/procgroup.go
+++ b/daemon/internal/procgroup/procgroup.go
@@ -1,0 +1,130 @@
+// Package procgroup runs an *exec.Cmd in its own process group so that
+// cancelling the command's context terminates the whole process tree rather
+// than just the direct child.
+//
+// The problem it solves: exec.CommandContext's default cancellation kills only
+// the process it started. Anything that process forked survives, gets reparented
+// to PID 1, and — because nothing calls wait() on a process it did not spawn —
+// becomes a permanent zombie once it exits. For `git` over SSH the forked child
+// is `ssh`, which is how a container accumulated 13 [ssh] + 2 [git] zombies in
+// 23h of uptime (theburrowhub/heimdallm#665). The same shape was already
+// diagnosed for the AI CLI launchers in #614.
+//
+// Killing the group is only half of the fix. A descendant that is already
+// orphaned still needs PID 1 to reap it, and the kernel does not do that
+// automatically for PID 1 — hence the init process in docker/Dockerfile. This
+// package stops the daemon from leaking live processes; the init stops the dead
+// ones from piling up as zombies.
+//
+// Note for future readers: internal/executor carries a richer variant of this
+// logic (it also registers groups so TerminateAll can reach agents at shutdown,
+// and tolerates ErrWaitDelay when the payload is already complete). That was
+// deliberately left in place rather than folded in here — see #614.
+package procgroup
+
+import (
+	"errors"
+	"log/slog"
+	"os"
+	"os/exec"
+	"sync/atomic"
+	"syscall"
+	"time"
+)
+
+const (
+	// termGrace is how long the group gets to exit after SIGTERM before the
+	// escalation to SIGKILL. Short on purpose: cancellation here means a
+	// timeout has already elapsed, so the caller is past waiting politely.
+	termGrace = 2 * time.Second
+	// waitDelay bounds Wait once the direct child has exited but a descendant
+	// still holds the inherited stdout/stderr pipes. Without it, Wait blocks
+	// until every copy of those descriptors is closed, which lets a stalled
+	// grandchild hang the caller indefinitely.
+	waitDelay = 2 * time.Second
+)
+
+// Run starts cmd in its own process group, waits for it, and on context
+// cancellation signals the entire group (SIGTERM, then SIGKILL after
+// termGrace).
+//
+// cmd MUST have been created with exec.CommandContext. Run installs a
+// cmd.Cancel hook, and os/exec refuses to start a command that has one without
+// an associated context — so misuse fails immediately at Start with an explicit
+// error rather than silently losing the group kill. The requirement is
+// semantic, not incidental: with no context there is nothing to cancel and
+// nothing for this package to do.
+//
+// Run replaces cmd.Run and takes ownership of cmd.SysProcAttr.Setpgid,
+// cmd.Cancel and cmd.WaitDelay. It deliberately wraps Start+Wait instead of
+// exposing a "configure this cmd" helper: the escalation goroutine must be told
+// when the child has been reaped, and a caller that forgot to do so would
+// eventually signal a pgid the kernel had already recycled onto an unrelated
+// process group.
+//
+// The residual race is the same one documented in #614 and is narrowed, not
+// closed: `reaped` is set the instant Wait returns and is checked immediately
+// before the kill, so the window is the gap between those two statements.
+// Closing it entirely needs a wait-without-reap (wait4(WNOWAIT) or pidfd).
+func Run(cmd *exec.Cmd) error {
+	// Setpgid makes the child the leader of a new group whose ID is its own
+	// PID, which is what lets a single negated PID address every descendant
+	// that has not created a group of its own.
+	if cmd.SysProcAttr == nil {
+		cmd.SysProcAttr = &syscall.SysProcAttr{}
+	}
+	cmd.SysProcAttr.Setpgid = true
+
+	runDone := make(chan struct{})
+	var reaped atomic.Bool
+
+	cmd.Cancel = func() error {
+		err := signalGroup(cmd, syscall.SIGTERM)
+		go func() {
+			select {
+			case <-time.After(termGrace):
+				if reaped.Load() {
+					// The pgid may already have been handed to somebody else;
+					// a late signal would hit an unrelated group.
+					return
+				}
+				if killErr := signalGroup(cmd, syscall.SIGKILL); killErr != nil {
+					slog.Debug("procgroup: group already gone before SIGKILL",
+						"path", cmd.Path, "err", killErr)
+				}
+			case <-runDone:
+			}
+		}()
+		return err
+	}
+	cmd.WaitDelay = waitDelay
+
+	if err := cmd.Start(); err != nil {
+		close(runDone)
+		return err
+	}
+	err := cmd.Wait()
+	reaped.Store(true) // before closing runDone: no signal may follow the reap
+	close(runDone)
+	return err
+}
+
+// signalGroup sends sig to the whole group led by cmd's child, reporting an
+// already-exited group as os.ErrProcessDone.
+//
+// That translation is required, not cosmetic: os/exec only tolerates a
+// cmd.Cancel error that is os.ErrProcessDone, so a raw ESRCH would surface out
+// of Wait and disguise a normal cancellation as a bogus "no such process"
+// failure on the caller's error path.
+func signalGroup(cmd *exec.Cmd, sig syscall.Signal) error {
+	if cmd == nil || cmd.Process == nil {
+		return os.ErrProcessDone
+	}
+	if err := syscall.Kill(-cmd.Process.Pid, sig); err != nil {
+		if errors.Is(err, syscall.ESRCH) {
+			return os.ErrProcessDone
+		}
+		return err
+	}
+	return nil
+}
--- a/daemon/internal/procgroup/procgroup.go
+++ b/daemon/internal/procgroup/procgroup.go
@@ -18,8 +18,10 @@
 //
 // Note for future readers: internal/executor carries a richer variant of this
 // logic (it also registers groups so TerminateAll can reach agents at shutdown,
-// and tolerates ErrWaitDelay when the payload is already complete). That was
-// deliberately left in place rather than folded in here — see #614.
+// and returns the collected payload alongside its ErrWaitDelay tolerance). That
+// was deliberately left in place rather than folded in here — see #614. The two
+// copies must agree on the timer ordering and on tolerating a clean exit whose
+// descendant holds the pipes; both are called out at their definitions.
 package procgroup
 
 import (
@@ -36,12 +38,22 @@ const (
 	// termGrace is how long the group gets to exit after SIGTERM before the
 	// escalation to SIGKILL. Short on purpose: cancellation here means a
 	// timeout has already elapsed, so the caller is past waiting politely.
-	termGrace = 2 * time.Second
+	//
+	// INVARIANT: termGrace must stay strictly — and comfortably — below
+	// waitDelay. On cancellation os/exec arms its own WaitDelay timer at the
+	// same moment Cancel fires, and when that timer expires os/exec kills only
+	// the DIRECT child and lets Wait return. If the two timers were equal (or
+	// termGrace were larger) the escalation goroutine would find `reaped`
+	// already true, skip kill(-pgid, SIGKILL), and leave alive precisely the
+	// descendant that ignored SIGTERM — reintroducing #665 on every timeout
+	// where the group does not die on the first signal. internal/executor
+	// encodes the same ordering as 1s vs 3s; keep both copies in step.
+	termGrace = 1 * time.Second
 	// waitDelay bounds Wait once the direct child has exited but a descendant
 	// still holds the inherited stdout/stderr pipes. Without it, Wait blocks
 	// until every copy of those descriptors is closed, which lets a stalled
-	// grandchild hang the caller indefinitely.
-	waitDelay = 2 * time.Second
+	// grandchild hang the caller indefinitely. See the termGrace invariant.
+	waitDelay = 3 * time.Second
 )
 
 // Run starts cmd in its own process group, waits for it, and on context
@@ -61,6 +73,9 @@ const (
 // when the child has been reaped, and a caller that forgot to do so would
 // eventually signal a pgid the kernel had already recycled onto an unrelated
 // process group.
+//
+// A clean exit whose descendant still holds the inherited pipes is reported as
+// success, not as exec.ErrWaitDelay — see the comment at that branch.
 //
 // The residual race is the same one documented in #614 and is narrowed, not
 // closed: `reaped` is set the instant Wait returns and is checked immediately
@@ -103,9 +118,38 @@ func Run(cmd *exec.Cmd) error {
 		close(runDone)
 		return err
 	}
+	pgid := cmd.Process.Pid // Setpgid makes the child its own group leader
 	err := cmd.Wait()
 	reaped.Store(true) // before closing runDone: no signal may follow the reap
 	close(runDone)
+
+	// WaitDelay also fires when the command itself exited cleanly but a
+	// descendant still holds the inherited pipes. That is common for git, not
+	// exotic: `git gc --auto` is spawned in the background after fetch/commit/
+	// merge, and an SSH remote using ControlMaster/ControlPersist leaves the mux
+	// master running on purpose. Returning ErrWaitDelay to the caller in that
+	// case would report a `git push` that actually succeeded as a failure and
+	// discard its output, so the pipeline would retry or fail an operation that
+	// already applied.
+	//
+	// Everything the command itself wrote is already buffered — it wrote it
+	// before exiting, and the copier drains concurrently — so the collected
+	// output is complete and the exit status is authoritative. Swallow the
+	// error and report success, matching internal/executor's handling of the
+	// identical case.
+	//
+	// The descendant is NOT swept here: the child has been reaped, so its pgid
+	// may already belong to another group and a signal would be aimed at
+	// strangers. Log its identity instead — this is the one place with positive
+	// evidence of a live leaked process, so it should be traceable while #614
+	// is open. With Setpgid the group ID is the leader's PID, so one number
+	// identifies both.
+	if errors.Is(err, exec.ErrWaitDelay) && cmd.ProcessState != nil && cmd.ProcessState.ExitCode() == 0 {
+		slog.Warn("procgroup: command exited 0 but a descendant held the output pipes; "+
+			"reporting success and leaving the descendant running",
+			"path", cmd.Path, "pgid", pgid)
+		return nil
+	}
 	return err
 }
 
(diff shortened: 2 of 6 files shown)


vbuenog (commented):
## 🤖 Heimdallm AI Review

Well-reasoned, unusually well-documented fix: git now runs in its own process group so cancellation reaps the forked ssh, and tini becomes PID 1 in both daemon images so already-orphaned grandchildren get reaped. Both git call sites are covered (gitops.captureGit and repoctx.execGit are the only exec.CommandContext("git") sites in the daemon) and both use CommandContext, so the Cancel-hook precondition holds. Two substantive issues remain, both of which the sibling implementation in internal/executor already got right and this new package diverges from: (1) termGrace equals cmd.WaitDelay, so the escalation SIGKILL to the group can be skipped exactly when it is needed; (2) exec.ErrWaitDelay is returned verbatim, so a git command that exits 0 while a descendant holds the pipes is reported to callers as a failure with its stdout discarded.

### Issues

⚠️ **daemon/internal/procgroup/procgroup.go:39** — termGrace (2s) is equal to waitDelay (2s), which makes the SIGKILL escalation racy in precisely the case it exists for. On cancellation os/exec calls Cancel (SIGTERM to the group) and simultaneously arms its own WaitDelay timer. If a descendant ignores SIGTERM, both timers fire at ~2s: os/exec kills only the direct child (Process.Kill), Wait returns, reaped.Store(true) and close(runDone) run — and the escalation goroutine then either takes the <-runDone branch or observes reaped==true and returns without ever sending SIGKILL to the group. The surviving descendant is exactly the leak #665 is about. internal/executor deliberately ordered these as processGroupTermGrace=1s < waitDelayAfterExit=3s; procgroup should keep termGrace strictly (and comfortably) below waitDelay.
⚠️ **daemon/internal/procgroup/procgroup.go:100** — Setting cmd.WaitDelay unconditionally means Wait returns exec.ErrWaitDelay even when the command exited 0 but a descendant still holds the inherited stdout/stderr pipes (an ssh ControlPersist mux master, a detached `git maintenance`/gc, a backgrounded credential helper). Run returns that error unchanged, so captureGit/execGit report a successful `git push`/`git fetch` as a failure and discard the already-buffered stdout — the pipeline then retries or marks the issue failed for an operation that actually succeeded. internal/executor treats this case explicitly (`errors.Is(runErr, exec.ErrWaitDelay) && ExitCode()==0` → return the collected output) and has a dedicated regression test (executor_waitdelay_test.go) calling out that "treating every non-nil Wait error as a failure throws away a complete" result. procgroup should apply the same tolerance (and, like executor, log the leaked pgid, which it currently drops silently).
🟡 **daemon/internal/procgroup/procgroup_test.go:79** — TestRun_KillsGrandchildOnTimeout reads the PID from cmd.StdoutPipe() on the test goroutine while another goroutine runs procgroup.Run → cmd.Wait(). os/exec documents that Wait closes the StdoutPipe read end and that reads must complete before Wait; with a 300ms context this is a real ordering hazard on a loaded CI runner — if the deadline fires first, Wait closes the pipe and ReadString returns os.ErrClosed, failing the test at t.Fatalf("read pid") for reasons unrelated to the behaviour under test. Give the context a more generous deadline (or gate cancellation on the PID having been read, e.g. read from an ExtraFiles pipe instead of stdout).
🟡 **daemon/internal/procgroup/procgroup_test.go:213** — TestRun_RejectsCommandWithoutContext asserts on the text of a stdlib error (strings.Contains(err.Error(), "CommandContext")). os/exec's message for this precondition is "exec: command with Cancel requires a Context", which does not contain "CommandContext" — so the assertion is either already failing or is coupled to an error string Go is free to reword. Assert only that Start failed (err != nil), or compare against the exact stdlib string with a comment pinning the Go version.
🟡 **daemon/internal/procgroup/procgroup.go:19** — The escalation path (a descendant that survives SIGTERM and must be SIGKILLed at termGrace) is untested — every test either exits promptly or is killed by the initial group SIGTERM. That is the one code path where the timing bug above lives. A test with a child that traps/ignores SIGTERM and a grandchild holding a probe pipe would pin it.
🟡 **daemon/internal/procgroup/procgroup.go:19** — The package doc acknowledges duplicating internal/executor's variant and defers unification to #614, which is a reasonable call, but the copy has already drifted in both constants and error handling (see the two medium findings). If the duplication stays, consider stating the invariant termGrace < WaitDelay in the const comments so the next editor of either copy cannot reintroduce the equal-timer race.

---

---
_Not approving: this review raised 6 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

Muriano (changes requested):
## 🤖 Heimdallm AI Review

El diagnóstico de #665 es correcto y la solución ataca la raíz por los dos lados: matar el grupo de procesos (para que `ssh` no quede huérfano) y añadir tini como PID 1 (para que los huérfanos ya existentes se cosechen). El paquete `procgroup` está bien documentado, la traducción de ESRCH a `os.ErrProcessDone` es necesaria y correcta (os/exec sólo tolera ese error desde `cmd.Cancel`), los tests cubren camino feliz, ExitError, fallo de Start, preservación de SysProcAttr y la regresión real con una sonda de pipe en vez de kill(pid,0) — ese detalle está especialmente bien pensado. Los Dockerfiles quedan consistentes entre sí y la decisión de no usar `-g` en tini está justificada (el daemon necesita gestionar su propio shutdown).

El problema es que `procgroup` copia la mecánica de `internal/executor` pero omite las dos calibraciones que allí sí existen: (1) executor.go:1435 tolera `exec.ErrWaitDelay` cuando `ProcessState.ExitCode() == 0`, con test de regresión dedicado (`executor_waitdelay_test.go`); aquí no hay nada equivalente, así que un `git` que sale 0 dejando un descendiente con los pipes abiertos (`git gc --auto`, que git lanza en background tras fetch/commit/merge, o un `ssh` con ControlMaster/ControlPersist) se convierte en error tras 2s. (2) executor usa `processGroupTermGrace = 1s` frente a `waitDelayAfterExit = 3s`, precisamente para que la escalada a SIGKILL del grupo ocurra antes de que os/exec corte el Wait; aquí ambos valores son 2s y compiten.

Nits para el summary, sin bloquear: los tests son específicos de Linux/macOS (syscall.Kill, Getpgid) y el paquete no tiene sufijo `_unix.go` ni build tag, por lo que un futuro build para Windows rompería en compilación — hoy irrelevante porque el daemon sólo se despliega en Linux/macOS. También sería útil, como hace executor, un `slog.Warn` con el pgid cuando se detecta un descendiente vivo, para poder rastrear la fuga mientras #614 siga abierto. Y merece la pena verificar en el runtime real que `apk add tini` deja el binario en `/sbin/tini` (así es en Alpine, pero el ENTRYPOINT es una ruta absoluta hardcodeada y un fallo ahí impide arrancar el contenedor entero).

### Issues

🔴 **daemon/internal/procgroup/procgroup.go:96** — `cmd.WaitDelay = waitDelay` sin tolerar `exec.ErrWaitDelay` en salida exitosa convierte comandos git correctos en fallos. Con WaitDelay activo, os/exec cierra los pipes heredados y devuelve `exec.ErrWaitDelay` desde Wait *incluso con exit code 0* si un descendiente sigue teniendo abiertos stdout/stderr. Escenario concreto: `git fetch` en repoctx dispara `git gc --auto` en background (git lo hace tras fetch/commit/merge) o el remoto SSH usa ControlMaster/ControlPersist y deja el mux `ssh` de fondo; git sale 0, pero 2 s después `procgroup.Run` devuelve ErrWaitDelay y tanto `captureGit` (gitops.go:388) como `execGit.Run` (repoctx/manager.go:1171) reportan el comando como fallido — un push que sí se completó se reporta como error, y el flujo de review/creación de rama falla o reintenta sobre estado ya aplicado. `internal/executor` ya resuelve exactamente esto en executor.go:1435 (`errors.Is(runErr, exec.ErrWaitDelay) && cmd.ProcessState.ExitCode() == 0` → devuelve la salida) con test de regresión propio; hay que replicar esa tolerancia en `procgroup.Run` (devolver nil cuando ProcessState existe y el exit code es 0) y cubrirla con un test análogo.
⚠️ **daemon/internal/procgroup/procgroup.go:61** — `termGrace` y `waitDelay` valen ambos 2 s, lo que pone la escalada a SIGKILL en carrera con el corte de Wait de os/exec y puede dejar vivo justo al nieto que este PR quiere matar. Al cancelarse el contexto, os/exec arranca su temporizador WaitDelay; cuando expira mata **solo al hijo directo** (Process.Kill) y hace que Wait retorne. La goroutine de escalada se despierta en el mismo instante y comprueba `reaped`: si Wait ganó la carrera, `reaped` ya es true, se omite el `kill(-pgid, SIGKILL)` y el `ssh` que ignoró/no procesó el SIGTERM sobrevive reparentado a PID 1 — es decir, se reproduce #665 en ~la mitad de los timeouts en que el grupo no muere con SIGTERM en 2 s. `internal/executor` evita esto con `processGroupTermGrace = 1s` frente a `waitDelayAfterExit = 3s`. Fix: hacer `termGrace` estrictamente menor que `waitDelay` (p. ej. 1 s / 3 s, alineado con executor) para que el SIGKILL al grupo se emita siempre antes de que os/exec pueda cosechar al hijo.

---

sergiotejon:
Both blocking findings were correct and are fixed in 50d2bd5. Thanks — the second one in particular I had reasoned about explicitly and got wrong.

**Timer ordering.** Fixed: `termGrace` 1s, `waitDelay` 3s, aligned with `internal/executor`. The invariant is now stated on both constants so neither copy can reintroduce it.

**`exec.ErrWaitDelay` on a clean exit.** Fixed, and the reasoning I'd used to justify propagating it was wrong. I argued that returning possibly-truncated output was worse than failing; the flaw is that **the command writes all of its output before exiting** and the copier drains concurrently, so the buffer is complete and the exit status is authoritative. The descendant only delays EOF — it does not truncate anything. Combined with `git gc --auto` (spawned in the background after fetch/commit/merge) and ControlPersist mux masters being routine rather than exotic, this would have turned successful `git push`es into reported failures. Now returns success and logs the leaked pgid, matching executor.

**Test hazard on `StdoutPipe`.** Fixed — the PID probe moved to an `ExtraFiles` pipe, so nothing races `Wait`.

**Escalation path untested.** Added, with a caveat worth recording. My first two attempts at that test were **vacuous**:

1. `trap '' TERM` in the parent shell only — the group SIGTERM still killed the grandchild directly, so the probe reached EOF and the test passed whether or not escalation fired.
2. Fixed that (both levels trap), and it now fails correctly when the escalation is removed entirely.

But it still **cannot pin the timer race**. I ran it against `termGrace == waitDelay == 2s` and it passes: with equal timers the outcome is a genuine race between this package's goroutine and os/exec's WaitDelay expiry, and the winner is scheduling-dependent — on this machine the escalation happens to win. So the ordering is enforced by the documented invariant on the constants, not by a test, and I've said so in the test's doc comment rather than let it imply coverage it doesn't have. Flagging in case you'd rather have a different mechanism there (a build-time assertion would be possible).

## One finding I'm disputing

> os/exec's message for this precondition is "exec: command with Cancel requires a Context", which does not contain "CommandContext" — so the assertion is either already failing or is coupled to an error string Go is free to reword.

The quoted string is not what Go emits. go1.25 `os/exec/exec.go:692`:

```go
return errors.New("exec: command with a non-nil Cancel was not created with CommandContext")
```

It does contain `CommandContext`, and the test was passing. The **second half** of the finding is fair though — coupling to stdlib wording is fragile regardless — so the assertion now checks only that `Start` failed.

## Not addressed

The Windows build-tag nit: `internal/executor` already uses `syscall.SysProcAttr{Setpgid: true}` unconditionally with no tags, so adding them here would leave the codebase inconsistent while the daemon still wouldn't build for Windows. Happy to do both packages in a follow-up if you want it.

`make test-docker` green across all 18 daemon packages. Re-requesting review.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #667 after 50d2bd5. Both blocking findings are genuinely fixed and correctly reasoned: (1) `termGrace` is now 1s vs `waitDelay` 3s, matching `internal/executor`, with the ordering invariant documented on both constants so the SIGKILL escalation always fires before os/exec can reap the direct child; (2) `Run` now swallows `exec.ErrWaitDelay` when `ProcessState != nil && ExitCode() == 0`, returning success and logging the leaked pgid — which is the right call, since the command wrote all of its output before exiting and the copier drains concurrently, so the buffer is complete and the exit status authoritative. The cancellation path is unaffected by that branch (a SIGTERM/SIGKILL'd child never reports exit 0). `signalGroup`'s ESRCH→`os.ErrProcessDone` translation still satisfies os/exec's requirement on `Cancel` errors, the escalation goroutine is bounded by `runDone` on every return path including the `Start` failure, and `reaped.Store(true)` before `close(runDone)` preserves the no-signal-after-reap ordering (residual window documented and accurately characterised). Test coverage is now meaningful rather than vacuous: the grandchild-liveness probe uses pipe EOF instead of `kill(pid,0)` (correctly, since the latter succeeds for zombies), the escalation test traps TERM at both levels so it fails when escalation is removed, and the new clean-exit-with-held-pipes test asserts elapsed >= waitDelay so it cannot silently stop exercising the branch. The stdlib error-string dispute is correct — go1.25 emits "exec: command with a non-nil Cancel was not created with CommandContext" — and dropping the string assertion in favour of "Start failed" is the better outcome. tini as PID 1 in both Dockerfiles, without `-g` so the daemon still owns its own shutdown sequence, is the right complement to the group kill. Nits, none blocking: `TestRun_ChildLeadsItsOwnGroup` still reads the PID from `cmd.StdoutPipe()`, contradicting `readPID`'s own doc comment; it is safe here only because `Wait` cannot return before the read (the kill happens after), so it is worth either moving to `ExtraFiles` for consistency or noting why this call site is exempt. `TestRun_ReportsSuccessWhenCleanExitLeavesADescendantHoldingThePipes` intentionally leaves a `sleep 30` behind; harmless but a `t.Cleanup` kill would keep the suite tidy. The duplicated logic against `internal/executor` remains a latent divergence risk, but the cross-references at both definitions are the pragmatic mitigation and #614 tracks the consolidation. On the timer-race coverage question: a compile-time assertion (`const _ = uint(waitDelay - termGrace - 1)`) would pin the invariant deterministically where a test cannot — cheap, and worth adding if you want it mechanically enforced rather than comment-enforced. The Windows build-tag point is fairly declined given executor's existing unconditional `Setpgid`. Nothing in this pass clears the reporting bar.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Re-review: both blocking findings are genuinely fixed and now match internal/executor's semantics exactly. termGrace is 1s vs waitDelay 3s with the invariant documented on both constants (verified against executor.go:24-31 which uses the same 1s/3s pair). The ErrWaitDelay-on-clean-exit branch (errors.Is(err, exec.ErrWaitDelay) && ProcessState.ExitCode()==0 -> return success, log the leaked pgid) is a faithful port of executor.go:1435 and is covered by a timing-guarded regression test. The StdoutPipe hazard is gone (PID now travels over an ExtraFiles pipe), and the escalation path has a non-vacuous test — trapping TERM at both levels is right, and SIG_IGN survives the inner `exec`, so `sleep` really does ignore SIGTERM and only the group SIGKILL can end it. The author's dispute is correct: go1.25 emits "exec: command with a non-nil Cancel was not created with CommandContext", so my earlier quoted string was wrong; dropping the message assertion is the right resolution anyway. Windows build tags are moot — daemon/.goreleaser.yml builds linux/amd64 only. Both git call sites (issues/gitops.go:377, repoctx/manager.go:1162) use exec.CommandContext, so Run's precondition holds, and those are the only two git exec sites in the daemon. tini is correctly wired in both Dockerfiles with no compose-level entrypoint override to bypass it. Remaining items are low-severity test-robustness and comment-accuracy nits; nothing blocking.

### Issues

🟡 **daemon/internal/procgroup/procgroup_test.go:100** — readPID has no deadline, and the test holds its own copy of the pid pipe's write end until after the read. If a loaded runner lets the 300ms context expire before the child's `echo $! >&3` runs, the child is SIGTERMed/SIGKILLed without ever writing, and ReadString blocks forever on pidR (EOF can never arrive because pidW is still open in the test process). The failure mode is a hang until go test's global 10m panic timeout rather than a clean t.Fatalf, in both TestRun_KillsGrandchildOnTimeout and TestRun_EscalatesToSIGKILLWhenGroupIgnoresSIGTERM. A read with a select/timeout (or a longer context, since cancellation is what the test controls) would fail loudly instead.
🟡 **daemon/internal/procgroup/procgroup.go:137** — The residual-race comment ("`reaped` is set the instant Wait returns and is checked immediately before the kill, so the window is the gap between those two statements") understates the window whenever WaitDelay is in play. cmd.Wait() reaps the leader internally via Process.Wait() and only then blocks on the pipe copiers until waitDelay expires, so reaped.Store(true) can trail the actual reap by up to ~3s — which spans the termGrace escalation at 1s. In practice this is still safe (the group cannot be recycled while any member lives, and an empty group yields ESRCH -> os.ErrProcessDone; reuse would need a full PID-space wrap within a second), and executor carries the same shape, but the comment as written would mislead the next reader reasoning about the bound. Worth restating as "between Process.Wait's internal reap and reaped.Store".
🟡 **daemon/internal/procgroup/procgroup.go:152** — The ErrWaitDelay comment cites `git gc --auto` as a routine descendant that "holds the inherited stdout/stderr". git's auto-gc detaches via daemonize(), which setsid()s and reopens fds 0/1/2 on /dev/null — so it neither holds the pipes nor stays in the process group the fix signals. The ssh ControlMaster/ControlPersist mux is the solid example and carries the argument on its own; the gc claim is worth verifying against the git version in the image or dropping, since the same assertion is repeated in the regression test's doc comment. No behavioural impact — the tolerance branch is correct either way.
🟡 **daemon/internal/procgroup/procgroup_test.go:302** — TestRun_ReportsSuccessWhenCleanExitLeavesADescendantHoldingThePipes deliberately leaves `sleep 30` running and never cleans it up, unlike the other tests which SIGKILL strays on the failure path. Harmless in CI, but it leaks a process for 30s after the suite finishes and is inconsistent with the hygiene the rest of the file maintains; capturing the sleep's PID (or killing its group after Run returns) would close it.

---

PR #668, opened 2026-08-05
fix(breaker): count review runs, not published reviews

### LINKED ISSUE #663: Pipeline failure releases the in-flight claim without a cooldown, causing an immediate duplicate review
## Symptom

Two full reviews were published for `theburrowhub/thaimaturgy#12` on the *same* HEAD SHA (`31615409a9152613021ddccfec2c6e2001d4575d`), 37 seconds apart, each running Claude end-to-end:

```
07:40:06  pipeline: starting review   pr=12 head_sha=3161540...
07:40:34  pipeline run failed         err="executor: parse JSON result: invalid character 'e' ..."
07:40:46  pipeline: starting review   pr=12 head_sha=3161540...   (same SHA, 11s after the failure's claim release)
07:44:25  pipeline: review done       severity=medium  github_review_id=4862190118
07:45:01  pipeline: review done       severity=medium  github_review_id=4862194642   <- duplicate
```

Both reviews were fully executed by Claude and published to GitHub (`4862190118`, `4862194642`), so this is a real double-spend, not just log noise.

## Root cause

`runReview` in `daemon/cmd/heimdallm/main.go` takes a persistent in-flight claim keyed on `(pr_id, head_sha)` before calling `p.Run`, and releases it unconditionally via `defer` (lines ~540-547):

```go
defer func() {
    if claimed {
        if err := s.ReleaseInFlightReview(claimPRID, claimSHA); err != nil {
            ...
        }
    }
}()
```

`pipeline.Run` only writes a `reviews` row on success (step 7, after Claude executes and returns a parseable result). When Claude's output fails to parse (as happened here — the diff contained JS template literals with backticks that broke JSON extraction), `p.Run` returns an error *before* `InsertReview`, so:

1. No review row exists for this `(pr_id, head_sha)`.
2. The in-flight claim is released by the `defer`.
3. The next review-worker tick (running every ~60s, but PR requeue can be much faster) sees no prior review for this SHA and no active claim — nothing stops it from starting a brand new, fully-priced review of the exact same commit.

This gap was introduced by #258 (persistent in-flight claims, commit `65557eb6`), which added the claim/release mechanism to survive daemon restarts but never paired it with a "failed attempt" cooldown — the claim's only job is deduping *concurrent* runs, not deduping a *fast retry after a hard failure*.

## Suggested fix

On `p.Run` error, persist a short-TTL "failed attempt" marker for `(pr_id, head_sha)` (e.g. reuse the `reviews_in_flight` table with a status column, or a new small table) so the next tick backs off for N minutes instead of retrying immediately. Should not apply to the explicit re-request path (`Force`), since that's operator-initiated.

## Repro data
- Repo: `theburrowhub/thaimaturgy` PR #12
- SHA: `31615409a9152613021ddccfec2c6e2001d4575d`
- Reviews: `github_review_id` 4862190118 and 4862194642, both `event=APPROVE severity=medium`, 37s apart
- Underlying parse failure: `executor: parse JSON result: invalid character 'e' looking for beginning of object key string` (diff contained `${esc(e.text||e.room_name||'')}` style JS in a Wails frontend PR)

sergiotejon:
Fixes #663.

## The bug is worse than the issue describes

The issue frames the impact as one duplicate review. It is **unbounded**.

`CheckCircuitBreaker` counted rows in `reviews`, which only exist once a run succeeded — `InsertReview` is step 7, after the CLI returns a parseable result. A run that died earlier left no trace anywhere:

| Layer | Keyed on | State after an early failure |
|---|---|---|
| circuit breaker | `SELECT COUNT(*) FROM reviews` | counter never advanced |
| in-flight claim | `reviews_in_flight` | released by the caller's `defer` |
| `PRAlreadyReviewed` | `PublishedAt` | never set |

All three blinded simultaneously, so the retry of the identical commit had **no upper bound**, each iteration paying full price. The incident stopped at 2 only because attempt 2 happened to parse — that's a property of the CLI's output, not of anything in the code. Same failure class as #243 (the €1,300 runaway).

`store/circuitbreaker.go` even stated the assumption without drawing the conclusion:

> *"Only reviews already persisted to SQLite count — an in-flight review that has not called InsertReview yet is gated separately via the inflight table (Task 5)."*

...and the inflight table is precisely what the `defer` deletes.

## Approach: count runs, not outcomes

Per the discussion, this does **not** add a cooldown. A cooldown limits *frequency*; against a deterministic failure it still retries forever, just slower. Counting attempts limits the *total*, which is what actually stops the bleeding — and it does so with the cap the operator already configured, so the existing 3/PR-SHA/24h bounds the loop at 3 with no new knobs.

New append-only `review_attempts` ledger, written after the breaker clears a run and before the CLI is invoked, never removed on failure.

## Decisions worth reviewing

**`max(attempts, reviews)` per axis, not either alone.** Attempts are authoritative going forward, but a DB upgraded in place has review history and no ledger rows — counting attempts alone would forget that history and grant a fresh full quota on the first tick after deploy. `max` rather than `sum` because a successful run writes one row of *each*; summing would double-count every normal review and halve the configured cap. Both directions are pinned by tests.

**Forced runs are ledgered even though `Force` bypasses the check.** A human clicking "Re-review" is deliberate intent, so the check is skipped — but the spend is real, and an unrecorded run would recreate the same blind spot for whatever runs next on that commit. Flagging it because it does mean an operator forcing N times consumes budget; in practice the automatic path wouldn't re-review that SHA anyway (a review row exists). Easy to invert if you disagree.

**Recording fails open**, matching the breaker check directly above it: a transient SQLite error must not block a legitimate review.

**Retention: 48h**, twice the widest breaker window (24h), so pruning can never reset a cap still in force. Added to the existing 5-minute sweep.

**Message wording changed** to "review runs" from "reviews". Reporting "3 reviews on this commit" when only one was published sent the operator hunting for reviews that don't exist. This required updating one assertion in `main_test.go` — flagging it since I touched an existing test.

**Why not the two cheaper options** I raised while analysing: not releasing the claim on failure (the 30-min sweep would act as a free cooldown) is ~3 lines but leaves the total unbounded and is wiped by `ClearAllInFlight` on restart; reusing `reviews_in_flight` with a status column breaks `ClaimInFlightReview`, which treats *any* row as in-flight.

## Tests

The end-to-end test was **verified to fail without the fix**: with attempt recording disabled, the 4th run invokes the CLI again instead of tripping — the unbounded loop, reproduced.

`daemon/internal/pipeline/pipeline_attempt_ledger_test.go`
- `FailedRunsCountTowardTheBreaker` — 3 runs all failing in `Execute` (real incident error string), 4th must return `CircuitBreakerError` with `Execute` calls still at 3; asserts the ledger carried the count and that no review row exists
- `ForcedRunIsRecordedEvenThoughItBypassesTheCheck` — documents the asymmetry

`daemon/internal/store/review_attempts_test.go` — 8 tests: failed-run counting, no double-counting, legacy DBs with no ledger rows, per-commit scoping (a follow-up commit stays reviewable), per-repo axis + repo isolation, window boundaries, retention contract, empty-SHA path.

`make test-docker` green across all 17 daemon packages (vet included).

## One thing found along the way

`PruneReviewAttempts(0)` does not delete same-second rows: `started_at` is second-granular and the comparison is strictly less-than — the same dead zone documented on `ClearAllInFlight` (#544). Harmless at the 48h the sweep uses, but I documented it on the method rather than leave it as a trap.

## Attribution correction

The issue blames `65557eb6` (#258). That commit only swapped a process-local `map[int64]bool` for the SQLite claim table, **preserving the existing release-on-defer semantics** — it also added persistence and SHA keying, both improvements. The unconditional release comes from `07df482c`, and the "persist only on success" half from `6f874cef` (18 minutes earlier, its direct DAG ancestor). Details in my comment on the issue. Worth correcting since #258 currently reads as the regression.

### DIFF
--- a/daemon/cmd/heimdallm/main.go
+++ b/daemon/cmd/heimdallm/main.go
@@ -778,6 +778,11 @@ func main() {
 			defer wg.Done()
 			const sweepInterval = 5 * time.Minute
 			const inflightSweepMaxAge = 30 * time.Minute
+			// The attempt ledger is only ever read through the breaker's
+			// windows (24h per-PR, 1h per-repo). 48h keeps a full margin over
+			// the widest of those: pruning anything still inside a window would
+			// silently reset a cap that is meant to be in force.
+			const attemptLedgerMaxAge = 48 * time.Hour
 			ticker := time.NewTicker(sweepInterval)
 			defer ticker.Stop()
 			for {
@@ -794,6 +799,11 @@ func main() {
 						slog.Warn("sweep: clear stale issue triage inflight failed", "err", err)
 					} else if n > 0 {
 						slog.Info("sweep: cleared stale issue triage inflight rows", "count", n)
+					}
+					if n, err := s.PruneReviewAttempts(attemptLedgerMaxAge); err != nil {
+						slog.Warn("sweep: prune review attempts failed", "err", err)
+					} else if n > 0 {
+						slog.Info("sweep: pruned review attempt rows", "count", n)
 					}
 				}
 			}
--- a/daemon/cmd/heimdallm/main_test.go
+++ b/daemon/cmd/heimdallm/main_test.go
@@ -1169,7 +1169,11 @@ drain:
 	if count != 1 {
 		t.Fatalf("circuit breaker events = %d, want 1", count)
 	}
-	if reason != "per-PR HEAD cap reached: 3 reviews on this commit in last 24h (cap 3)" {
+	// "review runs", not "reviews": the cap counts every run that spent CLI
+	// credits, including ones that failed before a review row was written
+	// (#663). Saying "3 reviews" when only one was ever published sent the
+	// operator looking for reviews that do not exist.
+	if reason != "per-PR HEAD cap reached: 3 review runs on this commit in last 24h (cap 3)" {
 		t.Fatalf("reason = %q", reason)
 	}
 }
--- a/daemon/internal/pipeline/pipeline.go
+++ b/daemon/internal/pipeline/pipeline.go
@@ -729,32 +729,6 @@ func (p *Pipeline) Run(pr *github.PullRequest, opts RunOptions) (*store.Review, 
 		}
 	}
 
-	// 4c. Record the attempt now: the breaker has cleared this run and every
-	// path from here leads to the CLI, so this is the moment the run becomes
-	// chargeable.
-	//
-	// It has to happen BEFORE Execute and must survive failure. `reviews` rows
-	// are only written on success (step 7), so a run that dies in between — a
-	// CLI whose output will not parse, say — used to leave no trace anywhere:
-	// the in-flight claim was released by the caller's defer, no review row
-	// existed, and the breaker's counter never moved. Every dedup layer keys off
-	// one of those two, so all of them were blind at once and the retry of the
-	// identical commit was unbounded, each iteration paying full price
-	// (theburrowhub/heimdallm#663).
-	//
-	// Forced runs are recorded too. Force bypasses the breaker CHECK because a
-	// human clicking "Re-review" is deliberate intent, but the spend is real and
-	// an unrecorded run would leave the same blind spot for whatever runs next.
-	//
-	// Fail-open on a write error, matching the breaker check just above: a
-	// transient SQLite blip must not block a legitimate review. The cost of
-	// being wrong is bounded by the caller's in-flight claim plus the dedup
-	// layers, which is the same posture the pre-existing code documents.
-	if err := p.store.RecordReviewAttempt(prID, pr.Head.SHA); err != nil {
-		slog.Warn("pipeline: could not record review attempt, proceeding",
-			"repo", pr.Repo, "pr", pr.Number, "head_sha", pr.Head.SHA, "err", err)
-	}
-
 	// All early-exit paths above are exhausted (gate, SHA-skip,
 	// legacy-backfill, circuit breaker): from this point we are committed
 	// to running the CLI and posting a real review. Both the desktop
@@ -771,6 +745,51 @@ func (p *Pipeline) Run(pr *github.PullRequest, opts RunOptions) (*store.Review, 
 	if stopped, err := p.stopIfRepoBecameIneligible(pr, opts.RepoEligible); stopped {
 		return nil, err
 	}
+	// 4c. Record the attempt. Every remaining path leads to Execute, which is
+	// what makes this — and not a line earlier — the point at which the run
+	// becomes chargeable.
+	//
+	// Placement is load-bearing. This deliberately sits AFTER
+	// stopIfRepoBecameIneligible: that guard is a zero-spend, explicitly
+	// retryable exit ("deferring review"), so recording before it meant a repo
+	// toggled off mid-poll burned a ledger row per tick without touching the
+	// CLI. After PerPR24h such ticks the commit was capped for 24h and the
+	// legitimate review was refused once the repo came back, recoverable only
+	// via Force.
+	//
+	// It has to happen BEFORE Execute and must survive failure. Rows in the
+	// reviews table are only written on success (step 7), so a run that dies in
+	// between — a CLI whose output will not parse, say — used to leave no trace
+	// anywhere: the in-flight claim was released by the caller's defer, no
+	// review row existed, and the breaker's counter never moved. Every dedup
+	// layer keys off one of those two, so all of them were blind at once and the
+	// retry of the identical commit was unbounded, each iteration paying full
+	// price (theburrowhub/heimdallm#663).
+	//
+	// Forced runs are recorded too. Force bypasses the breaker CHECK because a
+	// human clicking "Re-review" is deliberate intent, but the spend is real and
+	// an unrecorded run would leave the same blind spot for whatever runs next.
+	//
+	// Known trade-off: the ledger cannot tell a failure that spent credits from
+	// one that did not (provider auth rejection, an execution timeout, context
+	// cancellation on daemon shutdown). Those count too, so a restart loop can
+	// cap a commit for 24h and report "N review runs" for runs that cost
+	// nothing. Accepted deliberately — classifying CLI errors by whether they
+	// billed is guesswork, and under-counting is what #663 was. Force is the
+	// operator's escape hatch.
+	//
+	// Fail-open on a write error, matching the breaker check above: a transient
+	// SQLite blip must not block a legitimate review. Logged at Error, not Warn:
+	// this single write is what the entire cost brake now rests on, so a
+	// persistent failure (read-only DB, disk full) silently restores the
+	// unbounded retry loop and must be visible in the logs rather than inferred
+	// from the bill.
+	if err := p.store.RecordReviewAttempt(prID, pr.Head.SHA); err != nil {
+		slog.Error("pipeline: could not record review attempt — the per-PR/per-repo "+
+			"cost cap is not counting this run",
+			"repo", pr.Repo, "pr", pr.Number, "head_sha", pr.Head.SHA, "err", err)
+	}
+
 	p.notify.Notify("PR Review Started", fmt.Sprintf("%s #%d", pr.Repo, pr.Number))
 	p.publish(sse.EventPRDetected, map[string]any{
 		"pr_number": pr.Number,
--- a/daemon/internal/pipeline/pipeline.go
+++ b/daemon/internal/pipeline/pipeline.go
@@ -712,7 +712,7 @@ func (p *Pipeline) Run(pr *github.PullRequest, opts RunOptions) (*store.Review, 
 	}
 	slog.Info("pipeline: using CLI", "cli", cli)
 
-	// 4b. Circuit breaker: hard cap on review count per PR HEAD / per repo.
+	// 4b. Circuit breaker: hard cap on review runs per PR HEAD / per repo.
 	// Runs AFTER all dedup layers so it only fires when the dedup failed but
 	// the caller is about to spend Claude credits anyway. See
 	// theburrowhub/heimdallm#243.
@@ -727,6 +727,32 @@ func (p *Pipeline) Run(pr *github.PullRequest, opts RunOptions) (*store.Review, 
 				fmt.Sprintf("%s #%d: %s", pr.Repo, pr.Number, reason))
 			return nil, &CircuitBreakerError{Reason: reason}
 		}
+	}
+
+	// 4c. Record the attempt now: the breaker has cleared this run and every
+	// path from here leads to the CLI, so this is the moment the run becomes
+	// chargeable.
+	//
+	// It has to happen BEFORE Execute and must survive failure. `reviews` rows
+	// are only written on success (step 7), so a run that dies in between — a
+	// CLI whose output will not parse, say — used to leave no trace anywhere:
+	// the in-flight claim was released by the caller's defer, no review row
+	// existed, and the breaker's counter never moved. Every dedup layer keys off
+	// one of those two, so all of them were blind at once and the retry of the
+	// identical commit was unbounded, each iteration paying full price
+	// (theburrowhub/heimdallm#663).
+	//
+	// Forced runs are recorded too. Force bypasses the breaker CHECK because a
+	// human clicking "Re-review" is deliberate intent, but the spend is real and
+	// an unrecorded run would leave the same blind spot for whatever runs next.
+	//
+	// Fail-open on a write error, matching the breaker check just above: a
+	// transient SQLite blip must not block a legitimate review. The cost of
+	// being wrong is bounded by the caller's in-flight claim plus the dedup
+	// layers, which is the same posture the pre-existing code documents.
+	if err := p.store.RecordReviewAttempt(prID, pr.Head.SHA); err != nil {
+		slog.Warn("pipeline: could not record review attempt, proceeding",
+			"repo", pr.Repo, "pr", pr.Number, "head_sha", pr.Head.SHA, "err", err)
 	}
 
 	// All early-exit paths above are exhausted (gate, SHA-skip,
--- a/daemon/internal/pipeline/pipeline_attempt_ledger_test.go
+++ b/daemon/internal/pipeline/pipeline_attempt_ledger_test.go
@@ -2,6 +2,7 @@ package pipeline_test
 
 import (
 	"errors"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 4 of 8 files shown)


Muriano (commented):
## 🤖 Heimdallm AI Review

El diagnóstico y el diseño son correctos: `reviews` sólo se escribe en el paso 7, así que contar filas publicadas dejaba ciego al breaker ante runs que morían en `Execute` (#663). El ledger append-only, el `max(attempts, reviews)` (bien razonado: evita el doble conteo del run exitoso y preserva el historial de una BD actualizada in-place), el índice `(pr_id, head_sha, started_at)`, el prune a 48h y el mensaje "review runs" son todos acertados. La batería de tests es genuinamente buena: cubre el caso incidente, el no-doble-conteo, el upgrade sin ledger, el scoping por commit, el eje per-repo, la ventana y el prune (incluida la zona muerta de granularidad de segundo con maxAge negativo). Un único defecto real: el punto de inserción del `RecordReviewAttempt` está una línea antes de `stopIfRepoBecameIneligible`, por lo que la afirmación del comentario "every path from here leads to the CLI" no se cumple. Notas menores que no bloquean: (a) `review_attempts` no tiene FK ni ON DELETE hacia `prs`, así que si alguna vez se purgan PRs quedarán filas huérfanas que el eje per-repo dejará de ver (el JOIN las descarta) pero el eje per-PR seguiría contando — hoy inalcanzable, no es un issue; (b) el ledger contabiliza fallos transitorios (p. ej. un error de red del CLI) contra la cuota, de modo que tres fallos transitorios seguidos capan el commit 24h y el único recurso es Force — es el trade-off explícito y documentado del diseño, no un bug, pero conviene tenerlo presente en operación; (c) el diff llega truncado en `store.go`, así que no he podido verificar el resto del esquema ni si hay más ficheros tocados; reviso sólo lo visible.

### Issues

⚠️ **daemon/internal/pipeline/pipeline.go:729** — El attempt se registra ANTES de `stopIfRepoBecameIneligible`, que es una vía de salida temprana que no llega nunca al CLI. El comentario del bloque 4c afirma que "every path from here leads to the CLI", pero inmediatamente después el código hace `if stopped, err := p.stopIfRepoBecameIneligible(pr, opts.RepoEligible); stopped { return nil, err }`. Fallo concreto: si el repo se deshabilita (o deja de ser elegible) mientras el poll está en vuelo, cada tick consume una fila del ledger sin gastar un solo crédito; tras 3 ticks así el commit queda capado 24h y, cuando el repo se vuelve a habilitar, la review legítima de ese HEAD es rechazada por el breaker (sólo recuperable con Force). Fix: mover la llamada a `p.store.RecordReviewAttempt(prID, pr.Head.SHA)` justo después del bloque `stopIfRepoBecameIneligible` (y antes de `p.executor.Execute`), donde sí es cierto que todo camino restante llega al CLI; añadir un test que verifique que un run detenido por repo no elegible no deja fila en el ledger.

---

---

vbuenog (commented):
## 🤖 Heimdallm AI Review

Solid, well-motivated fix: the append-only `review_attempts` ledger correctly closes the #663 blind spot where a run that died before `InsertReview` advanced no counter, the `max(attempts, reviews)` blend sensibly handles in-place upgrades without double-counting, schema is created on every `Open` so existing DBs get the table, retention (48h) safely exceeds the widest breaker window (24h), and the test coverage is genuinely good (per-PR, per-repo, SHA scoping, window edges, prune, force, end-to-end). The main concern is that the ledger now charges the cap for failures that cost nothing: the record is written before `stopIfRepoBecameIneligible`, whose own log says it is *deferring* the review, and no distinction is made between a post-spend CLI failure (the incident) and a zero-cost one (provider auth error, timeout, daemon shutdown). With a typical cap of 3 that can lock a commit out for 24h and report "3 review runs" — the same class of misleading operator message this PR is fixing, inverted.

### Issues

⚠️ **daemon/internal/pipeline/pipeline.go:757** — The attempt is recorded before `stopIfRepoBecameIneligible`, so the comment "every path from here leads to the CLI, so this is the moment the run becomes chargeable" is not accurate. That helper (pipeline.go:201) returns `(true, nil)` and logs "repo no longer monitored, deferring review" — i.e. an explicitly retryable, zero-spend exit. If a repo is toggled off (or the eligibility check flaps) in the window between the breaker check and Execute, each poll tick burns a ledger row for a run that never touched the CLI; after `PerPR24h` such ticks the commit is capped for 24h and will be refused once the repo is re-enabled. Recording after that guard (it is only a few lines later, still before Execute) would preserve the fix's intent without charging deferred runs.
⚠️ **daemon/internal/pipeline/pipeline.go:736** — The ledger cannot distinguish a chargeable failure from a free one. `Detect` failures are caught earlier, but `Execute` can still fail without spending credits: provider auth/rate-limit rejection, an immediate non-zero exit, an execution-timeout kill, or context cancellation on daemon shutdown. All of these now count against the cap. With the default cap of 3, three consecutive misconfigured-CLI or shutdown-interrupted runs permanently block that commit for 24h and surface to the operator as "per-PR HEAD cap reached: 3 review runs on this commit", pointing at a cost runaway when the real cause is a broken CLI or a restart loop — the same misleading-message failure mode this PR sets out to fix. Consider only ledgering once the CLI has actually been invoked/produced output, or deleting the row for errors that provably precede invocation.
🟡 **daemon/internal/store/circuitbreaker.go:80** — `chargeableRuns` uses max(), which undercounts during the post-upgrade transition window. The doc comment only justifies max() against double-counting, but the other direction is unaddressed: a commit with N legacy `reviews` rows inside the live 24h window and M new ledger rows is charged max(N,M), not N+M. With cap 3, a commit that already had 2 legacy reviews still gets 2 more failing runs before tripping — up to 2*cap-1 chargeable runs in the worst case. Bounded and transient (one window after deploy), but worth stating in the comment since the current text reads as if max() were exact.
🟡 **daemon/internal/store/store.go:165** — `review_attempts.pr_id` has no `REFERENCES prs(id)` while `reviews.pr_id` does. Not load-bearing today (SQLite FKs are likely off anyway), but the asymmetry matters here: `CountReviewAttemptsForRepo` joins `prs`, so an orphaned attempt row silently vanishes from the per-repo axis while still counting on the per-PR axis. Adding the reference keeps the two axes reading the same data, as the doc on `CountReviewAttemptsForRepo` claims they do.
🟡 **daemon/internal/pipeline/pipeline.go:755** — Fail-open on `RecordReviewAttempt` is logged at Warn and nothing else. This is the single write that the whole cost brake now depends on, so a persistent failure (locked/read-only DB, disk full, a store opened before the migration ran) silently restores exactly the unbounded full-price retry loop #663 describes, with only a per-run Warn line to show for it. A rate-limited Error-level log or a counter surfaced to the UI would make that state observable rather than inferable from the bill.
🟡 **daemon/internal/pipeline/pipeline_attempt_ledger_test.go:110** — `if err == nil && rev != nil { t.Errorf(...) }` makes the test's own stated premise unverifiable: if `LatestReviewForPR` returns an error, the check that no review row exists is silently skipped rather than failing. Assert on the error explicitly (`t.Fatalf` on unexpected non-nil err, then require `rev == nil`) so the premise is actually pinned.

---

---
_Not approving: this review raised 6 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

sergiotejon:
The placement finding — raised independently by both reviewers — was correct and is fixed in ee65c58. My comment asserted "every path from here leads to the CLI" while `stopIfRepoBecameIneligible` sat on the very next line; that was straightforwardly wrong.

Moved to just after that guard, still before `Execute`.

**The new test needed two attempts to stop being vacuous**, which is worth recording because it shows the finding was real and easy to miss:

1. First version used `RepoEligible: func(string) bool { return false }`. That exits at the *first* eligibility guard (`pipeline.go:466`) and never reaches the record site — so it **passed against the buggy placement**.
2. Now it flips eligibility only at the final guard before `Execute`, with the number of expected checks pinned as a constant so a future guard being added or removed fails loudly rather than silently un-testing this.

With the record moved back, it fails with exactly the reported failure mode: `tick 3: deferring a review must not error: circuit breaker tripped: per-PR HEAD cap reached: 3 review runs on this commit in last 24h`.

## Other findings

**`chargeableRuns` undercount** — correct, and my doc only justified one direction. Now states that `max` is a floor: N legacy `reviews` rows plus M ledger rows charge `max(N,M)`, not `N+M`, so up to `2*cap-1` runs can be chargeable in the first window after an upgrade. Bounded and transient, and I'd still take it over permanently double-counting normal operation.

**FK on `pr_id`** — added, `REFERENCES prs(id)`, matching `reviews.pr_id`. Your reasoning is the convincing part: `CountReviewAttemptsForRepo` joins `prs`, so an orphaned row would silently drop off the per-repo axis while still counting on the per-PR axis, and the doc claims both axes read the same data.

**Fail-open logged at Warn only** — now Error, with the message naming the consequence. Agreed this is the one write the entire cost brake rests on.

**Test premise unverifiable** — fixed; the "no rows" error is now distinguished from an unexpected one instead of silently skipping the assertion.

## Deliberately not changed

**The ledger cannot tell a chargeable failure from a free one** (auth rejection, execution timeout, shutdown cancellation all count). Kept, and now documented at the record site with the operational consequence spelled out: a restart loop can cap a commit for 24h and report "N review runs" for runs that cost nothing.

Reasoning: classifying CLI errors by whether the provider billed is guesswork, and getting it wrong in the permissive direction is precisely what #663 is. `Force` is the operator's escape hatch. If it bites in practice, the cleaner fix is a signal from the executor about whether the provider was actually reached — a real capability rather than error-string heuristics.

You're right that it inverts the misleading-message problem this PR sets out to fix; I judged an over-strict brake to be the safer failure direction than an under-strict one. Happy to be argued out of it.

`make test-docker` green across all 17 daemon packages. Re-requesting review.

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-revisión: el hallazgo de placement está corregido. En el diff, el bloque 4c con `p.store.RecordReviewAttempt(prID, pr.Head.SHA)` queda ahora DESPUÉS de `stopIfRepoBecameIneligible` y antes de `Execute`, y el comentario ya no afirma que todo camino previo llegue al CLI (documenta explícitamente que la colocación es load-bearing). El test de regresión `TestRun_IneligibleRepoDoesNotConsumeLedgerQuota` no es vacuo: hace fallar la elegibilidad sólo en el último guard mediante `eligibilityChecksBeforeExecute`, verifica 0 filas en el ledger tras 5 ticks, 0 llamadas a Execute y que el breaker sigue intacto — exactamente el modo de fallo que reporté. También verificado: el log de fallo del write pasó a `slog.Error` nombrando la consecuencia; la doc de `chargeableRuns` ahora justifica ambas direcciones y acota el undercount a `2*cap-1` durante la primera ventana post-upgrade; el test de premisa distingue "no rows" de un error inesperado en lugar de saltarse la aserción; la FK `pr_id REFERENCES prs(id)` se añadió (fuera del diff visible, truncado). El sweep de poda a 48h mantiene margen sobre la ventana más ancha (24h) y el redondeo a segundos de `started_at` es irrelevante a esa escala.

Notas que no alcanzan el umbral de issue: (1) el ledger sigue sin poder distinguir un fallo que facturó de uno que no (auth rechazado, timeout, cancelación en shutdown) — declinado conscientemente con razonamiento explícito y `Force` como escape; coincido en que sobre-frenar es la dirección de fallo más segura y la solución limpia es una señal del executor sobre si se alcanzó al proveedor, no heurísticas sobre strings de error. (2) `max` en vez de suma deja el undercount transitorio ya documentado; aceptable frente a un doble conteo permanente. (3) El fail-open en el write del ledger significa que un DB read-only restaura silenciosamente el bucle ilimitado salvo por el log a Error — una métrica/alerta sobre ese contador sería el siguiente paso natural, no un defecto de este PR. Diff truncado: no revisé los hunks de migración de esquema ni el resto de `review_attempts_test.go`; no especulo sobre ellos. Converge.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

The re-review round addressed every actionable finding from the previous pass: the ledger write now sits after `stopIfRepoBecameIneligible` (verified against pipeline.go — the only things between the record site and `executor.Execute` are the notify/SSE emits and `applyStoredProfileCLIFlags`, neither of which returns early), `chargeableRuns` documents the max() undercount as a floor, the fail-open log is now Error with the consequence named, the FK was added, and the test premise around `LatestReviewForPR` is pinned. The remaining items are test-quality gaps in the two new pipeline tests, plus the deliberately-accepted trade-off about free failures counting — which is now documented at the record site and is weaker than I originally argued (the reason string says 'review runs', which is accurate, and the timeout case almost certainly did bill). No correctness defect found in the production code as it now stands.

### Issues

🟡 **daemon/internal/pipeline/pipeline_attempt_ledger_test.go:190** — TestRun_ForcedRunIsRecordedEvenThoughItBypassesTheCheck does not actually exercise the bypass. The comment claims "a forced run would trip it immediately if Force were checked", but with PerPR24h=1 and an empty ledger/reviews table the check evaluates n=0 >= 1, which is false — the run would proceed with Force=false too. The `fexec.calls != 1` assertion labelled "Force must bypass the check" therefore passes vacuously; only the recording half of the test is real. Seed one prior attempt (or a reviews row) for the same SHA before the forced run so the cap is genuinely saturated at check time.
🟡 **daemon/internal/pipeline/pipeline_attempt_ledger_test.go:274** — `eligibilityChecksBeforeExecute` does not give the "fails loudly" guarantee its comment claims, in the one direction that matters. Removing a guard is caught (Execute would run and `fexec.calls != 0` fires), but ADDING a guard before the record site is not: the fake flips to false at check #2, the run defers at the new earlier guard, and every assertion (runErr==nil, rev==nil, attempts==0, Execute not called) still holds — silently reverting the test to exactly the vacuous state v1 was in. Nothing asserts the deferral happened at the guard immediately preceding the record site. Deriving the check count dynamically (one all-eligible run to count calls, then flip on the last one) or asserting a marker that the record site was reached would close it.
🟡 **daemon/internal/pipeline/pipeline.go:757** — Accepted trade-off, noted for the record rather than as a change request: the ledger still charges runs where the CLI was invoked but nothing was billed (provider auth rejection, immediate non-zero exit), so a misconfigured CLI blocks a commit for 24h after PerPR24h attempts. The author's rationale holds and the documentation at the record site is honest about it; note also that two of the cases I originally cited are weaker than stated — Execute builds its context from context.Background() with a timeout (executor.go:1306), so daemon shutdown kills the process outright rather than producing a cancellable error, and a 5-minute timeout kill almost certainly did bill. The residual is the fast-fail auth case, mitigated by Force. If the executor ever gains a "provider was reached" signal, that is the clean place to fix it.

---

PR #687, opened 2026-08-06
test(cli): cover HTTP and SSE client contracts

### LINKED ISSUE #686: test(cli): cover HTTP and SSE client contracts
## Contexto

La auditoría de cobertura del CLI/TUI da 31,7 % global. El cliente HTTP/SSE compartido por todos los comandos solo cubre health; la mayoría de rutas, verbos, cabeceras, decodificación y errores no-2xx están sin protección de regresión.

## Alcance

Añadir tests de contrato en cli/internal/api/client_test.go, sin cambiar código de producción:

- Endpoints de lectura: método, ruta exacta, Accept, token y decodificación mínima.
- Mutaciones: método, ruta, token y propagación de status/cuerpo en errores no-2xx.
- SSE: ruta, cabeceras, comentarios, varios eventos y data multilínea.

## Criterios de aceptación

- Los tests fallan ante cambios incompatibles de método, ruta, autenticación o formato SSE.
- make -C cli test pasa.
- make -C cli test-race pasa.
- make -C cli lint pasa.
- La PR enlaza y cierra este issue.

## Relacionado

#632

ivanmunozruiz:
## Qué cambia

- Añade contratos table-driven para los 8 endpoints de lectura del cliente CLI.
- Cubre las 6 mutaciones con sus rutas, verbos, autenticación y propagación de errores no-2xx.
- Cubre el stream SSE: cabeceras, comentarios, múltiples eventos, datos multilínea y respuestas de error.
- Valida la URI completa para detectar parámetros de query inesperados.

## Por qué

La auditoría encontró que el CLI/TUI tenía un 31,7 % de cobertura global y que el cliente HTTP/SSE compartido por todos los comandos solo tenía cubierto `/health`. Una regresión de ruta, verbo, token o framing SSE podía pasar CI y romper simultáneamente comandos y dashboard.

## Impacto

No cambia código de producción. La cobertura de `cli/internal/api` sube de 29,9 % a 82,5 %, y la cobertura global del módulo CLI pasa de 31,7 % a 34,9 %.

## Validación

- `make -C cli test`
- `make -C cli test-race`
- `make -C cli lint`
- `go test -count=5 -race ./internal/api`
- `git diff --check`
- `gofmt -d cli/internal/api/client_test.go`

Closes #686
Related to #632

### DIFF
--- a/cli/internal/api/client_test.go
+++ b/cli/internal/api/client_test.go
@@ -10,6 +10,344 @@ import (
 	"time"
 )
 
+const apiContractToken = "contract-token"
+
+type contractRequest struct {
+	method string
+	path   string
+	accept string
+	token  string
+}
+
+func newContractServer(t *testing.T, status int, contentType, body string) (*httptest.Server, <-chan contractRequest) {
+	t.Helper()
+
+	requests := make(chan contractRequest, 4)
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		requests <- contractRequest{
+			method: r.Method,
+			path:   r.URL.RequestURI(),
+			accept: r.Header.Get("Accept"),
+			token:  r.Header.Get("X-Heimdallm-Token"),
+		}
+		w.Header().Set("Content-Type", contentType)
+		w.WriteHeader(status)
+		_, _ = fmt.Fprint(w, body)
+	}))
+	t.Cleanup(srv.Close)
+	return srv, requests
+}
+
+func assertContractRequest(t *testing.T, requests <-chan contractRequest, method, path, accept string) {
+	t.Helper()
+
+	var got contractRequest
+	select {
+	case got = <-requests:
+	case <-time.After(time.Second):
+		t.Fatal("client did not send a request")
+	}
+
+	if got.method != method {
+		t.Errorf("request method = %q, want %q", got.method, method)
+	}
+	if got.path != path {
+		t.Errorf("request path = %q, want %q", got.path, path)
+	}
+	if got.accept != accept {
+		t.Errorf("Accept header = %q, want %q", got.accept, accept)
+	}
+	if got.token != apiContractToken {
+		t.Errorf("X-Heimdallm-Token header = %q, want %q", got.token, apiContractToken)
+	}
+
+	select {
+	case extra := <-requests:
+		t.Errorf("client sent an unexpected additional request: %s %s", extra.method, extra.path)
+	default:
+	}
+}
+
+func TestReadEndpointContracts(t *testing.T) {
+	tests := []struct {
+		name string
+		path string
+		body string
+		call func(*Client) error
+	}{
+		{
+			name: "health",
+			path: "/health",
+			body: `{"status":"ok","version":"0.9.0"}`,
+			call: func(c *Client) error {
+				health, err := c.GetHealth()
+				if err != nil {
+					return err
+				}
+				if health.Status != "ok" || health.Version != "0.9.0" {
+					return fmt.Errorf("decoded health = %#v", health)
+				}
+				return nil
+			},
+		},
+		{
+			name: "PR list",
+			path: "/prs",
+			body: `[{"id":41,"repo":"acme/widget","number":7,"latest_review":{"id":5,"severity":"high"}}]`,
+			call: func(c *Client) error {
+				prs, err := c.ListPRs()
+				if err != nil {
+					return err
+				}
+				if len(prs) != 1 || prs[0].ID != 41 || prs[0].Repo != "acme/widget" ||
+					prs[0].LatestReview == nil || prs[0].LatestReview.Severity != "high" {
+					return fmt.Errorf("decoded PRs = %#v", prs)
+				}
+				return nil
+			},
+		},
+		{
+			name: "issue list",
+			path: "/issues",
+			body: `[{"id":42,"repo":"acme/widget","number":8,"latest_review":{"id":6,"action_taken":"review_only","triage":{"severity":"medium"}}}]`,
+			call: func(c *Client) error {
+				issues, err := c.ListIssues()
+				if err != nil {
+					return err
+				}
+				if len(issues) != 1 || issues[0].ID != 42 || issues[0].Repo != "acme/widget" ||
+					issues[0].LatestReview == nil || issues[0].LatestReview.ActionTaken != "review_only" {
+					return fmt.Errorf("decoded issues = %#v", issues)
+				}
+				return nil
+			},
+		},
+		{
+			name: "config",
+			path: "/config",
+			body: `{"server_port":7842,"repositories":["acme/widget"]}`,
+			call: func(c *Client) error {
+				cfg, err := c.GetConfig()
+				if err != nil {
+					return err
+				}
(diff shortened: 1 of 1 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured contract tests for HTTP and SSE client endpoints. Good use of table-driven tests, shared helpers, and both success/error paths. Only minor code-quality issues found.

### Issues

🟡 **cli/internal/api/client_test.go:257** — Inconsistent base URL construction: the error subtest uses `New(srv.URL, ...)` (no trailing slash) while the success subtest at line 249 uses `New(srv.URL+"/", ...)`. If the client handles these differently, the error path may hit a different route or produce a subtly different request path. Should be consistent unless both forms are intentionally tested.
🟡 **cli/internal/api/client_test.go:186** — `tc := tc` loop-variable rebinding is unnecessary in Go 1.22+ (where loop variables are per-iteration scoped). Harmless but should be removed if the module targets Go >= 1.22 to avoid confusing future readers.
🟡 **cli/internal/api/client_test.go:241** — Same `tc := tc` rebinding as above in the mutations test loop.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Well-constructed, table-driven test additions that pin down method, path, Accept and auth-token for every read/mutation endpoint plus the SSE stream, using a shared httptest harness. I traced every case against cli/internal/api/client.go and the expectations match the implementation (paths, `HTTP 409: ...` / `SSE HTTP 401: ...` error formats, multi-line `data:` joining with `\n`, comment-line skipping, `New()` trailing-slash trimming); the added identifiers don't collide and no import changes are needed. Note: the sandbox denied file writes and `gh`, so I reviewed by reading client.go rather than executing `go test` — the findings below are static observations. Remaining nits are test-robustness and coverage-symmetry issues, none blocking.

### Issues

🟡 **cli/internal/api/client_test.go:293** — TestStreamEventsEndpointContract calls StreamEvents with context.Background() and no deadline, on the test goroutine. StreamEvents uses `client := &http.Client{}` with no timeout (client.go:436), so any stall on the server side hangs until the 10-minute `go test` panic timeout instead of failing cleanly. The pre-existing TestStreamEventsReturnsWhenContextCanceledWithBlockedSend in this same file deliberately guards with context.WithCancel plus 1s `time.After` selects; the new test should use context.WithTimeout for consistency.
🟡 **cli/internal/api/client_test.go:291** — The events channel is sized `len(tc.wantEvents)+1`, which leaves room for exactly one unexpected event before a send blocks. Because StreamEvents' send is `select { case events <- ...; case <-ctx.Done(): }` and ctx is Background(), a parser regression emitting two or more extra events blocks forever rather than tripping the 'unexpected additional event' assertion — the test deadlocks instead of reporting the bug it was written to catch. Give the channel generous slack (or a bounded ctx) so over-emission surfaces as a failure.
🟡 **cli/internal/api/client_test.go:25** — newContractServer's handler does a blocking send into a 4-buffered `requests` channel before writing the response. If the client ever issues more than 4 requests (e.g. a redirect or retry loop regression), the handler goroutine blocks holding the connection. For do()-based calls the 30s HTTPClient timeout eventually breaks it, but for StreamEvents (no client timeout) it hangs the test run. A non-blocking send with `default:` plus an overflow counter, or a mutex-guarded slice, fails fast instead.
🟡 **cli/internal/api/client_test.go:61** — assertContractRequest hardcodes `got.token != apiContractToken`, so this harness structurally cannot exercise the empty-token path. Both newRequest (client.go:41) and StreamEvents (client.go:431) guard header emission with `if c.Token != ""`, and nothing in the suite asserts that X-Heimdallm-Token is absent when no token is configured. Parameterizing the expected token (or adding one case with `New(srv.URL, "")` asserting an empty header) would close an auth-adjacent branch this PR otherwise leaves untested.
🟡 **cli/internal/api/client_test.go:187** — Coverage is asymmetric: mutation endpoints get both success and error subtests, but the read endpoints get success only. The distinct decode-error messages in client.go ('parsing PRs', 'parsing issues', 'parsing config', 'parsing stats', 'parsing activity', 'parsing PR detail', 'parsing issue detail') and the non-2xx path through do() for reads are all unverified, so a swapped or dropped wrapper message would go unnoticed. For a PR framed as covering the HTTP client contract, at least one malformed-body and one non-2xx read case would round it out.
🟡 **cli/internal/api/client_test.go:240** — The error subtest builds the client with `New(srv.URL, apiContractToken)` while the sibling success subtest uses `New(srv.URL+"/", apiContractToken)`. Both work because New applies strings.TrimRight(host, "/"), but the divergence is unexplained. If it is deliberate coverage of the trailing-slash normalization it needs a one-line comment; otherwise make the two consistent so a reader doesn't hunt for significance that isn't there.
🟡 **cli/internal/api/client_test.go:78** — The 'health' read case substantially duplicates the existing TestGetHealthParsesDaemonVersion (same assertions on Status/Version), adding only the Accept/token checks. It also uses a plain 200, so it does not touch the behavior GetHealth actually special-cases — the 503-degraded decode path documented at client.go:135-143 — which remains covered only by the older dedicated tests. Consider trimming the duplicate or reusing the harness to also assert the 503 contract.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Test-only PR that adds contract coverage for the CLI API client: read endpoints (health, PR/issue lists, config, stats, activity, detail views), mutation endpoints (review triggers, promote/dismiss/undismiss, shutdown) with both success and 409 error paths, and the SSE stream (multi-line data, comment lines, 401 error). The assertions check method, path, Accept header and the X-Heimdallm-Token header on every call, plus that no extra request is sent — that last check is the most valuable part, since it pins the wire contract rather than just the decoded result. Decoding assertions are meaningful (nested latest_review, triage, by_severity map, truncated flag), and the error-path expectations ("HTTP 409: operation unavailable", "SSE HTTP 401: invalid API token") lock in the whitespace-trimming behaviour of the error formatting. Tests are deterministic: no sleeps, no shared global state, subtests are not parallel, and the only timing dependency is a 1s guard in assertContractRequest that only fires on a genuine failure.

Non-blocking observations, none of which meet the reporting bar:
- The error subtest builds the client with New(srv.URL, ...) while the success subtest uses New(srv.URL+"/", ...). This incidentally covers base-URL normalisation with and without a trailing slash, but it reads as an accident rather than intent; a short comment (or an explicit dedicated normalisation test) would make that coverage durable against a future cleanup that "fixes" the inconsistency.
- `tc := tc` loop-variable copies are redundant under the Go version this module targets (loopvar semantics changed in 1.22); harmless, just noise.
- The requests channel has a fixed buffer of 4. If a future client change ever issued more than 4 requests in one call, the handler would block on send instead of failing fast, and the test would hang rather than report. Latent only — nothing in the current client does this.
- The SSE success case reads events non-blockingly after StreamEvents returns, which is correct for the current synchronous implementation but couples the test to that property; if StreamEvents ever delivered asynchronously the assertions would flake toward "event not delivered". Worth a note if that method is ever reworked.
- Coverage gaps that could be added later: malformed JSON bodies on read endpoints, a non-2xx on a GET endpoint, and negative/zero IDs for GetPR/GetIssue path construction.

No correctness, security, or contract defects found in the diff.

---

PR #689, opened 2026-08-06
ci(coverage): enforce a 90% per-module ratchet

### LINKED ISSUE #688: ci(coverage): enforce a 90% floor per module with ratcheting
## Contexto

La auditoría actual mide daemon 68,37 % con coverpkg, Flutter 53,34 % y CLI 31,7 % en main. Activar directamente un floor global del 90 % dejaría CI permanentemente rojo y permitiría que módulos grandes ocultasen módulos pequeños sin cobertura.

## Objetivo

Fijar 90 % como mínimo por módulo (daemon, CLI y Flutter), exigirlo desde ahora sobre código de producción añadido o modificado y elevar la cobertura total mediante un ratchet que nunca permita retrocesos hasta alcanzar el 90 % global en cada módulo.

## Alcance de esta PR

- Medición reproducible de cobertura para daemon, CLI y Flutter.
- El daemon se mide exclusivamente dentro del target Docker canónico.
- Umbral de 90 % para statements/líneas ejecutables cambiadas en cada PR.
- Baseline total versionado por módulo que CI no permite reducir y que solo puede subir.
- Objetivo final de 90 % declarado y gate total automático al alcanzarlo.
- Exclusiones limitadas a código generado y justificadas en configuración.
- Tests unitarios del parser/gate: perfiles Go, LCOV, diff, líneas eliminadas, reportes vacíos y errores de formato.
- El mismo comando funciona localmente y en CI y publica un resumen legible.

## Criterios de aceptación

- Una PR con código nuevo sin tests falla si su cobertura de diff es menor de 90 %.
- Una PR que reduce la cobertura total de cualquier módulo falla aunque el diff supere 90 %.
- Una PR de tests puede elevar el baseline; bajarlo manualmente falla.
- Un reporte ausente, vacío o malformado falla de forma cerrada.
- Los tres módulos conservan sus suites verdes.

## Seguimiento

Las PRs posteriores añadirán tests por hotspot y elevarán los baselines hasta que los tres módulos alcancen 90 %.

Relacionado con #632 y #623.

ivanmunozruiz:
## Summary

- add a fail-closed coverage parser and policy gate for the daemon, CLI, and Flutter modules
- require at least 90% coverage on changed executable lines, exact non-decreasing total-coverage ratchets, and a hard 90% total floor once a module reaches it
- generate daemon coverage through the canonical pinned Docker target while keeping CLI and Flutter collection native
- measure the complete CLI module with `-coverpkg=./...` and exercise both the successful and failing entrypoint paths
- publish exact, reviewable baselines and justified frozen exemptions, with a Markdown CI summary
- pin Flutter 3.41.9 across test and release workflows, cache the containerized Go collector, and keep collection deterministic

## Why

The repository had coverage reports but no per-module policy that prevented regressions. A single aggregate threshold would allow one module to hide losses in another, and approximate percentages would permit small decreases. This gate compares exact covered/total pairs independently and fails closed on malformed or incomplete reports.

## Current ratchets

- daemon: `8437/12340` (68.37%)
- CLI: `947/2554` (37.08%), including the tests merged in #687
- Flutter: `4463/8427` (52.96%)

Until a module reaches 90% overall, its exact ratio cannot fall and changed executable lines must stay at or above 90%. Once it reaches 90%, the hard total floor also applies.

## Review follow-up

- initialize the Flutter test binding explicitly, keep real TabBar taps, scope the captured debug output, and document fatal hit-test warnings plus the coverage-universe contract
- tolerate harmless blank separators in both profile formats and treat LCOV `LF`/`LH` summaries as advisory while keeping `DA` authoritative
- document the deliberately independent universe-exemption allowlist and test its two-step policy
- reuse `GO_TEST_ARGS`, install Flutter dependencies explicitly, cache Docker Go data, and document why the atomic coverage job regenerates all profiles
- pin every test and release Flutter setup to 3.41.9
- fall back cleanly when a force-push's previous SHA is no longer fetchable, and cap the aggregate coverage job at 30 minutes

## Validation

- `make coverage-check` — PASS; CLI diff coverage `7/7` (100%)
- `make test-coverage-gate` — 29 tests
- `make coverage-daemon` and `make test-docker` — Go vet/tests in the pinned Docker image
- `make coverage-cli` and `make test-cli lint-cli`
- `make coverage-flutter` — 236 tests
- `cd flutter_app && flutter analyze`
- workflow YAML parsing, Makefile dry-runs, Python byte-compilation, and `git diff --check`

Closes #688


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured coverage ratchet with thorough input validation and defensive file I/O (atomic writes, temp files, fail-closed design). The main concern is a Flutter version inconsistency across CI jobs that could produce coverage numbers measured on a different toolchain than the primary test suite.

### Issues

⚠️ **.github/workflows/tests.yml:205** — Flutter version skew: the new `coverage` job and the updated `compose-isolation` job pin `flutter-version: '3.41.9'`, but the existing `flutter` job (line 195) still uses `flutter-version: '3.x'`. Coverage is measured on a potentially different Flutter version than the main test/analyze suite, which could cause the coverage gate to pass or fail based on instrumentation differences between versions rather than actual code changes.
🟡 **.github/workflows/tests.yml:176** — The `coverage` job does not run `flutter pub get` before `flutter test --coverage` (the existing `flutter` job has an explicit Dependencies step). While modern Flutter resolves packages implicitly, this makes the first test invocation slower and less deterministic in CI — an explicit `pub get` would also surface dependency resolution failures earlier and separately from test failures.
🟡 **Makefile:158** — GO_COVERAGE_ARGS hard-codes `-timeout 60s -count=1` independently of GO_TEST_ARGS. If someone overrides `GO_TEST_ARGS` to change the timeout (e.g. for a slow machine), the coverage path ignores it. Minor, but the parallel definitions could diverge silently.

---

vbuenog (commented):
## 🤖 Heimdallm AI Review

Solid, well-documented coverage-ratchet infrastructure (stdlib-only Python gate, containerized Go profile generation, fail-closed baseline schema with justified ignores). The Makefile/CI plumbing is careful. Main concerns are in the accompanying Flutter test changes: a new plain `test()` uses the test binary-messenger binding without any `TestWidgetsFlutterBinding.ensureInitialized()` in the repo, the globally-fatal hit-test warning changes semantics for every widget test, and the agents_test fix bypasses the tap path it used to exercise (a coverage/behavior regression rather than a fix). Also, the Flutter pin was applied only to tests.yml, leaving release.yml on '3.x'.

### Issues

⚠️ **flutter_app/test/features/dashboard_test.dart:95** — The new `SortNotifier handles preference load failures` case is a plain `test()` (not `testWidgets`) and is declared first in `main()`, yet it dereferences `TestDefaultBinaryMessengerBinding.instance`. Nothing in flutter_app/test calls `TestWidgetsFlutterBinding.ensureInitialized()` (grep finds zero occurrences, and flutter_test_config.dart only sets a static flag), so the binding may not be initialized when this test runs first and `.instance` will throw 'Binding has not yet been initialized'. Add an explicit `TestWidgetsFlutterBinding.ensureInitialized()` at the top of the test (or in testExecutable) so ordering cannot break it.
⚠️ **flutter_app/test/features/agents_test.dart:93** — Replacing `tester.tap(find.widgetWithText(Tab, ...))` with `DefaultTabController.of(...).animateTo(1)` silences the (now-fatal) missed-hit warning by removing the gesture entirely. The test no longer exercises TabBar hit-testing/tap handling — a real behavioral regression in test value driven by a coverage gate. It also hard-codes indices 1 and 2 without asserting the tab order matches the labels, so a reordering of tabs would silently assert the wrong screen. Prefer tapping the tab's actual hit target (e.g. `find.descendant(of: find.byType(TabBar), matching: find.text('Issue Triage'))` or the InkWell) rather than driving the controller.
⚠️ **flutter_app/test/flutter_test_config.dart:6** — `WidgetController.hitTestWarningShouldBeFatal = true` is applied repo-wide to every widget test. This turns previously-warning-only conditions into hard failures across the whole suite and for all future tests; the only in-PR remediation was to stop tapping altogether (see agents_test). This should be called out explicitly (docs/CONTRIBUTING) so contributors understand why taps start failing, and the flag's benefit weighed against the incentive it creates to bypass gestures.
⚠️ **.github/workflows/tests.yml:253** — Flutter is pinned to '3.41.9' in tests.yml only. `.github/workflows/release.yml` still uses `flutter-version: '3.x'` in three places, so release artifacts can be built with a different Flutter than the one the coverage baseline and tests were validated against. Pin release.yml to the same version (or centralize the version in one place) to keep the baseline meaningful.
🟡 **flutter_app/test/features/dashboard_test.dart:122** — `expect(messages, hasLength(1))` asserts on globally-captured `debugPrint` output. Any unrelated framework or provider debugPrint emitted during `pumpEventQueue()` (or a future one) will fail this test for reasons unrelated to SortNotifier. Filter to messages starting with 'SortNotifier:' and assert on that subset instead of the total count.
🟡 **scripts/coverage_gate.py:22** — `UNIVERSE_EXEMPT_PATHS` hard-codes two specific Dart paths in Python while the same information is also declared in `.github/coverage-baseline.json` (`scope: "universe"`). This duplicates policy in two places: adding a legitimate new platform stub requires editing the gate script itself, and the two lists can drift. Consider deriving the allowlist from a single source, or at minimum document the two-step process where the constant is defined.
🟡 **scripts/coverage_gate.py:300** — Both parsers reject blank lines outright (`blank lines are not valid`) and the LCOV parser hard-fails when declared LF/LH disagree with DA records. These are stricter than the LCOV format actually requires; a future Flutter/package:coverage formatter change (e.g. blank separator lines) would break CI with a parse error rather than a coverage verdict. Tolerating blank lines and treating LF/LH as advisory would make the gate more robust without weakening the policy.
🟡 **cli/cmd/heimdallm-cli/main_test.go:5** — `TestDefaultVersion` only asserts that the package-level `version` variable still equals its literal default. It exercises no behavior and will need updating if the default ever changes — it exists to add a test file to the package rather than to verify anything. Prefer a test that actually invokes the command's run path, or drop it.
🟡 **flutter_app/test/coverage_universe_test.dart:5** — This file keeps the coverage 'universe' complete by importing one screen, relying on transitive imports to pull in every other production library. Any newly added top-level library that no test imports will fail the gate with an error pointing at coverage rather than at the missing import, and the fix (adding an import here) is non-obvious. A comment explaining the maintenance contract — or generating the import list — would prevent confusing CI failures.
🟡 **.github/workflows/tests.yml:177** — The new `coverage` job re-runs the full daemon (in Docker), CLI, and Flutter suites that other jobs already run, roughly doubling CI wall-clock and cost for every PR, and it configures no Go build cache for the containerized daemon run (`/tmp/heimdallm-gocache` is not restored via actions/cache). Consider reusing profiles from the existing test jobs via artifacts, or caching the Go build dir.

---

---
_Not approving: this review raised 10 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

Muriano (commented):
## 🤖 Heimdallm AI Review

Well-engineered coverage ratchet: the gate is stdlib-only Python with strict fail-closed parsing (mode: atomic enforced, malformed/duplicate/overlapping inputs rejected, LF/LH cross-checked against DA records), the baseline schema validates exemptions hard (exact paths, no globs, mandatory reason + issue reference, `universe` scope restricted to an approved allowlist), and the Makefile refactor correctly reuses a single pinned, read-only, EDR-safe Docker boundary for the daemon profile instead of growing a second `docker run` that could drift. The recursive-make trick in `coverage-daemon`/`coverage-ci` is the right call for the `ifeq` on `GO_COVERAGE_PROFILE` and for ordering the gate after all three profiles under `-j`. Pinning Flutter to 3.41.9 in `compose-isolation` (was `3.x`) removes a real source of coverage drift. Note the diff is truncated (`scripts/coverage_gate.py` is 1009 lines and I can see roughly the first half, plus `scripts/tests/test_coverage_gate.py` is not visible at all), so the diff-coverage computation, the git-diff invocation, the universe/presence check and the summary rendering were not reviewable; I did not add findings to compensate for that.

Non-blocking observations for the summary: (1) `UNIVERSE_EXEMPT_PATHS` is hardcoded in the script and duplicates data that already lives in the baseline JSON — adding a new platform adapter stub will require a code change plus a config change, and the coupling is invisible from the baseline file. (2) `parse_lcov_text` rejects blank lines outright; package:coverage never emits them, but any future pass through `lcov`/`genhtml`-style merge tooling would hard-fail the gate rather than degrade. (3) The new `SortNotifier` test in `dashboard_test.dart` asserts `messages` has exactly length 1 while globally overriding `debugPrint`; any unrelated framework diagnostic emitted in that window turns it into a spurious failure — `messages.where((m) => m.startsWith('SortNotifier:'))` would be sturdier. (4) `flutter_test_config.dart` flips `hitTestWarningShouldBeFatal = true` for the entire suite; `agents_test.dart` was adapted, but this makes any future `tester.tap` on a non-hit-testable render object a hard failure repo-wide — worth a line in the contributing docs. (5) `ISSUE_RE` combines a trailing `$` with `fullmatch`, which is redundant. (6) The new `coverage` job has no `timeout-minutes`, and it runs three full suites plus a Docker pull; a hung Flutter run will sit for the 6h default.

### Issues

⚠️ **.github/workflows/tests.yml:210** — In 'Resolve coverage diff base', the push branch selects `base_ref="$BEFORE_SHA"` after checking only that it is non-empty and not all-zeros, never that the commit is actually present in the clone. `github.event.before` on a force-push is the *previous* head, which is orphaned on the remote and therefore not fetched by `actions/checkout` even with `fetch-depth: 0`. Concrete failure: force-push a branch covered by the workflow's push trigger, the `elif` matches, then `git rev-parse --verify "$base_ref^{commit}"` fails under `set -euo pipefail` and the job dies with a bare rev-parse error — even though the `origin/main` and `HEAD^` fallbacks written directly below exist precisely for this case and are unreachable because the earlier branch already matched. Fix: fold the existence check into the condition, e.g. `elif [[ "$GITHUB_EVENT_NAME" == "push" && -n "$BEFORE_SHA" && "$BEFORE_SHA" != "0000..." ]] && git rev-parse --verify --quiet "$BEFORE_SHA^{commit}" >/dev/null; then`, so a missing before-SHA falls through to `origin/main` instead of hard-failing.

---

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Re-review of #689 (coverage ratchet). The single previous MEDIUM finding — the push branch of 'Resolve coverage diff base' selecting `github.event.before` without verifying the commit exists locally — is unchanged in the diff. However, on re-examination the failure is not reachable in this repository: `.github/workflows/tests.yml` triggers `push` only on `branches: [main]`, and a force-push to main (the only way `github.event.before` can be orphaned and unfetched) is not a normal or permitted operation here. For ordinary main pushes the before-SHA is an ancestor present in the `fetch-depth: 0` clone, so `git rev-parse --verify "$base_ref^{commit}"` succeeds. I am therefore recording it as a latent risk rather than re-raising it as an issue: if the push trigger is ever widened to feature branches, or main force-pushes become possible, the `origin/main`/`HEAD^` fallbacks written right below become unreachable and the job dies with a bare rev-parse error. Cheap hardening if the author wants it: fold `&& git rev-parse --verify --quiet "$BEFORE_SHA^{commit}" >/dev/null` into the `elif` condition so a missing before-SHA falls through instead of hard-failing.

What I verified in the visible diff: the Makefile now has a single `GO_DOCKER_RUN` definition shared by test and coverage modes, preserving the read-only `/src` mount (the container profile is written to `/tmp` and streamed over stdout, which is consistent with that constraint); all profile writers use mktemp + `test -s` + atomic `mv` with an EXIT/HUP/INT/TERM trap, so an aborted run cannot leave a truncated profile that the gate would then read as real coverage; `coverage-ci` sequences the gate after generation via recursive make, which is correct under `-j`; `.coverage/` and `__pycache__/` are gitignored. The CLI `main.go` refactor into `run(args, stdout, stderr)` with an injectable `exitProcess` is a reasonable seam, and both new tests restore `os.Args`/`os.Stdout`/`os.Stderr`/`exitProcess` via `t.Cleanup`, so they are order-independent within the package. The `UNIVERSE_EXEMPT_PATHS` code-owned allowlist plus a required baseline reason/issue is a genuinely good two-key design: a baseline-only edit cannot exempt arbitrary product source. Baseline path validation rejects absolute paths, `..`, backslashes and NULs.

Nits and observations, none clearing the issue bar: pinning Flutter to `3.41.9` across release.yml/tests.yml removes patch drift but now needs manual bumps — worth a Renovate/Dependabot rule so it does not silently rot. `flutter_test_config.dart` making missed hit-tests fatal is a behavioural change for the whole existing widget-test suite; the two `find.descendant(of: TabBar)` fixes in `agents_test.dart` are presumably the fallout, and CI green is the evidence that nothing else regressed. The `SortNotifier` preference-failure test asserts on a `debugPrint` string prefix, which couples the test to log wording — acceptable, but it will need updating whenever that message changes. `flutter_app/lib/main.dart` line 375 being ignored for platform-dependent const-constructor attribution is the kind of exemption that should be periodically re-checked against the pinned Flutter version rather than left forever.

One caveat on completeness: the diff is truncated — `scripts/coverage_gate.py` (1039 lines) is cut off partway through baseline parsing, and `scripts/tests/test_coverage_gate.py` is not visible at all. I reviewed what was shown (module specs, ignore-rule semantics, percent/line-range validation, path safety) and found nothing that clears the bar; I am not speculating about the unseen hunks — notably `evaluate_gate`'s frozen-rule enforcement, the LCOV/Go profile parsers, and the diff-line extraction, which are where the real correctness risk in this PR lives and which the CI gate `make test-coverage-gate` is presumably exercising.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

All four previously-flagged MEDIUM issues are addressed: the binding is now explicitly initialized in the SortNotifier test, agents_test taps the real TabBar hit target instead of driving the controller, release.yml pins Flutter to 3.41.9 in all three places, and flutter_app/README.md documents the fatal hit-test-warning policy plus the coverage-universe maintenance contract. The LOW items also landed — debugPrint assertions are filtered to 'SortNotifier:' messages, the placeholder TestDefaultVersion was replaced with two real tests that exercise main()'s run/exit paths, UNIVERSE_EXEMPT_PATHS documents its deliberate two-part policy control, and the daemon Go cache is now restored via actions/cache. Remaining comments are minor robustness/maintenance points, none blocking.

### Issues

🟡 **.github/workflows/tests.yml:218** — The PR diff base is taken from github.event.pull_request.base.sha. On a pull_request event the checkout is the merge ref, whose first parent is the *current* base tip; base.sha can lag behind it when main advances between synchronize events. If coverage_gate.py diffs base_ref..HEAD directly (rather than computing a merge-base), unrelated commits merged into main since base.sha would be attributed to this PR's changed lines, so a PR could fail the diff-coverage gate on someone else's uncovered code. Using `git merge-base $PR_BASE_SHA HEAD` (or HEAD^1 of the merge commit) would make this deterministic.
🟡 **.github/coverage-baseline.json:46** — The line-scoped ignore for flutter_app/lib/main.dart pins line 375 by absolute number. Any edit above that point silently shifts the exemption onto an unrelated line, quietly excusing arbitrary code from the gate with no signal that the rule went stale. Consider anchoring these to a source directive (the parser already recognises `coverage:ignore`) or having the gate verify the exempted line still matches an expected snippet/pattern.
🟡 **Makefile:141** — The daemon coverage run reuses GO_TEST_ARGS verbatim, so it keeps `-timeout 60s` while adding `-covermode=atomic -coverpkg=./...`. Full-program instrumentation measurably slows every package; any suite currently close to the 60s ceiling will start flaking only in the coverage job, producing failures that look like gate failures rather than timeouts. Consider a separate, more generous timeout for the instrumented run.
🟡 **.github/coverage-baseline.json:5** — The per-module baselines are absolute covered/total counts committed to the repo, but nothing in the Makefile regenerates them — `make coverage` only produces profiles, and the comment tells contributors to update the ratchet by hand. Hand-transcribing six integers from three profile formats is error-prone and, if the gate requires the ratio to be non-decreasing, will need repeating on every coverage-improving PR. A `make coverage-baseline` target that rewrites the JSON from the generated profiles would remove that footgun.

---

PR #693, opened 2026-08-06
test(cli): cover issue detail command

ivanmunozruiz:
## Summary

- add end-to-end command coverage for `heimdallm-cli issue <id>` against an `httptest.Server`, including the HTTP method, path, token, and content negotiation contract
- cover human and JSON output, empty triage history, invalid IDs, HTTP failures, and malformed/empty JSON helper inputs
- keep output assertions semantic and independent of ANSI styling or map order
- advance the exact CLI coverage ratchet from `1062/2554` to `1160/2554` (45.42%)

## Validation

- `make test-cli`
- `make coverage-cli` (run twice with the same exact result)
- `make lint-cli`
- `make test-coverage-gate`
- `make coverage-check`
- `gofmt -d cli/internal/cli/issue_detail_test.go`
- `git diff --check`

### DIFF
--- a/.github/coverage-baseline.json
+++ b/.github/coverage-baseline.json
@@ -8,7 +8,7 @@
       "total": 12340
     },
     "cli": {
-      "covered": 1062,
+      "covered": 1160,
       "total": 2554
     },
     "flutter": {
--- a/cli/internal/cli/issue_detail_test.go
+++ b/cli/internal/cli/issue_detail_test.go
@@ -1,0 +1,297 @@
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/theburrowhub/heimdallm/cli/internal/api"
+)
+
+const issueDetailTestToken = "issue-detail-token"
+
+func TestIssueDetailCommandRendersCompleteIssue(t *testing.T) {
+	server := newIssueDetailServer(t, 17, http.StatusOK, `{
+  "issue": {
+    "id": 17,
+    "repo": "acme/platform",
+    "number": 81,
+    "title": "Handle offline startup",
+    "body": "first body line\nsecond body line",
+    "author": "carol",
+    "assignees": ["dave", "erin"],
+    "labels": ["bug", "help wanted"],
+    "state": "open",
+    "created_at": "2026-06-02T09:15:00Z",
+    "dismissed": true
+  },
+  "reviews": [
+    {
+      "cli_used": "codex",
+      "summary": "ready to implement\nwith safeguards",
+      "triage": {"severity": "high", "owner": "platform", "score": 4},
+      "suggestions": ["add a regression test", {"kind": "docs", "text": "update runbook"}],
+      "action_taken": "auto_implement",
+      "pr_created": 77,
+      "created_at": "2026-06-03T14:45:00Z"
+    }
+  ]
+}`)
+
+	out, err := executeIssueDetailCommand(t, server.URL, "issue", "17")
+	if err != nil {
+		t.Fatalf("issue command returned an error: %v", err)
+	}
+
+	assertIssueOutputContains(t, out,
+		"Issue #81 — Handle offline startup",
+		"Repo:", "acme/platform",
+		"Author:", "carol",
+		"State:", "open",
+		"Created:", "2026-06-02 09:15",
+		"Labels:", "bug, help wanted",
+		"Assignees:", "dave, erin",
+		"Dismissed:", "yes",
+		"Body", "first body line", "second body line",
+		"══ Triage 1 ══",
+		"Action:", "auto_implement",
+		"PR Created:", "#77",
+		"Date:", "2026-06-03 14:45",
+		"CLI:", "codex",
+		"Summary", "ready to implement", "with safeguards",
+		"Classification",
+		"severity:", "high",
+		"owner:", "platform",
+		"score:", "4",
+		"Suggestions",
+		"1. add a regression test",
+		`"kind": "docs"`,
+		`"text": "update runbook"`,
+	)
+}
+
+func TestIssueDetailCommandReportsNoTriages(t *testing.T) {
+	server := newIssueDetailServer(t, 9, http.StatusOK, `{
+  "issue": {
+    "id": 9,
+    "repo": "acme/api",
+    "number": 9,
+    "title": "Clarify health response",
+    "author": "bob",
+    "state": "closed",
+    "created_at": "2026-05-04T12:00:00Z"
+  },
+  "reviews": []
+}`)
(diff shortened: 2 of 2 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Well-structured test file covering the issue-detail CLI command. Tests are clear and cover the happy path, JSON output, error cases, and helper functions. Only minor style/robustness concerns.

### Issues

🟡 **cli/internal/cli/issue_detail_test.go:240** — captureIssueStdout mutates the global os.Stdout without any synchronization. While none of these tests call t.Parallel(), any future parallel test in this package that writes to stdout could race. Consider using cmd.SetOut() (cobra's writer redirection) or a pipe instead of swapping the process-wide file descriptor.
🟡 **cli/internal/cli/issue_detail_test.go:249** — os.Stdout is restored explicitly on line 249 and again by the deferred closure on line 245. The double-restore is harmless but the defer becomes dead code on the normal path; removing the explicit restore (or the defer) would be cleaner.
🟡 **cli/internal/cli/issue_detail_test.go:230** — captureIssueStdout and the assert helpers (assertIssueOutputContains, assertIssueOutputOmits) appear to duplicate similar helpers likely present in other test files in the same package (prefixed with 'Issue' to avoid name collisions). Consider extracting shared test utilities into a testutil file to avoid drift between copies.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Test-only PR adding cli/internal/cli/issue_detail_test.go (297 lines) plus the corresponding coverage-baseline bump (cli covered 1062→1160, total unchanged at 2554 — correct, since test files add no counted statements). I checked out the PR branch and read the implementation it exercises (cli/internal/cli/issue_detail.go, the shared helpers printJSONArray/printTriageMap in pr_detail.go, and api.Client.do/GetIssue) to validate the assertions rather than just reading the diff. Everything lines up: the HTTP error string really is "HTTP %d: %s" wrapped by "fetching issue: %w"; the invalid-ID path really returns "invalid issue ID: ...invalid syntax" and, with SilenceUsage:true on the root command plus cobra sending errors to stderr, stdout is legitimately empty; printTriageMap/printJSONArray really do early-return on "", "[]"/"{}", "null" and on unmarshal errors, so the table in TestDetailPrintHelpersIgnoreEmptyAndInvalidJSON is accurate; and score:4 rendered via %v on the float64 does print "4". Determinism looks fine: timestamps come back with a UTC location from the RFC3339 payload, so Format("2006-01-02 15:04") is timezone-independent; NO_COLOR is set the same way the existing config_test.go does; httptest servers are closed via t.Cleanup. Coverage is solid on both happy and sad paths (complete render, no-triages, --json, invalid ID, HTTP 500) and the fake server asserts method, path, token header and Accept header rather than blindly returning a body. Non-blocking observations for the summary only: (1) captureIssueStdout swaps the global os.Stdout, which is safe today because nothing in package cli calls t.Parallel() and t.Setenv would forbid it anyway — but it is a trap if a future test in this package is made parallel; a shared helper in util_test.go with a comment noting the constraint would be worth it, especially since pr_detail.go's near-identical renderer will presumably want the same harness. (2) The assertions match loose substrings of decorated output, so they will not catch label/ordering regressions (e.g. Labels and Assignees swapping) — acceptable for characterisation tests, just noting the limit. (3) printTriageMap iterates a map for non-severity keys, so multi-key triage output order is nondeterministic; the test only uses substring checks on "owner:" and "score:", so it is safe, but a future test that asserts a full block would flake. One caveat on my verification: I was not able to execute `go test ./internal/cli/` in this session (bash permission denied for the run), so the pass/fail claim rests on reading the implementation, not on observed output — CI should confirm. No unresolved reviewer concerns were visible.

---

vbuenog (approved):
## 🤖 Heimdallm AI Review

Test-only PR adding coverage for the `issue` detail command plus its JSON/print helpers, with the matching coverage-baseline ratchet bump. I verified every assertion against cli/internal/cli/issue_detail.go, pr_detail.go and cli/internal/api/client.go: fixture JSON field names match api.Issue/api.IssueReview, the request path/headers match Client.newRequest (X-Heimdallm-Token, Accept: application/json), the HTTP error string matches Client.do's "HTTP %d: %s", and the timestamp assertions are TZ-safe because CreatedAt stays UTC. The temp-file stdout capture avoids the os.Pipe deadlock risk of the existing pattern in cli/cmd/heimdallm-cli/main_test.go. Both flags (--host/--token) are always passed, so PersistentPreRun never touches the user config file or Docker discovery. I could not execute `go test`/`make coverage-cli` in this sandbox, so the 1062→1160 baseline number is unverified locally, but the CI gate is self-checking on that value. Remaining observations are test-quality nits, no correctness defects.

### Issues

🟡 **cli/internal/cli/issue_detail_test.go:281** — assertIssueOutputContains takes a flat variadic list of fragments and checks each independently, so label/value adjacency is never asserted. In TestIssueDetailCommandRendersCompleteIssue, swapping the Author and State values (or dropping a label while its value still appears elsewhere in the body/JSON) would still pass — e.g. printing `Author: open` / `State: carol` satisfies all four fragments. Consider asserting whole rendered lines (e.g. a `wantLines []string` of `"Author:  carol"`) or a fragment-pair type so the mapping is actually pinned.
🟡 **cli/internal/cli/issue_detail_test.go:169** — TestFormatJSONStringArray belongs here (formatJSONStringArray is in issue_detail.go), but TestDetailPrintHelpersIgnoreEmptyAndInvalidJSON exercises printJSONArray/printTriageMap, which are defined in pr_detail.go and shared by the PR detail command. There is no pr_detail_test.go yet, so a follow-up PR covering that file will either duplicate these cases or have its coverage silently pre-satisfied from an unrelated test file. Moving the shared-helper table (and the capture helper) into pr_detail_test.go or a detail_helpers_test.go would keep file-to-test locality.
🟡 **cli/internal/cli/issue_detail_test.go:254** — captureIssueStdout only captures stdout; cobra writes RunE errors to os.Stderr, so the invalid-ID and HTTP-500 tests leak "Error: fetching issue: HTTP 500..." into the `go test` output as noise even when passing. It also mutates process-global os.Stdout, which permanently forbids t.Parallel() anywhere in this package for tests that render output. Capturing stderr into the same helper (and asserting on it) would both silence the noise and let the tests verify the user-visible error text rather than only the returned error.
🟡 **cli/internal/cli/issue_detail_test.go:154** — Error-path coverage stops at ParseInt and HTTP >= 400. Two reachable failure branches are untested: a 200 response with a malformed body (GetIssue's "parsing issue detail: %w" wrap) and a wrong argument count (Args: cobra.ExactArgs(1)). Likewise, in text mode no review exercises the false side of the PRCreated/CLIUsed/Summary guards in issue_detail.go:80-95 — the only review rendered non-JSON has all three populated, so a regression in those conditionals (e.g. printing "PR Created: #0") would go unnoticed.

---

PR #694, opened 2026-08-06
test(daemon): cover launchagent lifecycle

ivanmunozruiz:
## Summary

- add black-box coverage for LaunchAgent install/uninstall behavior, plist content, filesystem permissions, and exact `launchctl` arguments
- exercise contextual failures for log directory creation, plist creation, bootstrap, and plist removal, plus uninstall idempotency
- isolate a `/bin/sh` fake `launchctl` behind an exclusive `PATH`, so tests can never invoke the host's real launch service
- advance the exact daemon coverage ratchet from `8437/12340` to `8470/12340` (68.64%); `launchagent/plist.go` reaches `33/35` statements

## Validation

- `make test-docker GO_TEST_ARGS="-timeout 60s -count=1 ./launchagent"`
- `make coverage-daemon` (full suite + vet in Docker; run twice with the same exact result)
- `make test-coverage-gate`
- `make coverage-check`
- `gofmt -d daemon/launchagent/plist_test.go`
- `git diff --check`

No daemon Go tests or builds were run on the host.

### DIFF
--- a/.github/coverage-baseline.json
+++ b/.github/coverage-baseline.json
@@ -4,7 +4,7 @@
   "diff_percent": 90,
   "modules": {
     "daemon": {
-      "covered": 8437,
+      "covered": 8470,
       "total": 12340
     },
     "cli": {
--- a/daemon/launchagent/plist_test.go
+++ b/daemon/launchagent/plist_test.go
@@ -1,0 +1,204 @@
+package launchagent_test
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/heimdallm/daemon/launchagent"
+)
+
+const plistName = "com.heimdallm.daemon.plist"
+
+func installFakeLaunchctl(t *testing.T, exitCode int, output string) string {
+	t.Helper()
+
+	binDir := t.TempDir()
+	capturePath := filepath.Join(t.TempDir(), "launchctl-args")
+	const script = `#!/bin/sh
+printf '%s\n' "$@" > "$HEIMDALLM_LAUNCHCTL_CAPTURE"
+printf '%s' "$HEIMDALLM_LAUNCHCTL_OUTPUT" >&2
+exit "$HEIMDALLM_LAUNCHCTL_EXIT"
+`
+	if err := os.WriteFile(filepath.Join(binDir, "launchctl"), []byte(script), 0o755); err != nil {
+		t.Fatalf("write fake launchctl: %v", err)
+	}
+
+	// Use an exclusive PATH so this test can never reach the real launchctl on
+	// the macOS CI runner. The fake uses only /bin/sh built-ins.
+	t.Setenv("PATH", binDir)
+	t.Setenv("HEIMDALLM_LAUNCHCTL_CAPTURE", capturePath)
+	t.Setenv("HEIMDALLM_LAUNCHCTL_OUTPUT", output)
+	t.Setenv("HEIMDALLM_LAUNCHCTL_EXIT", strconv.Itoa(exitCode))
+	return capturePath
+}
+
+func setupHome(t *testing.T) (home, plistPath, logDir string) {
+	t.Helper()
+
+	home = t.TempDir()
+	t.Setenv("HOME", home)
+	plistPath = filepath.Join(home, "Library", "LaunchAgents", plistName)
+	logDir = filepath.Join(home, "Library", "Logs", "heimdallm")
+	if err := os.MkdirAll(filepath.Dir(plistPath), 0o700); err != nil {
+		t.Fatalf("create LaunchAgents directory: %v", err)
+	}
+	return home, plistPath, logDir
+}
+
+func requireCapturedArgs(t *testing.T, capturePath string, want ...string) {
+	t.Helper()
+
+	data, err := os.ReadFile(capturePath)
+	if err != nil {
+		t.Fatalf("read captured launchctl arguments: %v", err)
+	}
+	if got, expected := string(data), strings.Join(want, "\n")+"\n"; got != expected {
+		t.Fatalf("launchctl arguments = %q, want %q", got, expected)
+	}
+}
+
+func requireMode(t *testing.T, path string, want os.FileMode) {
+	t.Helper()
+
+	info, err := os.Stat(path)
+	if err != nil {
+		t.Fatalf("stat %s: %v", path, err)
+	}
+	if got := info.Mode().Perm(); got != want {
+		t.Fatalf("mode for %s = %04o, want %04o", path, got, want)
+	}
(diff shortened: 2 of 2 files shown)


sergiotejon (approved):
## 🤖 Heimdallm AI Review

Clean, well-structured test-only PR that adds thorough lifecycle tests for the launchagent Install/Uninstall functions using a fake launchctl stub. The test design is solid: exclusive PATH prevents leaking to real launchctl, t.Setenv ensures proper cleanup, error paths and idempotency are covered. No correctness or security concerns.

---

Muriano (approved):
## 🤖 Heimdallm AI Review

Test-only PR (+204 lines) adding the first coverage for daemon/launchagent, plus the matching coverage-baseline bump (daemon covered 8437→8470, total unchanged, which is correct for a test-only diff). I verified every assertion against daemon/launchagent/plist.go: the plist content checks match the template (Label, ProgramArguments, StandardOutPath/StandardErrorPath under ~/Library/Logs/heimdallm, RunAtLoad, KeepAlive); 0600 for the plist and 0700 for the log dir match os.OpenFile/os.MkdirAll; the three error paths are triggered realistically (HOME pointing at a regular file → 'mkdir logs:', a directory at the plist path → 'create plist:' via O_WRONLY on a dir, fake launchctl exit 23 with stderr → 'launchctl bootstrap:' with the CombinedOutput text); Uninstall's ignored bootout error, its os.IsNotExist tolerance (idempotent second call) and the ENOTEMPTY 'remove plist:' path are all exercised; HOME="" makes os.UserHomeDir fail for both entry points. The fake launchctl is a good touch — replacing PATH entirely means the macOS CI runner can never reach the real launchctl and accidentally bootstrap a daemon, and the script uses only /bin/sh built-ins so the exclusive PATH is safe. Tests are deterministic: no t.Parallel (correct, since t.Setenv is used), no time or ordering dependencies, each subtest re-installs its own fake. Non-blocking observations: (1) the 'render plist' error branch is the only remaining uncovered path in the file, and it is hard to reach without an fs fault injector — fine to leave; (2) TestUninstallRemovesPlistWhenBootoutFails now pins the current semantics of Uninstall silently discarding the bootout error — reasonable to encode, but if that behaviour is ever tightened to surface the failure, this test will need updating, and it would read more clearly with a comment saying the swallow is intentional; (3) the second Uninstall in that test overwrites the capture file, so the idempotency assertion only checks the returned error, not the second invocation's arguments; (4) plist.go carries no darwin build tag, so these tests also run on Linux — they happen to be portable (os.Getuid, $HOME, /bin/sh), so that is fine today. I could not execute `go test ./launchagent/` locally in this session (bash was denied), so the review is static against the implementation source; nothing in the diff depends on unverifiable behaviour. No unresolved reviewer concerns visible. Nothing here clears the issue bar.

---

vbuenog (commented):
## 🤖 Heimdallm AI Review

Well-constructed test-only PR for daemon/launchagent. The fake-launchctl-on-an-exclusive-PATH approach is hermetic and correctly avoids reaching the real launchctl, and failure paths are provoked via ENOTDIR/EISDIR/ENOTEMPTY rather than permission bits, so the tests also pass when CI runs as root. Assertions cover plist contents, 0600/0700 modes, exact launchctl argv, Uninstall idempotency, and missing-HOME. Main concern is not a defect in the tests themselves but that the setupHome helper pre-creates ~/Library/LaunchAgents, which hides a real gap in Install: nothing in the codebase ever creates that directory, so a fresh macOS account would fail. A few smaller coverage gaps and fragility notes below.

### Issues

⚠️ **daemon/launchagent/plist_test.go:46** — setupHome pre-creates ~/Library/LaunchAgents via os.MkdirAll before every Install/Uninstall test. Install (plist.go:59) only OpenFile()s the plist and never MkdirAll's its parent, and scripts/macos-install.sh does not create it either (only ~/Library/Logs/heimdallm is created, and only by Install). On a fresh macOS user account ~/Library/LaunchAgents does not exist, so Install would fail with 'launchagent: create plist: no such file or directory'. The test helper masks this: there is no case exercising Install with the LaunchAgents directory absent. Either add MkdirAll(filepath.Dir(path), 0700) to Install and assert it, or add a test asserting the current (failing) behaviour so the gap is visible.
⚠️ **daemon/launchagent/plist_test.go:154** — TestUninstallRemovesPlistWhenBootoutFails locks in the fact that Uninstall discards the launchctl bootout error entirely (plist.go:89 calls .Run() with no error check and no output capture). Ignoring it is defensible for the 'not loaded' case, but the test as written also makes a genuine bootout failure (e.g. the agent is loaded and refuses to unload) indistinguishable from success while the plist is deleted anyway — leaving a running daemon with no plist. Consider asserting on a warning/log emitted for non-idempotent failures, or narrowing what the test blesses.
⚠️ **daemon/launchagent/plist_test.go:79** — The plist is rendered with text/template (plist.go:13), which performs no XML escaping. The happy-path test only uses an alphanumeric binaryPath, so a path containing '&', '<' or '>' — which would produce a structurally invalid plist that launchctl rejects — remains uncovered. A cheap extra case with such a path would either document the constraint or surface the bug.
🟡 **daemon/launchagent/plist_test.go:22** — The fake launchctl writes captured argv with '>' (truncate), so only the last invocation is ever observable. TestUninstallRemovesPlistWhenBootoutFails calls Uninstall twice and happens to assert between the calls, but any future test expecting two launchctl invocations would silently only see the second. Appending ('>>') plus an explicit invocation count would be more robust.
🟡 **daemon/launchagent/plist_test.go:126** — The 'logs directory' and 'plist creation' subtests install the fake launchctl but never assert it was NOT invoked. Install should fail before bootstrapping; asserting the capture file is absent would pin that ordering, which is the security-relevant part (do not bootstrap a plist that failed to write).
🟡 **daemon/launchagent/plist_test.go:14** — plistName is duplicated from the unexported constant in plist.go. Acceptable in an external _test package, but the two can silently drift; deriving the expected path from an exported helper (or adding an export_test.go) would keep them in sync.
🟡 **daemon/launchagent/plist_test.go:108** — Two branches remain uncovered: the 'launchagent: render plist' error path (plist.go:69) and Uninstall's success path with a bootout that exits 0. The latter is a one-line addition (installFakeLaunchctl(t, 0, "")) and would confirm the normal uninstall flow, which is currently only exercised through a failing bootout.
🟡 **.github/coverage-baseline.json:7** — The daemon baseline is bumped 8437 -> 8470 (+33) with total unchanged, which is consistent for a test-only change. Worth confirming the number came from an actual coverage run rather than an estimate, since the gate fails on any mismatch.

---

---
_Not approving: this review raised 8 findings above, and `never_approve_with_issues` is enabled for this repo, so it is posted as a comment instead of an approval. Address or dispute the blocking findings and re-request a review to get an approval._

